PR #72, opened 2015-02-02
Update GrblDriver.java

cri-s:
Grbl Driver update

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/GrblDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/GrblDriver.java
@@ -28,8 +28,6 @@ import java.util.Locale;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.TimeoutException;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import javax.swing.Action;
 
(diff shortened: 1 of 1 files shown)


vonnieda:
Can you please describe what the purpose of the changes are?

cri-s:
The purpose of this change is just update for Grbl Driver.
The status of the old Driver:
Homing not work, what instead of homing  work is just Go to pos 0 0 0
Only custom Grbl version work, no official version, even older works.
No support for switching to mm unit, it just use the configured unit.
Stop button don't interrupt anything, in fakt it emit nothing on the
driver side.
No support for 3 Axis Grbl

This modification cleanup with this issues and support all versions of Grbl.


2015-02-02 16:26 GMT, Jason von Nieda <notifications@github.com>:

cri-s:
I have just seen there are unicode probles, i have just copy and paste this code from editor.
I must manually check in, so wait pls because the code compiles and runs without problems,
but here sometimes ): is added.

src/main/java/org/openpnp/machine/reference/driver/GrblDriver.java
```diff
@@ -51,92 +49,84 @@
 /**
  * TODO: Consider adding some type of heartbeat to the firmware.  
  */
+
 public class GrblDriver extends AbstractSerialPortDriver implements Runnable {
+
```
vonnieda: These two variables do not appear to be used. Can they be removed?

src/main/java/org/openpnp/machine/reference/driver/GrblDriver.java
```diff
@@ -51,92 +49,84 @@
...
             @Override
             public void configurationComplete(Configuration configuration)
                     throws Exception {
                 connect();
             }
-        });
-	}
-	
-	@Override
-	public void actuate(ReferenceActuator actuator, boolean on)
```
vonnieda: Why was this removed?

src/main/java/org/openpnp/machine/reference/driver/GrblDriver.java
cri-s: do you mean
+//	private static final double minimumRequiredVersion = 0.75;
+//	private double connectedVersion;

This are commented out. For now i prefer leaving it as is until
jogging is implemented, because when implementing
Jogging it can be possible that it is advisable to limit version numbering.
I don't have checked it for now.
On jogging, it can be advisable to limit only firmware builds that
have lookahead builded in
for not decellerating and stop during moves or direction/speed changes.

vonnieda: I was referring to vacPumpPin and invertVacpump. If they are not needed then they should be removed.

src/main/java/org/openpnp/machine/reference/driver/GrblDriver.java
```diff
@@ -51,92 +49,84 @@
...
+            sb.append(String.format(Locale.US, "X%2.4f ", x));
+        }
+        if (!Double.isNaN(y) && y != this.y) {
+            sb.append(String.format(Locale.US, "Y%2.4f ", y));
+        }
+        if (!Double.isNaN(z) && z != this.z) {
+            sb.append(String.format(Locale.US, "Z%2.4f ", z));
+        }
+        if (!Double.isNaN(c) && c != this.c) {
+            sb.append(String.format(Locale.US, "E%2.4f ", c));
```
vonnieda: This should reference the C axis, not E.

vonnieda:
Changes look good overall. I added a few comments where I'd like to see the code improved before merging.

In the future, please try to limit pull requests to a single feature, fix or change. Having many different changes in a pull request makes it very hard to verify.

PR #79, opened 2015-02-05
Update Camera.java

cri-s:
Small changes to Camera SPI, for OpenCV
view = viewport for user
image = possibly rectified image
picture = for opencv, takes snapshoot and return image.

### DIFF
--- a/src/main/java/org/openpnp/spi/Camera.java
+++ b/src/main/java/org/openpnp/spi/Camera.java
@@ -67,7 +67,23 @@ public interface Camera extends Identifiable, HeadMountable, WizardConfigurable,
 	 * Immediately captures an image from the camera and returns it in it's native format.
 	 * @return
 	 */
-	public BufferedImage capture();
+ 	public BufferedImage capture();	// raw
+	/**
+	 * Immediately view an image from the camera and returns it in it's native format.
+	 * @return
+	 */
+ 	public BufferedImage view();	// return current image
+	/**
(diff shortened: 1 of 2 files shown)


vonnieda:
This change will need to pass the Travis CI build before it can be merged. A change to an SPI class will, at least, require changes to the base abstract classes. 

I'd also like some information as to the purpose of this change. What is it for and how is it used?

cri-s:
Visual processing during automatic placement access the picture() image.
This ensure that it gets possibly undistorted image that are granted to be taken after the procedure call.
Otherwise it happens that the image is not usable because taken in motion, and without global shutter
this is really problematic.

vonnieda:
This is the purpose of the capture() call. capture() is intended to take a
still frame using whatever method the camera driver prefers. This is why it
is used by the vision providers. Many of the existing drivers also use
capture to return a steam of images for video, but that is secondary to
it's real purpose.

Currently, if a vision operation requires a still frame it will pause
movement for a short period of time and then call capture() before moving
again. This is a bit scattered through the code. I would like for this to
eventually be part of the VisionProvider but that is going to require more
work.

cri-s:
Yes, actually streaming works with capture, and it will continue to work.
view is for streaming in order to have the possibility to add image enhanchments to video streaming.
it is open if switching to that in the future, defining it is not bad.
image is proposed for taking rectified and maybe scaled images.
and photo is for be sure that the image is not old.
stopping can be bad solution because depending on configuration, the snapshoot can be made during moving.

PR #122, opened 2015-07-02
Adding support for FireStep with rotational delta kinematics

FlyingLotus1983:
Other minor changes: 
* ReferenceMachineConfigurationWizard.java - Added FireStepDriver to combo box.  Modified the MarlinDriver to reflect new package name.
* Location.java - Added method for returning 3d distance between 2 locations
* Renamed / re-organized packages (renamed org.firepick to org.firepick.driver, and added other packages to keep everything organized.

### DIFF
--- a/src/main/java/org/openpnp/model/Location.java
+++ b/src/main/java/org/openpnp/model/Location.java
@@ -104,6 +104,10 @@ public class Location {
 		return (Math.sqrt(Math.pow(this.x - x, 2) + Math.pow(this.y - y, 2)));
 	}
 	
+	public double getXyzDistanceTo(Location location) {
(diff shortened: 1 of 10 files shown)


src/main/java/org/openpnp/model/Location.java
```diff
@@ -104,6 +104,10 @@ public double getLinearDistanceTo(double x, double y) {
 		return (Math.sqrt(Math.pow(this.x - x, 2) + Math.pow(this.y - y, 2)));
 	}
 	
+	public double getXyzDistanceTo(Location location) {
+		return (Math.sqrt(Math.pow(this.x - location.getX(), 2) + Math.pow(this.y - location.getY(), 2) + Math.pow(this.z - location.getZ(), 2)) );
```
vonnieda: This should convert units before calculation, otherwise an incoming Location could be of a different unit and the results won't make sense. I realize that some other methods in this class have the same problem, so I am mostly putting this here as a note to myself to fix em all.

PR #145, opened 2015-09-06
	modified:   src/main/java/org/openpnp/gui/MainFrame.java

DouglasPearless:
new file:   src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java
	modified:   src/main/java/org/openpnp/model/eagle/EagleLoader.java


DouglasPearless:
This pull request adds native Eagle PCB .brd support and removes the need for importing .mnt and .mnb files from the mountsmd.ulp programme.

PR #157, opened 2015-10-05
Develop

DouglasPearless:
Eagle Board importer, works for all footprints except polygons which are on the todo list, merged into the BoardPad change, includes SolderPaste generation.


vonnieda:
Hi @DouglasPearless, thanks for the pull request. I have a few notes:

1. Don't comment out imports. If you don't need them, just remove them. Leaving them in and commented creates clutter. See Footprint.java and EagleBoardImporter.java. If you are using Eclipse just use the Organize Imports command to clean everything up (after uncommenting the commented ones).
2. The integrated test suite did not pass. Please make sure you run `mvn test` before submitting a pull request. If it doesn't pass there is something wrong. In this case, the changes you made to Footprint are not compatible with any existing Board files and cannot be merged.
3. You added libraryId and accessors for it to Part, but you have commented out the `@Attribute` annotation, which means this data will not get saved in the configuration. If you don't want to save this data, do we need it?
4. I don't understand your changes to Footprint.java. This should not be used for solder paste at all. Footprint.java is currently only used for Fiducial finding. Your importer should just create BoardPad objects and add them directly to Board.solderPastePads. See https://github.com/openpnp/openpnp/blob/develop/src/main/java/org/openpnp/gui/importer/SolderPasteGerberImporterDlg.java#L192 for an example of how this is done in the Gerber importer.

DouglasPearless:
Hi,

I have been back and reviewed your comments:

1. fixed
2. I have found the error but don’t quite understand what is meant by "testSimpleJob(BasicJobTest): Error while reading packages.xml (Element 'pad' does not have a match in class org.openpnp.model.Footprint at line 8)”
3. fixed
4. The changes are to associate solder pads with board=>part=>package=>Footprint=>Solderpads, and not a disconnect jumble of pads like the Gerber import does.  This means that the user can select just the parts to PnP and to paste, which is impossible to do the way the Gerber import works. If there is a better way to associated tCream and bCream pads (and polygons) to allow this functionality them I am more than happy to change the code, but I think the logic is sound.  A good example would be certain parts will be hand assembled after the board has been reflowed, say for parts that would otherwise melt (certain connectors and such like spring to mind), wiring looms, etc.  My personal belief is that Gerber is fine for making PCBs but is going to be an issue for pasting and PnP where not all parts should be pasted and as importers for KiCAD, Altium etc are written, the gerber import will more or less not be used as much (YMMV).  Also, I believe that perhaps the “Solder Paste” tab can go and a “Solder Paste” column with a Y/N added to the “Pick and Place” tab would simplify the UI. (IMHO)  Thoughts??

Cheers
Douglas

vonnieda:
@DouglasPearless 

1. Ok
2. This is because you changed the format of Footprint and an older packages.xml will not load. This change would break every packages.xml currently in the wild and is not acceptable. As I said before, don't use Footprint at all. It's not used in the rest of the system, there is no support for it holding solder paste pads anywhere and the class itself will soon go away.
3. Ok
4. We discussed this at length in our email thread - please re-read it. The basic gist is that I agree that we will need to associate pads to placements in the future, but it's too early to do that now and in any case Footprint is not the right place to do it. Instead, Pads will likely be added directly to Placement, but I have not had time yet to design this.

In any case, if you do anything but add BoardPads to Board.solderPastePads your code won't actually do anything unless you also update the UI and the SolderPasteJobProcessor. If you want to take this on I am okay with reviewing it but as I said before I don't have the bandwidth right now to help with this design. Instead I suggest you just follow what the Gerber importer is doing and add the BoardPads to Board.solderPastePads. Yes, this is not an ideal user experience but it does *work* and it's a good start.

I disagree that Gerber will go by the wayside. Anyone who is using OpenPnP to assemble a board that is not of their own design is very unlikely to have access to the original CAD files and will be working only with Centroids and Gerbers. These are the lowest common denominator for PCB assembly.

robermeyer:
Douglas,
Doesn't your code assume the solder pad and paste mask are equal?
At least in Altium the solder paste can be any shape that works well for the package.  On simple small square or round pads it is typically just 3% smaller than the pad so it easy.  on more complex devices it can be way different.
If you don't want to include some parts for paste, don't include them in the gerber generation of the solder paste mask.  This means the gerber will always be usable.  IMHO
Rich

vonnieda:
The UI already has the ability to choose which pads get paste. It's not very friendly, because pads are not tied to placements, but it is usable and it only takes a little bit of touchup to identify which pads are which:

screen shot 2015-10-05 at 1 59 46 pm

vonnieda:
Actually, I just realized that this gives a good solution to the problem. You can set the name of a pad. @DouglasPearless, for the time being, just set the name of each pad to the placement name + an index. So, it would be like R1-1, R1-2, etc. Here is an example of how it would look:

screen shot 2015-10-05 at 2 02 01 pm

I think this is a very reasonable compromise that would not require any modifications to the model at all.

cri-s:
It is possible to get know why you have choose that solution and breaks a
lot of code?
Why you have added dispenser job.
Dispenser is just one actor with actuate(double) interface.
OK it was you design decision.
Do you are aware that the (ugly) Gerber
that you receive are already panelized?
And because that it is not associable to board?
And further board offset and rotation cannot be applied to that data , ie
solder paste is then extra board.
Another thing, solder layer miss the fiducials.
Because of this normally the copper layer is
used for PNP setup.



Il lunedì 5 ottobre 2015, Jason von Nieda <notifications@github.com> ha
scritto:

vonnieda:
@cri-s What broken code are you referring to?

If you want to discuss the solder paste feature in general, please open a new issue or discuss on the mailing list. This thread is for Douglas' Eagle importer.

DouglasPearless:
I have updated as requested, it now passes the maven test suite and footprint was back how it was :-)

vonnieda:
@DouglasPearless Make sure you push the changes to update the pull request. I don't see the new commits yet.

DouglasPearless:
Yep just trying to figure that our now, must be doing something wrong :-)

cri-s:
This patch remove height width from package. These two values are used for
nozzle tip selection. Further it seems package and part is used wrong. Part
is manufacture ,part having it's picture and feeders. Package is associated
with pads and shape and used for vision too.
It is more logical to associate package with eagle libs and not parts.
For eagle, gerbermerge or fab3000 is usually used to panelize it. If one
import the fiducials using the tab eagle brd, it gets name conflicts and
offset problems.
I don't want discuss if importing eagle native format is good or bad,
handling special cases as assembly variants,  ... just for the
convenience to not need launch one single ulp. Further it is easier to
write ulp that exports solder paste shapes to packages
If needed.


Il lunedì 5 ottobre 2015, Jason von Nieda <notifications@github.com> ha
scritto:

src/main/java/org/openpnp/model/Pad.java
```diff
@@ -3,6 +3,7 @@
 import java.awt.Shape;
 import java.awt.geom.Ellipse2D;
 import java.awt.geom.RoundRectangle2D;
+//import java.awt.geom.Path2D; //Used for Polygons as java.awt.Polygon only supports integer vertex
```
vonnieda: Again, please don't comment out imports. Just remove them if you aren't using them.

src/main/java/org/openpnp/model/Part.java
```diff
@@ -58,12 +60,13 @@ public void configurationLoaded(Configuration configuration)
...
                 if (getPackage() == null) {
                     setPackage(configuration.getPackage(packageId));
+//                    setLibrary(configuration.getLibrary(libraryId));
                 }
             }
 		});
 	}
 	
-	@SuppressWarnings("unused")
+//	@SuppressWarnings("unused")
```
vonnieda: Why is this commented out? It is there to avoid warnings about unused functions.

vonnieda:
@DouglasPearless Looking good. I made a few comments, but they are minimal. Fix them if you like, otherwise I will merge this later tonight.

@cri-s This patch doesn't make any changes to Packages at all, and it only *adds* to Part. It changes no existing functionality. I'm not sure what you are referring to.

DouglasPearless:
Hi,

It is late at night here so please accept the pull request & I will fix those with the polygon pad work I have started so they will be correct in my next pull request which will include the polygon functionality.

vonnieda:
@DouglasPearless I think this feature needs a bit more testing before it can be merged. There are at least two major bugs:

1. If you import a board and it's parts already exist, no pads are created. In other words, import a board, have it create parts, then try to import that same board again into a new job or otherwise. No pads will be created. I suspect this is an artifact of previously intending to tie Pads to Parts, but with the way the importer works now it's broken.

2. The Locations of the imported Pads are incorrect, making the resulting Pads unusable. To see this in action, try importing: https://github.com/openpnp/openpnp/raw/develop/samples/pnp-test/pnp-test.brd

Here is the result that I got:

screen shot 2015-10-05 at 10 40 38 pm

It's okay that not all the pads got imported, since this board has many polygons, but the ones that did get imported do not have the correct Location. The Pads for R1, for instance, should be closer to 37, 38 and 39, 38.

DouglasPearless:
Ok I will investigate this and fix it.

DouglasPearless:
bug 1, Investigating.
bug 2, FIXED, it was an omission on my part, I was only recording the offset and not the actual location (package x centre + x offset and package y + y offset).

DouglasPearless:
bug 1 fixed

DouglasPearless:
Code is now clean and should pass all test.

vonnieda:
Nice work @DouglasPearless! I am going to go ahead and merge this, but I do want to report an issue. Some of the Pad locations still seem to be wrong. I think it's probably due to rotation. Import the test board I referenced above and then take a look at the pads for R8.

screen shot 2015-10-06 at 10 15 37 am

DouglasPearless:
Ok I will have a closer look tonight.

vonnieda:
Hi @DouglasPearless, just letting you know I removed libraryId from Part. It was being set but never read, and it was not being stored in the parts.xml (no @Attribute) so it was not actually being used for anything. I realize that this is something you may want to use in the future, and I am okay with it being added back when the functionality is needed, but for now I don't want it there if it's not being used.

DouglasPearless:
Hi,

I was going to remove that in the interim until we all agree on having a proper library construct so all good.

I assume that if / when we add that to Part it will break every ones Part.xml file, so that leads me to ask how should OpenPnP handle versioning of such files, in the past projects I have put a version number in the XML file so programs can automatically upgrade XML files (usually only forwards and not for backwards compatibility) and / or handle different XML file versions as required.

This could be a useful discussion to be had on the forum as with a (hopefully) rapidly evolving system, it should really address this principle from the start.

Cheers
Douglas

vonnieda:
Hi @DouglasPearless,

I handle backwards compatibility a couple different ways, depending on what I am adding. In general, when I add something new I give it the required=false attribute, such as `@Attribute(required=false)`. What this does is allows the XML framework to load the old files and then when they are written back out the new parameters get included. This has worked very well for many, many iterations of the various XML files.

The XML framework also has the ability to version files which you can see an example of at https://github.com/openpnp/openpnp/blob/develop/src/main/java/org/openpnp/model/Placement.java#L52.

I don't use this feature very often because there is a bug in the framework that keeps it from working with subclasses of annotated files, but for simple models like Job, Placement, Part, Package, etc. it's fine.

I highly recommend reading through the tutorial at http://simple.sourceforge.net/download/stream/doc/tutorial/tutorial.php. This is the framework we use for XML serialization and the tutorial is short and very poignant. It will help you understand how the entire configuration system works.

So, in short, if you are adding a new parameter just mark it required=false and everything should work fine. If you want to ensure that it's valid, just check in `@Commit`.

PR #161, opened 2015-10-07
Fixed minor bug where the solder pads were only rotated relative to the package definition and did not take into account the package rotation itself

DouglasPearless:
Fixed minor bug where the solder pads were only rotated relative to the package definition and did not take into account the package rotation itself

### DIFF
--- a/src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java
+++ b/src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java
@@ -150,7 +150,7 @@ public class EagleBoardImporter implements BoardImporter {
 						mmMaxCreamFrame_number = Double.parseDouble(mmMaxCreamFrame_string) * mil_to_mm;
 					} else if (params.getValue().toUpperCase().endsWith("MM")) {
 						mmMaxCreamFrame_number = Double.parseDouble(mmMaxCreamFrame_string);
-					} // else throw an exception as can only be mm or mil BUT because we have already initialised these as 0, we don't care
+					} // TODO else throw an exception as can only be mm or mil BUT because we have already initialised these as 0, we don't care
 				}				
 			}
 			// Now we know the min and max tolerance for the cream (aka solder paste)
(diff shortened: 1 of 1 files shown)


vonnieda:
Still not working:

screen shot 2015-10-07 at 2 56 49 pm

To perform the same test I am doing:

1. Start OpenPnP with a fresh config: https://github.com/openpnp/openpnp/wiki/FAQ#how-do-i-reset-my-configuration
2. Create a new board.
3. Use the EagleImporter to import pnp-test.brd included in the OpenPnP samples directory.
4. Select Pad R8-1 from the Solder Paste pads list.
5. Click the square red position camera button above the list. 
6. The camera should move to be centered over the pad. Currently it does not.

To me it looks like the coordinate needs to be rotated before you store it. There are utilities in Utils2D.java to help with 2D rotation. Ideally, if you open the board in Eagle and place the mouse over the pad the coordinate shown in Eagle should be pretty close to what you provide to OpenPnP.

DouglasPearless:
OK, I am working on that, how did you turn on the blue/red cross-hairs?

vonnieda:
@DouglasPearless Right click in the camera view. There are a number of options you can use for crosshairs and targeting reticles.

DouglasPearless:
Ah hah!!!

now to check out what you mean...

DouglasPearless:
Now I see what you mean, the paste dispense is correct for 0 rotation, for 90 degree it is way off, other angles are off by varying amounts, time to think some more.

vonnieda:
Yep. I suspect that you need to perform a 2D rotation of the coordinates. You can think of the placement position as center for the pads. Then there are pads that are offset from center in X and Y and rotation. 

So, if there is a placement at 10, 10 and the two pads are at -1, 0 and +1, 0 from the placement, and the placement is rotated 90 degrees, the pads are actually at 0, -1 and 0, +1.

DouglasPearless:
I think you are correct, I will work on that now

cri-s:
check out the Location and Utils2D from github.com/cri-s/openpnp

But take care, i have removed the convertToUnits() calls, i always
disliked that.
I don't need it, because placements/boards/... location are stored
using mm only with the
exception of packages.

translate and translateLoc and eventually
calculateBoardPlacementLocation/Inverse
could be interesting for you. boardSide is the width/height/.... value
of the board needed
for assembling bottom side. It is double value.




2015-10-07 8:31 GMT, Douglas Pearless <notifications@github.com>:

DouglasPearless:
I have updated the pad rotation to correctly transform in a 2D space; it now passes my tests :-)

vonnieda:
Nice work Douglas! Passes my tests too :)

I'll merge this now, but I do have an unrelated bug to report: The importer doesn't seem to create the parts and packages even if the checkboxes are checked.

DouglasPearless:
OK, I will look at that,it might have been introduced when I reworked the logic to remove the Footprint logic, I will look into it.

DouglasPearless:
I m struggling to reproduce this issue, can you please show me the steps you performed to find this bug?

vonnieda:
Here you go: https://www.youtube.com/watch?v=4N-JfDJoJQ8

Gist is that after importing a board that has parts that are not already in
my Parts and Packages they are not created.

Jason

PR #162, opened 2015-10-09
Fixed bug in JobPanel and EagleBoardImporter

DouglasPearless:
Fixed bug in JobPanel where it would crash with NULL error if you "cancel" out of the board importer dialog.  Note this was not trapped as an exception by Java so I have to explicitly test for "null"

Hopefully fixed a bug where after importing a board that has parts that are not already in my Parts and Packages they are not created.

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -654,20 +654,22 @@ public class JobPanel extends JPanel {
         try {
             Board importedBoard = boardImporter
                     .importBoard((Frame) getTopLevelAncestor());
-            Board existingBoard = getSelectedBoardLocation().getBoard();
-            for (Placement placement : importedBoard.getPlacements()) {
-                existingBoard.addPlacement(placement);
+            if (importedBoard != null) {
+                Board existingBoard = getSelectedBoardLocation().getBoard();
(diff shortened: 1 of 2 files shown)


vonnieda:
@DouglasPearless Can you take care of the merge conflict and resubmit, please? You just need to pull openpnp/develop into your branch, take care of the conflict and commit the result.

DouglasPearless:
fixed the conflict

vonnieda:
Nice work @DouglasPearless, works great!

cri-s:
Just question about eagle importer.
How is calculated the centetoid of part and
What Standard do you use for part rotation
And do you use user supplied values or do you recalculate it?
How do you handle assembly variants,
TH and the parts assembled by smt pnp.
Do you consider that this importer should be
included to the program and do you
Feel that its more easy going that way instead of associating user supplied
packages names to system packages.
Similar to kicad association of schematic parts to PCB packages.


Il venerdì 9 ottobre 2015, Jason von Nieda <notifications@github.com
<javascript:_e(%7B%7D,'cvml','notifications@github.com');>> ha scritto:

DouglasPearless:
To try and summarize the answers

The centroid is supplied by eagle in the bed file.

I rotate the part from the rotation value for the part from the brd file.

The vales all come from the brd file.

User supplied variants appear in the brd file as part of the package definition for the part.

PR #187, opened 2016-01-10
LinuxCNC driver update

readerror67:
1) Updated to use protocol version 1.1
2) mdi commands issued will not respond until the movement is complete. Tested in LinuxCNC SIM only but seems to work fine.
3) Enabled verbose, ACK and NAK messages in the reply can be used to determine if the last command completed with or without errors.


vonnieda:
Thanks @readerror67!

Issue #189, opened 2016-01-11
Discussion for Driver refactoring, moveTo blocking, GCode, etc.


vonnieda:
Comments originally made on #183, moved to this issue for further discussion.

 cri-s commented 2 hours ago
Jason, i have extended the driver api a bit,
void dwell()  , wait until finished
void dwell(double)  wait time in seconds.
boolean GCode(string)
boolean GCode(string,time)

Actually the only not right is on firepick driver, where i probably
need community help if this
get accepted by you. Then i suggest to use GCode for driving it. I
suggest then to use GCode
for driving changer and i think, manual should be highest priority.
Currently i have tested only dwell(double) in order to go away with
thread sleeping and that
works.

2016-01-11 17:12 GMT+01:00, Jason von Nieda <notifications@github.com>:
…
 @vonnieda
  Owner
vonnieda commented an hour ago
void dwell() , wait until finished
void dwell(double) wait time in seconds.

What is the purpose of the dwell commands and how will they be used? I don't see this as being included because all of the driver methods are supposed to block until they are finished. If something is returning before it's complete then that is a bug.

boolean GCode(string)
boolean GCode(string,time)

These we will not include. The Driver interface is designed to be controller agnostic and this would make it specific to GCode. If you want GCode specific methods just extend the ReferenceDriver interface and implement them in your branch.
 @cri-s
  
 cri-s commented 44 minutes ago
dwell wait until all gcode are processed. No only some gcode are
blocking, the other return without being completed.
dwell(double seconds) wait seconds until and return after all previous
motion code are finished.
I have used it to replace    Thread.sleep(pickDwellMilliseconds);  as
otherwise it don't worked
because gcode and pc are never in sync.
GCode() would be nice for nozzle changer code and others, that are
always machine specific
and mostly have only moves inside it. You'r decision.

2016-01-11 18:04 GMT+01:00, Jason von Nieda <notifications@github.com>:
…
 @vonnieda
  Owner
vonnieda commented 32 minutes ago
@cri-s All driver calls should block until they are finished. This contract must be enforced or lots of other things will break. For instance, if moveTo() or place() does not block then the machine may try to move before it's finished placing. If you have driver calls that don't block, that is a bug and should be fixed. Just call dwell() internally in the driver, as many of the drivers already do.

We can't add GCode specific things to the driver interface because there are machines that don't use GCode. For instance, the TinyG and FirePick drivers use JSON for communication. Having a GCode command on it would be meaningless.

[...]

PR #192, opened 2016-01-15
added dwell on drivers for timing on gcode executor (microcontroller) instead of PC, 

cri-s:
I have not touched the driver behavior , it's still uses the "new" driver
model where the move must complete before call returns from moveTo .
I have only added code that if removing that, other timing code works.

Introduced driver dwell for not "old" driver model where commands
are asyncron and . 
The advantage is, that timing is not depend on PC, but are timed by microcontroller and this allow faster execution and less poblems with camera timing.
Further GCode interface and Msg interface was added. The intended purpose
beside switching coordinates on some JobPlanner is for test and setup purposes.
As example, automatic test sequences that move machine and test accuracy
increasing feed rate or acceleration automatically.
For drivers, as example the Firepick driver, the support is missing,
there have the own tests. For other drivers, it should work.
In the case you discover driver wait seconds instead of milliseconds,
the code inside dwell need to be changed. I have checked several driver docs,
but not double checked.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 doc
 bin
 target
+code
 
 installers
 .settings
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@ doc
 bin
 target
 code
+old
+dead
 
 installers
 .settings
@@ -9,4 +11,5 @@ installers
 .idea
 *.iml
 .classpath
+err
 src/main/resources/icons/src/Soft_Scraps_by_deleket
--- a/src/main/java/org/firepick/driver/FireStepDriver.java
+++ b/src/main/java/org/firepick/driver/FireStepDriver.java
@@ -33,6 +33,7 @@ import java.util.List;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.TimeoutException;
+import java.util.List;
 
 import javax.swing.Action;
 
@@ -81,6 +82,7 @@ public class FireStepDriver extends AbstractSerialPortDriver implements Runnable
 	private String connectedVersion;
 	private Queue<String> responseQueue = new ConcurrentLinkedQueue<String>();
 	
+	private boolean enabled;
 	@Override
 	public void setEnabled(boolean enabled) throws Exception {
 	    if (enabled) {
@@ -119,6 +121,7 @@ public class FireStepDriver extends AbstractSerialPortDriver implements Runnable
 	    		}
 	    	}
 	    }
+		this.enabled=enabled;
 	}
 	
 	@Override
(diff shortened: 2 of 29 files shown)


vonnieda:
There are a number of problems with this pull request:

1. I will not accept the GCode() API. I explained to you that if you want an API like this we can add something generic such as sendCustomCommand(Object) but I'm not adding a GCode specific command.
2. The Msg() API is not documented and does not follow the code style.
3. There are changes to many other unrelated files. Location and Part are examples.
4. I have explained repeatedly that the driver model as it is designed requires blocking commands. I understand that you want to change this but you have not provided good justification for doing so and you have not presented readable, documented code that shows why the change is useful.

As I said in our original discussion, I do not feel this is a useful change. You have not shown me what it is used for. You said "As example, automatic test sequences that move machine and test accuracy
increasing feed rate or acceleration automatically." but that is something that is easily done in your own driver class. It does not require an API change that changes how the entire driver model works.

vonnieda:
This pull request is in relation to the discussion on #189

cri-s:
As GCode is Industry standard, and basically all drivers that drive
carteisan machines
    can process it, some more , some less, but there is definitive
commune standard with
    exception of the dwell gcode parameter (seconds or milliseconds).
    If you prefer name it sendCustomCommand(String Object) for me it's same.

Msg() API , is not documented, right. On the other side, it should be
undocumented, as
    it returns the output of gcode controller, and it's output is
valid only if no other commands
    is given. Probably the only real usage could be a terminal window
to the controller for
    setting it up or diagnosting problems, or for loading and saving
configuration of the controller. Further it could be useful to display
the result lines of the gcode controller on the
   status bar. If Softlimit or Hardlimit happen, or if the controller
have some Alarm actually
   OpenPnP ignores that and it would by handy if the error or ok
status is displayed on status bar, at least for moves limited to n
Chars.

3) This is right, Part and location is from JobPlanner experiments,
location to for  get() i think. I remove it.

4) To clearify, Gcode could run without blocking on every command. I
don't want change
    the driver to non blocking for now , that can be done on custom
version, but if one want to
    drive it without blocking on every comand,
    then timing must be done on gcode microcontroller and not on PC.
    As the software grows, more timing are introduced, and i just want
ensure, that if one want
    run the driver where it's not blocking,  then
    it should work, and not need changing all drivers and changing all
code that is timing
    sensible, beacaus it's not small changes. This adds to the changes
need for camera,
    illumination, and other things.
    The reason i want run it without blocking on every command is
principally faster assembly. The most impact is when using uplooking
camera without closing loop.
    The head moves over camera, wait the time need for aquiring image,
about 35ms,
    then moves to intermediary destination, and during that the pc
calc the offsets and emit
    the final coordinates. If doing timing on PC, then ok, image calc
needs 350ms, aquisistion
    needs 350ms, just the image processing adds 0.8 second delay, not
calculated the delay
    for usb rs232 transmitter, delay on the gcode interpreter , pc
rs232 buffer, pc task switching.
    The problem is, if you develop control code without using dwell()
and using Thread.sleep()
    then one have to change all that code for changing to non blocking driver.
    On non blocking driver, pick and place are blocking, the rest
returns immediatly.

I have considered not submitting GCode and Msg , but the missing
status line, and
the usefullness of separate setup program have convinced me to tray to
submit it.

As alternative could you consider using the code inside MovingUtils
for dwell instead of
ThreadSleep inside timing code. On custom versions, one could do
  Configuration.get().getMachine.get()Driver.dwell(); if that is not a
problem with static  .



2016-01-15 16:53 GMT, Jason von Nieda <notifications@github.com>:

la3pna:
The problem as I see it is that you want to move from the "block untill complete" to a "checking for completion" sounds like its a major task, requiring a rewrite of quite a major part of the code. This has the posibility to break quite a lot of function and would require a large effort, for little gain. 

The code GCode("G4 P"+double); will return different number depending on locale, here it would be "G4 P1,0" instead of the intended "G4 P1.0". There is a lack of rigidity in the code that is going to be neccessary in order for this to be useable for more than a single person.

cri-s:
good catch. For such applications i prefer settings locale on main and
not use local locale.
Principally i need that timing is made by gcode microcontroller and not by pc.
That don't break a lot of code. Because of this, i have feelt it's
simpler to have
camera.dwell() too as everywhere call driver or annother internal
function, but as said , that
can be one alternative.



2016-01-15 19:41 GMT, Thomas S. Knutsen <notifications@github.com>:

vonnieda:
Hi @cri-s, I've given this some thought. I think the advanced functionality we can get by adding asynchronous moves is valuable, but I don't like the API being so Gcode specific. In addition, I want to have a better way to delineate when a move is complete. Let me think on this a few more days and I will let you know.

PR #195, opened 2016-01-22
fix bug for tray feeder

cri-s:
ReferenceTrayFeeder actually is updating wrong feedPosition or if don't seem to update it at all.
This is a working workaround about the issue.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceTrayFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceTrayFeeder.java
@@ -58,6 +58,7 @@ public class ReferenceTrayFeeder extends ReferenceFeeder {
 	
 	private Location pickLocation;
 
+
 	@Override
     public Location getPickLocation() throws Exception {
 	    if (pickLocation == null) {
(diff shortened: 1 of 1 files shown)


vonnieda:
@cri-s If you aren't going to read the [CONTRIBUTING.md](https://github.com/openpnp/openpnp/blob/develop/CONTRIBUTING.md) please don't submit pull requests. It's a waste of my time to have to keep reviewing these when you:
* Don't follow the code style.
* Don't follow *any* standard Java naming conventions.
* Have not described the issue or how you are fixing it.

src/main/java/org/openpnp/machine/reference/feeder/ReferenceTrayFeeder.java
```diff
@@ -58,6 +58,7 @@
 	
 	private Location pickLocation;
 
+
```
aep: unnessesary whitespace change

src/main/java/org/openpnp/machine/reference/feeder/ReferenceTrayFeeder.java
```diff
@@ -71,27 +72,32 @@ public void feed(Nozzle nozzle)
 			throws Exception {
 		logger.debug("{}.feed({})", getName(), nozzle);
 		int partX, partY;
+		double JavaBugWorkaroundX,JavaBugWorkaroundY,JavaBugWorkaroundC;
```
aep: JavaBugWorkaroundN doesn't tell us what this was supposed to do

src/main/java/org/openpnp/machine/reference/feeder/ReferenceTrayFeeder.java
```diff
@@ -71,27 +72,32 @@ public void feed(Nozzle nozzle)
...
 		    throw new Exception(String.format("Tray empty on feeder %s.", getName()));
 		}
         
+	JavaBugWorkaroundC=feedCount; JavaBugWorkaroundY=trayCountY; JavaBugWorkaroundX=trayCountX;
         if (trayCountX >= trayCountY) {
             // X major axis.
-            partX = feedCount / trayCountY;
-            partY = feedCount % trayCountY;
+            JavaBugWorkaroundX = JavaBugWorkaroundC / JavaBugWorkaroundY;
+            JavaBugWorkaroundY = JavaBugWorkaroundC % JavaBugWorkaroundY;
```
aep: that sounds like you actually just wanted to cast trayCountY to double here?

src/main/java/org/openpnp/machine/reference/feeder/ReferenceTrayFeeder.java
```diff
@@ -71,27 +72,32 @@ public void feed(Nozzle nozzle)
...
         }
         else {
             // Y major axis.
-            partX = feedCount % trayCountX;
-            partY = feedCount / trayCountX;
+            JavaBugWorkaroundY = JavaBugWorkaroundC / JavaBugWorkaroundX;
+            JavaBugWorkaroundX = JavaBugWorkaroundC % JavaBugWorkaroundX;
         }
+	partX=(int)JavaBugWorkaroundX;
```
aep: That's essentially the same as the original code. No need for the intermediate variable.

src/main/java/org/openpnp/machine/reference/feeder/ReferenceTrayFeeder.java
```diff
@@ -71,27 +72,32 @@ public void feed(Nozzle nozzle)
...
+	partY=(int)JavaBugWorkaroundY;
         
         // Multiply the offsets by the X/Y part indexes to get the total offsets
         // and then add the pickLocation to offset the final value.
         // and then add them to the location to get the final pickLocation.
+        
         pickLocation = location.add(
-                offsets.multiply(partX, partY, 0.0, 0.0));
+                offsets.multiply(partX, partY, 1.0, 1.0));
```
aep: That's a behaviour change here. Could you explain what its for, so we can check it doesnt break anything?

cri-s: I have traced it down, problem was upgrade to java8 have not
overridded all env varaibles
and so java7 core and java8 was mixed and resulted in integer modulus bug.
moving it on floating point domain resolved this issue apparently.

For the behaviour change.
It's for part stacks, and no, it don't break anything, as offsets is
always initialized with 0
and the wizard don't allow to change C and Z.
It is possible to change it on machine.xml without gui.
If you stacked parts, like metal dome caps for flexible keyboard, or
sim/sd cards, ...
with this modification is it possible to have one pick position, that
load parts having different
Z with each feed and thus having one part on top of the other.
Sometime parts need stacked with 180 degree offset because of pin
heigh differences of one side, that otherwise accumulated too mutch.
It makes only sense with one single pick position. If you have as
example two stacks, it's
not possible to configure this as single feeder, two separate feeder
must be configured for the same part.

2016-01-23 0:41 GMT+01:00, Arvid E. Picciani <notifications@github.com>:

PR #198, opened 2016-01-24
Fix coord

cri-s:
This fix coordinate bug for rotated boards and adds Board Dimension needed for bottom assembly and fiducials.
Currently only width is used, height is then needed for fiducials, but as there are not well defined,
currently not used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 doc
 bin
 target
+code
 
 installers
 .settings
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 doc
 bin
 target
-code
 
 installers
 .settings
--- a/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
@@ -375,12 +375,11 @@ public class JobPlacementsPanel extends JPanel {
         	UiUtils.messageBoxOnException(() -> {
             	HeadMountable tool = MainFrame.machineControlsPanel.getSelectedTool();
             	Camera camera = tool.getHead().getDefaultCamera();
-                Location placementLocation = Utils2D.calculateBoardPlacementLocation(
+                Location placementLocation = Utils2D.calculateBoardPlacementLocationInverse(
                 		boardLocation.getLocation(), 
                 		boardLocation.getSideWidth(),
-                        camera.getLocation().invert(true, true, true, true));
-                getSelection().setLocation(
-                        placementLocation.invert(true, true, true, true));
+                        camera.getLocation());
+                getSelection().setLocation(placementLocation);
                 table.repaint();
         	});
         }
@@ -398,11 +397,10 @@ public class JobPlacementsPanel extends JPanel {
         public void actionPerformed(ActionEvent arg0) {
             Nozzle nozzle = MainFrame.machineControlsPanel.getSelectedNozzle();
             Location placementLocation = Utils2D
-                    .calculateBoardPlacementLocation(boardLocation
+                    .calculateBoardPlacementLocationInverse(boardLocation
                             .getLocation(), boardLocation.getSideWidth(),
-                            nozzle.getLocation().invert(true, true, true, true));
-            getSelection().setLocation(
-                    placementLocation.invert(true, true, true, true));
+                            nozzle.getLocation());
+                getSelection().setLocation(placementLocation);
             table.repaint();
         }
     };
--- a/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
@@ -325,6 +325,7 @@ public class JobPlacementsPanel extends JPanel {
         	UiUtils.submitUiMachineTask(() -> {
                 Location location = Utils2D.calculateBoardPlacementLocation(
                 		boardLocation.getLocation(), 
+                		boardLocation.getSide(), 
                 		boardLocation.getSideWidth(), 
                 		getSelection().getLocation());
 
@@ -350,7 +351,7 @@ public class JobPlacementsPanel extends JPanel {
         public void actionPerformed(ActionEvent arg0) {
             Location placementLocation = Utils2D
                     .calculateBoardPlacementLocation(boardLocation
-                            .getLocation(), boardLocation
+                            .getLocation(), boardLocation.getSide(), boardLocation
                             .getSideWidth(), getSelection().getLocation());
 
             Nozzle nozzle = MainFrame.machineControlsPanel
@@ -377,6 +378,7 @@ public class JobPlacementsPanel extends JPanel {
             	Camera camera = tool.getHead().getDefaultCamera();
                 Location placementLocation = Utils2D.calculateBoardPlacementLocationInverse(
                 		boardLocation.getLocation(), 
+                		boardLocation.getSide(),
                 		boardLocation.getSideWidth(),
                         camera.getLocation());
                 getSelection().setLocation(placementLocation);
@@ -398,7 +400,9 @@ public class JobPlacementsPanel extends JPanel {
             Nozzle nozzle = MainFrame.machineControlsPanel.getSelectedNozzle();
             Location placementLocation = Utils2D
                     .calculateBoardPlacementLocationInverse(boardLocation
-                            .getLocation(), boardLocation.getSideWidth(),
+                            .getLocation(), 
+			boardLocation.getSide(),
+			boardLocation.getSideWidth(),
                             nozzle.getLocation());
                 getSelection().setLocation(placementLocation);
             table.repaint();
--- a/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
@@ -325,7 +325,7 @@ public class JobPlacementsPanel extends JPanel {
         	UiUtils.submitUiMachineTask(() -> {
                 Location location = Utils2D.calculateBoardPlacementLocation(
                 		boardLocation.getLocation(), 
-                		boardLocation.getSide(), 
+                		boardLocation.getSideWidth(), 
                 		getSelection().getLocation());
 
                 Camera camera = MainFrame
@@ -351,7 +351,7 @@ public class JobPlacementsPanel extends JPanel {
             Location placementLocation = Utils2D
                     .calculateBoardPlacementLocation(boardLocation
                             .getLocation(), boardLocation
-                            .getSide(), getSelection().getLocation());
+                            .getSideWidth(), getSelection().getLocation());
 
             Nozzle nozzle = MainFrame.machineControlsPanel
                     .getSelectedNozzle();
@@ -377,7 +377,7 @@ public class JobPlacementsPanel extends JPanel {
             	Camera camera = tool.getHead().getDefaultCamera();
                 Location placementLocation = Utils2D.calculateBoardPlacementLocation(
                 		boardLocation.getLocation(), 
-                		getSelection().getSide(),
+                		boardLocation.getSideWidth(),
                         camera.getLocation().invert(true, true, true, true));
                 getSelection().setLocation(
                         placementLocation.invert(true, true, true, true));
@@ -399,7 +399,7 @@ public class JobPlacementsPanel extends JPanel {
             Nozzle nozzle = MainFrame.machineControlsPanel.getSelectedNozzle();
             Location placementLocation = Utils2D
                     .calculateBoardPlacementLocation(boardLocation
-                            .getLocation(), getSelection().getSide(),
+                            .getLocation(), boardLocation.getSideWidth(),
                             nozzle.getLocation().invert(true, true, true, true));
             getSelection().setLocation(
                     placementLocation.invert(true, true, true, true));
--- a/src/main/java/org/openpnp/model/BoardLocation.java
+++ b/src/main/java/org/openpnp/model/BoardLocation.java
@@ -31,8 +31,6 @@ import org.simpleframework.xml.core.Commit;
 public class BoardLocation extends AbstractModelObject {
 	@Element
 	private Location location;
-	@Element(required=false) // for compatility
-	private Location dimension=new Location(LengthUnit.Millimeters);
 	@Attribute
 	private Side side = Side.Top;
 	private Board board;
@@ -73,13 +71,11 @@ public class BoardLocation extends AbstractModelObject {
 	}
 
 	public Location getDimension() {
-		return dimension;
+		return board.getDimension();
 	}
 
 	public void setDimension(Location location) {
-		Location oldValue = this.dimension;
-		this.dimension = location;
-		firePropertyChange("dimension", oldValue, location);
+		board.setDimension(location);
 	}
 
 	public Side getSide() {
@@ -87,7 +83,7 @@ public class BoardLocation extends AbstractModelObject {
 	}
 
 	public double getSideWidth() {
-		return dimension.getX();
+		return board.getDimension().getX();
 	}
 
 	public void setSide(Side side) {
--- a/src/main/java/org/openpnp/model/BoardLocation.java
+++ b/src/main/java/org/openpnp/model/BoardLocation.java
@@ -22,6 +22,8 @@
 package org.openpnp.model;
 
 import org.openpnp.model.Board.Side;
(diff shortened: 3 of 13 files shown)


.gitignore
```diff
@@ -1,6 +1,7 @@
 doc
 bin
 target
+code
```
aep: please commit unrelated additions to your .gitignore.

src/main/java/org/openpnp/gui/JobPlacementsPanel.java
```diff
@@ -375,12 +375,11 @@ public void actionPerformed(ActionEvent arg0) {
...
             	HeadMountable tool = MainFrame.machineControlsPanel.getSelectedTool();
             	Camera camera = tool.getHead().getDefaultCamera();
-                Location placementLocation = Utils2D.calculateBoardPlacementLocation(
+                Location placementLocation = Utils2D.calculateBoardPlacementLocationInverse(
                 		boardLocation.getLocation(), 
-                		getSelection().getSide(),
-                        camera.getLocation().invert(true, true, true, true));
-                getSelection().setLocation(
-                        placementLocation.invert(true, true, true, true));
+                		boardLocation.getSideWidth(),
```
aep: stray tab

src/main/java/org/openpnp/util/Utils2D.java
```diff
@@ -77,43 +77,33 @@ public static Point scalePoint(Point point, double scaleX, double scaleY) {
...
+	public static Location calculateBoardPlacementLocation(Location boardLocation, double bottom, Location placementLocation) {
         // We will work in the units of the placementLocation, so convert
         // anything that isn't in those units to it.
         boardLocation = boardLocation.convertToUnits(placementLocation.getUnits());
         
         // If we are placing the bottom of the board we need to invert
         // the placement location.
-        if (boardSide == Side.Bottom) {
-            placementLocation = placementLocation.invert(true, false, false, false);
+        if (bottom != 0.0) {
```
aep: this seems a bit strange. either .add with bottom == 0.0 does nothing in the first place, or i don't understand what this is supposed to do.

aep:
thanks for the change! Looks like a good idea. We'll probably just need some unit tests to make sure nothing breaks.

vonnieda:
@cri-s Can you explain what you mean by "Currently only width is used, height is then needed for fiducials, but as there are not well defined, currently not used."? Does this change work with the current fiducial capability?

src/main/java/org/openpnp/util/Utils2D.java
cri-s: It's the result of needing board width, and in spirit of the old API i have changed the enum to double.
It have two uses, it hold the board width (X) and at the same time is indicate if side is top or bottom.
If on layout software you place component at 10mm X and 20mm Y, and board width is 100mm,
The true coordinate for placing bottom side is X90 Y20 if the board don't have rotation.
Without having access to board size, it's not possible to place bottom from board-0 origin.
It could be more elegant instead of giving location of Board, Boardlocation is giving.
Having double is maybe more flexible, because the same need to be done on Panels,
and if converting position from camera or teaching, there is no BoardLocation available

cri-s:
This fixes several bugs of current fiducial capabity.
Having more boards, and using advanced job planners for assembling
higher components
the controller must know to what extension fiducial correction need to
be extend.
For this height and width is used. Not always it is possible to set
fiducials outside the placement area and for this the fiducial
correction need to be extended outside that area
in order to cover the whole board. Now if assembling only one board,
there is no problem.
Having one panel with more boards or several boards using fixtures,
then fiducial code need to
differentiate between the single area and different layers of fiducial
one on top of the other.

Said that, currently if there is no bottom side to assemble, no
widht/height info need to be filled in. If bottom side is present, the
width is requred. Exact for the case of no fiducials,
approximative if using fiducials.  In future this could eventually
change as eventually
fiducial code need both values.

2016-01-24 19:57 GMT, Jason von Nieda <notifications@github.com>:

src/main/java/org/openpnp/util/Utils2D.java
```diff
@@ -77,43 +77,33 @@ public static Point scalePoint(Point point, double scaleX, double scaleY) {
...
-                (placementLocation.getRotation() + boardLocation.getRotation()) % 360.0);
-        
-        // Set the final location's Z to that of the board location.
-        placementLocation = placementLocation.derive(null, null, boardLocation.getZ(), null);
-        
+	public static Location calculateBoardPlacementLocationInverse(Location boardLocation, double bottom, Location placementLocation) {
+	// inverse steps of calculateBoardPlacementLocation
+        boardLocation = boardLocation.convertToUnits(placementLocation.getUnits());
+        placementLocation = placementLocation.subtractWithRotation(boardLocation).rotateXy(-boardLocation.getRotation()).add(boardLocation);
+        if (bottom != 0.0)
```
vonnieda: if blocks must be surrounded with braces, such as:

if (condition) {
  doSomething();
}

src/main/java/org/openpnp/model/BoardLocation.java
```diff
@@ -68,10 +72,26 @@ public void setLocation(Location location) {
...
+		this.dimension = location;
+		firePropertyChange("dimension", oldValue, location);
+	}
+
 	public Side getSide() {
 		return side;
 	}
 
+	public double getSideWidth() {
+		if(side==Side.Bottom)
```
vonnieda: If blocks must be surrounded with braces.

src/main/java/org/openpnp/util/Utils2D.java
```diff
@@ -77,43 +77,33 @@ public static Point scalePoint(Point point, double scaleX, double scaleY) {
 		return new Point(point.getX() * scaleX, point.getY() * scaleY);
 	}
 	
-	public static Location calculateBoardPlacementLocation(Location boardLocation, Side boardSide, Location placementLocation) {
+
+	public static Location calculateBoardPlacementLocation(Location boardLocation, double bottom, Location placementLocation) {
```
vonnieda: I don't want Side removed. Just pass in the dimensions Location in addition to Side or add a new function that just takes the entire BoardLocation.

vonnieda:
Looks good overall. Please address the line comments and other comments and I think we can merge this. It's going to need at least a basic unit test to show it all works. Ideally the test would show the current system working and then change to show the new system working.

src/main/java/org/openpnp/model/BoardLocation.java
```diff
@@ -29,6 +31,8 @@
 public class BoardLocation extends AbstractModelObject {
 	@Element
 	private Location location;
+	@Element(required=false) // for compatility
```
vonnieda: Is there a reason you put this on BoardLocation instead of Board? The dimensions of a Board are fixed to the board. Do you intend to use it for something else?

vonnieda:
Just did a quick test with this and it doesn't seem to be calculating coordinates properly. See: https://www.youtube.com/watch?v=xw1knXE5lKg&feature=em-upload_owner

src/main/java/org/openpnp/model/BoardLocation.java
cri-s: > Is there a reason you put this on BoardLocation instead of Board? The
> dimensions of a Board are fixed to the board. Do you intend to use it for
> something else?
No, i have wanted use C/Z as temp variables during assembly, that is
why i have used
Location instead of Point, but it don't matter here.

my reason:
Job -> BoardLocation -> Board -> Placements
Board = Hold placement inside boad without knowing the position of board itself.
BoardLocation = Location and dimension for Boards.

Further value is inserted into BoardLocation Table.
If having Panels, that refer to the same Board (you don't have it),
Table and Data are inconsistent as it changes more row as the code
tell the UI to update.

vonnieda: Dimensions are an attribute of Board, not BoardLocation. They should be on
the Board. The user should not have to reenter this data each time they add
a Board to a Job.

cri-s: Ok, i change that if you want it,
Problems appear then if you implement step and repeat,
more board on same panel with same data . I have advised you.




2016-01-25 16:09 GMT, Jason von Nieda <notifications@github.com>:
>> @@ -29,6 +31,8 @@
>>  public class BoardLocation extends AbstractModelObject {
>>  	@Element
>>  	private Location location;
>> +	@Element(required=false) // for compatility
>
> Dimensions are an attribute of Board, not BoardLocation. They should be on
> the Board. The user should not have to reenter this data each time they add
> a Board to a Job.
>

vonnieda:
* You've added more commits but they also have coding style problems. See my previous comments about if blocks needing braces.
* You don't say whether this fixes the issue shown in the video. I'm not going to test this PR again until the issues are resolved.

I appreciate that you want to contribute to the project, but you are just creating heaps of work for me. Instead of spending time working on OpenPnP I am spending time testing and fixing your untested code and writing responses. Please either submit tested, working code or don't submit it.

cri-s:
I have corrected the if coding style after you have mentioned it.
If you eventually found uncorrected one, then please change it, if you
want commit this.
I have added test cases as you requested, but it's you'r decision to
commit it or not.
Yes, the issue reported on video is resolved.

2016-01-25 16:14 GMT, Jason von Nieda <notifications@github.com>:

src/main/java/org/openpnp/model/BoardLocation.java
```diff
@@ -68,10 +70,22 @@ public void setLocation(Location location) {
 		firePropertyChange("location", oldValue, location);
 	}
 
+	public Location getDimension() {
```
vonnieda: Dimensions should be a property of Board, not BoardLocation. Board contains properties of a single PCB and since the dimensions of the PCB are fixed there is no reason to have the user have to enter this for each Job.

src/main/java/org/openpnp/model/Location.java
```diff
@@ -256,6 +256,9 @@ public Location derive(Double x, Double y, Double z, Double rotation) {
 	 * @return
 	 */
 	public Location rotateXy(double angle) {
+	    if(angle==0.0) return this;
```
vonnieda: If block actions should always be surrounded by braces.

vonnieda:
Hi @cri-s, thanks for making the suggested changes and improvements. The PR is looking very nice now. There still seems to be a problem though. I tested it locally and I am getting what looks like angle related failures. I updated my test files to include width and ran this test. See video: https://www.youtube.com/watch?v=73EBJg_5qvk

It seems the board that has nearly 0 angle works fine, but the board with an angle of 74.646 goes off in the wrong direction.

This test board is checked in to develop now, so you can try this test yourself. You'll need to set the width of the board yourself since that is not checked in yet.

vonnieda:
@cri-s I have documented how to run the above test at: https://github.com/openpnp/openpnp/wiki/Developers-Guide#manual-testing-strongly-recommended

cri-s:
I have strange results for  Utils2D.calculateBoardPlacementLocation(),
difference of 0.2mm . Maybe that is correct, or some "rounding
errors", it looks just
strange to mee at first sight.

results (9.797509, 10.267162, 0.000000, -3.000000 mm)
results (9.588165, 10.272644, 0.000000, -3.000000 mm)

The fiducial code is supposed to not works as visual/FiducualLocator
is not updated.
It "seems", i'm not sure, that bottom and top fiducials are mixed up
and used at same time.
Normally one tray to design the board in order that the fiducials are
different for top and
bottom in order that if board side is wrong, the machine don't
populate wrong side.
If fiducials have the same locations, then the machine do fiducial
check and think, it's ok.
If fiducials are different, fiducial check gives error if board is
placed on wrong side.

code below:
    board  = new Location(LengthUnit.Millimeters, 5, 4, 0, -92);
        // determinate ideal locations
        idealA = Utils2D.calculateBoardPlacementLocation(board,
Side.Top, 100., idealA);
        idealB = Utils2D.calculateBoardPlacementLocation(board,
Side.Top, 100., idealB);

        // determinate actual locations
        actualA =
Utils2D.calculateBoardPlacementLocationInverse(board, Side.Top, 100.,
idealA);
        actualB =
Utils2D.calculateBoardPlacementLocationInverse(board, Side.Top, 100.,
idealB);
        actualA =
Utils2D.calculateBoardPlacementLocation(board.addWithRotation(offset),
Side.Top, 100., actualA);
        actualB =
Utils2D.calculateBoardPlacementLocation(board.addWithRotation(offset),
Side.Top, 100., actualB);
        results = Utils2D.calculateAngleAndOffset(idealA, idealB,
actualA, actualB);

        System.out.println("results " + results);
        check(results,9.797509, 10.267162, 0.000000, -3.000000 );


        board  = new Location(LengthUnit.Millimeters, 5, 8, 0, 92);
        // determinate ideal locations
        idealA = Utils2D.calculateBoardPlacementLocation(board,
Side.Bottom, 100., idealA);
        idealB = Utils2D.calculateBoardPlacementLocation(board,
Side.Bottom, 100., idealB);

        // determinate actual locations
        actualA =
Utils2D.calculateBoardPlacementLocationInverse(board, Side.Bottom,
100., idealA);
        actualB =
Utils2D.calculateBoardPlacementLocationInverse(board, Side.Bottom,
100., idealB);
        actualA =
Utils2D.calculateBoardPlacementLocation(board.addWithRotation(offset),
Side.Bottom, 100., actualA);
        actualB =
Utils2D.calculateBoardPlacementLocation(board.addWithRotation(offset),
Side.Bottom, 100., actualB);
        results = Utils2D.calculateAngleAndOffset(idealA, idealB,
actualA, actualB);

        System.out.println("results " + results);
        check(results,9.588165, 10.272644, 0.000000, -3.000000 );


2016-01-27 0:21 GMT, Jason von Nieda <notifications@github.com>:

vonnieda:
@cri-s I don't quite follow what you are saying here. If you see the video
I posted, I'm not using fiducials at all. Just board positions and trying
to move to a placement location. The first board with no angle works fine,
the second board is way off.

Jason

cri-s:
I have added test case for this :
  board = new Location(LengthUnit.Millimeters, 30.004, 31.386, -8, 74.646);
        loc1  = new Location(LengthUnit.Millimeters, 31,6,0,0);
        System.out.println("board: " + board + " 100mm" );
        System.out.println("place: " + loc1);
        loc=Utils2D.calculateBoardPlacementLocation(board, Side.Top,
100., loc1);
        System.out.println("results " + loc);
        loc1=Utils2D.calculateBoardPlacementLocationInverse(board,
Side.Top, 100., loc);
        System.out.println("inverse: " + loc1);
        check(loc,32.426392,62.868249,-8,74.646);
        System.out.println("results " + loc);
        System.out.println("test"+ ++i);
        check(loc1,31,6,0,0);

Output was calc with calculator, as previous, but if one have wrong
algorithm in head, it's
even manual calculating is wrong.

Now it passes the test.

FiducialLocate need to be rechecked.

2016-01-27 2:48 GMT, Jason von Nieda <notifications@github.com>:
> @cri-s I don't quite follow what you are saying here. If you see the video
> I posted, I'm not using fiducials at all. Just board positions and trying
> to move to a placement location. The first board with no angle works fine,
> the second board is way off.
>
> Jason
>

vonnieda:
Thanks @cri-s, looks like it's passing the tests now! Ignore the failure that GitHub is showing, that is my fault. There is a race condition in one of the new tests. I just ran the test manually and it looks like everything is working.

I will do a final review on this and if it looks okay I'll get it merged. Thanks for your diligence on this!

vonnieda:
Hi @cri-s, I have made some adjustments to the code, mostly just cleaning up syntax and formatting and changed a few names that didn't fit the naming schemes. I have pushed the branch with my updates to https://github.com/openpnp/openpnp/tree/cri-s-fix_coord.

I am going to post about this on the mailing list to get opinions about the change to bottom coordinates since this is a pretty major change which will break existing jobs.  I suspect the improvements are worth the pain though.

Thanks,
Jason

cri-s:
Jason, why it should break existing jobs? The dimension is optional, at
least on my code and default with zero location.
I have coded it that it don't break existing jobs.
The only real issue is that if having real panels with multiple same boards
the job file must be edited by hand and board file copied on command line
because of underlaying problem with board class.
Il 27/gen/2016 06:51 "Jason von Nieda" <notifications@github.com> ha
scritto:

vonnieda:
Hi @cri-s,

My apologies, I didn't realize you had handled this case. I just ran a test
and it does indeed work for old jobs. Nice work!

With that in mind, I think we can go ahead and merge this.

Thanks for all your effort on this.

Jason

PR #225, opened 2016-02-19
vision Utils.java stripped down

cri-s:
It allows to call external program , including python to find part.
usage example
bottomLocation = parseLoc(callImg("/usr/bin/python","bottom.py",part, toMat(camera.capture()), null);


cri-s:
Exec file is 
<cmd> <arg> <argc> image.png image.tpl
Example
/usr/bin/python bottom.py 1 image.png image.tpl
environment holds some info about part, if defined.

Return value:
exec status of 0 is expected
Return format it something like this:
found part at [324.33:432.33:33.22]  dimension 322:200 

The [::] is evaluated, if there is only one ':' inside square brace x and y is taken, otherwise rotation is taken too.

PR #232, opened 2016-02-24
Update OpenCvCamera.java

cri-s:
Added autorestart feature and changed Mat to be static instead of dynamically reallocated every image fetch

This patch is optional, it helps recovery from freezed camera, traced down to usb cable problems.


vonnieda:
@cri-s Thank you for the contribution, but I don't think I am going to merge this one. It's extremely environment specific and it sounds like it's specific to one user's hardware problems. In addition, there are numerous code quality and code formatting issues. 

If you'd like me to merge this, please improve the code formatting based on https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style and add some comments to the code explaining what it does and why. I'd also like to see it simplified a bit - maybe just loop a couple times with an increasing timer versus a number of hardcoded times.

cri-s:
It cannot simplified because of timings  requirement for webams and
drivers and nested Loops are needed.

2016-02-24 20:54 GMT, Jason von Nieda <notifications@github.com>:

PR #234, opened 2016-02-26
Update FluentCv.java

cri-s:
possible memory leak

### DIFF
--- a/src/main/java/org/openpnp/vision/FluentCv.java
+++ b/src/main/java/org/openpnp/vision/FluentCv.java
@@ -668,6 +668,8 @@ public class FluentCv {
 		if (tag != null && tag.length > 0) {
 			// Clone so that future writes to the pipeline Mat
 			// don't overwrite our stored one.
+			mat = stored.get(tag[0]);
+                        if(mat!=null) { mat.release(); }
 			stored.put(tag[0], this.mat.clone());
 		}
 		return this;


src/main/java/org/openpnp/vision/FluentCv.java
```diff
@@ -668,6 +668,8 @@ private FluentCv store(Mat mat, String... tag) {
 		if (tag != null && tag.length > 0) {
 			// Clone so that future writes to the pipeline Mat
 			// don't overwrite our stored one.
+			mat = stored.get(tag[0]);
+                        if(mat!=null) { mat.release(); }
```
vonnieda: The mat continues to be used in further pipeline calls. Won't this cause it to be released early?

cri-s: No. This.mat get the mat  value. normally this.mat should be released too,
but I'm not sure if that mat is used on some code
And because this I ignore it.
Mat have native pointer in it and this probably block normal memory GC or
something like that.

cri-s:
patch not approved.

vonnieda:
Why close this? I just haven't had time to test it yet.

cri-s:
The repostory it's not deleted.

2016-02-27 17:11 GMT, Jason von Nieda <notifications@github.com>:
> Why close this? I just haven't had time to test it yet.
>

PR #238, opened 2016-02-29
Add webcam support

cri-s:
All operating systems with possibility to select camera by name.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/camera/Webcams.java
+++ b/src/main/java/org/openpnp/machine/reference/camera/Webcams.java
@@ -1,0 +1,251 @@
+/*
+ 	Copyright (C) 2011 Jason von Nieda <jason@vonnieda.org>
+ 	
+ 	This file is part of OpenPnP.
+ 	
+	OpenPnP is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    OpenPnP is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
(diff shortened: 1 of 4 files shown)


src/main/java/org/openpnp/machine/reference/camera/Webcams.java
```diff
@@ -0,0 +1,251 @@
...
+	protected String deviceId = "###DEVICE###";
+
+	@Attribute(required=false)
+	private int preferredWidth=0;
+	@Attribute(required=false)
+	private int preferredHeight=0;
+	
+	protected Webcam webcam;
+	private Thread thread;
+	private boolean test;
```
vonnieda: Why is this called test? It seems to set the camera to gray.

src/main/java/org/openpnp/machine/reference/camera/Webcams.java
```diff
@@ -0,0 +1,251 @@
...
+			webcam.close();
+		}
+		try {
+		webcam=null;
+                for (Webcam cam: Webcam.getWebcams()) {
+                        if(cam.getName().equals(deviceId)) {
+				webcam=cam;
+			}
+		}
+		if(webcam==null) {	// Workaround ???
```
vonnieda: This comment is confusing. Is it a TODO? Can you explain or remove?

cri-s: Because i have used it during test, and at the end before submitting
changed it to gray,
same as on LtiCivilCamera, where the Wizard was copied. Should i
remove it completly,
or rename it to forceGray ?

For the comment:
       if(webcam==null) {      // Workaround ???
                        //webcam =  Webcam.getWebcamByName(deviceId);
                        return;
                }

It was a hint.
It's if the webcam is not found, should the driver open the first cam ,
   webcam = Webcam.getDefault()  ;

or should it as it works now, don't open the camera.




2016-02-29 17:18 GMT, Jason von Nieda <notifications@github.com>:

vonnieda: Yes, please change it to forceGrayscale. I think it's a useful feature, just a bad variable name.

And thank you for the explanation of the comment. I think the current method is good. If the camera is not found we don't want to confuse the user by opening a different camera. You can remove the comments.

Issue #228, opened 2016-02-21
Camera rotation crops frame


vonnieda:
In ReferenceCamera#transformImage we apply several user defined transformations, one of which is rotation. The current rotation system crops the frame, especially in cases where the aspect ratio is not 1:1. There is a TODO in the code to correct this, and there is a good example at: http://stackoverflow.com/questions/22041699/rotate-an-image-without-cropping-in-opencv-in-c

* Change the transformImage code so that it adjusts the size of the image before rotating so that the full rotated image will fit in the new frame.

cri-s:
Simple roation on 90 degree interval:

// rotate 90º counter-clockwise
Core.flip(img.t(), img, 0);

// rotate 90º clockwise
Core.flip(img.t(), img, 1);

// rotate 180º 
Core.flip(img, img, -1);

PR #239, opened 2016-02-29
Update ReferenceCamera.java

cri-s:
added zoom parameter and polish up rotation

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
@@ -1,5 +1,6 @@
 /*
  	Copyright (C) 2011 Jason von Nieda <jason@vonnieda.org>
+ 	Copyright (C) 2016 cri-s <phone.cri@gmail.com>
  	
  	This file is part of OpenPnP.
  	
@@ -28,9 +29,7 @@ import java.io.IOException;
 
 import org.opencv.core.Core;
 import org.opencv.core.CvType;
-import org.opencv.core.Mat;
-import org.opencv.core.Point;
-import org.opencv.core.Rect;
+import org.opencv.core.*;
 import org.opencv.imgproc.Imgproc;
 import org.openpnp.ConfigurationListener;
 import org.openpnp.model.Configuration;
@@ -75,9 +74,12 @@ public abstract class ReferenceCamera extends AbstractCamera implements Referenc
     
     @Attribute(required=false)
     protected int offsetX = 0;
-    
     @Attribute(required=false)
     protected int offsetY = 0;
+    @Attribute(required=false)
+    protected int sizeX = 0;
+    @Attribute(required=false)
+    protected int sizeY = 0;
     
     @Attribute(required=false)
     protected double zoom = 0;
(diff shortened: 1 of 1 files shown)


vonnieda:
@cri-s Hi Cri, this change seems to change the image results. Can you explain why and if this is as it should be?

Before:
screen shot 2016-02-29 at 9 08 49 am

After:
screen shot 2016-02-29 at 9 11 13 am

vonnieda:
@cri-s Is zoom intended to be set by the user? If so, it should be added to the wizard.

cri-s:
I have not discovered it when i have tested.
Now it's fixed, but i have added rotation limits to +-13 degree beside
90 degree interval.
Fip vertical and horizontal, as seen on you'r images is to be used
basically only when using
mirrors.

Yes, Zoom should be user settable, but i prefer to having it only on machine.xml
If you'r tell i should add it to xxx, i will do it.
Two uses cases:
 If camera is rotated and user need to rotate it and to remove black  border.
 To have additional configured camera, that uses the same hw camera,
but with different zoom level, and for uplooking cam, to remove
unwanted surrounding , that is not shielded.

vonnieda:
A few more questions:

* Why the limit of 13 degrees?
* Why do you prefer zoom to only be on machine.xml?

I like the idea, just needs a bit of cleanup.

src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
```diff
@@ -178,12 +181,37 @@ protected BufferedImage transformImage(BufferedImage image) {
         
         // apply affine transformations
         if (rotation != 0) {
-            // TODO: Fix cropping of rotated image:
-            // http://stackoverflow.com/questions/22041699/rotate-an-image-without-cropping-in-opencv-in-c
-            Point center = new Point(mat.width() / 2D, mat.height() / 2D);
-            Mat mapMatrix = Imgproc.getRotationMatrix2D(center, rotation, 1.0);
-            Imgproc.warpAffine(mat, mat, mapMatrix, mat.size(), Imgproc.INTER_LINEAR);
-	    mapMatrix.release();
+		double rot = rotation;
```
vonnieda: @cri-s Can you put some comments in this block to explain what is going on? I find this code to be very confusing as there is a lot of manipulation of the rot variable and no explanation as to why. 

In addition, can you explain the reason for the limit of 13 degrees?

cri-s: In short. It fixes rotation by 90 degree as commented on issue #228 .
affine rotation was leaved as it could be useful. Size was not calc on
affine rotation as  suggested by you because of fast template match
requirement constraints.
I don't see any utility in giving possibility to have rotation to 30
or 45 degree,  but understand that it could be necessary to rotate up
to 3 degree for removing mechanical misalignments. Because this I have
limited it to 13 degree giving additional 10 deg. Of rotation. If you
prefer I remove this limit. The variable rot is not more then variable
rotation, only that its normalized on positive domain for 90 deg
rotation.
45 degree is equivalent to 0.5 rounding on 90 deg rotation.
I know, the calc is wrong if more then +-360 deg is inserted.

made some tests:

640:480 resolution USB 2.1 Webcam , FPS calc every 100 images.

Rotation 133 degree (transpose+flip+warp-affine)
FPS: 66.403
FPS: 78.78
FPS: 91.117

Rotation 90 degree (transpose+flip)
FPS: 113.052
FPS: 104.015
FPS: 125.325

Rotation 3 degree (warp-affine)
FPS: 135.428
FPS: 147.6

Rotation 0 degree
FPS: 158.309
FPS: 169.418
FPS: 181.11

cri-s:
It's this better? what is the conflicts , reformatting ?

vonnieda:
@cri-s Yes, that's much better, thank you!

Don't worry about the conflicts. It's just reformatting. I'll handle it during the merge.

I appreciate your diligence, thanks again.

vonnieda:
Hi @cri-s: I'm working on merging this but running into a problem with the zoom function. I set zoom to > 1 (tried 1.1 and 2.0) and I get assertion errors of:
OpenCV Error: Assertion failed (0 <= _rowRange.start && _rowRange.start <= _rowRange.end && _rowRange.end <= m.rows) in Mat, file /Users/jason/Projects/opencv/opencv-2.4.11/modules/core/src/matrix.cpp, line 284

Have you seen this? Am I using the feature wrong?

cri-s:
No, please test it now. It's probably bug of newer version of java
submat implementation.
Have changed it to mat(img,Rect) and that seems to work on newer version.
I have not see it, because i'm using 2.4.8 externally, the precompiled
version quit with memory
dump on init. Have changed it, if zoom is negative, the image is
resized to orginal mat size.
Added ratio paramter too, it's the ratio of height, if someone needs
to change the ratio.


2016-03-02 6:10 GMT, Jason von Nieda <notifications@github.com>:

src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
```diff
@@ -205,12 +228,26 @@ protected BufferedImage transformImage(BufferedImage image) {
...
+        { double scale = zoom; if (scale<0) { scale=-scale; }
+	  if(scale>1.0) {
+		int w=(int)((double)mat.width()/zoom+.5);
+		int h=(int)((double)mat.height()/zoom*ratio+.5);
+		int x=(mat.width()-w)/2;
+                int y=(mat.height()-h)/2;
+          	Mat img = new Mat(mat,new Rect(x,y,w,h));
+          	if(zoom<0.) {
+    			Imgproc.resize(img, img, mat.size());
+          	}
```
vonnieda: @cri-s I am testing this now and the zoom doesn't seem to be doing anything. Looking at this code, I think that you are missing an else clause at this line. Did you want to resize mat at some point?

cri-s:
If it don't work for you, it's a bigger problem on Mac.
The zooming is in reality it's image cropping. The gui does the zooming.
I have added code to resize the image too, but zoom and scale need to be updated, i have omitted
it.
img.submat(x1,y1,x2,y2)  is a ROI inside orginal img. 
Mat(img,new Rect(x,y,w,h)) is equivalent to img.submat(x,y,x+w,y+h).clone()  , except that
code is similar to this.
mat = new Mat(new Size(w,h),img.type()) ; img.submat(x,y,x+w,y+h).copyTo(mat);
If both not works, it's strange.

I finish to merge latest change and uplooking vision to the branch for mojalova1 and then after physiotherapy i copy the code from that branch over here. 
If it's work on Mac, fine, otherwise close the pull request.

cri-s:
Checked in the code i have used close to official branch.

vonnieda:
@cri-s Can you explain what these new fields do? Have you tested this code in place and shown that it works? It takes a lot of my time to pull these changes and run through a bunch of testing and I don't want to keep doing it if this code is not being tested before being sent to me.

cri-s:
Yes, this is tested on Linux and is used by several users, one of them
use windows.
I don't know if it works on Mac
and if maybe the changed offset gives you or other problems. .
For example submat works on linux, and it seems to not work on Mac in
this particular
case.

zoom , set zoom level on camera, adjust offsetX/Y and sizeX/Y , ratio
is ratio of height for zoom level, just plain multiplicator. Zoom
takes precedence over other settings.
offsetX/Y and sizeX/Y is offset and size for cropping image.
If offsetX or offsetY is negative, then circular roi is used.
This is primary for uplooking cam.
Close this pull request.
The use is for uplooking camera. I include simple BottomOffset code,
you surely want use the FluentCv interface. On getCameraImage i do
gaussian blur and normalize .

  @Override
    public Location getPartBottomOffsets(Part part, Nozzle nozzle)
            throws Exception {
        if (camera.getLooking() != Looking.Up) {
            throw new Exception("Bottom vision only implemented for Up
looking cameras");
        }

        double speed=1.0;
	if(part!=null) {
		speed=part.getSpeed(); // TODO wrong speed, leave it as is until spi change
	}

	Location loc = camera.getLocation().
  		convertToUnits(LengthUnit.Millimeters);
        // Position the part above the center of the camera.
        // First move to Safe-Z.
        nozzle.moveToSafeZ(speed);
        // Then move to the camera in X, Y at Safe-Z and rotate the
        // part to 0.
        nozzle.moveTo(loc.derive(null,null,Double.NaN,null), speed);
  	double height = loc.getZ();
        if(part!=null) height+= part.getHeight()
                .convertToUnits(LengthUnit.Millimeters).getValue();
        else height+=1.;
        if(height>0.) height=0; // XXX HACK for other drivers
        // Then lower the part to the Camera's focal point in Z. Maintain the
        // part's rotation at 0.
        nozzle.moveTo(loc.derive(null, null, height, 0.0), speed);
        // Grab an image.
        // Return to Safe-Z just to be safe.
        nozzle.moveToSafeZ(speed);
        // Do OpenCV magic
       	List<MatOfPoint> contours = new ArrayList<>();
	RotatedRect rot = null ;
	double ang ;
	Imgproc.threshold(getCameraImage(0),tmp,240,255,0);
        Imgproc.findContours(tmp.clone(), contours,new Mat(),
Imgproc.RETR_EXTERNAL, Imgproc.CHAIN_APPROX_SIMPLE);
	if(contours.size()==0) return null;	// XXX Failed
	// join, get angle,
   	contours = joinContours(contours);
	rot=getRotatedRect(contours.get(0));
	ang = rot.angle % 45;
	// roate and repeat match, get center
	tmp = rotate(tmp,-ang); contours.clear();
        Imgproc.findContours(tmp.clone(), contours,new Mat(),
Imgproc.RETR_EXTERNAL, Imgproc.CHAIN_APPROX_SIMPLE);
   	contours = joinContours(contours);
	rot=getRotatedRect(contours.get(0));

    	return
	    loc.derive(rot.center.x,rot.center.y,0.,ang).subtract(loc.derive(gray.width()/2.,gray.height()/2.,0.,0.0)).multiply(camera.getUnitsPerPixel().derive(null,null,0.0,1.0)).multiply(1.0,-1.0,1.0,1.0);

    }

    private static RotatedRect getRotatedRect(MatOfPoint contour) {
	    if(contour==null) return  null;
      	    MatOfPoint2f contour_ = new MatOfPoint2f();
	    contour.convertTo(contour_,CvType.CV_32FC1);
            return Imgproc.minAreaRect(contour_);
    }

    private static List<MatOfPoint> joinContours(List<MatOfPoint> contours) {
        List<MatOfPoint> ret = new ArrayList<>() ;
	// filter it
        for (MatOfPoint contour: contours) {
            RotatedRect rotatedRect = getRotatedRect(contour);
            double area = rotatedRect.size.area();
            if (area <= 5.0) contours.remove(contour);
        }

	final ArrayList<Point> temp = new ArrayList<>();
        for (MatOfPoint contour: contours) {
            temp.addAll(((MatOfPoint)contour).toList());
	}

	contours.clear();
	if(temp.size()!=0) {
		MatOfPoint mat = new MatOfPoint();
		mat.fromList(temp);
		contours.add(mat);
	}
	return contours;
    }

static  public Mat rotate(Mat img, double angle)
{
    double radians = Math.toRadians(angle);
    Point center = new Point(img.size().width/2, img.size().height/2);
    Mat rotImage = Imgproc.getRotationMatrix2D(center, angle, 1.0);
    Imgproc.warpAffine(img, img, rotImage,img.size(),
Imgproc.INTER_LINEAR + Imgproc.CV_WARP_FILL_OUTLIERS);
    return img;
}





2016-03-04 17:26 GMT, Jason von Nieda <notifications@github.com>:

PR #245, opened 2016-03-18
Demo PNP board enhancements

anthonywebb:
Changed ruler to copper
Added .5mm spacing to ruler
Added ruler to Top Layer so now it is on top and bottom
Added 0201 components.
Created new components named BOTTOM_RULER_ORIGIN and TOP_RULER_ORIGIN. Now they will show up in job list so you can zoom to them.


vonnieda:
Great work @anthonywebb. Very nice improvements!

PR #256, opened 2016-04-07
Don't crop rotated frames in ReferenceCamera

### LINKED ISSUE #228: Camera rotation crops frame

richard-sim:
In ReferenceCamera#transformImage, adjust the size of the image before rotating so that the full rotated image will fit in the new frame. Fixes #228

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
@@ -82,6 +82,9 @@ public abstract class ReferenceCamera extends AbstractCamera implements Referenc
     private CalibrationCallback calibrationCallback;
     private int calibrationCountGoal = 25;
 
+    private Mat undistortionMap1;
+    private Mat undistortionMap2;
+
     protected ReferenceMachine machine;
     protected ReferenceDriver driver;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
```diff
@@ -206,14 +195,69 @@ protected BufferedImage transformImage(BufferedImage image) {
...
+                put(0, 0, 1, 0, offsetX);
+                put(1, 0, 0, 1, offsetY);
+            }
+        };
+        
+        Mat dst = mat.clone();
+        Imgproc.warpAffine(mat, dst, mapMatrix, mat.size(), Imgproc.INTER_LINEAR);
+
+        mapMatrix.release();
+
```
vonnieda: I think we're missing a mat.release() here?

richard-sim:
Erk, good catch - thanks!

vonnieda:
This all looks great aside from that one mat.release comment. If you'll fix that I'll get these merged.

richard-sim:
Should be good now - I also found another missed release() while I was at it. Double oops.

vonnieda:
Awesome work @richard-sim, thank you!

PR #309, opened 2016-08-03
Pull request

tracklogic:
Hi Jason

I've been slightly bad with my code management, not sure how to rectify at this point as we've got 4x things in the pull request. I've also got your loose-part-feeder in this as well. If its easier for you i'll break it up like i should of done in the first place. (Guess by doing separate feature branches?).

-> getFeederByName()
-> Vision interface change to allow for the centering stage as per our prior discussion
-> Fiducial homing
-> Paste dispenser support / initial groundwork for the glue dispenser


tracklogic:
Not happy with this mega-commit, have started splitting into feature-branches - much cleaner - should of done it in first place. Bad Matt!. Shouldn't take long.

PR #310, opened 2016-08-03
Added fiducial homing to gcode driver / fiducialLocator.

tracklogic:


vonnieda:
@tracklogic Does this code assume that the home pin is at the location where the controller homing stops? In other words, does it need to be at the home switch locations?

tracklogic:
Hi

Yes at the moment at least, it assumes that once the driver has homed the machine the fiducial is within the camera window. We could add an additional configurable move if this isn't the case though?

src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
```diff
@@ -188,6 +187,33 @@ public void home(ReferenceHead head) throws Exception {
...
         }
+
+        /*
+                The head camera for nozzle-1 should now be (if everythings homed correctly)
+                directly above the homing pin in the machine bed, use the head camera scan
+                for this and make sure this is exactly central - otherwise we move the camera
+                until it is, and then reset all the axis back to 0,0,0,0 as this is calibrated
+                home.
+         */
+        Part homePart=Configuration.get().getPart("FIDUCIAL-HOME");
```
vonnieda: I'd like to see this part be configurable, using a part dropdown like in the feeder configuration, for instance. If that is overly difficult, let's at least make it a configurable parameter in the XML and I will add the UI for it later. It should just be something like homingFiducialPartId, for instance.

src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
```diff
@@ -188,6 +187,33 @@ public void home(ReferenceHead head) throws Exception {
...
+        if (homePart != null) {
+            Location tmp = new Location(LengthUnit.Millimeters,0.0,0.0,0.0,0.0);
```
vonnieda: Let's make this location configurable in the XML. Just an @Element Location homingFiducialLocation would do the trick. Otherwise we're assuming that anyone who wants to use this will have their homing fiducial right under their endstops, and I think that situation will be rare.

PR #312, opened 2016-08-03
Feature paste dispense

tracklogic:


src/main/java/org/openpnp/spi/base/AbstractPasteDispenseJobProcessor.java
```diff
@@ -0,0 +1,90 @@
+package org.openpnp.spi.base;
+
+import org.openpnp.model.Configuration;
+import org.openpnp.model.Part;
+import org.openpnp.spi.*;
+import org.openpnp.util.MovableUtils;
+
+public abstract class AbstractPasteDispenseJobProcessor extends AbstractJobProcessor
+        implements PasteDispenseJobProcessor {
+
```
tracklogic: i'm thinking these funcs() can be removed - they are just left-over from when i took the abstractpnpjobprocessor class - not doing any harm, but may cause later confusion?

cri-s:
Position should be sorted based on distanc and eventually needle instad of using
it unsorted. For dispensing, instead of SafeZ the max height should be checked and used that.
That influence a lot the dispense speed.

tracklogic:
Hi Cri

I agree with you about sorting by distance, i had hoped to get this pushed in and then i (or someone else if they are feeling helpful hehe) will further refine it to do sorting, i was just hoping to get the paste feature at least back and enabled.

I'm not sure about not moving to safeZ though, I understand what you mean about speed, but is there a danger we are going to get a collision? Say for example the PCB we are dispensing is at a lower Z than the feeder - we could perhaps hit the feeder whilst moving about? Feeders sometimes have things sticking up (knock pins etc..) that make them taller than the PCB (and any max-height component you can calculate)?

cri-s:
On init and cleanup, clearly use moving safeZ, head.safeZ or just hardcoded 0.0 .
Normally pcb space is restricted in such way that no head accidentally touch feeder or something when moving annother head/nozzle/tip. There must be some clearance for safe opertion. 
Further for paste dispense, usually no other component is placed.
Sure, it is possible to do rework, and for that, probably one checkbox should be present in order to enable/disable safeZ optimisation. 

sorting:
Collections.sort(jobDispenses, (left, right) -> left.boardLocation.getLocation().getLinearLengthTo(calculateBoardPlacementLocation(left.boardLocation,left.placement)) -
right.boardLocation.getLocation().getLinearLengthTo(calculateBoardPlacementLocation(right.boardLocation,right.placement)));

I have realized now, it's glue dispenser, not all components are glued, only components that don't
are hold by superficial tension. Do you have made a dialog, that support selecting the parts to glue?

A rule of thumb to keep parts from falling off, is this one:

R = C/P (g/in) must be ˜ 30

Where
C = Component weight in grams
P = total pad area in square inches

tracklogic:
Hi Cri

I've sorted out the paste and i've started plugging in the glue one - its got a seperate tab list for glue dispense co-ordinates (so pick & place, paste dispense, glue dispense on the U.I.).

Matt

cri-s:
Just noted, probably is better using uplooking camera or
 right.placement.derive(0.0,0.0,0.0,0.0)  in order to have consistient distance from know point, even
it's 0 point.

src/main/java/org/openpnp/spi/base/AbstractPasteDispenseJobProcessor.java
vonnieda: Agreed, if they aren't used please remove them.

src/main/java/org/openpnp/gui/JobPanel.java
```diff
@@ -651,8 +651,10 @@ public void jobStart() throws Exception {
         if (title.equals("Solder Paste")) {
             jobProcessor = Configuration.get().getMachine().getPasteDispenseJobProcessor();
         }
-        else if (title.equals("Pick and Place")) {
-            jobProcessor = Configuration.get().getMachine().getPnpJobProcessor();
```
vonnieda: Correct me if I am wrong, but it looks like you removed the PnpJobProcessor entirely.

src/main/java/org/openpnp/gui/tablemodel/PlacementsTableModel.java
```diff
@@ -117,6 +117,16 @@ else if (columnIndex == 5) {
             else if (columnIndex == 6) {
                 placement.setType((Type) aValue);
             }
+            else if (columnIndex == 8) {
+                if (aValue.toString().compareTo("Yes")==0)
```
vonnieda: We typically render boolean options as a checkbox by making the column class Boolean. Is there a reason you used these text strings?

src/main/java/org/openpnp/gui/JobPanel.java
tracklogic: I had to look twice too :D hehe

src/main/java/org/openpnp/gui/tablemodel/PlacementsTableModel.java
tracklogic: Good spot, I'll flick it to being a bool, think i was just being lazy :)

PR #315, opened 2016-08-04
Mathematical sorting for LengthCellValue rather than string sorting.

tracklogic:


cri-s:
Could you please check if length.units are the same, and if not, convert the second unit to the same
as the first one. Further, it's possible to add something like this for identifier names?
It should sort names like usual, but if there is a number inside name, it should sort it numeric.

@Override
public int compare(String s1, String s2) {
{   if(s1==null||s2==null) return s1==s2?0:s1==null?-1:1;
    int i1=0,i2=0;
    for (;i<s1.length();i1++,i2++) {
       if(s2.length()==i) return -1;
       if(s1.charAt(i1) == s2.charAt(i2)&&!Character.isDigit(s1.charAt(i1))) continue;
       // digit now
       int d1=0,d2=0;
       for(;i1<s1.length()&&Character.isDigit(s1.charAt(i1));i1++) 
                  { d1*=10; d1+=Integer.parseInt(""+s1.charAt(i1));}             
        for(;i2<s2.length()&&Character.isDigit(s2.charAt(i2));i2++) 
                  { d1*=10; d2+=Integer.parseInt(""+s2.charAt(i2));}
       if(d1==d2) continue; return d1-d2;
      }
      if(i1<s1.length()) return -1;
      if(i2<s2.length()) return 1;
      return 0;
}

tracklogic:
Hi Cri

Ok @ length units - good spot. Also like the idea of the identifier name sorting, i'll sort it out later :)

Matt

tracklogic:
Cri -> i was struggling a bit to follow the code you wrote, i've quickly written this (though its in my native C++ hehe... but is this what you were thinking, i'll flip it to java later), think it does what we want

int numericPos(std::string str)
{
	int ret = -1;

	for (int i = 0; i < str.length(); i++)
	{
		if (isdigit(str.at(i)))
		{
			ret = i;
			break;
		}
	}
	return ret;
}

int getNumeric(std::string str)
{	
	std::string numeric;
	for (int i = 0; i < str.length(); i++)
	{
		if (isdigit(str.at(i)))
		{
			numeric+=str.at(i);
		}
	}
	return std::stoi(numeric);
}


int componentNameCompare(std::string s1, std::string s2)
{
	if (s1 == s2)
	{
		return 0;
	}
	if (numericPos(s1)!=-1 && numericPos(s2)!=-1)
	{
		std::string s1First = s1.substr(0,numericPos(s1));  // L123 this would contain L
		std::string s2First = s2.substr(0, numericPos(s2)); // L256 this would contain L
		if (s1First < s2First)
		{
			return -1;
		}
		if (s1First > s2First)
		{
			return 1;
		}
		if (s1First == s2First)
		{
			std::string s1Second = s1.substr(numericPos(s1), string::npos); // L123 this would contain 123
			std::string s2Second = s2.substr(numericPos(s2), string::npos); // L256 this would contain 256

			int num1;
			int num2;
			num1=getNumeric(s1Second);
			num2=getNumeric(s2Second);
			
			if (num1 < num2)
				return -1;
			if (num1 > num2)
				return 1;
			return 0;
		}
	}
}

cri-s:
probably not really, you simply compare string01 and string100 as i
could see, but not
resistor#220@0603%0.1,metal_foil
If after comparing numbers, you continue to compare strings and then
number, if there are
numbers present, it should work, but the -1 fool the comparsion between
foobar foo1bar foo2bar and fooabc
you code probably return
foo1bar
foo2bar
fooabc
foobar

the intended code should return
fooabc
foobar
foo1bar
foo2bar

The difference is default value, 0 or -1 .


Example:
res100%1
res1
res20%5

string sort:
res1
res100%1
res20%5

should be:
res1
res20%5
res100%1

Algorithm:
If one string contains numbers, parse number using temp int variable,
and compare int value.
Side effect:  foo0bar is same as foobar . foo007bar is same as foo7bar
For this, and because numbers can be different length, two position
iterators are needed.
Numbers need to be parsed, example:
   for(d1=0;isdigit(str1.at(i1));i1++) {d1*=10; d1+=str1.at(i1)-'0'; }
The result is the value, not the length of the numeric contents.
same for str2 using d2, and compare d1 and d2;
if(d1==d2) continue; else return d1-d2;
and after alphanumeric loop, one string is terminated, the other maybe not.
If both are iterated, return null, if str1 is iterated to end, return
-1 otherwise 1.


2016-08-04 13:28 GMT, tracklogic <notifications@github.com>:

tracklogic:
Hi Cri

Probably me complicating things doing it in C++ -> 

Your code
```java
 for(d1=0;isdigit(str1.at(i1));i1++) {d1*=10; d1+=str1.at(i1)-'0'; }
```

is same as C++ function, we loop through the string, any character that is a number is appended to a temporary string, and then at the end of the function the string is converted to an int, its not returning the length of the string, its returning the numeric value.

```C
int getNumeric(std::string str)
{	
	std::string numeric;
	for (int i = 0; i < str.length(); i++)
	{
		if (isdigit(str.at(i)))
		{
			numeric+=str.at(i);
		}
	}
	return std::stoi(numeric);
}
```

cri-s:
I have tested it in java, not perfect, but working as intended (it have a bug somewhere).
It's static because i have tested it with static main.

  public static int compare (String s1, String s2)
  {
    if (s1 == null || s2 == null)
      return s1 == s2 ? 0 : s1 == null ? -1 : 1;
    int i1 = 0, i2 = 0;
    for (; i1 < s1.length () && i2 < s2.length (); i1++, i2++)
      {
        if (s1.charAt (i1)== s2.charAt (i2))
          if(!Character.isDigit (s1.charAt (i1)))
                continue;
// digit now
        int d1 = 0, d2 = 0;
        if(Character.isDigit (s1.charAt (i1))) {
        for (; i1 < s1.length () && Character.isDigit (s1.charAt (i1)); i1++)
          {
            d1 *= 10;
            d1 += Integer.parseInt ("" + s1.charAt (i1));
          }
        }

        if(Character.isDigit (s2.charAt (i2))) {
        for (; i2 < s2.length () && Character.isDigit (s2.charAt (i2)); i2++)
          {
            d2 *= 10;
            d2 += Integer.parseInt ("" + s2.charAt (i2));
          }
        }

        if (d1 == d2)
          continue;
        return d1 - d2;
      }
    if (i1 < s1.length ())
      return 1;
    if (i2 < s2.length ())
      return -1;
    return 0;
  }

tracklogic:
Added part sorting (as per Cri suggestion) and also rotation sorting

src/main/java/org/openpnp/gui/support/RotationCellValue.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public void setDisplayNativeUnits(boolean displayNativeUnits) {
+        this.displayNativeUnits = displayNativeUnits;
+    }
+
+    @Override
+    public String toString() {
+        double r = Rotation;
+        /*if (l.getUnits() == null) {
```
vonnieda: If this commented code is not used, can it be removed? If not, please add a command explaining why it's there.

src/main/java/org/openpnp/gui/tablemodel/PlacementsTableModel.java
```diff
@@ -111,8 +112,9 @@ else if (columnIndex == 4) {
                 placement.setLocation(location);
             }
             else if (columnIndex == 5) {
-                placement.setLocation(placement.getLocation().derive(null, null, null,
-                        Double.parseDouble(aValue.toString())));
+                /* placement.setLocation(placement.getLocation().derive(null, null, null,
```
vonnieda: Is setting handled by the cell value class? If so, can this commented code be removed?

src/main/java/org/openpnp/gui/AlphaIcon.java
```diff
@@ -0,0 +1,93 @@
+/**
```
vonnieda: General comment: If you are using code that is from an external source, as some of these files appear to be, can you please include information in the header as to the source of the file and it's licensing terms? For instance, I will often include a link to a StackOverflow page where I found a snippet.

tracklogic: Hi Jason

No problem, I've checked the licence and put info in the commit comment, but having it in the comment section makes it easier to reference.

Matt

PR #316, opened 2016-08-06
Create Bank.java

cri-s:
Reduced bank functionality, reduced, because OpenPnP don't follow traditional Feeder coordinate convention and that raises up
some incompatibility. Sorry. 
If bank feeder id is "", then it hides or unhide any unused feeder. To be used 
Example:
     ```
    &lt;feeder class="org.openpnp.machine.reference.feeder.Bank" id="" name="Bank" enabled="false" part-id="FIDUCIAL-1X2-FIDUCIAL1X2" retry-count="3" orientation="0">
            &lt;location units="Millimeters" x="0.0" y="0.0" z="0.0" rotation="0.0"/>
            &lt;offset units="Millimeters" x="0.0" y="0.0" z="0.0" rotation="0.0"/>
            &lt;feeders/>
         &lt;/feeder>
```

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/Bank.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/Bank.java
@@ -17,7 +17,7 @@ modification, are permitted provided that the following conditions are met:
 THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
 ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
 WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
-DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER BE LIABLE FOR ANY
 DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
 (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
 LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
@@ -78,7 +78,8 @@ import org.openpnp.gui.support.Wizard;
 import org.openpnp.machine.reference.ReferenceFeeder;
 import org.openpnp.machine.reference.ReferenceMachine;
 import org.openpnp.spi.*;
-import org.openpnp.machine.reference.feeder.wizards.ReferenceFeederConfigurationWizard;
+import org.openpnp.machine.reference.feeder.wizards.
+  ReferenceFeederConfigurationWizard;
 
 import org.openpnp.spi.base.SimplePropertySheetHolder;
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by vonnieda** [comment]
What does this do and how is it used?

**Comment by cri-s** [comment]
when the feeder is "feed" it loads feeders depend on barcode. Barcode
size is hardcoded,
or call a script that do something for creating feeders, example is
tray feeder having several
components on it that are generated based on barcode.
If id="" , then when "feed", it hides all feeder that are not used on
current job and boards.
If enabled, it create feeders, with bank prefix, if disabled, it
removes that feeder.
As example it is possible to hide the unused feeders with the "unused"
feeder bank, and
then  move the machine feeder to feeder bank.
Bank feeder can be recursive, one can embedded annother bank feeder,
drag and drop on
tree view (Machine Panel) would be nice, and beside the feeder name,
all things can be
edited on that three without adding or removing feeders.
Automatic offset and rotation update was removed, because of some
incompatibility of coordinate system with usual pnp standard
coordinate system. I was not able to catch that.
It's a feeder for advanced users that as example want having auto
feeders with barcode
recognition just plugging in feeders and automounting by script, or
having project feeders triggered manually or by barcode keycard. It
allows have some Bank of feeders that are enabled or disabled
automatically. It's not a feeder for beginners, except if scripts are
used,
and it's not a feeder that can easily initial setup using gui.

2016-08-06 17:58 GMT, Jason von Nieda <notifications@github.com>:

File: src/main/java/org/openpnp/machine/reference/feeder/Bank.java [hunk 1]
```diff
@@ -0,0 +1,271 @@
...
+	if(ids!=null)
+	for(String id:ids.split("|")) { 
+		logger.debug("found feeder {} at bank {}",id,getId());
+		for(Feeder fdr:machine.getFeeders())  
+			if(id.equals(fdr.getId())) { feeder=fdr; break; }
+		if(feeder!=null) {
+			logger.info("found feeder {} at bank {}",feeder.getName(),getName());
+			// register feeder
+		} else { String dir = System.getProperty("user.dir");
+			System.setProperty("user.dir",  new File(Configuration.get().getConfigurationDirectory(), "script").toString());
```
vonnieda: Please do not modify user.dir. This can cause problems with the entire program. If your i.eval(id) fails, for instance, the old directory will not get set back correctly.

src/main/java/org/openpnp/machine/reference/feeder/Bank.java
```diff
@@ -0,0 +1,271 @@
...
+	for(Feeder fdr:feeders) {
+			if(machine.getFeeder(fdr.getId())!=fdr) 
+			try { machine.addFeeder(fdr); } catch(Exception e) {;}
+	}
+    }
+
+
+    @Override
+    public void feed(Nozzle nozzle) throws Exception {
+	if(getId().length()==0) { // hack
```
vonnieda: Why is this a hack? Please clean up or remove.

src/main/java/org/openpnp/machine/reference/feeder/Bank.java
```diff
@@ -0,0 +1,271 @@
...
+		setEnabled(true);
+		feeders.clear();
+		for(Board b:Configuration.get().getBoards())
```
vonnieda: Loops and conditionals must use braces. Please see https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style and clean up the code.

src/main/java/org/openpnp/machine/reference/feeder/Bank.java
```diff
@@ -0,0 +1,271 @@
+/*
+
+Copyright (C) 2016 AV (CS) <phone.cri@gmail.com>
```
vonnieda: If you wish to retain copyright, please remove the below license and add the standard OpenPnP license. You may leave the above copyright statement.

 * This file is part of OpenPnP.
 * 
 * OpenPnP is free software: you can redistribute it and/or modify it under the terms of the GNU
 * General Public License as published by the Free Software Foundation, either version 3 of the
 * License, or (at your option) any later version.
 * 
 * OpenPnP is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
 * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
 * Public License for more details.
 * 
 * You should have received a copy of the GNU General Public License along with OpenPnP. If not, see
 * <http://www.gnu.org/licenses/>.
 * 
 * For more information about OpenPnP visit http://openpnp.org

src/main/java/org/openpnp/machine/reference/feeder/Bank.java
```diff
@@ -0,0 +1,271 @@
...
+	double mmy=camera.getUnitsPerPixel().getY();
+	boolean ena = enabled;
+	// move camera
+	nozzle.moveToSafeZ();
+	camera.moveTo(location.derive(null,null,Double.NaN,0.0));
+	camera.moveToSafeZ();
+	// updte feeders
+	setEnabled(false);	
+	// changed from RTAG to barcode and to bufferedImage
+        CameraView cameraView = MainFrame.get().getCameraViews().setSelectedCamera(camera);
```
vonnieda: Feeders must not access the UI, otherwise they cannot be used in tests. This should be changed to simply call camera.capture() and then extract the part of the image you want.

tracklogic:
Hi Cri

This looks quite cool - i'm guessing (have you got any examples?) you create a QR barcode that has the various setup parameters on (pipe delimited) and you stick this on the feeder bank assembly. Camera then goes across, decodes the barcode and tokenises the string based upon | character, after which it setups the various individual feeders and/or executes a script for further setup?

src/main/java/org/openpnp/machine/reference/feeder/Bank.java
cri-s: That is true, good catch. The reason for that is , that the current
dir is from where the
user call the program, and the file inclusions don't work as is should
into scripting environment . I will do it otherwise.

2016-08-07 1:46 GMT, Jason von Nieda <notifications@github.com>:

src/main/java/org/openpnp/machine/reference/feeder/Bank.java
cri-s: This is a dual use of the feeder Bank.
If the id of the feeder is empty, special handling of the Bank feeder
is activated.
Only one feeder can exist with that id, globally.
It hiddes all feeders, that are not used inside actual Job, not matter
if feeder is enabled or not.
It's possible to define a Bank Feeder with name "Unused" as example,
and if feed, it loads
all unused feeder and is capable of hidding them, as there was never configured.

cri-s:
Yes, but there is "ONE" big difference. The script directory is named
"script" and not "scripts"
inside openpnp configuration. There was more functionality, as example
automatically
mapping and unmapping of locations including rotation relative to bank
location, but that was
incompatible with some feeder convention the OpenPnP uses, so i have
removed that.
Beside this configuration, the usage of Bank feeder is to hide feeder
and organize them into
Banks for better managability.

tracklogic:
The rotation sounds quite useful, is that for say my machine (Zevatech 460) which has a feeder bank on each side... dependant which feeder bank (front, left, back or right face). it goes in, you need to alter the rotation.

Do you have the original none-filtered code kicking about? Its a shame to strip all the functionality out....

cri-s:
Unfortunatly it's not compatible with Feeder Wizard (Drag+Strip) and
need small changes
to real feeder, so it's probably a no go. If you are willing to use
other feeder with different
setup, then it's possible. OpenPnP don't respect traditional feeder
coordinate convention, that this Bank feeder suppose to be. Hovewer
you can use a script to do that
without problems, you just need to hardcode some constants in case you
use barcodes.


2016-08-07 14:14 GMT, tracklogic <notifications@github.com>:

tracklogic:
Hi Cri/Jason

This doesn't quite work how i thought, though it might just be a design pattern thing. I thought that a feeder bank would own the feeders and the feeder bank class would be the interface to the feeders in the feeder bank. So when something wants to control a feeder it goes via the feeder bank class (its the interface).

What seems to happen though is the feeder bank class stores some feeders and adds it to the main machine feeder list, then later on it iterates through the main feeder list to do things on any feeder it added itself.

Maybe theres a good reason why Cri has adopted this model? 

Matt

tracklogic:
Thinking about it, I think I can see why Cri has done this... at the moment the feeder interface looks like its only setup to feed one specific part. 

 public void feed(Nozzle nozzle) throws Exception;

When feed is called, it has no knowledge of what part we actually want to feed only the nozzle?

Although I'm more confused than ever now as it appears when you do a feeder bank feed it uses the camera to setup the feeders?

cri-s:
Suppose you have 14 slots on two sides where you can place 8 or 12mm Feeder, or
16/24mm using 2 slots, or 3 slots for 32/48mm Feeder.
For the 14 slot feeder, there are defined 1+14+13+12 Bank feeder, for
loading the feeder,
and if the feeder is loaded two feeders are created, one tray feeder
with quantity 1 and
one AutoFeeder in order it does feeding after pickup and not before.

You could use bank feeder for projects too, and simply dont feed, but
enable and disable
feeder.  It's have multiple uses.

If there is one Bank feeder, with nested Bank feeders, it's possible
to physically load the feeders,
launch script to check all active Bank feeder, and it load all feeders
automatically based
on the barcode, and then hide the other bank feeders using the "unused" feeder.
Further its' nice for mixed trays where some strip feeders and other
tray feeders are present.
If you have as example 4 such trays, you can manually activate it, or
if there are
project dependent trays, the barcode set it up.


2016-08-08 7:38 GMT, tracklogic <notifications@github.com>:

src/main/java/org/openpnp/machine/reference/feeder/Bank.java
cri-s: The Licence is BSD 3 Clause type and compatible with GPL
and already used in this project.

Issue #276, opened 2016-05-24
Reset feed count


psykhon:
A very small but usefull feature, a button to reset Feed count on a tray feeder, better than to put 0 in the textbox and hit apply

psykhon:
the currently "reset" button should be renamed "Cancel"

vonnieda:
Agreed. We have this in the Strip Feeder, it should be added to the Tray Feeder as well.

vonnieda:
Resolved by #322

PR #322, opened 2016-08-10
Feature - Reset Feed Count

jjpickett:
This feature adds a "Reset" button to the Tray Feed count as requested by issue 276: [https://github.com/openpnp/openpnp/issues/276](url) 

The biggest change is from lines 121 - 130. This adds the buttons functionality and sets its design to be a simple button. Once the user selects the "Reset" button, it will then set the Feed Count to be 0 and then update the value.

Other lines had properties added to them to make their Grid width 2. This allows for the button to maintain a small size, while the other elements in the column are able to resize with the rest of the window.


vonnieda:
Looks good @jjpickett! One quite note: When you reference an issue / ticket in comments or commit logs, put a # in front of the number. This will cause Github to link it and it makes it easier to find things. For instance: #276.

jjpickett:
Thanks @vonnieda, I didn't realise GitHub had that feature. I'll make sure to use that going forward!

Cheers!

vonnieda:
Nice work @jjpickett, thank you!

Issue #321, opened 2016-08-10
Tray Feeder Count Not Updating


jjpickett:
**Summary:** The Tray Feeders feed count is not being updated when performing a feed and pick.

**Steps:**
1. Go to the bottom panel and select the tab Feeders
2. Create and set up a new Tray Feeder
3. Give the Tray Count some numbers (I used 1 and 3)
4. Select the "Perform a feed and pick on the selected feeder" button.

**Expected Result:** The Feed Count will update its number on the screen.

**Actual Result:** Notice that the Feed Count will display the same number as before.

**Note:** The process will still operate properly with the correct numbers. The issue is that this is not being displayed to the user.

[Video of Tray Feed Count Issue]
(https://github.com/openpnp/openpnp/files/411648/Feed.Count.Bug.Video.zip)

vonnieda:
See ReferenceStripFeederConfigurationWizard for how to set up the bindings to make this work right.

vonnieda:
Fixed by #323

PR #323, opened 2016-08-10
Bug Fix - Tray Feeders feed count was not updating

jjpickett:
This fix was in response to issue #321 

The major change is instead of the feeder increment on line 95, the code now uses the setFeedCount method to increase the number. The setFeedCount method takes in the new number given and then updates the screen accordingly.


vonnieda:
Thanks @jjpickett! Fixes #321.

Issue #319, opened 2016-08-08
Add option to check local fiducials on a board prior to placement


issus:
Add a column to the Job -> Boards -> Pick and Place table for checking fiducials before placing the part. If the column is checked, then the down camera will find the two (configurable?) closest fiducials to the part and adjust alignment before placing the part.

This will be useful for high pin count parts, oversize parts, or very fine pitch parts.

jjpickett:
After some initial research on this, I believe the best approach is to break this down into smaller proportions.

**Step 1:** Make the system have the option to do a fiducial check before placing a part on the board.
**Step 2:** Adjust the system so the fiducial check is only on the individual board before placing a part on it.
**Step 3:** Adjust the system so the fiducial check is on the closest fiducials to the part before placing the part on the board.

This should allow for a smoother transition and allow for any bugs to be managed and fixed much more efficiently.

vonnieda:
Step 1 and Step 2 completed by #327.

vonnieda:
Note that the work done in #327 really just added a way to re-check fids on the entire board. It didn't implement local fid checking. That code has now been removed and this feature request is still open.

PR #327, opened 2016-08-18
Feature - Check Fiducials Before Individual Placement

jjpickett:
This feature allows the user to do a Fiducial Check of an individual board prior to placing it. This is in regards to Issue #319.

In the PlacementTableModel.java there is an extra column added for the Check Fids which has a checkbox element within it.

In the Placement.java there is the Check Fids attribute. This will determine if the placement has requested a Check Fid for the board.

Lastly the ReferencePnpJobProcessor.java adds a method that will override the fiducial locations of the board being called. This method is then called in the doPlace() method just before a boardLocationFiducialOverrides check.

### DIFF
--- a/src/main/java/org/openpnp/gui/tablemodel/PlacementsTableModel.java
+++ b/src/main/java/org/openpnp/gui/tablemodel/PlacementsTableModel.java
@@ -38,11 +38,11 @@ public class PlacementsTableModel extends AbstractTableModel {
     final Configuration configuration;
 
     private String[] columnNames =
-            new String[] {"Id", "Part", "Side", "X", "Y", "ø", "Type", "Status", "Glue"};
+            new String[] {"Id", "Part", "Side", "X", "Y", "ø", "Type", "Status", "Glue", "Check Fids"};
 
     private Class[] columnTypes = new Class[] {PartCellValue.class, Part.class, Side.class,
             LengthCellValue.class, LengthCellValue.class, RotationCellValue.class, Type.class,
-            Status.class, Boolean.class};
+            Status.class, Boolean.class, Boolean.class};
 
     public enum Status {
         Ready,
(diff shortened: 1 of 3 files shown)


vonnieda:
HI @jjpickett, can you please explain in a bit more detail what this particular pull request accomplishes, and how it fits into the plan you described in #319? It's not clear to me if this is intended to be used for local fids, or if it's just causing the fid check to happen before certain placements. I'm confused because it appears that you are still updating the board fid offsets, and not an offset specific to a part.

jjpickett:
Hey @vonnieda, sorry for the confusion.

When I was working through the steps indicated in #319 I was able to do **Step 1** fairly easily. So instead of submitting that, I began working on **Step 2**.

To clear up your statement _"It's not clear to me if this is intended to be used for local fids, or if it's just causing the fid check to happen before certain placements."_  this code now causing a fid check to happen before certain placements. This way if there were any accidental shifting, or lost steps, or mechanical inaccuracies with the machine that may have occurred when the PnP was moving, the board location will be corrected.

You are correct that I am updating the board fid offsets. I ensure to do this before the code checks for the boardLocationFiducialOverrides.
```
if (boardLocationFiducialOverrides.containsKey(boardLocation)) {
                BoardLocation boardLocation2 = new BoardLocation(boardLocation.getBoard());
                boardLocation2.setSide(boardLocation.getSide());
                boardLocation2.setLocation(boardLocationFiducialOverrides.get(boardLocation));
                boardLocation = boardLocation2;
            }
```

This way the new board location gets applied to the part offsets. Once I tested this, I felt that I should submit it before I moved onto **Step 3**.

Please let me know if this is still unclear or if you would like me to make any changes or scrap it until I complete **Step 3**.

vonnieda:
Thanks @jjpickett! For anyone following on #319, this accomplishes @jjpickett's Step 1 and Step 2.

PR #334, opened 2016-08-29
Update Utils2D.java

cri-s:
Offsets center


vonnieda:
Why did you close this?

cri-s:
Because I cannot test it, your probably not too, and so its better to wait
a bit and have the possibility to test it .
If the other code don't work, probably this don't work too.
For the other code you have made the Tests, you don't have tested the
Testcase where I have made all the calcs and for the Testcase you have
tested I'm not able to get out the values from image without PC. Further
the numerical result was not given.

Il lunedì 29 agosto 2016, Jason von Nieda <notifications@github.com> ha
scritto:

Issue #244, opened 2016-03-08
Signaling when job complete or error


dandumit:
Hi,
~~Would be nice to have option to move placement head to a certain position when job complete~~ and also to signal machine status acustic and optic : 
[http://www.tme.eu/gb/details/ks-ad3dczp125ryg/signalling-columns/w2/ks-ad3d_cz_p_125rygm/](http://www.tme.eu/gb/details/ks-ad3dczp125ryg/signalling-columns/w2/ks-ad3d_cz_p_125rygm/)

Br,
Daniel

vonnieda:
The first part of this is a duplicate of #76. I've modified the issue to only include the unique part.

vonnieda:
This should probably be done with a boolean actuator. Let the user set an actuator name to trigger at job end. This makes it easy to add without driver modifications.

pfried:
I actually took a look at the code and wondered if i could contribute somewhere as this seems to be a reasonable starting point.

I would possibly have introduced a Signaler Class (and an implementation as a SoundSignaler or a driver output), because the Actuator interface implements the HeadMountable Interface and also seems no good fit to me. This would provide the possiblity to signal things like:

- Feeder close to empty
- Errors (like machine stop, etc.)
- Machine running and placing

But the major blocker i see here is that there is no central "error management". If an error occurs there is no way right now to detect the problem. The only thing that happens is a log entry which is helpful for  human, but no code-wise (Maybe it would make sense to introduce error codes or something similar).

I think it would be reasonable to extend the MachineListener Interface and introduce something similar to the `fireMachineHeadActivity` but for errors or at least a `machineError` function. Of course there are many places where one would need to insert it.

Any ideas?

vonnieda:
During a job run, all errors bubble up to the JobProcessor and are eventually passed on to the UI from there. The JobProcessor would be a good central place to handle this.

I'd be okay with a Signaler class so that we could have something like a SoundSignaler, but for anything related to communicating with the outside world, it should just hand it off to an Actuator. The Actuator interface is designed for handling this kind of stuff, and I don't think this feature is worth a Driver modification which will break all existing drivers.

So, there could be a Signaler interface with a variety of potential events, and there could be SoundSignaler where you configure different sounds to play for the events, or ActuatorSignaler where you configure Actuator names to handle the events.

pfried:
Ok i see, so i would catch the exception within the ReferenceJobProcessor where it calls next on the fsm and rethrow the exception?

[...]

PR #340, opened 2016-09-03
Introducing a Signaler interface and a SoundSignaler Implementation releated to #244

pfried:
This is a first idea of a Signaling interface. A `Signaler` should signal certain States of the Machine or the JobProcessor like errors or job finish.

Note: The is no machine error yet. The implementation does signal a job finish and a job error only right now. Befor i do the grunt work and implementing an `ActuatorSignaler` i wanted to first show a first draft of the Signaler interface.

Signaler has two  methods : `signalMachineState`, `signalJobProcessorState`

Added a State in the `AbstractJobProcessor` and in the `AbstractMachine` to signal states to the Signalers

One Problem here: The machine property is not available in the `AbstractJobProcessor`, so we need to call it from the Implementations with the signalers.

`SoundSignaler` plays .wav files (Files from https://www.freesound.org under Creative Commons, https://creativecommons.org/licenses/by/3.0/)

Note: Plays sound on Tests (how to disable this)?


pfried:
related #244

vonnieda:
@pfried I think this looks really good. A few small notes:

* WRT your comment "The machine property is not available in the AbstractJobProcessor, so we need to call it from the Implementations with the signalers.": You can always get the machine from Configuration.get().getMachine(), but the way you've done this is fine too.
* Can you create a short Wiki page explaining how to use and configure this feature? It doesn't need to be in depth. Basically just something I can link to people who ask about the feature. I suspect the two biggest questions will be "How do I disable the sounds?" and "How do I change the sounds that get played?". Page can go here: https://github.com/openpnp/openpnp/wiki/Signalers

Thanks!

pfried:
I actually tried to make the path configurable, but i kind of failed since accessing a path seems not be as easy as accessing a resource file across different OS. Do you know an easy way to do this? In the installation (packaged installation) there is no more resource folder, so it cannot be replaced there.

I also thought about removing the sound files completly to avoid licensing / attribution issues.

I created the Wiki Article altough there is not much to it.

tracklogic:
I like this... i need to enable this functionality with the feeder slot interface (other pull request), so can trigger an alert when a feeder needs changing - although as this is a feeder exception, it will probably all be magically caught and triggered

I've fancied putting one of those fancy traffic lights on my pnp like the bigger machines have, these normally have a built in buzzer, so having a signal interface that can trigger an output would be cool - probably via a gcode command that would enable a FET/relay to driver the buzzer.

vonnieda:
Take a look at ImageCamera.initialize and ImageCamera.sourceUri:
https://github.com/openpnp/openpnp/blob/develop/src/main/java/org/openpnp/machine/reference/camera/ImageCamera.java#L151

These show how to allow the user to set a classpath resource, Internet URI
or local File as a loadable resource. Should be able to use the same idea
here.

Jason

vonnieda:
@pfried Thanks for creating the Wiki! Please consider making the audio file paths configurable as I mentioned in the comment above, but for now this looks great and I will go ahead and merge it. Thanks for the contribution!

pfried:
I will, when i find some time i will make them configurable and i will also provide a ActuatorSignaler, so we can have the initial idea with external lights and horns

Issue #288, opened 2016-06-13
Logging Window or Tab


vonnieda:
Suggestion from @TomKeddie:

We should have a logging window or tab that can be shown or hidden from the view menu. This will be very helpful for users setting up new machines and helping with debugging. It will also be very important for the scripting feature #282.
- [x] Required: Add a new Log tab to the main tabs interface which shows the output from log4j.
- [x] Required: Make sure all UI settings in this feature are saved to preferences.
- [x] Required: Show stdout and stderr in log tab.
- [x] Nice to have: Add checkboxes or a dropdown menu in the tab to turn on and off stdout, stderr and common logging situations, e.g. driver, vision, gcode, etc.
- [x] Nice to have: Add a search box.
- [ ] Nice to have: Detach the tab to a floating window so the user can see the log messages while working in other tabs.
- [x] Nice to have: Copy to clipboard/paste buffer.
- [ ] Minor Improvement: Change Auto Scroll button to a checkbox. Current button is confusing.
- [x] System output is getting extraneous newlines. See comment: https://github.com/openpnp/openpnp/issues/288#issuecomment-287076122

See #333 for potential changes to logging to make this work better.

TomKeddie:
Nice to have: save to file.
Nice to have: copy to clipboard/paste buffer.

vonnieda:
First pass at this is in.

dzach:
Provide a way to clear the log widget (not the log file) and show only new entries from that point in time on. Could be a filter for time (i.e. filter out entries earlier than hh:mm:ss) with a button to set time to "now"

pfried:
I already fiddled around with the logging on the weekend for some hours, but i had no luck in redirecting the output stream (stdout, stderr) to the Logger and keeping the output on the console. Can anyone give a hint on this or has a link to a working implementation of such a logger redirect?

vonnieda:
@pfried 
Pseudocode:
PrintStream originalOut = System.out;
LoggingPrintStream out = new LoggingPrintStream(originalOut);
System.setOut(out);

The basic idea is to make a new implementation of PrintStream that sends it's output both to the original System.out and to the Logger. The same can be done for err with setErr.

pfried:
okay, same approach i had, but no luck. I cant remember the problem correctly since it was already some weeks ago. I will try again.

pfried:
just wanted to leave a quick working notice

I started working on the features above, some things to note:

[...]

PR #341, opened 2016-09-04
Saving UI Settings to preferences

pfried:
Saving log panel settings to preferences, related to #288 

Maybe the Hashmap solution for the line limit is not ideal, but adding an extra enum class maybe too much overhead.


vonnieda:
Thanks @pfried!

PR #344, opened 2016-09-04
Feature slot feeder

vonnieda:
@tracklogic Creating this as a pull request so we can discuss the changes.


src/main/java/org/openpnp/gui/FeedersPanel.java
```diff
@@ -166,6 +168,8 @@ public void propertyChange(PropertyChangeEvent evt) {
         feederSelectedActionGroup = new ActionGroup(deleteFeederAction, feedFeederAction,
                 pickFeederAction, moveCameraToPickLocation, moveToolToPickLocation);
 
+        JComboBox comboBox = new JComboBox();
```
vonnieda: This appears to be not used. Perhaps a relic from a previous iteration?

src/main/java/org/openpnp/gui/support/FeedersComboBoxModel.java
```diff
@@ -0,0 +1,53 @@
+/*
```
vonnieda: I suspect this class is no longer used in this iteration. Can it be removed?

src/main/java/org/openpnp/gui/tablemodel/FeedersTableModel.java
```diff
@@ -103,6 +116,9 @@ public Object getValueAt(int row, int col) {
             case 1:
                 return feeders.get(row).getClass().getSimpleName();
             case 2: {
+                if(feeders.get(row).getClass().getSimpleName().compareTo("ReferenceFeederSlot")==0)
```
vonnieda: We can't have implementation specific code in the generic UI. We're going to have to come up with another way to do this. Maybe an API change. Will comment further after the code review.

src/main/java/org/openpnp/gui/tablemodel/FeedersTableModel.java
```diff
@@ -111,8 +127,27 @@ public Object getValueAt(int row, int col) {
             }
             case 3:
                 return feeders.get(row).isEnabled();
+            case 4: {
+                if(feeders.get(row).getClass().getSimpleName().compareTo("ReferenceFeederSlot")!=0)
```
vonnieda: Just noting, more implementation specific code.

src/main/java/org/openpnp/spi/base/AbstractFeeder.java
```diff
@@ -28,15 +28,10 @@
 
     protected Part part;
 
-    public AbstractFeeder() {
+    public AbstractFeeder()
+    {
         this.id = Configuration.createId();
         this.name = getClass().getSimpleName();
-        Configuration.get().addListener(new ConfigurationListener.Adapter() {
```
vonnieda: Why was this removed? Won't this cause the part to be null after config loads?

tracklogic:
@vonnieda

vonnieda:
All ready to go @tracklogic? I'll look at this this evening and get back to you. About to run out the door.

tracklogic:
Hi Jason
No problem, it still needs some work but i think the overall class structure is about there?

Thanks

Matt

vonnieda:
Hi @tracklogic,

Overall, I think this is looking good. There are a few things I'd like to see a bit different, but I'd be okay with merging this one you feel good about it and making those changes myself. Most of them are just object hierarchy stuff that I am kind of nitpicky about:

* Don't change ReferenceFeeder to extend AbstractSlottedFeeder. Instead, create a new ReferenceSlottedFeeder which extends AbstractSlottedFeeder. 
* Don't create a new ReferenceAutoSlottableFeeder. Instead, change ReferenceAutoFeeder to extend ReferenceSlottedFeeder. The idea here is that ReferenceAutoFeeder can be used with a slot, or without. Conversely, a ReferenceStripFeeder continues to extend ReferenceFeeder because it cannot be used with a slot.
* Push the general slot functionality into ReferenceSlottedFeeder. This would be the stuff like finding a slot, checking if it's in a slot, etc.
* The feeder implementation cannot prompt the user in feed(). Instead it should throw an Exception that suggests what the user can do. For instance, the Exception message can be "Feeder X is required but not mounted. Please mount it in slot N.". The user will be prompted by the JobProcessor and they can resolve it how they want. When they hit continue it will try again. This keeps the UI separate from the base PnP functionality, which is important for testing and for future UI work.

So, the object hierarchy would look like this:

* SlottedFeeder extends Feeder: A feeder implementation can implement Feeder or SlottedFeeder. Implementing SlottedFeeder implies that the feeder implementation has slot functionality.
* AbstractFeeder implements Feeder: This is how the code works before your changes. This should stay the same. AbstractFeeder is just a handy base class.
* ReferenceFeeder extends AbstractFeeder: This is how the code works before your changes. Adds the "Reference" stuff needed by all reference feeders.
* ReferenceSlottedFeeder extends ReferenceFeeder implements SlottedFeeder: ReferenceSlottedFeeder becomes the base class for all reference feeders with slot functionality.
* ReferenceAutoFeeder extends ReferenceSlottedFeeder: The auto feeder has slot functionality, so it extends ReferenceSlottedFeeder instead of ReferenceFeeder.

What this boils down to is that a reference feeder can extends ReferenceFeeder if it has no slot functionality, or ReferenceSlottedFeeder if it does. Since ReferenceAutoFeeder is the only one that currently does (or will) then it should extend ReferenceSlottedFeeder.

The reason for the hierarchy working like this is that we don't want the slotted stuff to pollute the non-slotted feeder classes. This reduces our overall implementation complexity.


I know this sounds like a lot - but what you have is actually quite close to this. It's just a lot to explain it in text :) If you'd like, I'd be happy to make these changes and push it up to a branch for you to look at. Or, as I said, I can do the changes after the fact.

Thanks for all your work on this - it's looking really good and I think this will be a killer feature!

tracklogic:
Hi Jason

I've made the changes you suggested and it all fits together better, thank you.

I've got a problem though with 
   -> The feeder implementation cannot prompt the user in feed(). Instead it should throw an Exception that suggests what the user can do. For instance, the Exception message can be "Feeder X is required but not mounted. Please mount it in slot N.". The user will be prompted by the JobProcessor and they can resolve it how they want. When they hit continue it will try again. This keeps the UI separate from the base PnP functionality, which is important for testing and for future UI work.

Currently if a feeder raises an exception it is swallowed up - it does the logger.debug() and then sets the feeder to disabled. This isn't what I'm after for this case... i want to prompt the user by raising an exception in feed() - telling them "to put feeder foo into slot bar", after they hit ok, it should try and do a feed again. 

I think we should probably notify the user of any exception caused by any feeder anyway rather than just logging it away and disabling the feeder? But maybe you've done this for a reason :D I almost started thinking maybe we should have an exception type returned - so anything critical the plan class can just blow up - but minor things like misfeeds it could retry or something? 

```java
while (true) {
// Find a compatible, enabled feeder
Feeder feeder = findFeeder(machine, part);
plannedPlacement.feeder = feeder;
// Feed the part
try {
// Try to feed the part. If it fails, retry the specified number of times
// before
// giving up.
retry(1 + feeder.getRetryCount(), () -> {
fireTextStatus("Feeding %s from %s for %s.", part.getId(),
feeder.getName(), placement.getId());
logger.debug("Attempt Feed {} from {} with {}.",
new Object[] {part, feeder, nozzle});
feeder.feed(nozzle);
logger.debug("Fed {} from {} with {}.",
new Object[] {part, feeder, nozzle});
});
break;
}
catch (Exception e) {
logger.debug("Feed {} from {} with {} failed!",
new Object[] {part, feeder, nozzle});
// If the feed fails, disable the feeder and continue. If there are no
// more valid feeders the findFeeder() call above will throw and exit the
// loop.
feeder.setEnabled(false);
}
}
```

Thanks

Matt

vonnieda:
Matt,

It looks like I might have not finished a thought while writing that code.
The intent was that if no working feeder was found after the correct number
of tries then the original exception would be thrown, letting the user know
the reason for the failure. I'm not quite sure what happened there, and
will need to go back and see if I had this right previously and messed it
up, or if it just got lost in the work. In any case, one way or another,
we'll address the issue of the Exception being swallowed so that the
correct error is shown to the user.

The basic idea for feed will be:
1. Find a compatible/enabled feeder.
2. Attempt to feed, retrying up to retryCount times.
3. If all retries failed, save the final exception that was thrown and
disable the feeder.
4. If there is another compatible/enabled feeder, loop to 2.
5. Otherwise, throw the saved exception, notifying the user.

I should have some time tonight to look at the PR and to look at this feed
bug.

Jason

tracklogic:
Hi Jason

I've been giving this some more thought 

So say we've 3x refautofeeders 10k,4k7,1k and two slots defined with 10k (slot 1) and 1k (slot 2) installed

Placement 10k,4k7 then 1k

10k feeder feed gets called, it's in a slot, so feed succeeds and pickup location of slot 1 - all good

4k7 feeder feed called next, no slot, throw exception to user telling them to put in a slot, set retry count to 1 so it doesn't keep retrying? (Pointless as not in a slot and won't be until user moves it). User gets prompted with exception message and clicks continue once they've physically installed the feeder. We retry the feed  - but how's the feeder slot been updated? Do we let the user sort it on the gui? Not sure if the exception message is modal so they can't click anything behind it to change ? It would be nice to update it for the user - but as we've moved to the exception method of prompting we don't know what the suggested slot was unless I store it somewhere ? Obviously once it's in the slot when we retry it will be successful 

1k feed happens it's still in a slot, great, so feed them pick up, all good

Thanks :)

Matt

vonnieda:
@tracklogic This isn't quite how it would go down. What would happen is:

1. 10k feeder is in a slot, thus it is enabled, so feed gets called, all is well.
2. 4k7 feeder is not in a slot, so it is not enabled. Processor throws an error saying "No feeders found for Part 4k7, please fix." At this point it would be up to the user to figure out that they need to slot the feeder.

If you remember, I had suggested that the SlottedFeeder implementation override isEnabled() and return false if the feeder requires a slot but is not in one at the moment.

Now, I realize this does not quite meet the goals you wanted, which is the ability to guide the user through the process.

So, one way we can change it is that ignoring my suggestion about enabled. Then we end up with a process more like you described:

1. 10k feeder is slotted, feeds and is good.
2. 4k7 feeder is not in a slot, but is marked enabled.
3. We attempt to feed the feeder, but since it's not in a slot, it throws the error "Feeder X is not in a slot. Please mount in slot N."
4. User clicks "Pause Job" and slots the feeder. They set that slot in the feeder config.
5. User clicks the play button to continue the job. We try to feed the feeder again and this time it works because it's slotted.

The retry count doesn't really play into this because we re-enter the retry loop after the Pause / Continue. This all works because the first time we try to feed it fails, and the next time we try it works because it has been slotted.

The user *will* have to remember which slot to put it in between clearing the error and doing the work. There's not a great way to work around that right now, except for maybe adding a "Suggested Slot" field on the feeder config. You could remember the suggestion and show it there.

Does this make sense, and does it work for you?

tracklogic:
Hi Jason

I think the bottom solution works best for me, if your happy with that? Did you manage to have a think about that exception handling problem for when a feeder raises an exception? - i think the problem is in the retry() function - this seems to swallow up the exception retryCount times before raising it, but then it gets swallowed up again in the jobPlan class by just doing a logDebug() and setting the feeder to disabled.

I did have a quick look at a few past revisions of the classes but couldn't spot an iteration that was different to this.

Thanks again

Matt

vonnieda:
Matt, I went through the code and I think that I just thought about doing
it this way and then didn't do it, for whatever reason. I wrote that code
in a huge rush leading up to Maker Faire and just dropped the ball. So, go
ahead with your implementation as we've discussed and I will get this bug
fixed. It will work as I described above - we'll catch the last exception
thrown during retry and then if there are no more feeders available we'll
throw it. So the message from the feeder will get shown to the user. It
will probably be sometime next week when I get to this - about to start the
final leg of my move which will take until Tuesday.

Jason

vonnieda:
@tracklogic The feeder bug #398 is now fixed and committed. I'd like to work on getting this merged in now. Do you have any updates to it, or is this the most current code?

vonnieda:
Note: Closing this because it is too outdated to be merged. Matt doesn't have time at the moment to work on this so I am going to make an attempt at it with the ideas here in mind.

vonnieda:
Work being done in #399

PR #345, opened 2016-09-07
Bugfix fid home

tracklogic:
Fixing the issues reported - note, i've only managed to simulate this code as my pnp hardware is currently in bits.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
@@ -52,6 +52,7 @@ public class GcodeDriver extends AbstractSerialPortDriver implements Runnable {
         CONNECT_COMMAND,
         ENABLE_COMMAND,
         DISABLE_COMMAND,
+        POST_VISION_HOME_COMMAND,
         HOME_COMMAND("Id", "Name"),
         PUMP_ON_COMMAND,
         PUMP_OFF_COMMAND,
@@ -366,10 +367,28 @@ public class GcodeDriver extends AbstractSerialPortDriver implements Runnable {
                     .getHomeFiducialLocation(tmp, homePart);
 
             // homeOffset contains the offset, but we are not really concerned with that,
-            // we just reset everything back to 0 at this point.
-            for (Axis axis : axes) {
-                axis.setCoordinate(axis.getHomeCoordinate());
+            // we just reset X,Y back to the home-coordinate at this point.
+            double xHomeCoordinate=0;
+            double yHomeCoordinate=0;
+            for (Axis axis : axes)
+            {
+                if(axis.getType() == Axis.Type.X)
+                {
+                    axis.setCoordinate(axis.getHomeCoordinate());
+                    xHomeCoordinate = axis.getHomeCoordinate();
+                }
+                if(axis.getType() == Axis.Type.Y)
+                {
+                    axis.setCoordinate(axis.getHomeCoordinate());
+                    yHomeCoordinate = axis.getHomeCoordinate();
+                }
             }
+
+            String g92command = getCommand(null, CommandType.POST_VISION_HOME_COMMAND);
+            g92command = substituteVariable(g92command, "X", xHomeCoordinate);
+            g92command = substituteVariable(g92command, "Y", yHomeCoordinate);
+            sendGcode(g92command, -1);
+
         }
     }
 
--- a/src/main/java/org/openpnp/machine/reference/driver/NullDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/NullDriver.java
@@ -21,6 +21,7 @@ package org.openpnp.machine.reference.driver;
 
 import java.io.IOException;
 import java.util.HashMap;
+import java.util.Locale;
 
 import javax.swing.Action;
 import javax.swing.Icon;
@@ -37,6 +38,8 @@ import org.openpnp.machine.reference.ReferencePasteDispenser;
 import org.openpnp.model.Configuration;
 import org.openpnp.model.LengthUnit;
 import org.openpnp.model.Location;
+import org.openpnp.model.Part;
+import org.openpnp.spi.Camera;
 import org.openpnp.spi.Head;
 import org.openpnp.spi.PropertySheetHolder;
 import org.simpleframework.xml.Attribute;
@@ -58,7 +61,7 @@ public class NullDriver implements ReferenceDriver {
 
     private boolean enabled;
 
-    /**
+ /**
      * Gets the Location object being tracked for a specific Head. This is the absolute coordinates
      * of a virtual Head on the machine.
      * 
@@ -75,6 +78,7 @@ public class NullDriver implements ReferenceDriver {
     }
 
     protected void setHeadLocation(Head head, Location l) {
+
         headLocations.put(head, l);
     }
 
@@ -83,6 +87,28 @@ public class NullDriver implements ReferenceDriver {
         logger.debug("home()");
         checkEnabled();
         setHeadLocation(head, getHeadLocation(head).derive(0.0, 0.0, 0.0, 0.0));
+
+             /*
+         * The head camera for nozzle-1 should now be (if everything has homed correctly) directly
+         * above the homing pin in the machine bed, use the head camera scan for this and make sure
+         * this is exactly central - otherwise we move the camera until it is, and then reset all
+         * the axis back to 0,0,0,0 as this is calibrated home.
+         */
+        Part homePart = Configuration.get().getPart("FIDUCIAL-HOME");
+        if (homePart != null) {
+            Location tmp = new Location(LengthUnit.Millimeters, 0.0, 0.0, 0.0, 0.0);
+            Camera camera = Configuration.get().getMachine().getDefaultHead().getDefaultCamera();
+            // camera.moveTo(tmp);
+            Location homeOffset = Configuration.get().getMachine().getFiducialLocator()
+                    .getHomeFiducialLocation(tmp, homePart);
+
+            // homeOffset contains the offset, but we are not really concerned with that,
+            // we just reset X,Y back to the home-coordinate at this point.
+
+            logger.debug(String.format(Locale.US, "%f,%f", homeOffset.getX(), homeOffset.getY()));
+
+
+        }
     }
 
     /**
@@ -105,6 +131,7 @@ public class NullDriver implements ReferenceDriver {
             throws Exception {
         logger.debug("moveTo({}, {}, {})", hm, location, speed);
         checkEnabled();
+
 
         // Subtract the offsets from the incoming Location. This converts the
         // offset coordinates to driver / absolute coordinates.
--- a/src/main/java/org/openpnp/machine/reference/driver/NullDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/NullDriver.java
@@ -21,7 +21,6 @@ package org.openpnp.machine.reference.driver;
 
 import java.io.IOException;
 import java.util.HashMap;
-import java.util.Locale;
 
 import javax.swing.Action;
 import javax.swing.Icon;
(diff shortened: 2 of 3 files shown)


cri-s:
Having homing inside null driver don't make sense.

cri-s:
Homing dot

tracklogic:
Hi Cri

It does when you want to look for a homing dot on the simulated camera
screen and test that it picks it up...

Matt

cri-s:
Usage case for null driver is loading exported 600 DPI Gerber file or
scanned PCB image that usually have fiducial near zero position and same
part file as real machine.
It's used to check and tweak positions of centeroid files and part rotation
and to setup missed parts doing offline setup.

tracklogic:
Ah ok... i figured it was just a sample lets play with something type setup
which is useful when your away from the machine? I can remove from the null
if its causing headaches though

tracklogic:
Actually on that note... do you the process of how Jason made the original
png that gets loaded in?

vonnieda:
I just export the board from Eagle as a PNG, at 600 DPI.

Jason

vonnieda:
@tracklogic I think the fixes look good, but I agree with @cri-s that we probably don't want this in the NullDriver. And in any case, unless I am mistaken, it looks like the NullDriver code doesn't actually do anything with the results? If you can remove that part we can merge this.

cri-s:
Eventually the code should go to moveutils without g92 part in order it can
be effectively shared with several drivers and named homingDot(Location loc)
Further homing dot can be difficult at 0 position using machine coordinate,
because any move over zero triggers soft/hardlimit, even maybe moving
camera over pos0.
Il 08/set/2016 05:41 "Jason von Nieda" <notifications@github.com> ha
scritto:

cri-s:
Another thing. Special package should have special naming rules, I suggest
lisp style.
Instead homing-dot it should named *homing-dot*  And further homingDot
should check *homing-dot1* and
*homing-dot2* package
and driver should apply scaling factor for thermal expansion. Maybe
homingDot(boolean home) should go inside AbstractMachine.java returning
current absolute position and all driver should be updated for using scale
factor with default value 1000000000 for x and y coordinate on x/y delta
only.
Scale factor is real nm-per-mt travel and correction need avoid doubles.
Even that could go into machine , double scale( double diff) .  HomingDot
should store last homingDot time in order job processor call it on start
and if homing-dot-interval is passed from last homing, it's repeated. This
if there is need for precision related to homing dot.

tracklogic:
Hi Cri

Do we need to worry about thermal expansion? How much difference will it
actually make

Matt

cri-s:
6 degree changes 0.1 mm.
As said, it depend what minimal component you do, if 0603 and 0.5mm pitch
is what you do, don't worry, 3 recalibration/year is enought.
The error depend further on camera position related to homing (dot) as if
there are distant,
Up looking can could multiplicate that error, especially on rotation.

tracklogic:
@vonnieda  - hi jason i've removed from null driver - thanks Matt

vonnieda:
Merged, thanks @tracklogic!

tracklogic:
Hi Cri

Just so I'm following what you mean - what material changes 0.1mm with a 6
degree change?

I'm guessing what your planning is say you have a fid. footprint defined as
6mm, then you'd define 2x more 6.1mm and a 5.9mm one - when the software
finds the homing fid, it finds which of the three are the closest fit. If
its say the 6.1mm one, then from what your saying everythings likely to be
0.1mm larger - so feeders etc.. will be 0.1mm away from home

Matt

cri-s:
I'm guessing what your planning is say you have a fid.


> 3x fid

Can be the same, example 1x homing and up looking camera.

I use hough circle instead of template match and shapes.

In this case fid0 is same as fid1 and fid2 is up looking camera lenses.



> Fid0 is for homing. Fid1 and fid2 are distant as example 770mm

at setup and if there are found at 770.6mm , scaling is uptaded to
999221386 .
And driver multiply x and y with 0.999...
If driver want move 200mm it send 199.844mm on CNC driver without modify
steps/mm setting, but use 200
to actualize coordinate.



>

> footprint defined as
> 6mm, then you'd define 2x more 6.1mm and a 5.9mm one - when the software
> finds the homing fid, it finds which of the three are the closest fit. If
> its say the 6.1mm one, then from what your saying everythings likely to be
> 0.1mm larger - so feeders etc.. will be 0.1mm away from home
>
> Matt
>
> On Thu, Sep 8, 2016 at 5:31 PM, cri-s <notifications@github.com

cri-s:
Sorry, in the previous example, distance at setup must be 770.6 and at
check 770mm .


Il venerdì 9 settembre 2016, Cri S <phone.cri@gmail.com> ha scritto:

>
>
> I'm guessing what your planning is say you have a fid.
>
>
>> 3x fid
>
> Can be the same, example 1x homing and up looking camera.
>
> I use hough circle instead of template match and shapes.
>
> In this case fid0 is same as fid1 and fid2 is up looking camera lenses.
>
>
>
>> Fid0 is for homing. Fid1 and fid2 are distant as example 770mm
>
> at setup and if there are found at 770.6mm , scaling is uptaded to
> 999221386 .
> And driver multiply x and y with 0.999...
> If driver want move 200mm it send 199.844mm on CNC driver without modify
> steps/mm setting, but use 200
> to actualize coordinate.
>
>
>
>>
>
>> footprint defined as
>> 6mm, then you'd define 2x more 6.1mm and a 5.9mm one - when the software
>> finds the homing fid, it finds which of the three are the closest fit. If
>> its say the 6.1mm one, then from what your saying everythings likely to be
>> 0.1mm larger - so feeders etc.. will be 0.1mm away from home
>>
>> Matt
>>

tracklogic:
I suppose it depends if everything is going to linearly expand at the same
rate.

What about putting fiducials around the machine? So one next to the feeder
bank, one in each corner - and whenever it needs to do something it goes to
the nearest fid. My thoughts being, if its close enough to a fid, then any
error due to expansion will be tiny.

issus:
Aluminium will expand 0.0041mm with a 6c temperature rise, if you are
working over a 300mm length (~12"). How your machine is constructed however
will determine how much this really affects you. Most people would just see
the tiniest change in belt tension, as they have wood beds with (or
without) steel tops. The extrusion itself changing in length is irrelevant
mechanically - whether you have steel rail or you use roller bearings on
the extrusion, the change in the length of the extrusion has no impact on
steps per mm - you're just moving your pulleys at each end closer or
further apart, hence the minute change in belt tension. The thermal
expansion of wood is about 8 times lower than aluminium, and steel about
half of aluminium.

FR4 is pretty similar to that of steel and copper.

If we were using ballscrews or rack and pinion, then we can start talking
thermal expansion, but not so much for belts. Actually, your belt tension
will slacken.. as rubber (unreinforced) will expand 0.14mm over the same
30cm distance with a 6c rise in temperature. Again, I don't really see this
affecting your accuracy overly much as the pulley doesn't change in
diameter very much at all (0.00014% on a 16mm pitch diameter.)

cri-s:
Supposing 40x65cm movement area.
Further homing located back right and homing dot front left , 5cm from
front and 5mm from left with up looking camera near that position.
At 6 degree temperature change the homing dot moves 0.1mm, mathematically
0.099...
mm . The 32 mm circumference pulley is already subtracted from the thermal
expansion of supporting structures because step with increase  reduce
thermal expansion error. Wood, you are right, it have 8x lower expansion in
one direction,  the other direction present more expansion as aluminium.
MDF have equal expansion on all sides and its a bit more as half of that
from aluminium. Normally structured aluminium elements are used to build
the type of PNP used for openpnp.
Maybe you own RedFrog PNP. If you have converted the design from Vpapanik
low budged manual p&p to openPnP I'm highly interested to see it, otherwise
I'm no see the point about wood.

issus:
This is getting pretty offtopic... but anyway.

As I said in my last post, the size and change of the aluminium frame has nothing to do with the steps/mm. You wouldn't recalculate your steps per mm because you went from a 400mm rail to a 600mm rail. The frame material for anything other than a rack and pinion or ballscrew (at which point only the rack and pinion or ball screw material should be considered) matters. The frame is just the surface you're moving on.

Pretty much everyone seems to use a wood bed, some with steel on top for its magnetic properties. Anthony and all building machines like him, liteplacer, jason's reference design. The only machines I havent seen with a wood bed are converted commercial machines.

The only thing that affects your steps per mm in a belt system is the pulley diameter. The belt changing temperature will change your backlash. 

The surface the parts are sitting on matter, as the position of the board relative to other components will change - however we have vision systems to find our parts within the strip feeders every pick, and the board has fiducials. If you have a temperature rise that is high enough to cause inaccuracies in placements across a single pcb from the time the fiducials are checked to when the board is finished being populated, I can only assume your house/office is on fire?

The worst case scenario I can think of is at my cities local hackerspace in winter, even in Australia the diurnal temperature swing in the midst of summer isnt likely to cause as much temperature change so quickly. At the hackerspace they set the heating to 10c overnight to keep the heating cost down, then the first person in will change the temperature to about 22c after they arrive. It takes around 1.5hrs for the space to reach that temperature if the outside temperature isn't much cooler than 0c. So we assume the machine has completely cooled to 10c overnight, and the person starts to use it right after they have turned up the heat. If we assume they have a very slow pick and place, or a very large board, it takes them 30mins to fully populate the board. Lets say they had a very large board, 400mm x 400mm, the copper/mdf/steel/fr4 are all going to be pretty similar in expansion, with a 4c temperature rise over just 30 mins you'll see a deviation from one corner of the board to the other of 0.02mm. This is still more accurate than I can hand place, and I do 0.4mm pitch parts and 0402s every week. BUT! We're talking about a shift across the entire board, so really we're looking at just 0.01mm average per part, and thats a HUGE board. On the same size board, to get the point you're having issues with a 0.4mm pitch part, or 0201 on a IPC standard pad (medium density) you're going to have to see a 100c/hr temperature rise, assuming it takes you 30mins to place the board.

Thread successfully derailed.

mlyle:
> The only thing that affects your steps per mm in a belt system is the pulley diameter. The belt changing temperature will change your backlash.

I'm confused by this.  Aren't people using timing-type belts?  So the pulley diameter shouldn't matter if those rungs engage at the same place-- there's still the same number of belt-rungs per revolution.

tracklogic:
I think reading all of the below makes me think that theres no need for any more complication in the fid. home code - i’m thinking lets get a few machines running on it as it stands, and we can always do a v2 version if we need uber precision?

issus:
Agreed @tracklogic.

The pulley diameter for the drive one does matter, the fact the belts are
toothed just means they don't slip - the pulley is a wheel, and the pitch
diameter of the wheel (ie: Pi*D = circumference) determines how far the
belt moves in linear terms per revolution :) This is why we try to use the
smallest pulleys possible, so we can have the least movement per rotation
possible (ie: most steps/mm).

mlyle:
> The pulley diameter for the drive one does matter, the fact the belts are
toothed just means they don't slip - the pulley is a wheel, and the pitch
diameter of the wheel (ie: Pi*D = circumference) determines how far the
belt moves in linear terms per revolution

No, the belt moves an integral number of teeth per revolution, and the belt stretches while on the wheel to conform.  This ratio will remain constant as long as no teeth are skipped.

mlyle:
E.g. if the wheel has 20 teeth and is 20mm in circumference nominally.. and the belt has teeth at perfect 1mm intervals.. what happens when I rotate the wheel 5000 times to go a km-- it advances 100,000 teeth.   Now assume it expands to 20.01mm.

If I rotate it 5000 times, you'd expect it to go 1.0005 km--  are you expecting it to skip 5 teeth somewhere?  Does tension modulate and get tighter and tighter until we skip one?

I'd expect it to go 1 kilometer, not skipping any teeth.

tracklogic:
It might be a poor analogy but i guess a cambelt / timing belt on a car
engine might be a good way of thinking about this.

The belt stretches when you apply load (i.e its idling and you floor the
throttle), and stretches as it ages, yet you still always maintain the
timing of crank / camshaft - otherwise you'd end up with nasty valve train
damage?

issus:
Belts are relatively squishy, The circumference of the pulley determines
the distance travelled, period. You didnt lose 5 teeth, in 5000 revolutions
you stretched the "tooth" 0.00005% each time.

mlyle:
@issus Read the source material from the textbook.   Positional variation Delta S7.

mlyle:
@issus Also see this timing belt manufacturer source.

http://www.pfeiferindustries.com/glossary.html#g

> Counting the teeth calculates the exact gear ratio, **regardless of any variations in the diameter measurement.   **
...
> Timing belt pulleys coupled with timing belts have teeth that behave just like meshing gear teeth and exact counting of teeth and revolutions can be applied with these machines.  Chain sprockets coupled with chains work exactly the same way.

PR #350, opened 2016-10-23
Actuator Signaler Class

pfried:
As earlier promised i created an Actuator Signaler
Uses a machine actuator to signal machine or job states like error or run

As is do not have a machine yet i was not able to test it.

Also for some Machine Actuators (like a horn) a Timeout would be nice, but i wonder how to implement that.

Thoughts on this?

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/signaler/ActuatorSignaler.java
+++ b/src/main/java/org/openpnp/machine/reference/signaler/ActuatorSignaler.java
@@ -1,0 +1,96 @@
+package org.openpnp.machine.reference.signaler;
+
+import org.openpnp.ConfigurationListener;
+import org.openpnp.machine.reference.ReferenceMachine;
+import org.openpnp.model.Configuration;
+import org.openpnp.spi.Actuator;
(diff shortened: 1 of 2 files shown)


pfried:
Related #244

vonnieda:
Looks good! Thanks for the nice feature @pfried!

For the timer, check out http://docs.oracle.com/javase/7/docs/api/java/util/Timer.html. You can just set a timer for N seconds when the actuator fires and when the timer expires you can turn it back off. I'd suggest making the timer value configurable in the signaler XML.

pfried:
Thanks, i just noticed that i should not have set default values. Is it possible to set null values in the xml? Otherwise i should remove the default values as this leads to the situation where the actuator always reacts to machine and job errors even if i do want it only for one of them.

vonnieda:
Null values would not be appropriate here. I will remove the defaults and add the new configuration information to the wiki so that people can easily find it.

vonnieda:
Wiki updated and default signalers removed
https://github.com/openpnp/openpnp/wiki/Signalers

PR #361, opened 2016-11-03
Update Placement.java

cri-s:
Added todo
Mod for feeder split, missing is button with rewind that changed placed to place.
Step1

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -676,8 +676,9 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
 
             plannedPlacement.stepComplete = true;
 
+            placement.setType(Placement.Type.Placed);
             Logger.debug("Place {} with {}", part, nozzle.getName());
-        }
+            
 
         clearStepComplete();
     }
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -678,7 +678,7 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
 
             placement.setType(Placement.Type.Placed);
             Logger.debug("Place {} with {}", part, nozzle.getName());
-            
+            }
 
         clearStepComplete();
     }
--- a/src/main/java/org/openpnp/model/Placement.java
+++ b/src/main/java/org/openpnp/model/Placement.java
@@ -35,7 +35,7 @@ import org.simpleframework.xml.core.Persist;
  */
 public class Placement extends AbstractModelObject implements Identifiable {
     public enum Type {
-        Place, Fiducial, Ignore
+        Place, Fiducial, Ignore, Done, Todo, Todo1, Todo2, Todo3, Done1, Done2, Done3, Placed
     }
 
     /**


vonnieda:
@cri-s I'm not sure what this is intended to do. These new enums are not descriptive enough. Please describe the issue and what this does. Also, it looks like you submitted two pull requests that rely on each other. These should be a single pull request so that the code works together and the tests pass.

cri-s:
The button is missing, that can be present instead of stop button, when program is not running.
Rewind checks the placement inside the job and change placement from Placed to Place .
The benefit that this mod brings:
Interrupted jobs can be saved and later resumed without problems.
Job can be breaked down on several batches in order to populate a board that need many feeders and in real less are available, either physically, or that can be placed on PNP.
Menu entry or script can change Place to Todo as change Placed to Done.
It can move at init Done to Done1 and Todo to Todo1  too, ...
This allows to incrementally build board. I don't know if you want use scripts, place this into menu or use mouse right menu.

vonnieda:
@cri-s Okay, I understand now. I think this is a good idea but it needs some adjustments. I don't like the Done1, Done2, Todo1, Todo2 stuff. It's not very descriptive and it will be confusing for users.

I think we should only add Placed. Place already takes care of Todo - if a Placement is set to Place then it is set to be processed. When it finishes we set it to Placed. This will let people save the state of the job and they can continue where they left off.

We can add a new button to the right of Stop called Reset Job, which will change all Placed to Place.

Will this work for you? If not, can you come up with some more descriptive names for the other values? It would be helpful if you would list each value and describe what it means.

Thanks,
Jason

cri-s:
Then you don't have understand or you either have plenty of feeder, space
or only small feeder counts on assembly.
If you need to make 50 boards with 3x16mm feeder and 6x 12mm feeder, 1x48mm
feeder,
1xtray and 26x8mm feeder, but you machine can mount 17x8mm feeder, 1x16mm ,
2x12mm and tray/strip feeder.

With todo/1/2/3 and done/1/2/3 it is no problem splitting the job to 3
batches and assembly it as needed. It is never be possible doing it using
place as if there is a problem, the placement  is skipped and instead
annother placement is used.
Todo and Done are the correct keyword, same as rewind and not reset. Reset
imply zeroes the counter for assembled boards, ...
In this case the 50 boards need to pass 5x the pnp each in order to
assembly it, 3x for top and 2x for bot.

If you just make 2 prototype that is later assembled externally, you have
right, place is enought. If externally assembly need minimum 200 piece to
get competitive pricing and 6week, and doing internally you could do it.
You are right, it is possible to split the assembly as 3 separate jobs.
That is error prone but possible. Todo/done is not only for that. It allows
having as example a job with 3/4  assembly variation or jobs with 3
optional sections placed.

PNP is not easy,  you must understand the processes and tools incl pnp
software.
Same with eagle, altrium, mentor, zuken, ....
You cannot play without traying it out, make some projects for introducing
yourself into sw, reading manuals and asking when you don't know how to
proceed. You could reduce the needed skills with wizards, as you have done,
the drawback is, one loose time and flexibility with wizards. Good example
is the strip feeder wizard as fail example. I know some people love it even
if fails 4 of 5 times.

cri-s:
I'm fine if you add a config inside machine.xml where it's possible to enable additional enums, or define it, or set a flag, to enable additional ones. I suspect the limits is on table, can you show how to do that, or implement this you.

vonnieda:
@cri-s I have grown tired of your rudeness. The last thing I need is a lecture from you. I am closing this pull request and all future ones you submit until you are able to be civil. I have spent hundreds of hours responding to your half written pull requests, bad code and rude emails, but now I am finished. If you wish to participate in this project you need to learn to communicate without being insulting.

PR #364, opened 2016-11-05
Feeder panel layout in line with other panels

dave-prosee:
Changed "Configuration" border into a tab like other panels.
Before: 
feeder panel before
After: 
feeder panel after


vonnieda:
Look good, thanks @dave-prosee!

Issue #172, opened 2015-11-16
Camera wizard (and others) tab title is incorrect in machine setup tab


JacobChrist:
On the Machine Setup Tab in the ReferenceMachine tree under the Heads -> ReferenceHead H1 -> Cameras, if you select a camera the panel to the right has two tabs.  The tab text is in correct.

The tab heading when selecting a camera read "Untitled" and "Untitled".  They should probably read "General Configuration" and "Camera Specific".

JacobChrist:
This is coming from approximately line 169 of this file:
openpnp/src/main/java/org/openpnp/gui/MachineSetupPanel.java

It also affect every panel not just cameras.

vonnieda:
Thanks for the bug report. I'm aware of this and it's something on my list
to fix, but it's good to have an issue to track it anyway.

JacobChrist:
What's your plan on fixing this?  I'm trying to fix it (mostly to understand the code base not because I think it needs to be fix).  If I can understand how you might go about fixing this this I might be able to patch.  From my preliminary looking, it seems as if you need to have each panel implement the PropertySheet class so that the title is accessible when rendering the control.  I though this would be a simple fix three hours ago.

vonnieda:
Yep, that was the basic idea but I have had a few false starts on this. I
actually think this interface is going to get redesigned before this gets
fixed because as it is right now there are some interlocking dependencies
that I hate. So, as of now I don't really have a plan. It's part of the
much bigger task of finishing up the Machine Setup tab which has been in
progress for quite some time.

vonnieda:
See #103

dave-prosee:
@vonnieda : **you are probably aware of this, but I use this as a way to get familiar with the code and appreciate some feedback a.k.a. insight.** 
The configuration wizards for Jpanels don't set the tab title. Tab title is set by e.g. their grandparents, i.e. the CamerasPanel, FeedersPanel or MachineSetupPanel. The first two have code with specific text to do that. The MachineSetupPanel tries to get the title based on the object (type) but that info is not available. Could it be a solution to add a <confWizardTitles> attribute to the machine.xml. At least for now. So we can access titles via the configuration class?

dave-prosee:
Or, like the Gcodedriver, add the title in the getPropertySheets method like for OpenCvCamera:
 @Override
    public PropertySheet[] getPropertySheets() {
        return new PropertySheet[] {
[...]

PR #366, opened 2016-11-05
Configuration tabs within "Machine Setup" panel now with titles (partly #172)

dave-prosee:
Changed for available camera objects but with some remarks:

- ImageCamera, IPCamera, OpenCVCamera, SimulatedUpCamera and Webcams
work fine

- Webcams give however a warning in the log file
  log4j:WARN No appenders could be found for logger
  (com.github.sarxos.webcam.Webcam).
   log4j:WARN Please initialize the log4j system properly.
   log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for
   more info.

- Lticivil cam give, when trying to add on the cameras panel:
Caused by: com.lti.civil.CaptureException:
java.lang.UnsatisfiedLinkError: no civil in java.library.path

- Vfw cam gives, when selecting in the machine setup tree and also after
returning to Cameras panel and selecting it:
The vonnieda.org Java video capture kit is unable to load it's native
layer.
Please make sure you have org_vonnieda_vfw_CaptureDevice.dll in
yourPATH, Windows directory or current directory.
If you do not have this file, you should be able to find the latest
version at http://www.vonnieda.org/software.html.
Load failed due to:
java.lang.UnsatisfiedLinkError: no org_vonnieda_vfw_CaptureDevice in
java.library.path
camera’s are not updated to the machine setup tree, you have to reload
and crashes the app.

The remarks however, I cannot relate to the changes made. Hence still a
commit and PR.

A second commit covers the feeders. I found no problems there.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
@@ -30,10 +30,13 @@ import org.opencv.core.Rect;
 import org.opencv.core.RotatedRect;
 import org.opencv.imgproc.Imgproc;
 import org.openpnp.ConfigurationListener;
+import org.openpnp.gui.support.PropertySheetWizardAdapter;
+import org.openpnp.gui.wizards.CameraConfigurationWizard;
 import org.openpnp.model.Configuration;
 import org.openpnp.model.Length;
 import org.openpnp.model.LengthUnit;
 import org.openpnp.model.Location;
+import org.openpnp.spi.PropertySheetHolder.PropertySheet;
 import org.openpnp.spi.base.AbstractCamera;
 import org.openpnp.util.OpenCvUtils;
 import org.openpnp.vision.LensCalibration;
@@ -384,6 +387,13 @@ public abstract class ReferenceCamera extends AbstractCamera implements Referenc
 
     @Override
     public void close() throws IOException {}
+
+    @Override
+    public PropertySheet[] getPropertySheets() {
+        return new PropertySheet[] {
+                new PropertySheetWizardAdapter(new CameraConfigurationWizard(this), "General Configuration"),
+                new PropertySheetWizardAdapter(getConfigurationWizard(), "Camera Specific")};
+    }
 
     public interface CalibrationCallback {
         public void callback(int progressCurrent, int progressMax, boolean complete);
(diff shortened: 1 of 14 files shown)


vonnieda:
@dave-prosee This looks great, and I will be happy to accept it as is, but I want to mention a possible improvement and see what you think. On all the cameras, this code is duplicated. Since it's all the same I think we could push it into the base class ReferenceCamera and eliminate all the duplicate code. Any class down the road that has specific requirements can just override the method. What do you think?

Again, if you'd like me to just merge it as is for now I'd be happy to. This is already a great improvement over what we have.

dave-prosee:
Thanks for asking, but I am the novice ;-)

I tend to say that better is the enemy of good. Sometimes we postpone to create something better while in the meantime little happens. #172 e.g. is open for almost a year now. You are right wanting things to go up base classes and I am open to assist but really a novice, ...meaning that going for base class will most certainly require, at the moment, some more support from you to me. 

Further more, this specific solution has been copied from the Drivers and so in line with what you were thinking yourself, some time ago.

On the other hand again, this issue is not on its own. The feeders (I used "Configuration" to be compliant with the feeders panel) and remaining objects could do with just "Configuration" so another solution would be to leave Drivers and Cameras as is and like above and for the rest rely on "if title = null then title = "Configuration".

It is your pick I'm afraid ;-) but I make one last statement like you did. Choosing one option does not have to mean you might not do the other, some day.

dave-prosee:
Gave it another thought. If you accept these I will look into changing the base class. It's a nice exercise for me. OK?

dave-prosee:
Couldn't sit still,.. was easier than I thought. Shouldn't be that cautious. 
For Cameras: will referencecamera do? That is one class higher. My feeling says that AbstractCamera should remain abstract and that specific titles do not have a place there.
For feeders we could go to AbstractFeeder (two classes up) because "Configuration" is quite abstract.
Your @vonnieda opinion please.

vonnieda:
@dave-prosee Very nice work. This looks really good. Sorry for the long response time, was out and about. I agree with your choices of classes, and will merge this now.

dave-prosee:
Took the decision to move the generic code for feeders to the AbstractFeeder class.
I think all you wanted regarding the code in the first two commits has been done.
Please giver your view on the tab names for the remaining objects.
By the way, the drivers have to be checked too

dave-prosee:
Ah, messages crossed. Thanks! You're very welcome. Glad to do something in return for this nice piece of software.
Cheers!

PR #371, opened 2016-11-15
Feature post feed

tracklogic:


vonnieda:
I'm good with this feature, but I think the name of the method should be changed to postPick. If I understand how the feature works then this more accurately reflects when it will be called.

tracklogic:
Hi Jason

No problem, i'll rename it your right postPick makes more sense

Matt

cri-s:
The convenience to have feed and preFeed , postFeed is that feed can invoke either pre- or postFeed presenting the operator one unique interface, feed . As ReferenceFeeder preFeed calls feed the old feeder interfaces are still valid.

vonnieda:
@tracklogic Fixed this up as discussed and merged it.

PR #372, opened 2016-11-15
Feature error regex

tracklogic:


vonnieda:
Note to self: After merging this, update the docs at https://github.com/openpnp/openpnp/wiki/GcodeDriver#regular-expressions-receiving-responses.

vonnieda:
@tracklogic I went ahead merged this. I made a few changes:
1. Fixed the typo in raised.
2. Changed the order of the error and timeout checks. Errors were being reported as timeouts due to the ordering. I think this is working correctly now. I tested all functions in my test harness. Let me know if there are any problems.

vonnieda:
Docs updated: https://github.com/openpnp/openpnp/wiki/GcodeDriver#command_error_regex

PR #373, opened 2016-11-15
Feature vacuum detect

tracklogic:


vonnieda:
@tracklogic This build failed. Looks like a wayward logger call. It might be that you didn't pull before pushing this, as it looks like it's using the old logger code. Details: https://travis-ci.org/openpnp/openpnp/builds/175975218

vonnieda:
* Fixed Logger errors.
* Renamed the vacuum variables to use proper capitalization.

vonnieda:
Docs updated for VACUUM_REQUEST_COMMAND and VACUUM_REPORT_REGEX: https://github.com/openpnp/openpnp/wiki/GcodeDriver:-Command-Reference

vonnieda:
@tracklogic Looking at this code again, I think I may misunderstand how it's supposed to be used. You are checking that the value is within the range defined on the NozzleTip for both pick and place, but since it's the same range both times what is it actually checking? Am I mistaken in thinking this was supposed to be used to check for part on and off the nozzle?

tracklogic:
Hi Jason

I was pondering that myself... currently it checks when the part is picked up and then just before its placed to make sure its still on the nozzle after transport alignment etc.. we could also check after the placement to make sure the vacuum has gone away, that would indicate the part has been placed ?

Matt_

vonnieda:
@tracklogic Oh, I see! I completely missed that in place it was checking that it was STILL on the nozzle, not that it had been placed. That makes more sense.

I wonder, though, would it make more sense if we just had two values that are used as checks for part on and part off? For part on the value needs to be >= than the set value, and for part off it needs to be <= the set value?

tracklogic:
Maybe, i guess it depends how much the vacuum differs dependant on whats being picked up, thats why i thought having it on nozzle size kind of makes sense... large ics will have a large nozzle fitted so they can have different vacuum propertys to a small 0402 with a tiny nozzle?

tracklogic:
The Zevatech originally just had a vacuum switch that you set and this convered all parts, but you were forever tweaking it dependant on what your picking up - thats what i swapped (well in the process of) to a vacuum sensor

vonnieda:
@tracklogic Sorry, yes, I still man having it on the nozzle tip, but instead of vacuumLevelMin and vacuumLevelMax we would have vacuumLevelPartOn and vacuumLevelPartOff. The logic would be something like:

pick(NozzleTip nt) {
  doPick();
  vacuumLevel = readVacuum();
  if (vacuumLevel < nt.getVacuumLevelPartOn()) {
    throw error(Vacuum level too low, pick failed)
  }
}

place(NozzleTip nt) {
  doPlace();
  vacuumLevel = readVacuum();
  if (vacuumLevel > nt.getVacuumLevelPartOff()) {
    throw error(Vacuum level too high, place failed (part is stuck))
  }
}

tracklogic:
Ah i see, yes I think thats probably a better way of doing it - its clearer.

vonnieda:
@tracklogic Okay, thanks. If that will also work for your use case I will make the change.

tracklogic:
Np, thanks.

crinq:
use something like "vacuumLevel < nt.getVacuumLevelPartOn() * part.minVacuumLevel"
0.0 <= part.minVacuumLevel <= 1.0

and the same for place "vacuumLevel > nt.getVacuumLevelPartOff() * 0.9" 
or better "vacuumLevel > (nt.getVacuumLevelPartOff() * (1.0 - safetyFactor) + nt.getVacuumLevelPartOn() * (1.0 + safetyFactor)) / 2.0"
0.0 <= safetyFactor < 2.0

vonnieda:
@crinq I agree, we need part specific values or modifiers eventually. That's going to be in a future change.

vonnieda:
@tracklogic Okay, I have reworked this feature as described above. It's in a branch, let me know what you think: https://github.com/openpnp/openpnp/compare/feature/vacuum-sense?expand=1

tracklogic:
Looks good to me :)

vonnieda:
@tracklogic Okay, thanks. I have not fully tested it yet. I will do that tonight and then merge it.

vonnieda:
@tracklogic This code is now pushed up. Note, I had to remove the first check in place (checking if the vacuum level was still high enough) because it caused discards to fail. The issue was that if you had a failed pick (vacuum level not high enough) then when you try to discard to make sure the nozzle is clean the discard place would fail for the same reason.

So, now we only check after pick and after place. I am open to ideas if you'd like the other check added back in.

PR #375, opened 2016-11-18
Feature gcode dispense

tracklogic:


tracklogic:
Hi Jason

It was the best way to do the below, well that i could think of anyway.. :)

Driver 1 - pneumatic-Z down
Driver 2 - auger valve control
Driver 1 - pneumatic-Z up

Matt

vonnieda:
@tracklogic Okay, that makes sense. Can you fix up the conflicts and change the values to true that I mentioned and we'll merge this.

tracklogic:
As per e-mail i think this can be scrapped Jason as I've reimplemented this in a slightly different way

Issue #376, opened 2016-11-25
independant Z axis impacts Gcodedriver (when XY move to subdriver)


dave-prosee:
A 2nd Z axis gives double tool offset in action. The reason why and first actions are mentioned below.

Having a second independent Z axis (5th axis) often means having a second controller i.e. interface because not all controller boards support 5 axis.
Homing then becomes something you have to manage with the main GcodeDriver and a subdriver.
Because you want to home all Z axis before any X or Y move this means that the second Z axis is driven by the main driver and the other axis by the a subdriver. This is because the main driver actions are executed first before any subdriver action.

The Gcode driver is one driver used as main and subdriver. It first estimates the desired location for the given headmovable by correction for the headmovable offset. Then it decides whether actions are needed based on machine.xml, given location and current location. If needed it executes the gcode actions and calls itself for all subdrivers it has.
  
Currently, if no processing is needed the Gcode driver return without giving control to subdrivers.
This happens at two points: when no axis is defined for current headmovable or when no move is necessary for current headmovable.
Tweeking machine.xml could force the program beyond these points but reveals a second problem that during this proces the desired location to move to has already been corrected for the offset of the headmovable. So when reaching the next execution of the Gcode driver (e.g. in a subdriver) the desired location is again corrected for the headmovable offset. Any switch between headmovables (e.g. camera and nozzle) managed by different (main or sub) drivers will thus result in a physical move that is a multiple of the defined offset. With more (sub)drivers this error increases.

The fix this seemed to be twofold at first:
- allways give control to subdrivers, even when no action is necessary in the current/main driver 
- don't build offset on offset, use the original desired location or build on top of executed increments per axis so that machine.xml really controls, see additional steps below. Another solution is using a main subroutine, holding location and calling main and subdrivers (for now I do the first)
Not that theoretically subdrivers could call other subdrivers. The code calls itself and is not aware of how deep subdrivers are subbed.

This fix makes it possible to only have a second Z axis definition on the main driver. But removing X and Y (and Z1 and C) reveals something new: the coordinates are not updated in the user interface neither under the hood. Any new action is based on the home location (0,0,0,0). This means an increment of 1mm on e.g. X makes the machine jump to 1,0,0,0. 
Restoring X and Y (and Z1 and C)back creates a workable solution but is not ideal.
@vonnieda can you give me a hint as to how results are stored and retrieved for displaying. Further more I propose to cover this issue as a second step, i.e. second pull request so we have some progress.

[...]

PR #377, opened 2016-11-25
Improve Gcode driver (see #376)

dave-prosee:
Gcode driver used to:
1) add headmovable offset to target location on every entry --> giving twice or more the physical move
2) stop processing when there is no action thus preventing subdrivers to take control --> unnecessary tweaks in machine.xml 

This PR:
1) keeps original target location aside for every next iterative call --> offsets are used correctly
2) just skips part of the code to always reach the point where subdrivers take control --> no tweaks necessary in machine.xml for this bug HOWEVER same tweaks still necessary for other bug (see text in #376)

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
@@ -469,8 +469,10 @@ public class GcodeDriver extends AbstractSerialPortDriver implements Runnable {
     @Override
     public void moveTo(ReferenceHeadMountable hm, Location location, double speed)
             throws Exception {
+        //keep copy for calling subdrivers as to not add offset on offset
+        Location locationOriginal = location;
+        
         location = location.convertToUnits(units);
-
         location = location.subtract(hm.getHeadOffsets());
 
         double x = location.getX();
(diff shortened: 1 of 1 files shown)


vonnieda:
Hi @dave-prosee,

Will fully review this later today, but in the mean time, for the second issue, I think we need to modify getAxis() to search sub-drivers as well. The idea is that when we call getLocation() it finds all the axes needed to make up a full location. Since some of those axes might be a sub-driver we need to search them too.

Jason

dave-prosee:
@vonnieda 
Hi Jason,
I haven't seen anything indecent yet on getAxis(). I have seen moves completing ok now but where is the update of the screen done? And where do you keep the (target) location. I presume you use a listener since i have seen asynchronous updates of the screen. Can you direct me a little.

vonnieda:
@dave-prosee What I'm saying is that the modification of getAxis will fix the problem of coordinates not updating, and with having to duplicate the axis definitions.

When the coordinate display (DROs) needs to update it calls Nozzle.getLocation() which eventually makes it's way to GcodeDriver.getLocation(). GcodeDriver.getLocation() calls getAxis() to find an Axis definition for each of the four displayed axes (X, Y, Z, Rotation), applies transformations and offsets and returns the value, which is displayed.

The problem you are seeing is that since some of your axes are defined in the sub-driver, when getLocation() calls getAxis() on the main driver it doesn't find the X and Y axes since these are defined on the sub-driver.

The solution, I think, is to have getAxis also search sub-driver Axes so that the full set of four axes is found in getLocation(). Then you will be able to define only your X and Y on the sub-driver and X on the main driver. The filtering that happens in getAxis() by head-mountable-id will still apply, so the correct set of axes will be returned for the tool that getLocation is being called for.

dave-prosee:
@vonnieda
Thank you. That's very clear. 
I will have a look later this weekend.
Cheers,
Dave

vonnieda:
Thanks @dave-prosee! This has been merged.

dave-prosee:
@vonnieda 
Hi Jason, 
I looked into the matter. I think getAxis() should stay the same as it is used within moveTo(). Changing it like you described will return axis all the time within moveTo(). I went that way like: 
```javascript
 public Axis getAxis(HeadMountable hm, Axis.Type type) {
    	Axis AxisFound = null;
    	//search in main driver
        for (Axis axis : axes) {
            if (axis.getType() == type && (axis.getHeadMountableIds().contains("*")
                    || axis.getHeadMountableIds().contains(hm.getId()))) {
                AxisFound = axis;
            }
        }
        
        //search in subdrivers
        for (ReferenceDriver driver : subDrivers) {
        	GcodeDriver d = (GcodeDriver) driver;
            if (d.getAxis(hm, type) != null){
            	AxisFound = d.getAxis(hm, type);
            }
        } 
        return AxisFound;
    }
```
.... and running into more problems (moving on the main driver and thus updating new location, making the subdriver think that no physical move is not necessary anymore).

So I think GcodeDriver.getLocation() should handle the fact that location info may come from subdrivers.  Going to try that way now.

PR #385, opened 2016-12-22
add 2 more windows

ldpgh:
**MainFrame.java modified**
- added `Windows` menu with one item to control window layout style
- added method `splitWindows()`, called at the end of the class constructor
- added preferences property `PREF_WINDOW_STYLE_MULTIPLE` to store the current window layout style + default value set to single window mode
- elevated `panelCameraAndInstructions `& `panelMachine `to become class member variables

[
](url)
**Operation mode**
- Main-Window / Windows / Multiple Window Style ... enable OR disable
- With the window-layout/managment in discussion the current implementation requires to restart the application to get the different mode active.

[
](url)
**Further changes if it fits the needs.**
- Apply the setting immediately, if "Multiple Window Style" is changed.

[
](url)
**Issue**
- If one of the new windows is closed (e.g. by |x|), there is nothing added yet to re-open the window. Restart of the application is required.

### DIFF
--- a/src/main/java/org/openpnp/gui/MainFrame.java
+++ b/src/main/java/org/openpnp/gui/MainFrame.java
@@ -100,6 +100,8 @@ public class MainFrame extends JFrame {
     private static final int PREF_WINDOW_HEIGHT_DEF = 768;
     private static final String PREF_DIVIDER_POSITION = "MainFrame.dividerPosition";
     private static final int PREF_DIVIDER_POSITION_DEF = -1;
+    private static final String PREF_WINDOW_STYLE_MULTIPLE = "MainFrame.windowStyleMultiple";
+    private static final boolean PREF_WINDOW_STYLE_MULTIPLE_DEF = false;
 
     private final Configuration configuration;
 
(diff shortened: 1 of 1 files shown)


ldpgh:
Travis-CI seems to pass the required `mvn test -B`

vonnieda:
Thanks @ldpgh, I have merged this with a few minor changes:
* Changed name of menu from Windows to Window to match standard app menus.
* Added a message box after changing multi window mode so user knows something happened.
* Used pack() to resize machine controls window instead of hardcoded values.
* Reformatted code to match project spec.

Please make sure you pull this code before making any further updates.

If you find the time, I'd like to see a few other changes made so this feature can be complete:
* Each window should remember it's size and position on screen. Look at the componentListener on line 716 to see how this is done.
* Currently on Mac the menu is not visible unless the main window is selected, and it should be visible when any window is selected. I think using dialogs instead will fix this but I have not tested it.
* If a user closes one of the windows there is no way to get it back without restarting the program. I suggest modifying the Window menu to have three options: Cameras, Main and Machine Controls and make clicking on any of these options bring the window up.

Thanks,
Jason

PR #389, opened 2016-12-26
Import support for DipTrace PCB

BendRocks:
Hi, first pull request ever. Please scrutinize as needed. Wiki on the topic of first pull request  was awesome. 

What it does
This allows native DipTrace PCB pick and place exports to be imported without change. 

Why not use the default 'NamedCSVImporter'?

This could be done (see below). But the error messages and log output were hard to decipher and pretty much required the dev environment to be installed to make heads or tails of why the CSV was being rejected. In order to use the default importer, the CSV exported by DipTrace needs the line 0 header item changed from 'Name' to 'Footprint'. There is no way to change this header name in the DipTrace tool.  It must be done outside of Diptrace PCB.

Why not add the option to use 'name' as a footprint column in the 'NamedCSVImporter'?

This could be done, and this would allow Diptrace to be imported with a few lines changes to the NamedCSVImporter. But I thought it better to add support for Diptrace as a full-fledged client on par with Eagle and KiCad. Plus, DipTrace is under active development with a strong user community, features like 3D have been added and presumably at some point a wealth of other info could be exported from the PCB and used by OpenPNP in part creation. 

What was done?

The EagleBoardImporter.java source was copied to create DiptraceImporter.java. The code that considered top and bottom board files was collapsed in DiptraceImporter to consider a single file that has top and bottom data. The parsing logic was changed for DipTrace's native CSV format. 

A changed was made in MainFrame.java File->Import Board menu option to insert the reference to DipTraceImporter class.

What has been tested?

As change is isolated, I've verified that the code correctly imports the data, and correctly rejects and alerts the user for malformed Diptrace CSV files. I will happily support issues from other Diptrace users related to import. I have a TVM920 machine and am looking forward to helping.

### DIFF
--- a/src/main/java/org/openpnp/gui/MainFrame.java
+++ b/src/main/java/org/openpnp/gui/MainFrame.java
@@ -69,6 +69,7 @@ import org.openpnp.gui.importer.BoardImporter;
 import org.openpnp.gui.importer.EagleBoardImporter;
 import org.openpnp.gui.importer.EagleMountsmdUlpImporter;
 import org.openpnp.gui.importer.KicadPosImporter;
+import org.openpnp.gui.importer.DipTraceImporter;
 import org.openpnp.gui.importer.NamedCSVImporter;
 import org.openpnp.gui.importer.SolderPasteGerberImporter;
 import org.openpnp.gui.support.HeadCellValue;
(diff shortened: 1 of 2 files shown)


vonnieda:
@BendRocks Github should use this as a model of how to make a good PR. Bravo, well done and thank you! This a great pull request and I very much hope to see more from you in the future!

PR #400, opened 2017-01-03
Backlash compensations capabilities added for X/Y Axis

sp-apertus:
The idea is to always move to a point defined by BacklashOffsetX / BacklashOffsetY from the target coordinates first and then move slowly (speed multiplied by the BacklashFeedRateFactor) to the target coordinates.

without backlash compensation:
backlash-off

with backlash compensation:
backlash-on

The Gcode MOVETO command with backlash compensation is meant to look like this:
G0 {BacklashOffsetX:X%.4f} {BacklashOffsetY:Y%.4f} {Z:Z%.4f} {Rotation:A%.4f} F{FeedRate:%.0f}
G1 {X:X%.4f} {Y:Y%.4f} {Z:Z%.4f} {Rotation:A%.4f} F{BacklashFeedRate:%.0f}

This does not compensate for backlash on axes A and Z as I did not consider that too relevant but if people see a real world benefit I am happy to reconsider adding those as well.

I did not get maven to install so I cant run mvn test from inside eclipse yet but I tested the changes with a fresh configuration directory manually.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/camera/OpenCvCamera.java
+++ b/src/main/java/org/openpnp/machine/reference/camera/OpenCvCamera.java
@@ -154,6 +154,10 @@ public class OpenCvCamera extends ReferenceCamera implements Runnable {
             
             fg.open(deviceIndex);
             
+            for (OpenCvCaptureProperty property : OpenCvCaptureProperty.values()) {
+                Logger.trace("{} {} = {}", this, property, getOpenCvCapturePropertyValue(property));
+            }
+            
             for (OpenCvCapturePropertyValue pv : properties) {
                 if (pv.setAfterOpen) {
                     Logger.debug("Setting property {} on camera {} to {}", pv.property.toString(), this, pv.value);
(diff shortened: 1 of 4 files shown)


vonnieda:
@sp-apertus This looks very good! I have one tiny change request and then we can merge this. Can you please move the three new fields in the wizard to a single line? Vertical space in the wizards is at a premium so I'd like to minimize how much of it we use.

Thanks,
Jason

sp-apertus:
I rearranged everything into two grid rows to save vertical space, its a
bit wide now because the labels are long and the textfields are a lot wider
than their content...


[image: Inline image 1]

so I reduced the textfield widths a bit:

[image: Inline image 2]

git commited.

Regards Sebastian

vonnieda:
Thanks @sp-apertus, but what I meant is just to put the three new fields themselves on a new line instead of on two new lines. Don't worry about it, though. I will merge this and rearrange the fields a bit. Thank you for adding this!

vonnieda:
Updates to the wizard pushed. Thanks again @sp-apertus!

sp-apertus:
My pleasure :D

PR #408, opened 2017-01-13
Replaced Rotation Symbol 

sp-apertus:
Replaced Rotation Symbol with UTF8 (CLOCKWISE RIGHT AND LEFT SEMICIRCLE ARROWS) http://www.fileformat.info/info/unicode/char/1f5d8/index.htm


sp-apertus:
need to test first

vonnieda:
Looks like this character is too rare:

screen shot 2017-01-13 at 9 39 47 am

We may be better off just using "Rot."

PR #409, opened 2017-01-14
Rotation Symbol replaced with "Rot."

sp-apertus:


vonnieda:
Looks good @sp-apertus, thank you!

PR #410, opened 2017-01-15
Making it clearer what "height" and "length" refer to

sp-apertus:
Making it clearer what "height" and "length" refer to


vonnieda:
Thanks @sp-apertus, this is a good improvement. I'm also going to change the name of the field above from Body Height to Body Length.

PR #414, opened 2017-01-15
added a light "Orange" to the crosshair color selection

sp-apertus:


vonnieda:
Looks good, thanks @sp-apertus!

Issue #421, opened 2017-01-23
Icon Improvements


dzach:
(EDIT: See related discussion and work in   #426, #429, #434)

On most cases the _plus sign_ icon means `Create a new ...`, as indicated by the help popup on the button. The only exception I noticed (which always confuses me) is its use in the `Job` panel, where the popup says `Add an existing board...`. Here are some example screen shots:

In `Packages`:

![pnp icon3](https://cloud.githubusercontent.com/assets/1109829/22226532/7b67c27a-e1cf-11e6-9b0c-da1e60bcd68e.png)

In `Solder Paste`:

![pnp icon2](https://cloud.githubusercontent.com/assets/1109829/22226534/7b73cf0c-e1cf-11e6-8b81-810dff8f6a68.png)

However, in `Jobs` a `pencil` has replaced the meaning `create a new` of the `plus sign` icon, which now means `add an existing` :

![pnp icon1](https://cloud.githubusercontent.com/assets/1109829/22226535/7b767202-e1cf-11e6-8781-7f0aceb0f81c.png)

![pnp icon](https://cloud.githubusercontent.com/assets/1109829/22226536/7b8092d2-e1cf-11e6-88d5-039f5ac0c2cf.png)

Is there a reason for that, or it's just an oversight?

vonnieda:
@dzach The thinking was more that the plus sign is the way to add something to the list. In many cases the only way to add to the list is to create something new, while in the job panel there are two ways to "add" something - one of which adds existing and one of which creates new.

I find it confusing, too, and I put it there. I'm open to suggestions on how to improve it.

dzach:
In this case, `add an existing board` is equivalent to `import a board`. My suggestion would be to use an `import` icon, which is, usually, an arrow pointing down or into a circle or rectangle. The `plus sign` icon then keeps its `create a new ...` meaning.

Import icon example:

https://d30y9cdsu7xlg0.cloudfront.net/png/65557-200.png

vonnieda:
Ah, in looking at the Material Design icons (which is what most of the
OpenPnP icons are) I see that the pencil is labeled "create" and the plus
sign is labeled "add", so that is probably why I did it that way. It's been
several years...

Anyway, if your suggestion is that we change the New Board action to use
the plus icon and the Add Board action to use the Import icon, I am fine
with that change. We'll need to find a Material Design icon that fits, or
another free icon that is similar in style.

The Material Design icons are at https://material.io/icons/. Do you feel
[...]

PR #423, opened 2017-01-24
Adding new icons and suggesting some visual clues

dzach:
Adds new icons to keep visual consistency, as per discussion in #421.
Not all new icons have currently a use in OpenPnP.


vonnieda:
Thanks @dzach! I have merged this, plus a few minor changes of my own. I made changes to use the new icons you created as well as renaming and clarifying a few. 

I do have one thought on the new nozzle tip add and remove icons. These put the + and x in a different position than the other icons where we do this. Was it intentional, or should we standardize on a location? See:

screen shot 2017-01-24 at 11 16 32 am

Reference: #421

vonnieda:
WRT to the above, I like your positioning better ,so if we standardize I think we should move the + and x on the nozzle icons (right most) to the bottom like your icons have it.

dzach:
That was intentional. I was trying to apply the principle laid out above, i.e. provide a non-color clue together with a color one. So, close to the (grayed) tip, means 'tip' while close to the head means 'nozzle'. Hopefully, it makes sense.

vonnieda:
@dzach Okay, that makes sense now. I personally find it visually a bit glaring, but I think it's functionally better. In the future it may be better to have a completely different icon for nozzle and nozzle tip so that we can standardize on the position of the + and x.

Issue #329, opened 2016-08-19
Fiducial recognition should use the Pipeline system, configured in Machine XML.


issus:
Currently the fiducial recognition does not use the OpenCV pipeline as other parts of the software does. 

The fiducial recognition should be ported over to use CvPipeline, with the default pipeline configuration matching the current class. 

Users will then be able to experiment with custom pipelines for improved fiducial matching where their machine requires it.

visualecho1:
If a button is added to the part alignment config wiz to use the default fiducial camera, that config can be used for the fiducial pipeline editing and storage (you might also have the Edit button switch cameras if the name is like FIDUCIAL-*).  How would I get an accessor to that pipeline data from within ReferenceFiducialLocator?  It looks like it's only accessible from ReferenceBottomVision, is there a way to link up to that instance?

Question from the other side: is there an example of template matching in a pipeline staged vision operation?  I'm getting better at finding circles, but this is a little harder than sprocket holes.

vonnieda:
@visualecho1 The pipelines would not be shared - they are too different. When we convert the fid system to use pipelines a pipeline will be added to the fid locator and it will be customized similar to how bottom vision.

There's currently no template matching examples with the pipeline system, though there is template matching available. See https://github.com/openpnp/openpnp/wiki/CvPipeline for information about how to experiment with pipelines outside of OpenPnP.

visualecho1:
Hi Everboddy! "When we convert the fid system" - is that on a schedule?  Is this where that's being discussed?  Would you mind if I tried to help with this?  The code is really close to working on my machine, so I've been making fixes and then I kinda noticed that a few of the things I've fixed were on the issues list.  I'm glad to hear that there's a team working on this.

Fiducials are parts (footprint to camera) which don't need to be aligned with bottom vision.  It definitely makes sense to keep the pipeline with the part, one per atomic footprint-camera.  A part is a fiducial when specially marked as a part (aka "FIDUCIAL-HOME") or only when it's attached to a board?  The model doesn't seem to work well with this.  I think that any part can be a fiducial, that being a fiducial (sh/c/w)ould be a boolean attribute of a part-package relationship.  As much as I think this world needs more "bottom vision", I think "part vision" might be a better term, select a required camera.

Okay, so I've got some definite ideas about the fiducial vision pipeline, could you tell me more about the conversion?  Where is that being discussed?  I'm somewhat eager to see this code working better in many respects, but I'm very new here, only been debugging this app for a week or so.

[...]

PR #425, opened 2017-01-24
Issue #329 - fiducial pipeline

visualecho1:
This is a first draft of an enhancement to allow editing of a fiducial part pipeline.  It stores a pipeline XML config in the Part Alignment section of machine.xml.

The only change to the GUI is a checkbox on the (Part Tab) -> (Pick any part) lower panel "Alignment", which makes the existing "Edit" button use the default (top) camera instead of the bottom vision camera.

The fiducial locator logic now uses that pipeline from the bottom vision config data to find keypoints under "results" of processing the pipeline and uses the closest one.  This needs to be upgraded to using a template match, but that will require a little more research by me.  This is a demo to see if everyanybody want to continue in this direction.

How to test this:
1. Start machine and Home
2. Load a job with a board with a fiducial, put board on table
3. move camera to fiducial
4. In "Part" tab, find the fiducial part.
5. Note new checkbox at bottom of Alignment panel.  Try Editing the pipeline with the box checked/unchecked.  The only difference is the dialog name and the camera used, the XML data is per part.
6. set up a pipeline which returns keypoints named "results" and tags the fiducial well
7. go to the "Jobs" tab and click the magic button for a fiducial check and position update.

The error I always get at this point (if the pipeline is working) is something about outside 1% of expected, so the pipeline has been processed for the two most distant fiducials.  That's all I want to do for this first draft, as I'm looking to make this work right, and I'm very very new around here.

Okay, so what questions need to be asked here? 
* I don't think this is what somebody originally had in mind, but per part pipeline is already supported by the rest of the code here.  Is this acceptable?  Will this change or prevent any planned model schema changes?  Should this be tweaked, scrapped, or is it almost good to go?
* While parts have a top and bottom, fiducials are virtual and only have a shadow.  There's no need to store a bottom vision pipeline for a fiducial, but I can see why a separate top vision pipeline for a part might be useful.  This isn't a fixed-length storage system, and this can be added if needed to the XML as an additional pipeline (or an array, if a pipeline per part per camera).  
* The checkbox is automatically populated if the lower-case part ID starts with "fiducial", but is otherwise manually operated (not persistent). How's the wording on that?  Does it take you less than 15 seconds to figure out what it does just based on it's appearance?

I have to figure out how to do the template match thing, not expecting that to be too hard.

--visualecho1


vonnieda:
@visualecho1 Thank you for the pull request! Unfortunately it looks like the tests didn't pass. You can see the errors at https://travis-ci.org/openpnp/openpnp/builds/194979510#L1523. You can run the tests locally with `mvn test`.

visualecho1:
Yeah, I was looking at that output... I couldn't figure out if that was 
just me or what was causing that.  I'll look closer.

vonnieda:
Yep, the same tests that run when you do a mvn test are run on every pull
request, against the code in the pull request. If the tests fail the pull
request fails.

You should get the same error if you run the tests locally and you can
iterate on that to get them fixed.

Jason

visualecho1:
So what what error is generated with this test?  It's running the NullMachine?  I can't remember how to set up a test target for maven in Eclipse so I can debug this.  Like I said, I get an exception for the fiducials being more than 1% away from expected when checking fiducials at the start of a process.  If the unattended test is failing on that (and does appear to be the case), it's a non-error at this time.

It's also not actually necessary to merge this into development, I 'm asking very specific questions about usability of the change to these two files.  I implore you to look at them briefly.

visualecho1:
Looking at this... not making much sense...

vonnieda:
Hi @visualecho1, my apologies, I didn't quite read your message well enough to realize that you were just looking for validation on the idea. I'll do a full review of the code tonight and post my comments.

vonnieda:
@visualecho1 Okay, I've had some time to look through the code. I like the premise, but the implementation does not quite work with how this system is designed to work. First, let me explain a bit how the bottom vision pieces all fit together and then I will explain why it doesn't make sense to put the fiducial pipeline in with them. Then I'll suggest some changes.

Due to personal design goals in the software, I do not believe it's appropriate to store machine specific data with parts. Bottom vision pipelines fall into this category. I can go into why if you'd like, but for this discussion, suffice to say that I want pipelines stored in machine.xml and not parts.xml. This becomes important further down.

ReferenceBottomVision is an implementation of PartAlignment, which is an interface that basically can tell you the orientation of a part on a nozzle. ReferenceBottomVision relies on a pipeline to do it's work.

I wanted to be able to store a customized pipeline per part for ReferenceBottomVision, but as I said above, I didn't want this in parts.xml. The solution, then, was to store the pipelines with the ReferenceBottomVision configuration in machine.xml and link them to a part by ID.

Since ReferenceBottomVision is just one of potentially many PartAlignment implementations, it was important that you be able to configure whatever PartAlignment you happened to be using - not just ReferenceBottomVision.

So, when you open the configuration panel for a Part, what happens is we look up the PartAlignment implementation you are using and then call PartAlignment.getPartConfigurationWizard(Part). This gives us an abstract way to configure the data for a part from the perspective of the PartAlignment that that data is used with.

So, in summary, ReferenceBottomVision is a PartAlignment, and when you call ReferenceBottomVision.getPartConfigurationWizard(Part) you get a ReferenceBottomVisionPartConfigurationWizard for that part and for that instance of ReferenceBottomVision. The ReferenceBottomVisionPartConfigurationWizard is then shown to the user in the context of the Part.

Now, with the above, I hope it's clear why the fiducial pipeline does not belong in ReferenceBottomVisionPartConfigurationWizard. I do like the idea of having part specific fiducial pipelines, but that system should mirror the one described above, not glom onto it. Fiducials have nothing to do with bottom vision.

So then, what I would suggest for your code is:

* Add getPartConfigurationWizard(Part) to the FiducialLocator interface, in the same way that it exists on PartAlignment. These interfaces are analogous to each other. One is for fiducial finding, one is for part alignment.
* Implement a similar system for returning a panel that contains a pipeline editor. This can probably be pretty much a copy and paste of ReferenceBottomVisionPartConfigurationWizard.
* In PartsPanel.java, look at line 203. That is where all the above actually happens. It grabs the PartAlignment defined on the machine, asks for it's configuration wizard and displays it.
* Copy the above chunk of code around line 203 and do the same thing now for the FiducialLocator defined on the machine. Throw a new tab on the panel and drop the wizard into that tab. It would be called "Fiducials" or something. 

All in all, it should be a fairly easy set of changes and it keeps like with like. This makes it possible for people to drop in entirely new FiducialLocator implementations. In fact, you can even implement this whole system without changing the existing ReferenceFiducialLocator. You can have VisualEchoFiducialLocator with all the same first class capabilities as the reference one and easily swap between the two just by changing the name of the class in machine.xml.

Hope this is helpful. Please ask any further questions that you need to!

Jason

visualecho1:
Yay!  This is perfect, exactly what I needed, Thank you!  I think I can deliver this, working on it now.

Thanks again for glancing at this without it passing the tests, I am having some trouble with tuning the default stages to recognize the sample PNG for the NullDriver, but I am beginning to understand how that works.

vonnieda:
Glad to hear it helped. Looking forward to seeing you work.

vonnieda:
Just realized in my previous comment I omitted an "r" and it changed the whole tone of the message. The message was intended to be "Looking forward to seeing youR work", not "you" work :)

I saw that you added a commit today. Is this ready for review, or still have some work to do?

Jason

visualecho1:
Actually I liked that spelling.  You spent some good time writing me a good tech brief and I'm gonna try and act on that, so nope don't mind, and thanks again.  Have I mentioned you have a fabulous application here, sir?

I've got a lumberjacked set of files committed here which appear to be working properly.  They interface the vision system in a way I'm not sure about (memory and .clone() and such), and I'm trying to work with the config system for reduced footprint of the fiducial-locator data in machine.xml.  There's still a lot of inherited junk that needs to be cleaned up, and I need to beat up on it (basic testing/debugging).  It's using a default file-resource fiducial pipeline and custom per-part pipelines, control widgets added to the parts panel with a wizard.  Logic only locates the closest key point, not yet pipeline matching a template actually based on the footprint.  If I load up the sample job and null machine, the default template nails the position/rotation on each board.

Definitely take a look if you wish, if you're busy skip it and I'll have better code committed in a few days.  I need to reread the note and see what I missed.

p.s. Not yet looking at other tech note about nozzle tip loader actuator, wanna make sure I understand git and github and get this one clear before I confuse myself any further :-)

visualecho1:
To do:
* class, method, and variable naming - read standards doc and review silly names I wrote quickly while programming
* reduce complexity of data access, clean up parameter passing, especially settings, camera, default pipeline, and per part pipelines
* review user facing text - remove stevemartinesque word usements
* look at split between a default only pipeline class and additional class which allows the per-part pipeline

vonnieda:
Hi @visualecho1,

Took a brief look at the code and the current functionality. Looking pretty good. One thing I noticed is that the fiducial wizard is being added to the alignment panel, which overrides the bottom vision wizard. This can be resolved with the following small diff:

screen shot 2017-01-30 at 10 09 32 pm

I would also probably suggest dropping the enabled flag from this as it's not really needed for fids. It's handled at the higher level of selecting a placement as a fid.

Other than that, it looks like it's coming along well!

visualecho1:
Oh, that's good, yep, I put that right in.  Thank you!

The 'enabled' checkbox nulls the pipeline when cleared, so no (very little) pipeline data is stored in the machine.xml for a part that isn't enabled.  I wrote this as a way to keep the machine.xml tidy, but putting the controls on a different tab makes casual perusal of parts that aren't fiducials less likely, and it doesn't appear you consider this an issue.

Looking at the template CvVision stuff now, this week I may not have much time to spend working on this.

vonnieda:
My only concern with the enabled checkbox is that I don't want to add a stumbling block to people who just want to use fiducials. If it defaults to enabled, that's fine. Most people will never need to edit the pipeline. If they do we need a better default pipeline.

One thing we could do to improve this situation is make is so that if the per part pipeline is null we use the default. Only populate the per part pipeline the first time someone opens the editor. Then, for the majority of cases, there will be no per part pipeline and the machine.xml stays clean.

visualecho1:
Yes, that's what I thought you meant and it's more consistent with the rest of the way the option work in other panels.  I'm famous for overcomplex GUIs and this is not a difficult programmatical issue, so moving slowly as this evolves.

vonnieda:
I think this would be a good change both in your code and in the bottom vision code. I'm actually not sure why I didn't do it that way in the first place. I wouldn't sweat it though, it's an easy enough change for me to make after the fact. You can focus on what you are adding.

vonnieda:
Hi @visualecho1, this is looking pretty solid. Are you ready for a full review, or do you still have some work to do? I noticed a few minor things I would typically point out in a review that I would ask to have changed before merging, so I can either start making those comments now or wait until you are finished. Whichever you prefer.

visualecho1:
I gotta do some cleanup, and should get some time today.  I still feel like some of the objects and data flow is clumsy, but not sure if my eyes can see that stuff right now.  The wizard needs the most work.

visualecho1:
Okie dokes! All the options are in play that matter, fiducials are getting matched by template of the part footprint by a user controlled vision pipeline.  This is the current logic: 
* Pipeline will accept the template image via the new ImageInput class (extension of CvStage) on a stage named "template", if stage is removed default should drop to keypoint detection.  Capture image can be modified before named "match" stage, but seems to work great right out of the box.
* should behave as before without any additional config changes
* wizard checkbox to use template can be cleared to force keypoint detection (with results named "results").  Both the keypoint "results" and the template "match" result can be generated, or those stages of the pipeline can be disabled.
* Templates are rotated for matching to the board. It's a quick operation and doesn't waste much  CPU on the parts that are fiducials and don't need it, and helps matching of parts that aren't and do.
* I've forgotten to ensure that parts with a nonzero center (geometric center of all pads gives X or Y not zero) work correctly, but single pad parts (like regular fiducials) won't see a problem with this.
* Excellent prior work on this set of classes made my work easy.  Some of the logic is really quite good. All in all, I think I expanded and fluffed up about 5 lines of existing code.

I'm expecting a laundry list of things I need to change, resubmit, think, and change again.  Hopefully we can keep that loop to just one or two iterations.  Okay, I need beer, doing Dilbert's engineer victory dance over here.

vonnieda:
This is awesome work @visualecho1! I will review it either tonight or tomorrow. I'm impressed how much shorter the code is now from when I looked at it yesterday. I *love* shorter :)

visualecho1:
Take your time.  Look closely.  There's secret roosky checkista LOLCat trojan virusez sugar coated with super mega ultra happy fun time Japanese schoolgirl kawaii cuteness in there.  All seriousness aside, take a look and lemme know what I need to change for it all to be easy and proper.

src/main/java/org/openpnp/machine/reference/vision/wizards/ReferenceFiducialLocatorConfigurationWizard.java
```diff
@@ -0,0 +1,233 @@
+package org.openpnp.machine.reference.vision.wizards;
```
vonnieda: The labels in the wizard make it too wide for a common laptop screen. Screenshot:
screen shot 2017-02-06 at 7 23 38 pm

You can just add some newlines to the labels to get the total width down. Shoot for the program to look good on a 1440x900 screen.

src/main/resources/org/openpnp/machine/reference/vision/ReferenceFiducialLocator-DefaultPipeline.xml
```diff
@@ -0,0 +1,20 @@
+<pipeline>
```
vonnieda: It's not quite clear to me how this pipeline is working. It seems like it has the ability to do both template matching and blob detection? Would you mind just explaining the way this works a bit?

src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
```diff
@@ -208,56 +264,152 @@ private static Location getFiducialLocation(BoardLocation boardLocation, Placeme
...
 
         return location;
     }
 
-    private static Location getBestTemplateMatch(final Camera camera, BufferedImage template)
-            throws Exception {
-        VisionProvider visionProvider = camera.getVisionProvider();
+    // here's the big magic
+    // use default or custom pipeline for part to pinpoint fiducial location
+    private Location findFiducial(final Camera camera, Part part, double rotation) throws Exception {
```
vonnieda: This method has become too long to readily understandable. Please break it up into a few smaller methods.

src/main/java/org/openpnp/machine/reference/vision/wizards/ReferenceFiducialLocatorConfigurationWizard.java
vonnieda: I'd like to see the wizard simplified a bit. I think we can remove the "Use custom pipeline?" checkbox and just use the custom one if it's been created, otherwise use the default. If the user wants to go back to using the default they can just reset it. Resetting it should just set the custom one to null.

vonnieda: I find the template option confusing. I think this needs to be simplified a bit too. We may need to discuss this further, but here is my thinking:

* The code that processes the pipeline should detect a stage called "template" and add the template. I think this is happening now, so that's good.
* Instead of having distinct code paths for template matching and blob detection, which limits the ability to customize the pipeline, just specify that certain result stages will be processed in different ways. i.e. if after processing, the pipeline contains a stage called "resultTemplates", process the results as template matches. If it contains a stage called "resultKeypoints" process the results as keypoints. Even better would be to just look for "results" and make a decision based on it's data type.

In general, try to present as few options as possible and do as much for the user as you can.

src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
```diff
@@ -208,56 +264,152 @@ private static Location getFiducialLocation(BoardLocation boardLocation, Placeme
...
+        BufferedImage template = null;
+        Object result = null;
+        
+        PartFiducialPipeline settings = getFiducialSettings(part);
+        CvPipeline pipeline = createPartPipeline(settings);
+        if (settings.isUseTemplateMatch()) {
+            // here we either need a footprint or a previously created template
+            Footprint fp = null;
+            if (rotation == settings.getTemplateRotation())
```
vonnieda: What is the purpose of this line? It seems like if the rotation is different the template doesn't get created? In what case is there a template stored on settings?

src/main/java/org/openpnp/machine/reference/vision/wizards/ReferenceFiducialLocatorConfigurationWizard.java
visualecho1: No argument, I was not expecting this format to be perfect.  More beauty newlines coming right up.

visualecho1: Yes, can do this.

visualecho1: Right now template detection is working so well, I don't know if we need the keypoint code.  The code is intended to allow maximum flexibility in setting up a pipeline, but provide sensible working defaults if not needed, so yeah, user shouldn't worry about this (but some of us do).

Yes, single set of results can be named 'results', but then can't show an example of both matching types inside the configuration, which is fine.  If there were a CvStage class which amalgamated the different detection types, it would be different.  I can do this, second bullet is right idea.

src/main/resources/org/openpnp/machine/reference/vision/ReferenceFiducialLocator-DefaultPipeline.xml
visualecho1: That's it, yes.  I hadn't quite determined which if either was better, if we should favor the new code with the template match, or how anyone felt about this.  From above, I think I'm going to preserve the code which matches keypoints (or circles for that matter) but clean up the default template match code.

The new CvStage extension class receives template image prior to processing.  The template is matched with result in 'match', then captured image is recalled and blob keypoints saved in 'results'.  Template match checkbox on wizard selects between the two, but not sure I did that exactly right.  One or both can generate locations, closest location to expected location wins, and it tries three times to find one (could be changed to something much more robust).

src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
visualecho1: Okay!  This one might need an aspirin.

src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
visualecho1: Did I get it backwards?  It's supposed to not need re-creating the exact same template as required for the same part at the same rotation, trading a few CPU for memory.  It is an optimization, it can be removed.
However, I also allowed for not updating that template when viewing the fiducial pipeline, so it can be used as a model for the last part the camera viewed as a fiducial.  I think it works, and if there's no template yet created it makes one with angle 0.0D.  This was the spooky part I was joking about.  See if it works, but yeah, no trouble cutting it if it's unusual.

visualecho1:
Thank you again for the positive, constructive feedback.  I'm working on these as I can, and hope that my reasoning and logicses aren't confusing.

visualecho1:
I think I crossed off this list...
* needed to add ImageInput class to CvPipeline editor dialog so it could be added to pipelines
* cleared some chaff, streamlined defaults: wizard is a lot cleaner, code a lot more pliable (processes template matches, keypoints, and circles together), big honking findFiducial method busted into pieces.

Regarding the wizard, "Resetting it should just set the custom one to null.", I think what is here now will work. Or maybe add another button to null out the custom pipeline, or detect that the user has deleted all stages after dialog is closed, or something else.

Regarding multiple detection types, since the pipeline only returns one stage named "results", any (and only) one of (circle, keypoint, template) match can be used.  Locations are generated from the result, sorted by distance from expected location (camera center), and the first (closest) one used.  The current default factory preset XML uses template detection with very little (no) pre-processing, additional XML examples showing KeyPoint and Circle matching could/should be shown somewhere.

I am really only testing single pin round fiducials, I need to see if I can pinpoint a 28-TSSOP  using Circle matching on LitePlacer hardware, then I'll be satisfied it works :-)

src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
```diff
@@ -390,4 +604,84 @@ public Icon getPropertySheetHolderIcon() {
...
+        @Element(required = false)
+        protected CvPipeline pipeline;
+        
+        protected BufferedImage template;
+        protected double templateRotation;
+
+        public PartFiducialPipeline() {
+            pipeline = null;
+            template = null;
+            templateRotation = 99999.9D;
```
vonnieda: What is the purpose of the 99999.9D here? Could it be changed for something more self evident? If not, can you add a comment explaining why it is the value that it is?

src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
```diff
@@ -390,4 +604,84 @@ public Icon getPropertySheetHolderIcon() {
...
+        }
+
+        public PartFiducialPipeline(CvPipeline pipeline, boolean useCustomPipeline, boolean useTemplateMatch) {
```
vonnieda: What do the useCustomPipeline and useTemplateMatch parameters do? It looks like they aren't used.

src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
```diff
@@ -390,4 +604,84 @@ public Icon getPropertySheetHolderIcon() {
...
+            this.pipeline = pipeline;
+            this.template = null;
+            templateRotation = 99999.9D;
```
vonnieda: Another magic 99999.9D. Please comment or remove.

vonnieda:
Hi @visualecho1, just wanted to check in and see if you had time to look at the last few comments I made. Pretty simple stuff, so if you are low on time just let me know and I will finish it up and get it merged.

visualecho1:
Yeah, sorry about this... the computer for my PnP machine failed, I haven't had a chance to repair it yet.  I have the data, but having difficulties with time, space, winter temperatures, and Oxford commas.

vonnieda:
@visualecho1 Thank you for your work on this, but I decided to merge a different version of this in. More information about that is here: https://github.com/openpnp/openpnp/issues/329#issuecomment-312158337

Thanks,
Jason

PR #426, opened 2017-01-25
Icons

dzach:
Branch with ongoing work in icons. See src/main/resources/icons/. See #421 . Should be deleted as soon as icons become final.

PR #429, opened 2017-01-26
Icons

dzach:
Finalized new icon proposal. See #421 and #426. The following mockups show how they look:
.

screenshot-2 Pick and Place panel
.

job-mockup Job panel 
.

feeder-mockup Feeder panel
.

icons-alt Alternative icons (see below)
.

nozzle-tip-changer-mockup Nozzle Tip Changer Panel

The alternative icons are an improvement of the previous ones I made, but are not proposed to be used, because they are necessarily more complex and do not allow for easy abstraction of their shapes. The ones I propose are much simpler.

Some effort was put into making colors more suitable for color blind people. This made it necessary to alter certain colors to a different luminosity to make them distinguishable from the background color.


vonnieda:
@dzach These look fantastic! I really like the additional feed icons that include the tape motif. So much more clear as to what they do. A couple thoughts:

* I think the pick icon (blue nozzle with green up arrow) should have the nozzle being red, since it's a movement operation and not a capture operation. 
* I like the two concentric blue circles to represent the position rather than the large blue dot. Was there is a reason you changed to the dot?
* Would it be possible to update the other uses of red in the icons to the new red you've picked so that they all match? Or would that cause problems for color blindness?

dzach:
The large blue dot alone is already used for an actuator I believe, and I tried to re-employ it in the icon, because of its simplicity.

Regarding the concentric vs large blue dot, I tried to simplify and abstract the icons so that I can remove some elements without loosing the general meaning. E.g. for the camera location, remove the dot and leave just the **camera frame** as a representation of the camera in various, like the feeder and the board ones. If you prefer the concentric, I have no problem with that, that's why I've included them as alternatives.

The existing colors offer better contrast with the white background, and are preferable. However, when they are mixed with other colors of similar contrast they don't work properly for color blindness. I believe this doesn't present a problem, as the basic color is kept mostly the same and only the luminosity is changed.

I didn't change the pick nozzle to red because the operation there is `pick` and not `look`. I hope this is consistent in all icons. I might write a small document with all the conventions used, just for reference or guidance, when time permits.

With the new icons for the feeder and the board (and possibly others, as I discover them going through the program in depth) we depart from the simple style that was adopted until now. I did this intentionally, because I caught myself frequently resorting to the help popup to find out what an icon meant. I think this might be a better icon style, because it offers direct and specific iconic meaning to the buttons, something very desirable it terms of HMI for moving machinery, I believe.

I hope the above make sense.

dzach:
There is this project [https://github.com/nvkelso/color-oracle-java](url) which offers help to make visual designs color blindness aware, but I couldn't use it in my machine due to some unknown problems (with the kde installation ?). If I get rid of the problems I'll pass the icons through it and see what comes out.

vonnieda:
Hi @dzach, I agree with all of the above except for the pick icon. I may be misunderstanding the design goals. Previously I have always used red to show that the selected object will move and blue to show that it will capture. (Capture meaning read coordinates)

So, prior to your changes, if a button was going to move the nozzle, the nozzle would be red, indicating "danger, the machine is going to move". Same with the camera, the camera square would be red. You've retained this language throughout the new icons except in the case of the pick icon. 

I didn't notice this when I checked in the pick icon the other day, but I should have. I think that the pick icons, because it is going to move the nozzle, should be red. 

I have these ideas outlined in the readme for the icons at https://github.com/openpnp/openpnp/blob/develop/src/main/resources/icons/README.md#capturing.

I don't quite understand your comment of "I didn't change the pick nozzle to red because the operation there is pick and not look.", so perhaps I am missing something. If you disagree that the pick nozzle should be red, can you help me understand why?

Other than that, I agree with everything you said above. I do prefer the concentric rings but if you think the design language is better with the dot I'm happy to take your guidance.

Jason

dzach:
I have no strong opinion for the colors and I love the discussion because it clarifies matters. Thanks for the link to the readme for the icons, I missed that completely. I've myself thought about the color of the nozzle being blue, and actually tried some other colors for it, but left it blue based on the assumption that we color only the main operation and not the side effects; which in this case is `Pick` and not `Look`, meaning that during a pick we not actually interested in what the camera does but in whether a component gets picked correctly. Hence the black bar at the bottom, denoting the feeder/tape. But this is by no means cast in stone; it's just a helper to keep consistency throughout a design.

Actually changing it to red makes more sense, because I cannot justify it's color being blue; according to my reasoning, it could be anything. So will change it to red. What red, though?

I saw in some pages in the wiki some icons with a brighter red and it doesn't look bad at all. The pink I've used was chosen for its luminosity value, looking at a gray tone render of the icon and trying to make the icon elements stand out above their background. Here is a pink pick icon:

screenshot-3


It doesn't really look bad. So if you agree, I'll visit all icons again and change all red colors to this pink. What do I do next? Another pull request?

vonnieda:
I think that looks great! I'm fine with changing all uses of red to pink through the program if that helps with color blind folks. The color looks good to me. I like the idea of consistency. 

You can just make more commits to your branch and this PR should automatically update. No need to make a new PR.

dzach:
Well, I'm not sure the pinkish red is as good as the original. If not, then we should revert to the previous version and only keep the pick icon change from blue to red. I don't mind using any, would prefer the original red everywhere there is no need for better contrast with the background.

vonnieda:
@dzach I agree, I guess I misunderstood. I thought the pink was needed for contrast, but it sounds like you are saying it's only needed in certain icons? Is that the tape ones? I assume that since the red is dark it does not show well on the black tape to a color blind person?

vonnieda:
I converted your mockups to gray and now it's quite clear what the issue is. I see how the additional contrast is needed for darker backgrounds. 

91e9b5cc-e3fe-11e6-92d1-923e4c5a4b15
91d61df0-e3fe-11e6-88c1-e521ea17a7fc
5be3a52c-e3ff-11e6-8af1-ef36fe58ad70
91e45708-e3fe-11e6-91bc-88f7f34dc971

dzach:
Yes, that's exactly what I did with those icons. I wish I could run color-oracle on all icons, but have no time to debug my problems with it right now. I'm by no means an expert on this issue, just discovered it a long time ago in one presentation for an important project, where the person responsible for implementing the work didn't get a clue of what I was presenting. I redid all colored graphs and succeeded with the project, and  since then I try to pay some attention to these matters.

The difficult part is to stick to the given colors of a limited palette. So I tried to only change some of the color parameters to make it somewhat easier to distinguish. It would be easier if we could use two shades for each color, depending on the backgrounds.

vonnieda:
@dzach So are you happy with the PR as it stands? Should I go ahead and merge?

dzach:
I'd rather use the original reds where there is no problem with the background contrast. If it's on me, I'll change most reds back and commit once more. Is that ok?

dzach:
Went ahead and changed all but the tape reds back to the original. I feel this is the best at this time.

vonnieda:
Merged with very minor updates. Thank you @dzach! This is a very nice improvement and it makes navigating OpenPnP significantly easier.

dzach:
Thank you Jason. Let's start using it and see how it goes, what needs to be corrected. Please let me know if a need for more icons arises during the development.

vonnieda:
@dzach One thing I noticed is that some of the new icons appear a little blurry in some of the features. I was going to look into this over the weekend, but do you have any ideas?

screen shot 2017-01-27 at 11 24 44 am

The outer blue circle looks a little misshapen and blurry.

dzach:
It's the sub-pixel alignment. I don't know how java renders the svg content, whether it needs the svg paths on integer pixel values or on half pixels. I could try it with all icons but compiling openpnp for each icon change is very tedious and time consuming. If there were some java utility to render svg on the fly I could make them all crispy, I suppose.

vonnieda:
It looks like the same is visible in my SVG editor. I use AutoDesk Graphic.

You are right, it's subpixel issues. I made a minor change to align
everything to pixel boundaries and it looks a bit clearer.

[image: Screen Shot 2017-01-27 at 12.03.52 PM.png]

I will do some research and see if there is hints I can give the renderer
to improve this. The renderer we use is called Batik.

On that topic, I've been thinking it would be really helpful to have a way
to easily view all the icons in one place. I may take a couple hours this
weekend to make a little program that shows all the icons in a panel and
make them refresh when changes are made on the filesystem. This will make
it much easier to see how things will work without having to wait for
OpenPnP each time.

Jason

dzach:
Well, they were supposed to be aligned on pixel boundaries, at least that's what I've tried. Unfortunately at present I'm lacking the capability to snap lines on construction guidelines in Inkscape, so I might have missed some trying to do it manually.

Batic is an old timer and a proven renderer I think. An SVG file with all icons will do fine then, there is no need to write a special program. I would have already done this **all-icons.svg** were it not for the strange problem I'm  having with the construction guidelines in Inkscape. I need to fix that first.

Stroke thickness plays a role together with alignment. An 1.5px line doesn't align on pixel boundaries, so some choices may be limited, or one must live with some blurriness. A 1.5px stroke thickness circle might look nicer on the icons but I found the alignment difficult and opted for the 2px that is used in the original icons. We can change that in a future revision, despite the misalignments it looks crispier.

Google has done a superb job documenting Material Design. Their [guidelines for icons](https://material.io/guidelines/style/icons.html#icons-system-icons) are very useful and almost guarantee a uniformity in UIs, if  followed closely. I need to remember to have a look there more often.

I the icons were larger, say 32x32 or 48x48px then things would be a lot different. OpenPnP's icons do have enough white space around them, so changing them might not require other changes to the GUI. Just an idea.

vonnieda:
@dzach That guidelines link is very good reading!

If there is significant value in increasing the icon size I'd be okay with it. Maybe not as high as 48, but 32 would be worth a try.

PR #438, opened 2017-02-01
Added a Gcode console to the GcodeDriver

dzach:
Added a Console for the Gcode driver, with browse-able history, 50 commands deep. Commands typed at the command line are sent verbatim to the Gcode controller of the machine. Timeout time is fixed at 5000ms.


src/main/java/org/openpnp/gui/GcodeDriverConsole.java
```diff
@@ -0,0 +1,200 @@
...
+        gcodeConsole.add(cmdLineTextField,"4, 4");
+        
+        cmdLineTextField.addKeyListener(new KeyAdapter() {
+            @Override
+            public void keyPressed(KeyEvent e) {
+            
+              switch (e.getKeyCode()) {
+                case 0:  return;
+                
+                case 38: // arrow key UP, older history
```
vonnieda: Instead of hardcoding key codes, please use the constants defined on KeyEvent. See, for instance: https://docs.oracle.com/javase/7/docs/api/java/awt/event/KeyEvent.html#VK_UP

src/main/java/org/openpnp/gui/GcodeDriverConsole.java
```diff
@@ -0,0 +1,200 @@
...
+        // shift history one place to the past. Copy is done in-place.
+        System.arraycopy(history, 0, history, 1, moveLen);
+        // store current command at 0
+        history[0] = cmd;
+      }
+      // reset historyCursor to the present
+      historyCursor = 0;
+      
+      // most controllers prefer commands to be in upper case.
+      cmd = cmd.toUpperCase();
```
vonnieda: I don't see any reason to force upper case. I don't know of any controllers that require it and it could restrict someone from using a command that requires mixed case.

src/main/java/org/openpnp/gui/GcodeDriverConsole.java
```diff
@@ -0,0 +1,200 @@
...
+import javax.swing.JTextField;
+import org.pmw.tinylog.Logger;
+
+public class GcodeDriverConsole extends AbstractConfigurationWizard {
+    private final GcodeDriver driver;
+
+    public GcodeDriverConsole(GcodeDriver driver) {
+        this.driver = driver;
+        
+        historyLen = 0;
```
vonnieda: In the future I'd suggest using a `List<String> (List<String> history = new ArrayList<>())` rather than an array. It removes the history limit, makes it easier to shift elements around and makes it less likely to run into a programming error. No need to change it here and now, just something to consider for the future.

dzach: I tried that at the beginning, gave me errors and didn't quite understand what it was asking me to do, so I preferred a well known path. Please go ahead and fix it, if it's better.

src/main/java/org/openpnp/gui/GcodeDriverConsole.java
dzach: Will do.

src/main/java/org/openpnp/gui/GcodeDriverConsole.java
dzach: Let me check again.

Nop. Marlin doesn't like low case gcode:

```
> g28
echo:Unknown command: "g28"
ok
```

dzach: So, should we leave it to the user, or force it to upper case?

src/main/java/org/openpnp/gui/GcodeDriverConsole.java
vonnieda: I'm fine with it. It's working and tested. That's better than new and untested :)

src/main/java/org/openpnp/gui/GcodeDriverConsole.java
vonnieda: Wow, that's interesting! Okay, let's leave it uppercase and if anyone complains we'll address it then.

dzach: Checking _position-nozzle.svg_.

dzach: I don't know where these changes to position-nozzle.svg came from. My last change to that file was in commit 0f7f801498c878bd7b6730e1c9fed6b63fcd7414 on 2017-01-29 21:59:31 with message "More icon refinements"

vonnieda: Don't sweat it. I'll compare after merge and if it needs to be reverted I'll take care of it.

One suggestion for the future: Do your work in a new branch instead of develop. Then when you want to send a PR you can still have a clean develop branch to compare against without having to constantly pull in upstream changes. 

That aside, it looks like this is ready to go. Shall I go ahead and merge it?

dzach: Yes, go ahead, merge.

I do work in other branches locally, but have not managed to do a merge with the remote, without all these pulling of upstream changes.

vonnieda:
Thanks @dzach! This fixes #438

dzach:
Thanks @vonnieda ! Time to start thinking about the `@macro`s now.

vonnieda:
Please make sure to pull changes from develop before working on new changes. I made a few very minor updates.

dzach:
Sorry for the extra work you had to put into this. 

I had a look into the _position-nozzle.svg_ file. The file came from a test I made a few days ago with Inkscape 0.92, which defines 96dpi as the base instead of 90dpi. Doesn't seem to affect anything.

vonnieda:
Nothing to apologize for. It was a solid PR and just needed a little light touchup. Really appreciate the time and effort you put into this!

vonnieda:
Just an FYI: The upper case thing became a minor issue so I added an option for it. It defaults to on, so nothing really changes except that you can turn it off if you want: https://github.com/openpnp/openpnp/blob/develop/CHANGES.md#2017-03-05

dzach:
@vonnieda That's the best solution. 
On a side note, I've been busy designing a low entry pnp machine for myself. I abandoned the idea of using the 3D printer for this task because it involved mounting and unmounting heads, cameras and cables all inside a temperature controlled chamber, and the hassle was way more than the gain. I hope to start testing it soon.

PR #441, opened 2017-02-02
Merge remote-tracking branch 'refs/remotes/openpnp/develop' into develop

sp-apertus:


vonnieda:
Nice work @sp-apertus, looks very good! I have a couple suggestions:

1. Change case of class name to HttpActuator. I know it's picky, but that's the convention we use.
2. Instead of duplicating all the code from ReferenceActuator, just change your class to extend ReferenceActuator, then you could remove all of the duplicated code.  You would just need to override actuate(boolean) and getConfigurationWizard() and include the properties specific to your class.

vonnieda:
Merged this after a bit of clean up in the identifiers and formatting. One note, @sp-apertus, I noticed that you have the "on" property marked as an @Element. This means the state of the field will be stored in the machine.xml and will be retained between runs. Meaning that if the actuator is set to on and OpenPnP is exited, next time it started it will still think it's on. Is this what you intended?

sp-apertus:
Thanks!

It's intended this way but I am open to community feedback if this makes sense.
For us the feeders that will get activated this way will will be toggled on and off for every feed operation but I imagine if other people have some mechanical actuators, like a drag pin then its important that the machine remembers if it had be retracted already after software restart....

vonnieda:
@sp-apertus My concern is that the class can get "stuck" by being in an on state, even if the hardware has been reset. In addition, I think that in the majority of cases it doesn't matter if you send the "on" command when the device is already on, and likewise off. Although I assume that since you included this functionality you must have had a reason for it. Can you give me some details on that?

I guess you can always fix the stuck case by sending the off command, so maybe this isn't an issue.

sp-apertus:
I will test it with our feeder system in about a week and report back if we see any issue with the way it is now.

vonnieda:
Sounds good, I won't worry about it for now unless someone else has an issue with it.

PR #453, opened 2017-03-01
Adding a SystemLogger class for redirecting stdout and stderr to tinylog, related to #288

pfried:
Hey there,

i managed to redirect the output, but there are some things to know about this:

1. No overwrite of the `System.out.write(int b)` primitive char printing as this does not make sense in the LogTab.
2. I had to change the addWriter to writer because the consoleWriter which is default attached by tinylog will create a circular dependency: System.out=> SystemLogger => Logger (Console Writer) => System.out
3. Initialization in Main class as the first logging is done there and i do not want config changes in between setup logging and setup of Log Panel.

Class was named SystemLogger because tinylog uses reflection and the class name will be used for tagging the LogEntry. I found no way to overwrite this in a simple way.

 Unfortunately there is no generic log function where one could pass a LogLevel, therefore it is used in the constructor of the SystemLogger to be evaluated on logging.

Best, Friedrich

### DIFF
--- a/src/main/java/org/openpnp/Main.java
+++ b/src/main/java/org/openpnp/Main.java
@@ -25,8 +25,10 @@ import java.io.File;
 import javax.swing.UIManager;
 
 import org.openpnp.gui.MainFrame;
+import org.openpnp.logging.SystemLogger;
 import org.openpnp.model.Configuration;
 import org.pmw.tinylog.Configurator;
+import org.pmw.tinylog.Level;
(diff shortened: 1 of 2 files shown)


pfried:
related to #288

vonnieda:
Very nice work @pfried! I had tried to get this done before but I couldn't figure out the change of addWriter to writer. That seems to have done the trick. Thank you!

Issue #382, opened 2016-12-19
Wiki Images 2 - Ignore


vonnieda:

vonnieda:
![screen shot 2016-12-18 at 8 37 41 pm](https://cloud.githubusercontent.com/assets/1182323/21299497/e30e6d82-c561-11e6-8f42-1663c2f994d0.png)

vonnieda:
![screen shot 2016-12-18 at 8 39 04 pm](https://cloud.githubusercontent.com/assets/1182323/21299519/0c806c56-c562-11e6-8a51-5e4df3092a5d.png)

vonnieda:
![screen shot 2016-12-20 at 5 53 55 pm](https://cloud.githubusercontent.com/assets/1182323/21372675/562c4ae6-c6de-11e6-8071-86b126f78b95.png)

![screen shot 2016-12-20 at 5 37 16 pm](https://cloud.githubusercontent.com/assets/1182323/21372503/1c1ac8f6-c6dd-11e6-89ee-64e922fbcdcf.png)

![screen shot 2016-12-20 at 5 40 57 pm](https://cloud.githubusercontent.com/assets/1182323/21372502/1c11244a-c6dd-11e6-9f4a-fb37ba47c5e0.png)

vonnieda:
![screen shot 2017-01-09 at 7 59 34 pm](https://cloud.githubusercontent.com/assets/1182323/21791132/78884716-d6a6-11e6-8579-fe2021c927b5.png)

![screen shot 2017-01-09 at 7 59 00 pm](https://cloud.githubusercontent.com/assets/1182323/21791153/962a6e0c-d6a6-11e6-82b7-ba252e94cc79.png)

vonnieda:
![screen shot 2017-01-09 at 8 03 00 pm](https://cloud.githubusercontent.com/assets/1182323/21791175/b900e7da-d6a6-11e6-8979-17308c278d51.png)

vonnieda:
![screen shot 2017-01-09 at 8 08 20 pm](https://cloud.githubusercontent.com/assets/1182323/21791296/674a164a-d6a7-11e6-82ef-b38fa4edb9c6.png)

vonnieda:
![screen shot 2017-01-09 at 8 15 26 pm](https://cloud.githubusercontent.com/assets/1182323/21791442/6253b302-d6a8-11e6-806b-222fec908b91.png)

vonnieda:
[...]

PR #456, opened 2017-03-02
Autopanelize 

BendRocks:
Finally completed verification of panel fiducials.  Discussion on feature is at the link below.

https://groups.google.com/forum/#!msg/openpnp/_ni0LK8LR8g/5u-0-P-1EwAJ;context-place=forum/openpnp


cri-s:
Hi, just looked over, i think you have a problem with boadLocation. I don't know if you have tested it.
The problem presumly is this, you copy boardlocation, but FILE remain the same. If it save the state after
partial assembly, or all assembly, but not all boards, and later reload, all boards have the same and wrong placed parts.
Usually it's required that at least on scripting i have implemented it, for every board, there is a dedicated file.
Otherwise if assembly is not 100% complete or not done, there are problems.
later i'm passed to @@3:2:1 notation, that resolve the problem differently.

Regards
Chris

vonnieda:
@cri-s What you are describing is not functionality that is present in OpenPnP. OpenPnP doesn't currently do state in BoardLocation. That is only in your branch. Having multiple BoardLocations with the same Board is fine. You can see an example of this in the file that is included with OpenPnP and used for all automated tests: https://github.com/openpnp/openpnp/blob/develop/samples/pnp-test/pnp-test.job.xml

cri-s:
You have absolutly right. I store the info if component is placed or
not and you don't do that.


2017-03-03 16:35 GMT+01:00, Jason von Nieda <notifications@github.com>:

vonnieda:
Adding some notes here while I work on merging this:
* Remove board button enabled state is being managed incorrectly now. The remove button is enabled when no board is selected. Related to code around JobPanel#929.

src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
```diff
@@ -177,40 +183,49 @@ public Location getHomeFiducialLocation(Location location, Part part) throws Exc
...
+	private static Location getFiducialLocation(BoardLocation boardLocation, Placement fid, Length fidDia) throws Exception {
         Camera camera = Configuration.get().getMachine().getDefaultHead().getDefaultCamera();
 
         Logger.debug("Locating {}", fid.getId());
 
+		Footprint footprint;
+
+		// Careful...magic number below. Also used in JobPanel. This are for
+		// global panel fiducials, which do not have an entry in the PCB BOM.
+		if ((fid.getId().equals("PanelFid1_a322") || fid.getId().equals("PanelFid2_g301"))) {
```
vonnieda: Are these special IDs intended to be left in the code? What is their purpose?

BendRocks: Philosophically, you probably need to answer the question on how the user should enter panel fiducials. To me, __PCB__ fids as part of the BOM make 100% sense, and thus entering them as parts and having corresponding  part which then has a corresponding package makes sense.  In a panel, they show up in every board, just as you'd expect (because the fids are part of the BOM).

__Panel__ fiducials, however, are treated differently in every layout package I've worked on. As you see the screen capture from the Altium package that was posted, there a separate GUI to specify the panel. And that's true in Mentor, diptrace, etc. You have a dedicated dlg to enter all the panel info. 

Now, I opted to add a dlg to handle the panel parameters and part of that was specifying the panel fids locations (pcb fids stay the same). 

if you decide you don't like the panel fids being specified in a separate dlg, then it seems (not certain, however) that you need to deal with specifying panel fids in just one of the boards in the array. That would preserve the way all the software in openpnp works today, but you still need the dlg, and then you have to explain to users that panel fids need to be in one of the panels....in other words, the rules get complicated.

And you already need a dlg to specify the panel parameters....

Re: special IDs....the problem I ran into was that we have the global panel fids the user entered into the panel specification. How to get those all the way down to the RefFidLoc() routine. They were passed as placements (but not part of the list of placements) with special names. And when those special names are encountered down in RefFidLoc(), the method knows to create the footprint with the given diameter. 

If you opt to specify panel fids some other way that that avoids this, I could certainly re-work. But think about the user implications of breaking apart where all this info sits. With packages.xml and parts.xml being treated as global databases, it already gets to be a bit messy because fids (and parts) on Project A that was created by engineer X could very well be overwritten the next day by engineer Y working on Project B (and when engineer A returns to the machine, his work is polluted).  Long term, if packages and parts are moved to per-job, then this concern is lessened.  But right now it gets quite messy even for me on the few boards I've been through thus far. I think it much better to have all the panel details held in the board file instead of spread across 3 files (board, package, part), 1 of which is project-based and the other 2 which are global. 

The other items I can certainly tidy up, but first please share your thoughts on how you'd like to see the user enter the location of the panel fids

vonnieda: Hi @BendRocks, thanks for your detailed response. I agree with most of your points. Here is my current thinking on this:

Biggest issue is the fiducial handling:
* Current code only supports round fids. Certainly the most common, but it's not rare to have others.
* An in-process PR adds pipeline configuration for fids, allowing vision customization. Happens at Part level, like bottom vision. Skipping Part/Package for panel fids will make this feature not work for panels.
* The hardcoded placement IDs are a brittle way to handle the problem. Gut feeling is that we should just pass the BoardLocation like we do now and if the fid locator sees that the BoardLocation is a panel it uses panel fids instead of board fids.
* Fids are not checked during job startup, and there's no way to indicate that they should be. This could be considered an enhancement down the road - not required for v1.

Here is roughly how I'd like to see this feature change in the code:
* Panel class (rename of PCBPanel):
    * Contains panel info such as columns, rows, xGap, yGap.
    * Adds IdentifiableList<Placement> placements for panel fids. Can also be used in future for non-board placements, if needed.
    * Adds "Check Fids?" checkbox which indicates the panel fids should be checked at job start.
* Job gets IdentifiableList<Panel> panels.
* BoardLocation gets panelId which refers to a Panel in Job.panels. We'll also materialize a BoardLocation.panel property during job load so that BoardLocation references can get you back to the Panel. This will be used in the fid checker.

And here is how it would all work:
* Dialog and buttons all remain basically the same as they are now.
* Dialog can look up the panel info based on a BoardLocation selection via BoardLocation.panel.
* Remove the fid diameter and add fid part dropdown. The dialog saves two placements into the Panel.placements with the specified coordinates and part.
* FiducialLocator API can just get a boolean checkPanel which tells it to do a panel check rather than an individual board check. It uses BoardLocation.panel to find the panel fids and then everything else works basically the same.

I think this will make the feature much more flexible and allow it to be grown in the future as needed. It will also make it compatible with the upcoming fiducial vision changes, which is important. And I don't actually think it is a very large set of changes from what you have now.

I know you said you have some other pressing work at the moment, so if this is too much to get to right away we can either sit on it for a while or I can work on it. I think I can probably refactor it into the above in a night or two and I'd be happy to do that to get the feature merged in before it gets too stale.

Let me know what you think. This is just my thinking on it and it's not a dictation. Completely open to discussion.

BendRocks: This is very sensible and clear to me. I think I can do on Saturday no problem. 

So, the only change to the fidloc API would be to revert back to the old interface (before this PR) and then add a bool to indicate panel, right? And then I'd find all the places that used the existing pre-PR interface and make the change there, right? I ask because (from memory) there are drivers that are calling this interface. Just making sure I understand your direction here.

Thanks again for the well reasoned response.

vonnieda: Right, so the interface goes from:
```
public Location locateBoard(BoardLocation boardLocation) throws Exception;
```
to 
```
public Location locateBoard(BoardLocation boardLocation, boolean checkPanel) throws Exception;
```

and in the implementation, if that boolean is true, we just check panel fids instead of board fids.

Another option is to just duplicate that method in the interface, add the parameter and in the implementations default to false. This would mean that no existing code has to change. You would just add the additional call into your button and to the ReferenceJobProcessor at the start of the job. This may be a better option.

So, the interface becomes:
```
public interface FiducialLocator extends PropertySheetHolder {
    public Location locateBoard(BoardLocation boardLocation) throws Exception;
    public Location locateBoard(BoardLocation boardLocation, boolean checkPanel) throws Exception;
    public Location getHomeFiducialLocation(Location location, Part part ) throws Exception;
}
```

And in ReferenceFiducialLocator:
```
    public Location locateBoard(BoardLocation boardLocation) throws Exception {
        return locateBoard(boardLocation, false);
    }

    public Location locateBoard(BoardLocation boardLocation, boolean checkPanel) throws Exception {
        // Find the fids in the board
        IdentifiableList<Placement> fiducials = getFiducials(boardLocation);
...
```

BendRocks: I'm stuck on this part that you wrote: "BoardLocation gets panelId which refers to a Panel in Job.panels. We'll also materialize a BoardLocation.panel property during job load so that BoardLocation references can get you back to the Panel."

This suggests to me that BoardLocation will have both a panelId string and a panel property and that at load time the panelId will be turned into a panel instance, is that correct? If so, where precisely does that take place during job load? 

What I'm a bit confused on is if the BoardLocation knows the panel ID, it could just find the panel instance at when asked to locateboard(), yes? But how precisely?

Thanks

vonnieda:
I think your suggestion of doing it on the fly is a completely reasonable
solution for this.

Jason

vonnieda:
Hi @BendRocks, with this latest commit are you ready for this to be merged, or are you still working?

BendRocks:
I think I have about an hour left. I'll aim to finish tonight or tomorrow night. Sorry for the delay.

BendRocks:
Hi Jason, here it is pretty much complete. At this stage, I'd hope you could make a pass through and say "yay or nay" that this captures your plans above. There's still more testing on my side and some clean up related to coding styles etc. I also need to get the fid check checkbox in the dlg.

Question 1: Around line 1446 in JobPanel, I have a part or partid but I cannot figure out how to get that to auto select the same in the combo box. Can you suggest a way? IOW, I need the combo box to remember the last setting for part. 

Question 2: What is your preferred etiquette on pull requests: Should I deliver something that can be automatically merged with dev branch? On my end, this would mean pulling in stuff that is unrelated to the pr. I'm not sure if that would confuse you ("wth does this code do?") OR if it just makes things easier on your end because it doesn't even show up in the diff

I've attached sample project xml. It's probably easier to look at that and know for sure if this is what you outlined or not. 

Note that the auto-panelize only works if you have a single board in the board list. When you selected more than 1 row or column in the dlg, then the feature becomes active and you'll see the other boards are autopopulated and cannot be edited. When active, you can only edit line zero of the table and all the other lines follow. if you go back and change the row and col to 1, then the feature turns off. When the feature is on, you'll see the extra stuff in the xml. When it's off, it should be just like it was before. 

Also, there are a few places in the code where you'll see it's assumed there's a single panel. The plumbing is there to allow N panels, but the UI is not. 




[TestJob2.job - Copy.xml.txt](https://github.com/openpnp/openpnp/files/846820/TestJob2.job.-.Copy.xml.txt)

vonnieda:
Hi @BendRocks, I'm reviewing this now. First, answering your questions above:
1. The core issue here is that you are using a Part that isn't in the Part database. The PartsComboBoxModel only includes Parts from parts.xml, so when you try to select the Part you've defined in the Panel it can't find it in the list so there is nothing to select. More about this below...
2. Yes, please pull develop in and resolve conflicts if possible. The end result is that the PR will only include the diffs from develop. All the stuff between your branch and the develop branch stuff will be seen as "unchanged" when the comparison is made and will not be included in the PR.

A note before I forget: One of the conflicts includes a really important bit of code that might be easy to miss. In the boardLocations selection listener, there is this line:
`Configuration.get().getBus().post(new BoardLocationSelectedEvent(boardLocation));`
which in develop is:
`Configuration.get().getBus().post(new BoardLocationSelectedEvent(boardLocation, JobPanel.this));`

We need to make sure that doesn't get missed when this is merged. That is a fix for a bug that causes massive performance problems and crashes.

Okay, general thoughts:

Overall, looks very good. It's exactly what I was thinking and it's very, very close to being ready to merge, I think.

The Part in Panel should go away. The Placements in Panel should just reference a Part from parts.xml. There should be no reason to define a Part for a fiducial differently for a panel versus a board. A 1mm dot is a 1mm dot, basically. This is what I meant by "Remove the fid diameter and add fid part dropdown. The dialog saves two placements into the Panel.placements with the specified coordinates and part." although maybe I wasn't clear there.

The idea is that in the Panel Placements there would just be a dropdown to select the Part that is used for the fiducial. It appears you've already done this but instead of setting the partId on the placement, you are storing the part itself. There's no need to do this. So, remove Part from panel and just set the partId on the placements. The Part comes from the combo box, which gets it's list from the parts database. Basically, this models exactly the way placements work in the job, you just have a second list of placements in the panel.

If that's still not clear, let me know. This is close enough that I can probably fix it up in an hour.

vonnieda:
I'm going to spend a few minutes and see if I can fix the above issues.

vonnieda:
Pushing commits to https://github.com/openpnp/openpnp/tree/BendRocks-develop_autopanelize_PR. Conflicts are resolved.

BendRocks:
Thanks Jason, much appreciated. Your point on the parts versus placements makes sense. The architecture is starting to gel in my head ;)

BendRocks:
I have pushed up the to-do items here https://github.com/BendRocks/openpnp/tree/BendRocks-develop_autopanelize_PR

What I did was pull the branch you made above down to my PC, made the changes there, then pushed to my new fork on github. I get the feeling that if I do another pull request it will mess up this thread continuity, so I trust you can take the branch linked above. But if there's something else I need to do just let me know.

One area is not correct that I'd like your comment on  is how the panel rotation (determined in RefPNPJobProc) gets propagated to the entire panel.  Around line 357 of RefPnpJobProc you'll see I copy the board code, which will adjust that one board. But how to adjust the other boards in the panel?

You'll see around lines 906 in JobPanel that is where I adjust each derived board location based on the root panel. For example, you have a root panel, and you panelize to 3x3. Then you run the manual panel fid check and it gets the adjustment (offset and angle) and applies to each board. 

So, the JobPanel method PopulatePanelSettingsIntoBoardLocations() needs to run. What I am thinking is that perhaps that adjust code should move to Panel or, preferably, Job, and then when using panels, after board0 has been updated, the same adjustments could be done to all other boardlocs in the job array.

Does that seem OK?

vonnieda:
@BendRocks Do you see this working like the follow?
1. Job is started.
2. Panel fid check is performed, giving us the location of the panel.
3. Each board in the panel has it's location updated based on the panel location + derived offsets.
4. Fid checks are then performed on each board individually to zero in on that specific board?

If so, then yes, I think what you are suggesting makes sense. I think what I would do is add setLocation(Location, Job) to Panel. Then after running a panel fid check  you would call Panel.setLocation(newLocation, job) and this method would update each board in the panel.

BendRocks:
Yes, that's right. 

OK, done and pushed up to the link below and it can be merged without conflict. I didn't do a PR because I think it would screw things up and give you a second PR. Is that in fact the case?

https://github.com/BendRocks/openpnp/tree/BendRocks-develop_autopanelize_PR

I've attached a rotated panel if you want to try this. 

X spacing is 0, Y spacing is 1. 3x3 array. Board size is 50x20mm. Fid1.X is 3, Fid1.Y is -5, Fid2.X is 147, Fid2.Y is 67. Fid diameter is 1mm

1) Load this file as your imagecamera. Cal is 0.109 x 0.108
2) Create a new job. Add a board to the job
3) Select panalize option and enter the data above
4) Pick the panelize fid button and it should populate each board with X,Y and rot

srwp3_rot3

BendRocks:
Jason, if you'd like a doc written for this, just let me know where you'd like to see it in the wiki user manual and I'll add it.

vonnieda:
Hi @BendRocks, you can add documentation to https://github.com/openpnp/openpnp/wiki/Panelization. Once it's there I'll link in various places that makes sense.

For follow up PRs, what's best is to do the work in a feature branch (which it looks like you did) and then just add and push additional commits to that branch. Github with automatically update the PR. You can see a note about this right above the comment box:

screen shot 2017-04-01 at 9 30 16 pm

I'm looking through this now and will work on getting it merged.

vonnieda:
Hi @BendRocks,

First, thank you for all the hard work and perseverance you put into this. I really appreciate your professional attitude and your handling of criticism and feedback.

I went ahead and merged this (with some changes) as I felt it was to the point where people could start using it, but I want to point out some changes that I'd like made fairly rapidly. If you are still up for working on this and want to address these that would be awesome, but if you are sick of working on this then I'll be happy to make these changes.

I've added the changes to a new ticket at: https://github.com/openpnp/openpnp/issues/501

The changes I made before merging were mostly with regards to formatting, variable and method naming. I also adjusted the dialog size as the OK and Cancel buttons were hidden on my screen and the part dropdown was getting cut off too much.

Thanks again for all your work on this!

BendRocks:
> First, thank you for all the hard work and perseverance you put into this. I really appreciate your professional attitude and your handling of criticism and feedback.

And thanks to you for making the feature better and explaining things so clearly. 

> but if you are sick of working on this then I'll be happy to make these changes.

Not sick of it at all, but limited in time with some other features I need to start on, so if you see a path, please do it! Java UI is really, really slow going for me, and I think the stuff you'd like to see (support for N panels, especially) is UI heavy.

Thanks again for your guidance.  I look forward to adding more.

vonnieda:
@BendRocks I'm not sure if you are aware, but I use [WindowBuilder](https://eclipse.org/windowbuilder/) for doing all the GUI design. If you are working without it you are probably working too hard :) By moving the dialogs into their own files, which I just finished, you can use WindowBuilder with these too. Just right click the file and Open With -> WindowBuilder Editor. It makes everything a lot easier, and is basically drag and drop.

BendRocks:
> For follow up PRs, what's best is to do the work in a feature branch (which it looks like you did) and then just add and push additional commits to that branch. Github will automatically update the PR. You can see a note about this right above the comment box:

But in this case the initial PR was on one branch, then you did an intermediate merge and it changed to another branch name. I pulled that branch down, modified and I pushed back to my repo...should I have pushed that back to the original branch name (if even possible?) or should I have submitted that as a pull request?

vonnieda:
Yea, in this case it was more complex. Best would have been to pull my
changes into your original branch and then continue pushing changes to that
same branch. As the commits were pushed the PR would update.

BendRocks:
>, you can add documentation to https://github.com/openpnp/openpnp/wiki/Panelization. Once it's there I'll link in various places that makes sense.

Where do you want art assets? Looking just now, some dump them all in a message and link to that because you can drag and drop. The art in your wiki posts point to github cloud, but I can't see how to get stuff there. Thanks

vonnieda:
@BendRocks What I do is create an issue named like "Wiki Images", close it and then drag and drop images as comments to that issue. They will get converted to a URL that you can use. You can add as many to an issue as you like. For example, here is one of mine: https://github.com/openpnp/openpnp/issues/382

BendRocks:
OK, thanks. Wiki page first version done

https://github.com/openpnp/openpnp/wiki/Panelization

vonnieda:
Thanks @BendRocks, I've linked it in the [[User Manual]].

PR #457, opened 2017-03-03
Center camera and move to next part

BendRocks:
This was sketched out here with a picture: https://groups.google.com/forum/#!topic/openpnp/SpG09TKEVWg. The goal is to augment the "move camera to board" and "move camera to part" buttons with similar buttons with an arrow that performs the same action BUT then it advances the board or part table to the next line. The buttons also keep focus, allowing you to quickly move the camera through the PCB origins or part origins using the space bar so that you can verify the board origins in a panel, for example, are correct. And in part mode, this allows you to quickly zip from part to part and using the rotating gratical you can quickly see if part rotations are as expected.

Without this feature, you had to bounce between "move camera" button and then go click the next line in the table, and then back to "move camera" button. Very tedious. 

Changes needed were minor. One challenge was in order to ensure focus isn't lost on table row change, the focus had to first be captured, then the table changed, then the focus restored. Otherwise the focus bounced to seemingly random controls and the spacebar advance couldn't work. In the JobPanel this was straightforward to fix. In the JobPlacement panel, there wasn't a way to get to the frame, and so I added the frame arg to the constructor of the JobPlacementPanel.

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -283,7 +283,7 @@ public class JobPanel extends JPanel {
         pnlRight.add(tabbedPane, BorderLayout.CENTER);
 
         jobPastePanel = new JobPastePanel(this);
-        jobPlacementsPanel = new JobPlacementsPanel(frame, this);
+        jobPlacementsPanel = new JobPlacementsPanel(this);
 
         add(splitPane);
 
@@ -1040,15 +1040,16 @@ public class JobPanel extends JPanel {
 				// Need to keep current focus owner so that the space bar can be
 				// used after the initial click. Otherwise, button focus is lost
 				// when table is updated
-				Component comp = frame.getFocusOwner();
+				Component comp = MainFrame.get().getFocusOwner();
 				HeadMountable tool = MainFrame.get().getMachineControls().getSelectedTool();
 				Camera camera = tool.getHead().getDefaultCamera();
 				MainFrame.get().getCameraViews().ensureCameraVisible(camera);
 				Location location = getSelectedBoardLocation().getLocation();
 				MovableUtils.moveToLocationAtSafeZ(camera, location);	
 				Helpers.selectNextTableRow(boardLocationsTable);
-				if (comp!=null)
+				if (comp!=null){
 					comp.requestFocus();
+				}
 			});
 		}
 	};
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -21,6 +21,7 @@ package org.openpnp.gui;
 
 import java.awt.BorderLayout;
 import java.awt.Color;
+import java.awt.Component;
 import java.awt.FileDialog;
 import java.awt.Frame;
 import java.awt.Rectangle;
@@ -282,7 +283,7 @@ public class JobPanel extends JPanel {
         pnlRight.add(tabbedPane, BorderLayout.CENTER);
 
         jobPastePanel = new JobPastePanel(this);
-        jobPlacementsPanel = new JobPlacementsPanel(this);
+        jobPlacementsPanel = new JobPlacementsPanel(frame, this);
 
         add(splitPane);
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/openpnp/gui/JobPanel.java
```diff
@@ -1024,6 +1028,30 @@ public void actionPerformed(ActionEvent arg0) {
...
+				// used after the initial click. Otherwise, button focus is lost
+				// when table is updated
+				Component comp = frame.getFocusOwner();
+				HeadMountable tool = MainFrame.get().getMachineControls().getSelectedTool();
+				Camera camera = tool.getHead().getDefaultCamera();
+				MainFrame.get().getCameraViews().ensureCameraVisible(camera);
+				Location location = getSelectedBoardLocation().getLocation();
+				MovableUtils.moveToLocationAtSafeZ(camera, location);	
+				Helpers.selectNextTableRow(boardLocationsTable);
+				if (comp!=null)
```
vonnieda: Please always use optional braces: https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style

src/main/java/org/openpnp/gui/JobPlacementsPanel.java
```diff
@@ -72,7 +75,8 @@
     private static Color statusColorReady = new Color(157, 255, 168);
     private static Color statusColorError = new Color(255, 157, 157);
 
-    public JobPlacementsPanel(JobPanel jobPanel) {
+    public JobPlacementsPanel(JFrame frame, JobPanel jobPanel) {
+    	this.frame = frame;
```
vonnieda: I've been moving to MainFrame.get() rather than passing the frame around everywhere. You can also use getTopLevelAncestor() in UI classes. No need to change this, just keep in mind for the future.

src/main/java/org/openpnp/gui/JobPlacementsPanel.java
```diff
@@ -325,6 +332,30 @@ public void actionPerformed(ActionEvent arg0) {
...
+				// used after the initial click. Otherwise, button focus is lost
+				// when table is updated
+        		Component comp = frame.getFocusOwner();
+                Location location = Utils2D.calculateBoardPlacementLocation(boardLocation,
+                        getSelection().getLocation());
+                Camera camera = MainFrame.get().getMachineControls().getSelectedTool().getHead()
+                        .getDefaultCamera();
+                MovableUtils.moveToLocationAtSafeZ(camera, location); 
+                Helpers.selectNextTableRow(table);
+                if (comp!=null)
```
vonnieda: Optional braces: https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style

src/main/java/org/openpnp/gui/support/Helpers.java
```diff
@@ -68,6 +68,21 @@ public static void selectLastTableRow(JTable table) {
         index = table.convertRowIndexToView(index);
         table.addRowSelectionInterval(index, index);
     }
+    
+    public static void selectNextTableRow(JTable table){
+    	 int index = table.getSelectedRow();
+    	 
+    	 if (index == -1)
```
vonnieda: Optional braces: https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style

src/main/java/org/openpnp/gui/support/Helpers.java
```diff
@@ -68,6 +68,21 @@ public static void selectLastTableRow(JTable table) {
...
+    		 index = 0;
+    	 
+         table.clearSelection();
+         if (++index > table.getRowCount()-1)
```
vonnieda: Optional braces: https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style

BendRocks:
This should address your issues. Sorry about the coding style points, I should have remembered.

vonnieda:
Thanks @BendRocks, this is a really nice addition!

vonnieda:
Just noticed a minor bug in this. It doesn't respect table ordering when sorting has been applied. I'll see if I can figure out the fix real quick.

vonnieda:
Fixed.

BendRocks:
Good catch, thanks. Sorry I missed. There's also an issue where it "stalls" (stops advancing) if you were on top parts and it makes its way to bottom parts in the placement list. I guess it could wrap around back to the top parts (or vice versa), but I opted to leave it as is.

PR #458, opened 2017-03-12
Adding more functionality to the Logging Panel

pfried:
Related to #288 

# General Changes

- Adding a ConsoleWriter which will print to the original stdout stderr PrintStreams, removing this in the new SystemLogger
- Changing LogPanel to a rendered JList which is now filterable and has references to the original LogEntries
- Removed Line Limit, lines limited to 10k entries in the LogEntryListModel

# UI Changes

- Split Global Settings and Filter Settings into two Panels
- Changing popup menus to select boxes for logging levels
- Removed toolbar elements in Logging (not floatable anymore), replaced by Panels

# Custom Filters

- Adding a Search TextField, searches full error message, case insensitive
- Adding a Checkbox to enable System.out and System.err
- Adding a button to enable auto scrolling on the LogPanel
- Adding a button Copy to Clipboard of filtered Items

# Notes
I did not want to save all settings to the preferences as i think that might not be good as some a too volatile and the danger is high that someone forgets to reset his filters.


pfried:
Preview:

image

vonnieda:
This is awesome! Nice work! I'll do a full review this evening and work on getting this merged. Thanks!

vonnieda:
Noting a few small issues:

* The panel is a little wide and does not fit well on a laptop screen. Several buttons are clipped. Maybe just put the buttons on the next line:
screen shot 2017-03-12 at 9 51 13 pm
* Selecting several lines and hitting Cmd-C (on Mac) copies...
```
org.pmw.tinylog.LogEntry@75e7b5f4
org.pmw.tinylog.LogEntry@75a1d560
org.pmw.tinylog.LogEntry@26fb5c7f
org.pmw.tinylog.LogEntry@9dbcca8
```
...instead of the actual log entries. This can probably be resolved by placing a custom object with an overridden toString() on it in the log list instead of the LogEntry objects.
* Auto Scroll should be a checkbox instead of a stateful button. The stateful buttons have proven confusing for users over the years due to poor rendering on Windows and Linux.

Other than that, the functionality seems great. It's a very nice improvement! I'm going to review the code itself now, too.

src/main/java/org/openpnp/gui/LogPanel.java
```diff
@@ -1,175 +1,289 @@
...
+        filterPanel.add(filterContentPanel, BorderLayout.WEST);
+
+        JPanel filterControlPanel = new JPanel(new FlowLayout(FlowLayout.RIGHT));
+
+        JButton btnClear = new JButton("Clear Panel");
+        btnClear.addActionListener(e -> logEntries.clear());
+
+        filterControlPanel.add(btnClear);
+
+        JButton btnCopyToClipboard = new JButton("Copy to clipboard");
```
vonnieda: This could just use the Icons.copy icon and skip the text to save some space.

src/main/java/org/openpnp/gui/LogPanel.java
```diff
@@ -1,175 +1,289 @@
...
+        btnCopyToClipboard.addActionListener(e -> {
+            StringBuilder sb = new StringBuilder();
+            logEntries.getFilteredLogEntries().forEach(logEntry -> sb.append(logEntry.getRenderedLogEntry()));
+            copyStringToClipboard(sb.toString());
+        });
+
+        filterControlPanel.add(btnCopyToClipboard);
+
+        JToggleButton btnScroll = new JToggleButton("Auto Scroll");
```
vonnieda: Maybe make this a new icon. Just noting it here, if you don't want to mess with icons I can always throw that in later.

pfried:
Added a commit, some notes

The autoscroll functionality might not be the best, if autoscroll is enabled it is fine, but if turned off i feel that the behavious is not very good, it will scroll to the top instead of staying where it was. This is tricky in general, as the list will be clipped at 10k entries and behaviour needs to adapt then.

Also the button thing breaks with the Current style, maybe remove it and replace it with a Checkbox with Text ("Autoscroll").

vonnieda:
@pfried Sorry for the delay in reviewing this. Been a busy couple days. I have not forgotten it and will get it done as soon as I can.

vonnieda:
@pfried Thanks for this work. This is a really great improvement on the logging system!

I did notice a couple minor issues, so I will add those to the original ticket, but they are minor enough that I didn't think it was worth delaying this anymore.

vonnieda:
@pfried Please see https://github.com/openpnp/openpnp/issues/467#issuecomment-287779796 for comments related to a performance problem from this PR. Note that while the filtering is the obvious place to look for what is going on, I think it might be worth checking out how auto scroll is handled as well, as I saw the list jumping around quite a bit while running a job.

pfried:
I just tested it, i am pretty sure it is the autoscroll functionality, if i disable it, even during a run it is stable again

PR #459, opened 2017-03-13
Update AbstractPnpJobProcessor.java

cri-s:
Updated nozzleCanHandle for Nozzles without autochanger,
Updated compare on feeder to compare only getId() , different part of same components are usual, different height, speed, reference
image.
Updated function requiring head, that if called with head=null searches all heads.

### DIFF
--- a/src/main/java/org/openpnp/spi/base/AbstractPnpJobProcessor.java
+++ b/src/main/java/org/openpnp/spi/base/AbstractPnpJobProcessor.java
@@ -8,6 +8,7 @@ import org.openpnp.spi.Machine;
 import org.openpnp.spi.Nozzle;
 import org.openpnp.spi.NozzleTip;
 import org.openpnp.spi.PnpJobProcessor;
+import org.openpnp.machine.reference.ReferenceNozzle;
 import org.openpnp.util.MovableUtils;
 
 public abstract class AbstractPnpJobProcessor extends AbstractJobProcessor
@@ -22,7 +23,7 @@ public abstract class AbstractPnpJobProcessor extends AbstractJobProcessor
      * @throws Exception
      */
     public static void discardAll(Head head) throws Exception {
-    	for (Head hd : machine.getHeads()) {
+    	for (Head hd : Configuration.get().getMachine().getHeads()) {
 			if(head==null||hd==head) {
 				for (Nozzle nozzle : head.getNozzles()) {
 					discard(nozzle);
@@ -70,7 +71,7 @@ public abstract class AbstractPnpJobProcessor extends AbstractJobProcessor
 	public static boolean nozzleCanHandle(Nozzle nozzle, Part part) {
 		if(nozzle==null||nozzle.getNozzleTip()==null) return false;
 	 
-		if(nozzle.isChangerEnabled()) {
+		if(((ReferenceNozzle)nozzle).isChangerEnabled()) {
 			for (NozzleTip nozzleTip : nozzle.getNozzleTips()) {
 				if (nozzleTip.canHandle(part)) {
 					return true;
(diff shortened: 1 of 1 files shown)


src/main/java/org/openpnp/spi/base/AbstractPnpJobProcessor.java
```diff
@@ -36,7 +49,12 @@ public static void discard(Nozzle nozzle) throws Exception {
                 Configuration.get().getMachine().getDiscardLocation());
         // discard the part
         nozzle.place();
+		// repeat the air puff to clean filters over discard location
```
vonnieda: This is too machine specific and should be handled at the driver level. It will cause solenoids to trigger repeatedly and will confuse users who do not have this filter.

src/main/java/org/openpnp/spi/base/AbstractPnpJobProcessor.java
```diff
@@ -84,8 +113,9 @@ public static NozzleTip findNozzleTip(Head head, Part part) throws Exception {
      * @throws Exception If no Feeder is found that is both enabled and is serving the Part.
      */
     public static Feeder findFeeder(Machine machine, Part part) throws Exception {
+		if(part!=null)
         for (Feeder feeder : machine.getFeeders()) {
-            if (feeder.getPart() == part && feeder.isEnabled()) {
+            if (feeder.getPart().getId().equals(part.getId()) && feeder.isEnabled()) {
```
vonnieda: What is the purpose of this change? There should never be more than one instance of Part with the same ID in the configuration, so this should not be needed.

src/main/java/org/openpnp/spi/base/AbstractPnpJobProcessor.java
cri-s: It cannot handle by the driver as it is allowed only on discard location, it is for machines with
air puff  for 40ms as example. If machine don't have it, there is only a M9 statement
or similar, and any repeated call don't actuate any solenoid in this case after the first place.
If the nozzle hold one component, the solenoid is deactivated, otherwise it remains deactivated.
I can remove it.

src/main/java/org/openpnp/spi/base/AbstractPnpJobProcessor.java
cri-s: Sure that this exist. Example are parts with different height. 
Different batches have 0.3mm height differences as allowed from datasheet,
i never seen more then 0.2mm. As example Murata, if part comes from 
Germany or Hungary the part thickness is a lot different, millimeters, not thents of millimeters.
For resistor arrays, concave and convex as example have different footprint, or if part are
from differnt manufacture, description holds order part number, .... , parts are always different.

cri-s:
Why it's needed in first place, the nozzleCanHandle function assume that nozzle use always autochanger,
and return that it can handle the part even it's not allowed to change NozzleTip.

cri-s:
>Please also review https://github.com/openpnp/openpnp/blob/develop/CONTRIBUTING.md and explain 
what this change does and why it is needed
If this is too difficult to understand it makes no sense going further.
Most important change is this, but there are other problems with planner on machine with more nozzles.
https://groups.google.com/forum/#!topic/openpnp/2sNx7RmkRvM

In case you don't think code could works wrong, here machine.xml and logs of exceptions made with
unmodified demo pnp job and 3 nozzles, 2x no nozzle changer and the third have a nozzle changer enabled. Nozzle tip  loaded was N1-NT2 and N2-NT5, the third nozzle had no loaded nozzle at job start.
On Error, X3 was loaded on third nozzle.

[OpenPnP.txt](https://github.com/openpnp/openpnp/files/841499/OpenPnP.txt)

[machine.txt](https://github.com/openpnp/openpnp/files/841513/machine.txt)

vonnieda:
Why did you close this?

Issue #444, opened 2017-02-07
NamedCsvImporter Improvements


vonnieda:
The NamedCsvImporter needs some cleanup and improvements. Daniel from IRC reports that an Altium exported file fails to load. The files are attached below along with some comments from the reporter.

I think we should probably take a close look at rewriting the importer from scratch. The code is very rough and there are a lot of hardcoded values and difficult to understand code paths. I think the whole thing could be simplified greatly with the following ideas in mind:

* Importing of delimited files in general. Allow user to specify delimiter.
* Ignore non-matching lines, like the preamble visible in the attached files.
* Allow user to specify field names and formats for each of the required fields.
* The import dialog should save it's settings between sessions.
* Stretch Goal: Dropdown with some common file format profiles to autofill the dialog.

It may make sense for this to be a new importer caller DelimitedTextImporter.

Notes below:

> attached you will find my recreated files.
> One as a .csv and one text file, as i can export both.
> Note that my values are with a comma and not a dot. 

> I tried a couple of more files, and it seems like my .csv has problems. 
> The importer works fine if i copy and paste the content into a .txt file and delete the overhead and fix the field-names. 
> 
> All values are imported correctly, with reference to values having a comma instead of a dot. 
> My Values: 2,500mm
> Example Values: 2.500mm
> 

[csv_altium.txt](https://github.com/openpnp/openpnp/files/756285/csv_altium.txt)

[altium.txt](https://github.com/openpnp/openpnp/files/756287/altium.txt)

cri-s:
simplified greatly with the following ideas in mind:
Delimiters are automatically detected.
Non matching lines are automatically ignored.
It imports CSV files, not Excel files. 
The second file is Exel file, it have excel ""delimiter"" instead of CSV "delimiter" .
Some limits was chosen, format specs need to be within the first 10 lines, and it need to be present 6 valid fields in order file is imported. In this file format specs is on the eleven line. Its possible to change this, changing this for, its increases wrong import risk too.
for (int i = 0; i++ < 10 && (line = reader.readLine()) != null;) {

vonnieda:
Note that work done in #461 should resolve the Altium issue.

vonnieda:
Recent changes have improved this enough. If additional changes are needed, new issues can be opened.

PR #461, opened 2017-03-13
#444 Fixed NamedCSV importer for Altium R14 

Misfittech:
I have added in a fix for importing Altium R14 files, currently only supports mm output from altium

### DIFF
--- a/src/main/java/org/openpnp/gui/importer/NamedCSVImporter.java
+++ b/src/main/java/org/openpnp/gui/importer/NamedCSVImporter.java
@@ -240,7 +240,6 @@ public class NamedCSVImporter implements BoardImporter {
 	 * csvParser.lastLineNumber() + " " + t );
 	 */
 	//////////////////////////////////////////////////////////
-
 	private static List<Placement> parseFile(File file, boolean createMissingParts)
 			throws Exception {
 		BufferedReader reader =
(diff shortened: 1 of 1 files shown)


vonnieda:
Hi @Misfittech, it appears you have reformatted the entire file so I can't tell what the changes you made are. Can you please re-submit with just the changes? I suspect that your editor may have replaced spaces with tabs or something like that. You can see the diff at https://github.com/openpnp/openpnp/pull/461/files.

cri-s:
I'm against allowing 1000 lines for the format line.

Misfittech:
I am trying to figure out how to fix that problem...

I often do a "ctrl-a" followed by "ctrl-I" to clean up indents, I guess
eclipse has this set wrong for the project.

Misfittech:
Sorry about that I downloaded the eclipse formatter and applied to the file
and recommitted.

Misfittech:
Cri-s,

why was there a 10 in there?  I figured it was some sort of time out, but
not sure why....

Thanks
Trampas

cri-s:
The 10 was to look for the format specification on the first 10 lines of the file.
Maybe 10 is a bit less, i agree that someone want more comment on the file and that the number could
be increased, 1000 seems too permissive. 

It's for the line "Designator","Footprint","Mid X","Mid Y" ...
that define the csv contents, As this should be on the file header, if it's not found on the first 10 lines,
it is not a file thats can be read with this importer. Think further that it's possible to open binary files where
one line is several tens of Megabytes. If you increase it to 40 lines, 80, 100, ... it search then the first
N lines of file for the format specifier.

Misfittech:
Since in AD14, the column headers are on line 13, I have updated the code
to stop at row 50, instead of 1000.

I have also added optional part height processing, and option check box to
GUI to update part height if part is already in OpenPnP.

The dialog box form is not spaced the best, row spacing seems large for
check boxes, but I am not sure how to fix. I will read up on it some more.

I have pushed the latest version, again I am just learning so I hope it is
all good, if not please let me know.

Thanks
Trampas

vonnieda:
@Misfittech Sorry for the delay in reviewing this. Been a busy couple days. I have not forgotten it and will get it done as soon as I can.

Misfittech:
No worries... I certainly am not in a hurry.

vonnieda:
Thanks @Misfittech!

PR #463, opened 2017-03-15
corrected bug about rotation

cri-s:
It happens that angle result is found not only at 90 or 270 degreee but 180 too.
This correct the missplacing


vonnieda:
Thanks @cri-s. Can you provide a sample image where this happens? I'd like to add a test for it.

cri-s:
I check it out (i must ask, not having saved image).
Do you remember the longer discussion from bottomvision that im sayed:
Bottomvision x/y offsets must be recalculated considerating part angle rotation, either on image and
repeating the vision process or rotating it matematically, that sometime returns different results.
Hovever it was commone consense that both the results was valid and more correct as the unrotated.

That piece of code seems to have merged out of the actual codebase.

PR #464, opened 2017-03-15
Multiple part alignments, paste dispenser additional logic

tracklogic:
Mainly for Jason....

I've done a few little things here and there, not sure what you think - if your happy with them I'll split them up into individual commits / branches

Obviously ignore the zevatech type classes as these are machine specific.

All probably needs a tidy etc.. but I'm hoping it all slots in a Jason-friendly fashion!

Thanks in advance

Matt


vonnieda:
Notes from review:

# Changes Made
* Adds a dispense button to jog controls.
* Looks like some kind of change to alignment config panel setup, but not sure what is happening here. Line 201-217 in PartsPanel.
* Adds support for multiple part alignment options.
* Changes to paste job processor involving some type of transform. Not sure what it does.
* Adds actuator up down support for paste dispenser. Uses hardcoded actuator names which will need to be configurable. Not sure why this is needed actually. Can probably be handled with head mountable mapping but might need support in the controller.
* Adds a new moveTo method to GcodeDriver which appears to be in support of dispense. Not added to interface, so should be made private.
* NullDriver update for dispense, looks fine.
* ReferenceBottomVision adds support for multiple alignments. Seems to add an "allowIncompatibleParts" field but I don't see a UI for that.

# Thoughts
Overall looks like very positive changes and I don't see many problems with it. Will have some more specific questions once features are broken into individual PRs but overall it looks clean and well considered. Really only major concern is the actuator stuff added to the dispense functionality. We may need to think that through a bit more.

tracklogic:
Hi Jason

Thanks for looking at this.
 ** Looks like some kind of change to alignment config panel setup, but not sure what is happening here. Line 201-217 in PartsPanel. **

On the parts panel for every part aligner specified on the system, it creates a tab, this allows you per part to enable or disable each part aligner. The part aligner wizards have a getName() method which gets called per aligner, and the tab title is set to this.

**Changes to paste job processor involving some type of transform. Not sure what it does.**

This might be crazy, although made sense to me... i get the pad centroid, draw a line from left to right and then transform this by the pad rotation, this is the path the dispense needle will go to cover the pad. I intend to extend this in case of an extra wide pad by introducing multiple lines, so the dispense needle will do multiple lines across the large pad, filling the area with paste.

    **Adds actuator up down support for paste dispenser. Uses hardcoded actuator names which will need to be configurable. Not sure why this is needed actually. Can probably be handled with head mountable mapping but might need support in the controller.**

This is more me experimenting, i'll refactor this and make the code more flexible - your correct though it can probably just be done with the head mount mapping.

    **Adds a new moveTo method to GcodeDriver which appears to be in support of dispense. Not added to interface, so should be made private.**

Ok.

   **NullDriver update for dispense, looks fine.**

I did this for when I'm playing about on the train... probably useful to have for simulating anyway :-)

    **ReferenceBottomVision adds support for multiple alignments. Seems to add an "allowIncompatibleParts" field but I don't see a UI for that.**

I'll double check this

vonnieda:
Thanks for the responses @tracklogic. All sounds good. If you can break this up into feature PRs I'll get it merged. Seems like a PR for the multiple alignment stuff and one for the dispense stuff would do the trick.

vonnieda:
Closing this PR as it needs to be broken up into feature PRs.

PR #465, opened 2017-03-17
added Part to HeadMountable

cri-s:
This is  for the visionOffsets(HeadMountable hm)  that are callable with nozzle for bottom vision and with
camera for top vision. The reason for prefer this over 
Location visionOffsets(Nozzle nozzle, Part part) 
where in case of topvision nozzle is given as null is that adding part to camera gives vision pipeline the
access to part data and preserve the possibility to have more then one downlooking camera for the case
of multiple heads or for two distinct cameras with different fields of view.


cri-s:
Forget about actuators be part on head on official branch. Get back.

PR #468, opened 2017-03-18
Added histogram equalization and doing circle detection for fiducials

trampas:
The histogram equalization fixed problems with strip tape feeders not
being detcted by the image processing. 

The circle detection for the fiducials is done if the fiducial template
match fails and the fiducal is 100% round package. This works well for
using through hole parts as fiducials


vonnieda:
@trampas Hi Trampas, I don't know if Github notified you or not, but the automated tests did not pass. It looks like there is a problem running the strip feeder tests. See https://github.com/openpnp/openpnp/wiki/Developers-Guide#testing for how to run the built in tests.

vonnieda:
Additional note: The tests are run automatically on every PR. If they don't pass the PR can't be merged. See below where it says "All checks have failed". If you click the Details on the travis CI section you can see what failed. In this case it's because a strip hole was not found during the internal automated test. This means there is a regression in the way the strip feeder code works and it is no longer finding holes that it used to.

vonnieda:
Please see the above comments, fix the test issues and submit a new PR.

PR #470, opened 2017-03-20
Logging ui issues

pfried:
This is a reworked version of the logging functionality as there were some issues where the ui did not respond when the ui tried to scroll the logging panel.

@vonnieda Could you run a test on this?


src/main/java/org/openpnp/gui/LogPanel.java
```diff
@@ -157,7 +158,16 @@ public void contentsChanged(ListDataEvent e) {
...
+            JScrollBar verticalBar = scrollPane.getVerticalScrollBar();
+            AdjustmentListener downScroller = new AdjustmentListener() {
+                @Override
+                public void adjustmentValueChanged(AdjustmentEvent e) {
+                    Adjustable adjustable = e.getAdjustable();
+                    adjustable.setValue(adjustable.getMaximum());
+                    verticalBar.removeAdjustmentListener(this);
+                }
+            };
+            verticalBar.addAdjustmentListener(downScroller);
```
vonnieda: Won't this cause the downScroller Listener to be added to the verticalBar every time a line is added to the log?

pfried: That is true, the problem if you keep the Listener attached is that it does not allow scrolling up if autoscroll is enabled. I will check if i can improve this behaviour during the weekend, maybe it is possible to distinguish user events and data added events

pfried:
I changed the behaviour see commit for details. Still not ideal, but compared to the current logging its better, also the problems with performance are fixed, so it runs quite smooth now

vonnieda:
Looks good to me. There is a UI flash happening once the list hit it's max but I suspect that is due to the jump to top you mentioned. I think we can live with it and this is definitely an improvement.

PR #489, opened 2017-03-27
Change hard-coded for loop to timeout after 50ms instead

iAmRoland:
Hello!

I made a small change to CSV Importer. Instead of the for loop which only checks the first 50 lines, I added in a while loop and made it timeout after 50ms. Also added a custom timeout message, not sure tho if it's clear enough and easy to understand. 

**A little bit of good to know info:** 
We are a 6 interns divided into 2 groups at a company in Sweden called Pluspole.  The first group of 4 people is going to work on the documentation part, which in the beginning will mostly consist of creating class-diagrams and alike, while the other group (which i'm in) of 2 will be further developing the OpenPNP software. You might have already received a message or email about this from the group working on documentation. 

We are all currently studying for software development, so any and all feedback is greatly appreciated 😄

### DIFF
--- a/src/main/java/org/openpnp/gui/importer/NamedCSVImporter.java
+++ b/src/main/java/org/openpnp/gui/importer/NamedCSVImporter.java
@@ -267,28 +267,18 @@ public class NamedCSVImporter implements BoardImporter {
         ArrayList<Placement> placements = new ArrayList<>();
         String line;
 
-        // Set the timeout limit, currently it will timeout after 50 milliseconds.
-        // 50ms seems like it should be more than enough.
-        long timeoutLimit = System.currentTimeMillis() + 50;
         while((line = reader.readLine()) != null) {
             line = line.trim();
             if (line.length() == 0)
-                continue;
-            if (checkLine(line) || System.currentTimeMillis() > timeoutLimit)
+            	continue;
+            
+            if (checkLine(line))
                 break;
         }
-           
+                   
         if (Len == 0) {
-            reader.close();
-            
-            String errorMessage = "Unable to parse CSV File Names";
-            if (System.currentTimeMillis() > timeoutLimit) {
-                // Is this message clear enough?
-                // It throws this when the lineReadThread fails to find the CSV proper headers.
-                errorMessage = "Timout while parsing CSV file. File may contain too much overhead.";
-            }
-            
-            throw new Exception(errorMessage);
+            reader.close();            
+            throw new Exception("Unable to parse CSV File Names");
         }
 
         // CSVParser csvParser = new CSVParser(new FileInputStream(file));
--- a/src/main/java/org/openpnp/gui/importer/NamedCSVImporter.java
+++ b/src/main/java/org/openpnp/gui/importer/NamedCSVImporter.java
@@ -267,19 +267,28 @@ public class NamedCSVImporter implements BoardImporter {
         ArrayList<Placement> placements = new ArrayList<>();
         String line;
 
-        //TODO why only check first 50 lines? If it is a time out would time based time out be better? 
-        for (int i = 0; i++ <50 && (line = reader.readLine()) != null;) {
+        // Set the timeout limit, currently it will timeout after 50 milliseconds.
+        // 50ms seems like it should be more than enough.
+        long timeoutLimit = System.currentTimeMillis() + 50;
+        while((line = reader.readLine()) != null) {
             line = line.trim();
             if (line.length() == 0)
                 continue;
-            if (checkLine(line))
+            if (checkLine(line) || System.currentTimeMillis() > timeoutLimit)
                 break;
         }
-
-
+           
         if (Len == 0) {
             reader.close();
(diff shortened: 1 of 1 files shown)


vonnieda:
Hi @iAmRoland, welcome to the project! This is the first I've heard about your group, or the documentation group, but I am happy to have your help!

A couple things:
* Have you tested this with various length files? Can you supply the test files?
* @cri-s and @Misfittech, since you two have done the most work in this code, do you have any thoughts on this?

cri-s:
Why you do a errorprone 50ms timeout instead of line size timeout.
The file could be on network mounted file that have some delay, or there could be annother task that
interrupt openpnp, and the timeout trigger.
The orginal centeroid specifications used two lines that was discarded from file and contained the heading columns, no other lines could be present. Later comment lines beginning with # was introduced.
The second line was mostly empty or contained unit hint. 
I have initially set a limit of 10 lines for the header describing the format. As Altium on default exporter
used 15 lines, the limit was raised up to 50.
Annother guy have want raising the lines to 1000 not understanding what hi do, 
but changing the limit from search for the header lines
at the beginning from the file limiting the search to the first 50 lines to 50ms don't make any sense.
Before doing this, remove the limitations, but beware that drill file legend file as example could be parsed
as valid header files, still as other.  This my 2cents.
Cri

robermeyer:
Curious, has anybody seen a legal CSV formatted file where header is not in the first 50 lines?
If so who generated it.  If not why the change?

Misfittech:
Using Altium 17 the header line is on line 17 (if I recall correctly). The
original code only checked the first 10 lines, so it did not work with
AD17.  I did not understand the need for the check at first, and changed to
1000, Cri-S and Jason recommended 50 as a safe number.

Since humans are time based machines, and computers are getting faster all
the time I personally prefer time based time outs.  For most people 1-2
seconds is about the limit after clicking a button before people expect
something to happen. Thus it might be better to have a time out of 1 second
and after that open a dialog asking for user intervention like:

"We have parsed xxx lines but have not found header, do you want to
continue to process until end of file which could take a while, or give up?"

Trampas

iAmRoland:
Oh wow, a lot of great responses.

I made this change when I had an older version which still had the 10 or 12 line limit. I then updated my OpenPNP version and saw the limit was raised to 50, but still thought that a bit better solution could be implemented instead. It is probably highly unlikely a CSV file will contain more than 50 lines in the header, I have not found one yet. But no limit or time based limit feels like a bit better solution than a set limit.

I went with 50ms timeout because it seemed like enough, it was able to go through waay more lines than needed. But as mentioned, this would probably fail in some scenarios.

BUT, I removed the timeout completely, just to see how it would do, so now it's only a while loop until the file end or until it finds something. I did a very extreme-case-most-likely-not-going-to-happen test, I just threw in thousands of lines in the header, the importer was able to go through all of it pretty much instantly. So based on that test, I would say that no set limit would be needed at all. It will probably at most only have to read through 20-30 lines in the header, even on slower systems that would go by in an instant. 
So it would still work just as fast as before and with no set limit. Or am I missing to think of a scenario were that while loop would fail as well, where it would get stuck loading or alike?

I will commit and push the changes right after i've sent this away.

cri-s:
Can you explain please why you want remove a check that it only purpose is to valid that the file is a valid
csv file where you itself admit to apparently not have any problem with that check ? The last push because not good implemented had as consequence that wrong data get imported as valid data for AD15+ users. Previously it gives error, there noticed it and used external converter to convert format in oder to be compatible, now it import the data with wrong coordinates and that is sometime a bigger problem as raising the error because before, if it was imported, it works, now it must be checked explicitly.

iAmRoland:
Sorry but i'm slightly confused by what you mean exactly. I pretty much only changed the for loop, which only checks the first 50 lines of a CSV file to a while loop which has no 50 line limit, or no limit at all.
It still validates the CSV file just like before.

So basically, after removing the timeout part, only this line is changed now:
`for (int i = 0; i++ <50 && (line = reader.readLine()) != null;) {`
to this:
`while((line = reader.readLine()) != null) {`

vonnieda:
Let's not get too far into the weeds here. The change @iAmRoland made is basically just the one that @Misfittech suggested in the comments of their last commit.

I agree that a time based solution is not a good one, for the various reasons listed.

The reason given previously for having a check at all was to keep the program from looping endlessly on an invalid file, e.g. someone has clicked on a random 200MB Excel CSV for instance. I think that a check of 50 lines, as @Misfittech did is the right solution for now, and we can revisit later if we find that Altium moves their header to line 51.

I am going to close this as I think that no changes are needed for now. Sorry for the wasted effort @iAmRoland, I hope it doesn't dissuade you from trying again. If you are looking for places where you can help with the code, please see the issues list filtered by the "help wanted" tag. These are issues I've identified that I feel people new to the project can make meaningful contributions:

https://github.com/openpnp/openpnp/issues?q=is%3Aissue+is%3Aopen+label%3Ahelp-wanted

Thanks,
Jason

PR #491, opened 2017-03-28
Add arguments to ScriptRun OpenCV pipeline stage

dzach:
It is useful to have arguments passed to the ScriptRun OpenCV pipeline stage. This PR adds a string parameter `args` to the stage, which can then be parsed from inside the script.

Example parsing, from inside a ScriptRun stage with `args=size:[45,60],tol:5` 
```
eval("var args={" + args + "}");
for (var i in args)
  print(i,args[i]);
```
LOG window shows:
```
2017-03-29 01:33:01 SystemLogger INFO: size 45,60
2017-03-29 01:33:01 SystemLogger INFO: tol 5

```


vonnieda:
Looks good, thanks @dzach!

PR #492, opened 2017-03-29
Update ReferenceNozzleTip.java

cri-s:
Speed issue because of logging calling.
If call it present, on 400 components and 2 nozzles, it pauses over 7 seconds after having placed
the parts from nozzle, removing the call resolved the issue.


vonnieda:
Looks good, thanks @cri-s!

PR #493, opened 2017-03-29
Job placement status and progress on Id cell

iAmRoland:
Hello again!

I added in displaying of status on the job placements, something we found to be useful. Makes it easy to follow the progress on the placements. Question is if you consider this correctly implemented or not? 

Once the start button is pressed, it will mark all pending placements with a yellow color. When it's processing a placement it will display a blue color on the Id cell. Once it's done with all placements with the same id, it then marks that cell with a green color. If no placement is going to be done then the cell is left white.

How it looks in action:
alt text

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
@@ -26,6 +26,7 @@ import javax.swing.JScrollPane;
 import javax.swing.JTable;
 import javax.swing.JToolBar;
 import javax.swing.ListSelectionModel;
+import javax.swing.UIManager;
 import javax.swing.border.LineBorder;
 import javax.swing.event.ListSelectionEvent;
 import javax.swing.event.ListSelectionListener;
@@ -43,7 +44,6 @@ import org.openpnp.gui.support.PartCellValue;
 import org.openpnp.gui.support.PartsComboBoxModel;
 import org.openpnp.gui.tablemodel.PlacementsTableModel;
 import org.openpnp.gui.tablemodel.PlacementsTableModel.Status;
-import org.openpnp.machine.reference.ReferencePnpJobProcessor;
 import org.openpnp.model.Board;
 import org.openpnp.model.Board.Side;
 import org.openpnp.model.BoardLocation;
@@ -55,6 +55,7 @@ import org.openpnp.model.Placement.Type;
 import org.openpnp.spi.Camera;
 import org.openpnp.spi.HeadMountable;
 import org.openpnp.spi.Nozzle;
+import org.openpnp.spi.PnpJobProcessor;
 import org.openpnp.util.MovableUtils;
 import org.openpnp.util.UiUtils;
 import org.openpnp.util.Utils2D;
@@ -74,7 +75,7 @@ public class JobPlacementsPanel extends JPanel {
     private static Color statusColorWarning = new Color(252, 255, 157);
     private static Color statusColorReady = new Color(157, 255, 168);
     private static Color statusColorError = new Color(255, 157, 157);
-    private static Color cellColorSelected = new Color(51, 153, 255);
+    private static Color cellColorSelected = UIManager.getColor("Table.selectionBackground");
     private static Color jobColorProcessing = new Color(157, 222, 255);
     private static Color jobColorPending = new Color(252, 255, 157);
     private static Color jobColorComplete = new Color(157, 255, 168);
@@ -551,7 +552,7 @@ public class JobPlacementsPanel extends JPanel {
     }
 
     class IdRenderer extends DefaultTableCellRenderer {
-        // This is used just to set backgroud color on Id cell when selected.
+        // This is used just to set background color on Id cell when selected.
         // Could not find another way to do this in.
         public Component getTableCellRendererComponent(JTable table, Object value,
                 boolean isSelected, boolean hasFocus, int row, int column) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/openpnp/gui/JobPlacementsPanel.java
```diff
@@ -39,9 +39,11 @@
 import org.openpnp.gui.support.IdentifiableListCellRenderer;
 import org.openpnp.gui.support.IdentifiableTableCellRenderer;
 import org.openpnp.gui.support.MessageBoxes;
+import org.openpnp.gui.support.PartCellValue;
 import org.openpnp.gui.support.PartsComboBoxModel;
 import org.openpnp.gui.tablemodel.PlacementsTableModel;
 import org.openpnp.gui.tablemodel.PlacementsTableModel.Status;
+import org.openpnp.machine.reference.ReferencePnpJobProcessor;
```
vonnieda: UI classes like JobPlacementsPanel cannot access implementation classes in this way. You must go through the SPI interface http://openpnp.github.io/openpnp/develop/org/openpnp/spi/PnpJobProcessor.html. The reason for this is that users can change which job processor they use by changing the class in machine.xml. This would make the feature fail to work in any setup that did not use the default.

Instead, add getJobPlacements() to the PnpJobProcessor interface and then access it via machine.getPnpJobProcessor(). Then do the various bits of filtering in UI code rather than in the JobProcessor so that we don't have to expose UI specific methods in the SPI class.

src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
```diff
@@ -141,7 +141,7 @@ public String toString() {
 
     protected Head head;
 
-    protected List<JobPlacement> jobPlacements = new ArrayList<>();
+    protected static List<JobPlacement> jobPlacements = new ArrayList<>();
```
vonnieda: This, and the static methods added below should not be static and instead accessed via the interface.

src/main/java/org/openpnp/gui/JobPlacementsPanel.java
```diff
@@ -72,6 +74,10 @@
     private static Color statusColorWarning = new Color(252, 255, 157);
     private static Color statusColorReady = new Color(157, 255, 168);
     private static Color statusColorError = new Color(255, 157, 157);
+    private static Color cellColorSelected = new Color(51, 153, 255);
```
vonnieda: This color ends up clashing with the system color for a selected row. Instead, use UIManager to get the correct Swing color. See http://alvinalexander.com/java/java-uimanager-color-keys-list

Example:
screen shot 2017-03-29 at 9 38 39 am

iAmRoland:
Thanks for the feedback! In process of fixing those things now. 

> Since we already have a status field that shows the status of a placement before the job starts, I think the status as the job progresses should go here as well. Then, when the job starts, the status can change from Ready to Processing (2/4), Finished, etc. What do you think?

Yeah that sounds good to me. 
Also the status field text is set with `fireTextStatus` right? So displaying of progression on the status field should be done from ReferencePnpJobProcessor?

vonnieda:
@iAmRoland fireTextStatus triggers the TextStatusListener in JobPanel. You could add code here to update your status as well since it is called regularly through the process. I don't think any changes will be needed to ReferenceJobProcessor. Probably just a table refresh from that method would do the trick. Basically I'm just suggesting moving your code over to the Status column instead of the ID column.

iAmRoland:
Alright I see, I'll fix the status text tomorrow.
Here comes the latest changes, have fixed most of the things you mentioned, hopefully the implementation is more correct now.

src/main/java/org/openpnp/spi/PnpJobProcessor.java
```diff
@@ -1,4 +1,13 @@
...
 
+import java.util.List;
+
+import org.openpnp.machine.reference.ReferencePnpJobProcessor.JobPlacement;
+
 public interface PnpJobProcessor extends JobProcessor {
+
+    public List<JobPlacement> getJobPlacementsById(String id);
+
+    public List<JobPlacement> getJobPlacementsById(String id, String status);
```
vonnieda: This should take the enum value rather than a String for status. This enforces type safety.  So, just change it to `public List<JobPlacement> getJobPlacementsById(String id, JobPlacement.Status status);`

src/main/java/org/openpnp/spi/PnpJobProcessor.java
```diff
@@ -1,4 +1,13 @@
 package org.openpnp.spi;
 
+import java.util.List;
+
+import org.openpnp.machine.reference.ReferencePnpJobProcessor.JobPlacement;
```
vonnieda: This is a use of a concrete implementation class in a SPI class, which is not allowed. The solution is to just move JobPlacement class into the PnpJobProcessor interface, like this:

```
package org.openpnp.spi;

import java.util.List;

import org.openpnp.model.BoardLocation;
import org.openpnp.model.LengthUnit;
import org.openpnp.model.Placement;

public interface PnpJobProcessor extends JobProcessor {

    public List<JobPlacement> getJobPlacementsById(String id);

    public List<JobPlacement> getJobPlacementsById(String id, String status);

    public static class JobPlacement {
        public enum Status {
            Pending,
            Processing,
            Skipped,
            Complete
        }

        public final BoardLocation boardLocation;
        public final Placement placement;
        public Status status = Status.Pending;

        public JobPlacement(BoardLocation boardLocation, Placement placement) {
            this.boardLocation = boardLocation;
            this.placement = placement;
        }

        public double getPartHeight() {
            return placement.getPart().getHeight().convertToUnits(LengthUnit.Millimeters)
                    .getValue();
        }

        @Override
        public String toString() {
            return placement.getId();
        }
    }
}
```

iAmRoland:
Fixed those two small things now, thank you for the feedback! For the status text, I think i'm gonna leave it as it is for now.

vonnieda:
Nicely done @iAmRoland! The changes look great. I am going to pull this in, add a note to the CHANGES.md and merge this.

vonnieda:
Merged, thank you @iAmRoland!

PR #495, opened 2017-03-29
Add FilterRects stage to OpenCV pipeline to filter rects based on width and height

dzach:
Original image:

orig

`MinAreaRectContours` OpenCV pipeline stage selects rects based on contour area:

contours-both

The contours above have approximately the same area, which makes it difficult to distinguish within a safe margin, although the have quite different dimensions.

It would be useful to be able to filter objects based on their expected dimensions, and not only on their contour area. The proposed OpenCV pipeline  stage adds this functionality.

With FilterRects and the example image above, specifying `width = 60, height = 50, tolerance = 5` yields the following rect:

result_soic-8

while specifying `width = 85, height = 80, tolerance = 5` yields the following rect:

result_to-252

### DIFF
--- a/src/main/java/org/openpnp/vision/pipeline/stages/FilterRects.java
+++ b/src/main/java/org/openpnp/vision/pipeline/stages/FilterRects.java
@@ -17,24 +17,24 @@ import org.pmw.tinylog.Logger;
 public class FilterRects extends CvStage {
     @Attribute
     @Property(description="Width of filtered rects. Filtering takes tolerance into consideration.")
-    private double width = 50.0;
+    private double rectWidth = 50.0;
     
-    public double getWidth() {
-      return width;
+    public double getRectWidth() {
+      return rectWidth;
     }
-    public void setWidth(double width) {
-      this.width = Math.abs(width);
+    public void setRectWidth(double rectWidth) {
+      this.rectWidth = Math.abs(rectWidth);
     }
     
     @Attribute
     @Property(description="Height of filtered rects. Filtering takes tolerance into consideration.")
-    private double height = 50.0;
+    private double rectHeight = 50.0;
 
-    public double getHeight() {
-      return height;
+    public double getRectHeight() {
+      return rectHeight;
     }
-    public void setHeight(double height) {
-      this.height = Math.abs(height);
+    public void setRectHeight(double rectHeight) {
+      this.rectHeight = Math.abs(rectHeight);
     }
     
     @Attribute
@@ -49,31 +49,31 @@ public class FilterRects extends CvStage {
     }
     
     @Attribute
-    @Property(description="Aspect ratio for selecting rects, used if one or both of width and height are 0. If both width and height are 0, then any rect with size satisfying the aspect ratio and tolerance requirements will be selected.")
-    private double aspect = 0.0;
+    @Property(description="Aspect ratio for selecting rects, used if one or both of rectWidth and rectHeight are 0. If both rectWidth and rectHeight are 0, then any rect with size satisfying the aspect ratio and tolerance requirements will be selected.")
+    private double aspectRatio = 0.0;
     
-    public double getAspect() {
-      return aspect;
+    public double getAspectRatio() {
+      return aspectRatio;
     }
-    public void setAspect(double aspect) {
-      this.aspect = Math.abs(aspect);
+    public void setAspectRatio(double aspectRatio) {
+      this.aspectRatio = Math.abs(aspectRatio);
     }
 
     @Attribute
     @Property(description="Tolerance is taken to be a percentage, e.g. 5 = 5%.")
-    private boolean toleranceIsPerc = false;
+    private boolean toleranceIsPercentage = false;
     
-    public boolean gettoleranceIsPerc() {
-        return toleranceIsPerc;
+    public boolean getToleranceIsPercentage() {
+        return toleranceIsPercentage;
     }
 
-    public void settoleranceIsPerc(boolean toleranceIsPerc) {
-        this.toleranceIsPerc = toleranceIsPerc;
+    public void setToleranceIsPercentage(boolean toleranceIsPercentage) {
+        this.toleranceIsPercentage = toleranceIsPercentage;
     }
     
     @Attribute
     @Property(description="Show selection ranges in the Log.")
-    private boolean logSelectionRanges = true;
+    private boolean logSelectionRanges = false;
     
     public boolean getLogSelectionRanges() {
         return logSelectionRanges;
@@ -99,6 +99,9 @@ public class FilterRects extends CvStage {
        Logger.info(s);
       }
     }
+    // assign values here to minimize storage allocation costs in time, and use pointers to these values.
+    private static String[] verdict = {"[REJ]","[PASS]"};
+    
     @Override
     public Result process(CvPipeline pipeline) throws Exception {
         if (rectStageName == null) {
@@ -121,25 +124,25 @@ public class FilterRects extends CvStage {
         List<RotatedRect> results = new ArrayList<RotatedRect>();
                 
         // we need to sort sizes to do comparisons. The following assures w is always greater than h
-        double w = Math.max(width,height), ww, upper_w;
-        double h = Math.min(width, height), hh, upper_h;
+        double w = Math.max(rectWidth,rectHeight), ww, upper_w;
+        double h = Math.min(rectWidth, rectHeight), hh, upper_h;
         double rw, rh, tol, extUpper;
         String pass;
         
         // check the validity of the input
         if (w == 0.0) {
           // this means h is also 0
-          if (aspect == 0.0) {
+          if (aspectRatio == 0.0) {
             // cannot calculate absolute dimensions
-            throw new Exception("Width, height, or width and aspect, or aspect, must be non zero.");
+            throw new Exception("rectWidth, rectHeight, or rectWidth and aspectRatio, or aspectRatio, must be non zero.");
             
-          } else if (!toleranceIsPerc) {
+          } else if (!toleranceIsPercentage) {
             // bad input
-            throw new Exception("If only aspect is specified, tolerance must be expressed as a percentage.");
+            throw new Exception("If only aspectRatio is specified, tolerance must be expressed as a percentage.");
           }
         }
         // parse tolerance
-        if (toleranceIsPerc) {
+        if (toleranceIsPercentage) {
           tol = tolerance / 100.0;
         } else {
           tol = tolerance;
@@ -154,11 +157,11 @@ public class FilterRects extends CvStage {
           tol = Math.abs(tol);
         }
         
-        if (h == 0.0 && aspect != 0.0) {
-          // derive height from aspect
-          h = w / aspect;
+        if (h == 0.0 && aspectRatio != 0.0) {
+          // derive rectHeight from aspect ratio
+          h = w / aspectRatio;
           if (w != 0)
-            logInfo("derived height parameter: " + h);
+            logInfo("derived rectHeight parameter: " + h);
         }
         // iterate over input rects
         for (RotatedRect rect : rects) {
@@ -166,16 +169,16 @@ public class FilterRects extends CvStage {
           rw = Math.max(rect.size.width,rect.size.height);
           rh = Math.min(rect.size.width,rect.size.height);
           
-          if (width == 0.0 && height == 0.0) {
+          if (rectWidth == 0.0 && rectHeight == 0.0) {
             // Any size:
             // Select rects based only on the aspect ratio of each input rectangle
             w = rw;
-            h = rw / aspect;
+            h = rw / aspectRatio;
           } 
           
           // trick to avoid lengthy checking of tolerance criteria bellow
-          ww = toleranceIsPerc ? w : 1;
-          hh = toleranceIsPerc ? h : 1;
+          ww = toleranceIsPercentage ? w : 1;
+          hh = toleranceIsPercentage ? h : 1;
           upper_w = tol * ww * extUpper;
           upper_h = tol * hh * extUpper;
 
@@ -184,15 +187,15 @@ public class FilterRects extends CvStage {
           if (rw >= w - tol * ww && rw <= w + upper_w && rh >= h - tol * hh && rh <= h + upper_h) {
             // rect passes criteria
             results.add(rect);
-            pass = "[PASS]";
+            pass = verdict[1];
           } 
           else {
-            pass = "[REJ] ";
+            pass = verdict[0];
           }
-          logInfo(pass + " widths  range: " + (w - tol * ww) + " <= " + rw + " <= " + (w + upper_w));
-          logInfo(pass + " heights range: " + (h - tol * hh) + " <= " + rh + " <= " + (h + upper_h));
-          if (aspect != 0) {
-            logInfo(pass + " aspect ratio range: " + (aspect * (1 - tol)) + " <= " + (rw / rh) + " <= " + (aspect * (1 + tol)));
+          logInfo(pass + " width  range: " + (w - tol * ww) + " <= " + rw + " <= " + (w + upper_w));
+          logInfo(pass + " height range: " + (h - tol * hh) + " <= " + rh + " <= " + (h + upper_h));
+          if (aspectRatio != 0) {
+            logInfo(pass + " aspectRatio range: " + (aspectRatio * (1 - tol)) + " <= " + (rw / rh) + " <= " + (aspectRatio * (1 + tol)));
           }
         }
         // deside what type to return
--- a/src/main/java/org/openpnp/vision/pipeline/stages/FilterRects.java
+++ b/src/main/java/org/openpnp/vision/pipeline/stages/FilterRects.java
@@ -107,13 +107,13 @@ public class FilterRects extends CvStage {
             // cannot calculate absolute dimensions
             throw new Exception(dimErr);
             
-          } else if (!tolerance.matches("[\\+\\-0-9]*\\.?[0-9]+%")) {
+          } else if (!tolerance.matches("[\\+\\-0-9]*\\.?[0-9]+%*")) {
             // bad input
             throw new Exception("If only aspect is specified, tolerance should be expressed as a percentage, e.g. 5%");
           }
         }
         // parse tolerance
-        if (tolerance.matches("[0-9]*\\.?[0-9]+%*")) {
+        if (tolerance.matches("[\\+\\-0-9]*\\.?[0-9]+%*")) {
           tol = Double.parseDouble(tolerance.replaceAll("%","")) / 100.0;
           tolIsPerc = true;
         } else {
--- a/src/main/java/org/openpnp/vision/pipeline/stages/FilterRects.java
+++ b/src/main/java/org/openpnp/vision/pipeline/stages/FilterRects.java
@@ -13,43 +13,54 @@ import org.openpnp.vision.pipeline.Stage;
 import org.openpnp.vision.pipeline.Property;
 import org.pmw.tinylog.Logger;
 
-@Stage(category="Image Processing", description="Filters rotated rects based on given size and tolerance. Orientation of width and height does not matter.")
+@Stage(category="Image Processing", description="Filters rotated rects based on given size and tolerance. Orientation of width and length does not matter.")
 public class FilterRects extends CvStage {
     @Attribute
     @Property(description="Width of filtered rects. Filtering takes tolerance into consideration.")
-    private double rectWidth = 50.0;
+    private double width = 50.0;
     
-    public double getRectWidth() {
-      return rectWidth;
+    public double getWidth() {
+      return width;
     }
-    public void setRectWidth(double rectWidth) {
-      this.rectWidth = Math.abs(rectWidth);
+    public void setWidth(double width) {
+      this.width = Math.abs(width);
     }
     
     @Attribute
-    @Property(description="Height of filtered rects. Filtering takes tolerance into consideration.")
-    private double rectHeight = 50.0;
+    @Property(description="Tolerance for the width of filtered rects.")
+    private double widthTolerance = 5.0;
+    
+    public double getWidthTolerance() {
+      return widthTolerance;
+    }
+    public void setWidthTolerance(double widthTolerance) {
+      this.widthTolerance = widthTolerance;
+    }
+    
+    @Attribute
+    @Property(description="Length of filtered rects. Filtering takes tolerance into consideration.")
+    private double length = 50.0;
 
-    public double getRectHeight() {
-      return rectHeight;
+    public double getLength() {
+      return length;
     }
-    public void setRectHeight(double rectHeight) {
-      this.rectHeight = Math.abs(rectHeight);
-    }
-    
-    @Attribute
-    @Property(description="Tolerance of specified sizes. If negative, sizes are maximum values.")
-    private double tolerance = 5;
-    
-    public double getTolerance() {
-      return tolerance;
-    }
-    public void setTolerance(double tolerance) {
-      this.tolerance = tolerance;
+    public void setLength(double length) {
+      this.length = Math.abs(length);
     }
     
     @Attribute
-    @Property(description="Aspect ratio for selecting rects, used if one or both of rectWidth and rectHeight are 0. If both rectWidth and rectHeight are 0, then any rect with size satisfying the aspect ratio and tolerance requirements will be selected.")
+    @Property(description="Tolerance for the length of filtered rects.")
+    private double lengthTolerance = 5.0;
+    
+    public double getLengthTolerance() {
+      return lengthTolerance;
+    }
+    public void setLengthTolerance(double lengthTolerance) {
+      this.lengthTolerance = lengthTolerance;
+    }
+
+    @Attribute
+    @Property(description="Aspect ratio for selecting rects, used if one or both of width and length are 0. If both width and length are 0, then any rect with size satisfying the aspect ratio and tolerance requirements will be selected.")
     private double aspectRatio = 0.0;
     
     public double getAspectRatio() {
@@ -73,29 +84,29 @@ public class FilterRects extends CvStage {
     
     @Attribute
     @Property(description="Show selection ranges in the Log.")
-    private boolean logSelectionRanges = false;
+    private boolean enableLogging = false;
     
-    public boolean getLogSelectionRanges() {
-        return logSelectionRanges;
+    public boolean getEnableLogging() {
+        return enableLogging;
     }
 
-    public void setLogSelectionRanges(boolean logSelectionRanges) {
-        this.logSelectionRanges = logSelectionRanges;
+    public void setEnableLogging(boolean enableLogging) {
+        this.enableLogging = enableLogging;
     }
 
     @Attribute
-    @Property(description="Pipeline stage that outputs rotated rects.")
-    private String rectStageName = null;
+    @Property(description="Previous pipeline stage that outputs rotated rects.")
+    private String rotatedRectsStageName = null;
     
-    public String getRectStageName() {
-        return rectStageName;
+    public String getRotatedRectsStageName() {
+        return rotatedRectsStageName;
     }
 
-    public void setRectStageName(String rectStageName) {
-        this.rectStageName = rectStageName;
+    public void setRotatedRectsStageName(String rotatedRectsStageName) {
+        this.rotatedRectsStageName = rotatedRectsStageName;
     }
     private void logInfo(String s) {
-      if (logSelectionRanges) {
+      if (enableLogging) {
        Logger.info(s);
       }
     }
@@ -104,11 +115,11 @@ public class FilterRects extends CvStage {
     
     @Override
     public Result process(CvPipeline pipeline) throws Exception {
-        if (rectStageName == null) {
+        if (rotatedRectsStageName == null) {
             return null;
         }
         
-        Result result = pipeline.getResult(rectStageName);
+        Result result = pipeline.getResult(rotatedRectsStageName);
         if (result == null || result.model == null) {
             return null;
         }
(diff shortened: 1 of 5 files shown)


dzach:
@cri-s 
You can follow this stage with a DrawRotatedRects and have it display the results. This is how I did it above. But you probably don't mean this. Can you elaborate? 

Do you mean to check the aspect ratio of the rects and select the ones that match a given one? This can be done, in addition with the existing parameters. Please explain.

cri-s:
DrawRotatedRect can display results from drawRotatedRect and from list of
drawRotatwdRect object. It uses instanceof for differentiate the two cases.
The filter should be capable of the same flexibility.

For the aspect ratio, I mean similar to maskCircle if value is negative it
have a slightly different behavior.
Instead of checking size, it checks ratio.
For that changing int to double for parameters is required.
I don't think it makes sense to check both, size and ratio, only one of
that.

Il giovedì 30 marzo 2017, Dimitrios <notifications@github.com> ha scritto:

dzach:
@cri-s  The aspect ratio sounds like a good idea, I'm working on it. For the DrawRotatedRect, I may be lacking some knowledge of the internals. Will have a look.

vonnieda:
Hi @dzach, this all looks great. Ready to merge when you are. What @cri-s is referring to is this bit of code: https://github.com/openpnp/openpnp/blob/develop/src/main/java/org/openpnp/vision/pipeline/stages/DrawRotatedRects.java#L64

Basically, it checks if the model is a RotatedRect or a List<RotatedRect> and operates equally well on either. It just allows the stage to accept a single value instead of a list. It can actually be simplified by doing something like List<RotatedRect> results; if (result.model instanceof RotatedRect) results = Collections.singletonList((RotatedRect)result.model); else results = (List<RotatedRect>)result.model;

dzach:
Thanks @vonnieda 
Aspect is now working nicely. I'm looking into tolerance, to make it more flexible, and also account for aspect when specified without width and height, as @cri-s suggested, in which case tolerance needs to be a percentage.

I was thinking of interpreting positive tolerances as -tolerance/2+tolerance/2 and negative tolerance values as -tolerance+0, the latter making the dimension, e.g. width, a max. Also tolerances with absolute value less than 1 will be interpreted as a percentage = tolerance * 100% without need for further parsing. 

I hope the above make sense.

cri-s:
Unrelated to this PR, what do you intend to realize ?

dzach:
@cri-s 
With regard to this PR, I'll include your good suggestion regarding aspect ratio,  and for RotatedRect will use the code from @vonnieda 's comment above.

Aspect ratio has now its own field, since I don't like hidden meanings in parameters, except when they are obvious. So for negative tolerance I described my thoughts above.

vonnieda:
@dzach Sounds good. I agree with not hiding meaning in params. I prefer verbosity and additional params to make things easier to understand at a glance.

cri-s:
I wanted ask not releated to this pr, instead related to the image that you have published in this pr, what you tray to do ? It seems you want tray to setup a pipeline for loose part feeder with not so ideal background and high vignetting optics.
Example of pipeline from openpnp using this image.
pipeline

dzach:
@cri-s  
 :-)
Yes, my testing requires conditions that are not ideal. Otherwise all kinds of problems hide in my best of intentions. Fortunately my setup can offer plenty of those non-ideal conditions, since it is all a 3D printed, low cost PnP. I might show it publicly when I manage to pnp a decent pcb. 

The "bed" on the picture above is a 3D printed plate in gray ABS, which I sanded to reduce some of its glare and the camera is way out of focus, with this "bed" being aprox 7mm below the focus level. 

The cameras are Philips SPZ5000 webcams, which have manual controls for exposure etc, and I set with some external commands before OpenPnP starts. With the new ScriptRun argument capability I might even try to set some camera parameters from within a pipeline stage.

Lighting comes from 4 PLCC-2 LEDs and is currently set at about 100/255 of max, while camera's exposure is set to 80, to achieve fast responses, but more importantly, because at this setting **there is no interference** from the PWM that drives the LED lighting.

Your pipeline above though looks very good! Would you share it?

I'm about ready to push a new commit. Tolerance can now be expressed as a percentage with a ` %` at the end, and is therefore parsed as a string. Negative tolerances ares as I describe above. Here is a diagram describing how tolerance works, for e.g. `width`:

tolerance

I'll do a little more testing and then proceed to a new commit. I'd much appreciate it if more testing can be done by others before this finds its way in OpenPnP.

cri-s:
Tollerance sounds good, i don't know, but maybe this should be in
length units and not pixel units. On the other side, pixel works fine,
but the pipeline is stored in parts, and can use different cameras
with different pixel/mm . I know this implies other stages too, you'r
image have contributed to raise up the question.

For the pipeline, i have done it fast, without using external hsv
color charts and tools.
Just for the black IC parts,
<cv-pipeline>
   <stages>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRead"
name="0" enabled="true"
file="C:\Users\Gast\Downloads\1197bc7a-14b4-11e7-99ca-d95acf04bfb5.png"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.ConvertColor" name="6"
enabled="true" conversion="Rgb2Gray"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.BlurGaussian" name="1"
enabled="true" kernel-size="13"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.Normalize"
name="2" enabled="true"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.Threshold"
name="11" enabled="true" threshold="70" auto="true" invert="true"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.MaskCircle"
name="10" enabled="true" diameter="360"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.MaskRectangle" name="12"
enabled="true" width="400" height="300"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.FindContours" name="14"
enabled="true" retrieval-mode="External" approximation-method="None"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.FilterContours" name="15"
enabled="true" contours-stage-name="14" min-area="500.0"
max-area="-1.0"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.MinAreaRectContours"
name="19" enabled="true" contours-stage-name="15"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRecall"
name="17" enabled="true" image-stage-name="0"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.DrawContours" name="13"
enabled="true" contours-stage-name="15" thickness="2" index="-1">
      </cv-stage>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.DrawRotatedRects" name="16"
enabled="true" rotated-rects-stage-name="19" thickness="2">
         <color r="0" g="0" b="255" a="255"/>
      </cv-stage>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageWrite"
name="3" enabled="true" file="C:\Users\Gast\Downloads\pipeline.png"/>
   </stages>
</cv-pipeline>


and the pipeline you have seen the result image:
<cv-pipeline>
   <stages>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRead"
name="0" enabled="true"
file="C:\Users\Gast\Downloads\1197bc7a-14b4-11e7-99ca-d95acf04bfb5.png"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.Normalize"
name="2" enabled="true"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.ConvertColor" name="8"
enabled="true" conversion="Bgr2HsvFull"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.Normalize"
name="9" enabled="true"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.Threshold"
name="11" enabled="true" threshold="70" auto="false" invert="false"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.ConvertColor" name="4"
enabled="true" conversion="Bgr2HlsFull"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.MaskHsv"
name="1" enabled="true" hue-min="80" hue-max="0" saturation-min="50"
saturation-max="255" value-min="200" value-max="255"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.MaskRectangle" name="12"
enabled="true" width="400" height="300"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.ConvertColor" name="6"
enabled="true" conversion="Rgb2Gray"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.FindContours" name="14"
enabled="true" retrieval-mode="External" approximation-method="None"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.FilterContours" name="15"
enabled="true" contours-stage-name="14" min-area="500.0"
max-area="-1.0"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.MinAreaRectContours"
name="19" enabled="true" contours-stage-name="15"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRecall"
name="17" enabled="true" image-stage-name="0"/>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.DrawContours" name="13"
enabled="true" contours-stage-name="15" thickness="2" index="-1">
      </cv-stage>
      <cv-stage
class="org.openpnp.vision.pipeline.stages.DrawRotatedRects" name="16"
enabled="true" rotated-rects-stage-name="19" thickness="2">
         <color r="0" g="0" b="255" a="255"/>
      </cv-stage>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageWrite"
name="3" enabled="true" file="C:\Users\Gast\Downloads\pipeline.png"/>
   </stages>
</cv-pipeline>

2017-03-30 18:20 GMT+02:00, Dimitrios <notifications@github.com>:

dzach:
@cri-s  Thank you! I've got quite a few tricks to learn from these pipelines. I edited the first post to include the original image.

dzach:
Ok. New commit is in. Please test and comment.
The descriptions of parameters tell what values are expected and their function. The can possibly be improved.

dzach:
Fixed a typo in a regexp parsing negative tolerances.

If all goes well, I'll write a section in the relevant wiki page explaining the workings of the parameters of this stage.

cri-s:
I have intendend aspect ratio width is minimum, height = maximum if tolerance is negative.
Using width/height as median aspect value and half tolerance for the swing, one needs a calculator to
know what is configured. Ok, it is better as not having it and i'm not saying you should change it.
Remove the regex for the % sign, and use boolean flag for percentage indicator, as you already do inside
code. 
The calc is wrong , specify 330 and 220 for dimension and 40 pixel tolerance, it does the calc wrong.
Percentage , i check it after you have changed the code.
rectStageName should be initialized with null or the check should be rectStageName.length==0 instead
of ==null for premature exit.
Maybe for wrong parameters, instead of thrown errors, it should just return with null, same as the case
with rectStageName.
Return value at the end , i write the better code as it should return null if no valid result is present and
not should modify return type:
  if(results.size()==0) {
      return null;
  }
 if(result.model instanceof RotatedRect)  {
      return new Result(null, results.get(0));
 }
        return new Result(null, results);

dzach:
@cri-s Thank you for the comment. I'll check all points and go through them one by one.

dzach:
@cri-s 

> The calc is wrong , specify 330 and 220 for dimension and 40 pixel tolerance, it does the calc wrong.

Please provide rect data to test the calculations, e.g. 
```
{{334.6418151855469 313.45587158203125} {66 66} -17.281497955322266} 
{{251.12911987304688 238.4286651611328} {22 41} -79.11447143554688 }}
```

dzach:
@cri-s Thanks again for the tests and the comments. I'll go through them one by one.

> 1. I have intendend aspect ratio width is minimum, height = maximum if tolerance is negative.

`width` and `height` are only naming conventions for the user. The stage has no use for the 'orientation' of the rects, i.e. it doesn't care which is width and which height.  What difference would it make to provide means for having them the other way around?

> 2. Using width/height as median aspect value and half tolerance for the swing, one needs a calculator to know what is configured. Ok, it is better as not having it and i'm not saying you should change it.

That's true for the calculator, but is the usual way tolerances are expressed. The alternative, for which much code went into this opencv stage, is to use the negative notion to define the maximum dimension expected, which does not need a calculator. Is there another way to do it?

> 3. Remove the regex for the % sign, and use boolean flag for percentage indicator, as you already do inside code.

Yes, I didn't think of that. It's better to just click a button than have to `shift` type a key, and even worse if you need to correct it. Will do.

> 4. The calc is wrong , specify 330 and 220 for dimension and 40 pixel tolerance, it does the calc wrong.

Please provide rect data to test the calculations, **and reproduce the errors**, e.g. 
```
{{334.6418151855469 313.45587158203125} {66 66} -17.281497955322266} 
{{251.12911987304688 238.4286651611328} {22 41} -79.11447143554688 }
```

> 5. rectStageName should be initialized with null or the check should be rectStageName.length==0 instead of ==null for premature exit.

Yep. Will do.

> 6. Maybe for wrong parameters, instead of thrown errors, it should just return with null, same as the case with rectStageName.

I did this as a way to let the user know what went wrong. Since this stage would most of the time be the one to provide the results, returning just null does not give any indication of the problem, when there is one, nor does it yield any results. I don't know what the consensus is.

> 7. Return value at the end , i write the better code as it should return null if no valid result is present and not should modify return type

This sounds right. Will do.

cri-s:
2017-03-31 16:15 GMT+02:00, Dimitrios <notifications@github.com>:
>
> @cri-s Thanks again for the tests and the comments. I'll go through them one
> by one.
>
>> 1. I have intendend aspect ratio width is minimum, height = maximum if
>> tolerance is negative.
>
> `width` and `height` are only naming conventions for the user. The stage has
> no use for the 'orientation' of the rects, i.e. it doesn't care which is
> width and which height.  What difference would it make to provide means for
> having them the other way around?
What i intended was, for aspect , if one search a square , then the
aspect could be from
0.97 to 1.235  . If you need to fine tune that, doing it at 1.235
aspect parameter and tolerance of -26.5% is after some time hard to
understand, with half tolerance even more, specially if
one does this infrequently.  Maybe i personally prefer doing aspect as
a hack  in this way specifying min and max value only when tolerance
is set at -0 and document it on wiki, it is possible to add this
behavier, if it's not too mutch
code ? It should not be work when tolerance is 0, only on -0 that is a
different number.

Aspect ratio, i'm proposing of removing that from UI and adding
boolean variable/flag to.,
computing the aspect ration from width and height, what do you think ?

>> 2. Using width/height as median aspect value and half tolerance for the
>> swing, one needs a calculator to know what is configured. Ok, it is better
>> as not having it and i'm not saying you should change it.
>
> That's true for the calculator, but is the usual way tolerances are
> expressed. The alternative, for which much code went into this opencv stage,
> is to use the negative notion to define the maximum dimension expected,
> which does not need a calculator. Is there another way to do it?
>
specify the min/max as on -0 hack as example.

!!! 5.
The calc is correct, sorry for that, really sorry and for the time you loose.

>> 6. Maybe for wrong parameters, instead of thrown errors, it should just
>> return with null, same as the case with rectStageName.
>
> I did this as a way to let the user know what went wrong. Since this stage
> would most of the time be the one to provide the results, returning just
> null does not give any indication of the problem, when there is one, nor
> does it yield any results. I don't know what the consensus is.
>
That is not always true. As example, if you want filter out everything
is less then 10 pixel, you setup
10, 10  for widht/height and tolerance -100 as example. Then you draw
the blobs with black color in image. This is a very usual filtering
stages. Actually i approximate it with gaussian blur and threshold
modifying the two values, but that modify the other drawing causing
slight offsets.

dzach:
Ok. Here is a list of changes I'm pushing in the new commit:

- [x] Added checkbutton for percent tolerance and removed the % symbol
- [x] Added option to show selection ranges in the Log window
- [x] Fixed premature return type;
- [x] Aspect ratio ranges now show up in the Log window e.g
`2017-04-01 03:49:19 FilterRects INFO: [PASS] aspect ratio range: 1.26 <= 1.4 <= 1.54`

When aspect ratio enters the calculation then tolerance applies to it the same way it applies to size. Therefore one can adjust the tolerance to achieve a desired aspect ratio. The results show up in the Log window if 

When there is something wrong with the data in the parameters of the stage an exception is still thrown. I see no purpose in returning null when there is misconfiguration in the parameters. One can always disable the stage. In the case where there is no detection of rects, then a null is returned.

I like it a lot more now. No calculator is needed any more.

Please test and comment!

dzach:
@cri-s would you care to explain what you mean? Aspect ratio is not required to be expressed as a percentage in any case.

cri-s:
I refer to this code:
   // check the validity of the input
        if (w == 0.0) {
          // this means h is also 0
          if (aspect == 0.0) {
            // cannot calculate absolute dimensions
            throw new Exception("Width, height, or width and aspect, or aspect, must be non zero.");
            
          } else if (!toleranceIsPerc) {
            // bad input
            throw new Exception("If only aspect is specified, tolerance must be expressed as a percentage.");
          }
        }

For aspect, w(idth) need to be zero and if aspect is not zero and no percent is used, further processing
is quit with a big error.


Anoother thing reviewing the source,
  settoleranceIsPerc(boolean toleranceIsPerc)  and gettoleranceIsPerc
need to be
    public void setToleranceIsPerc(boolean toleranceIsPerc)  and getToleranceIsPerc

It's confusing that width and height is distant, maybe param1 and param2 or rect_width and rect_height is better.

Please set logging as default off, the reason is, when there are 1000 rectangle, logging consumes some excessive time.
Time goes from 48ms to 580ms when enabling logging for a lower end count contours.

As a note to @vonnieda , the description of fields that are filled in here, it would be nice if that comes out as tooltip.

ToleranceisPerc maybe it is better if it's not abbreviated, toleranceIsPercent , other names are already longer

StringWriter consumes a lot of time, i presume is that, because filterContours need 1.5ms and
this need 22 ms for the same contours doing basically the same.

Please check you'r timings, otherwise it looks good.

dzach:
@cri-s Thank you for your comments.

That code excerpt that you mention says **`tolerance`** must be expressed as a percentage, not `aspect`, as you commented. And tolerance expressed as pixels has no meaning when there is no size expressed in pixels, while tolerance expressed as a percentage applies to the aspect ratio for any size, as I mentioned above.

Also, if width == 0 (this means height == 0 because of w = Math.max(width,height)) **AND** aspect == 0, then there is nothing that this stage can do, period. If user enters no useful data or bad configuration data, user gets an exception thrown. The stage can be disabled if it is not in use, or deleted.

I was also considering changing width and height to `rectWidth` and `rectHeight` so that they have clearer meaning and also appear together in the list, since sorting is in alphabetic order. I will change this as well as `toleranceIsPerc`, which reminds me of "perch" and sounds a little fishy.

Will turn default logging off.

cri-s:
Il sabato 1 aprile 2017, Dimitrios <notifications@github.com> ha scritto:

> @cri-s <https://github.com/cri-s> Thank you for your comments.
>
> That code excerpt that you mention says *tolerance* must be expressed as
> a percentage, not aspect, as you commented. And tolerance expressed as
> pixels has no meaning when there is no size expressed in pixels, while
> tolerance expressed as a percentage applies to the aspect ratio for any
> size, as I mentioned above.
>

If you have w=0 and aspect is not zero, there is a check that force you to
use percentage.
That percentage is used only for tolerance is irrelevant here. We refer to
the same thing.
Specify aspect ratio 2 and limit 20 pixels is the same as saying aspect
ratio of 2 with tolerance +-0.333mm  if camera is near 30 pixel/mm.  When
doing it on percent, bigger is the component more dimensional error is
allowed, for real component the inverse is true, smaller the component,
more error there have related to there size.

Also, if width == 0 (this means height == 0 because of w =
> Math.max(width,height)) *AND* aspect == 0, then there is nothing that
> this stage can do, period. If user enters no useful data or bad
> configuration data, user gets an exception thrown. The stage can be
> disabled if it is not in use, or deleted.
>

I have not commented about that.

> I was also considering changing width and height to rectWidth and
> rectHeight so that they have clearer meaning and also appear together in
> the list, since sorting is in alphabetic order. I will change this as well
> as toleranceIsPerc, which reminds me of "perch" and sounds a little fishy.
>
Do the same for aspect.

dzach:
@cri-s  Have you looked at the stage's math and tested your proposal?  If you allow tolerance not to be a percentage this is what happens with your suggestion (aspect=2 and tolerance = 20, not percentage) on the following rect:

`{ {323.0456848144531, 190.13958740234375} 16x28 * -49.08561706542969 }`

Log output:

```
FilterRects INFO: [PASS] width  range: 18.363529205322266 <= 28.363529205322266 <= 38.363529205322266
FilterRects INFO: [PASS] height range: 4.181764602661133 <= 16.474021911621094 <= 24.181764602661133
FilterRects INFO: [PASS] aspect ratio range: -18.0 <= 1.7217124851165873 <= 22.0
FilterRects INFO: [REJ] width  range: 25.527176284790038 <= 28.363529205322266 <= 31.199882125854494
FilterRects INFO: [REJ] height range: 12.763588142395019 <= 16.474021911621094 <= 15.599941062927247
FilterRects INFO: [REJ] aspect ratio range: 1.8 <= 1.7217124851165873 <= 2.2

```

Can you make any sense out of such an output? Cause I can't. Aspect ratio range is way off anything logical, and so are the limits for the rectangle dimensions, let alone that the wrong aspect ratio limits pass the rect while it should be rejected, as happens with the proper tolerance expressed as a percentage, shown in the last 3 lines.

I think we have exhausted the subject of tolerance.  Take it as an engineer's perversion, I'd like to keep things with units clear, and avoid hidden meanings, as much as possible. Keeping units in tact is of utmost importance to any engineering problem.

dzach:
Changed some parameter names. 
I think I'm done with this PR.

cri-s:
Yes, I have done the math.
It does it right, only the debug output is wrong because it assume
percent as it does (1-tol).
To avoid this:
          if (aspect != 0) {
			double low,high;
		    if(toleranceIsPerc) {
				low=aspect * (1 - tol); high=aspect * (1 + tol);
			} else {
				high= (w/(h - tol * hh));				
				low=(w/(h + upper_h));
			}
            logInfo(pass + " aspect ratio range: " + low + " <= " +
(rw / rh) + " <= " + high);
          }

dzach:
Oops! Thanks for catching that @cri-s .
Will fix it and recommit.

cri-s:
Use
                                high= (w/(h - tol * hh));
                                low=(w/(h + upper_h));

for both, percent and not, because then it display -33% correct,
otherwise actually it print that out wrong.

2017-04-02 1:48 GMT+02:00, Dimitrios <notifications@github.com>:

vonnieda:
Is this ready to go, or is there a bit more to do?

dzach:
No, there is more to do.

dzach:
I'm doing once again a change of `rectWidth` and `rectHeight` to `width` and `length`, to be consistent with the terminology used in the Packages panel. _Height_ and _length_ are used differently by different application, so keeping the same terminology within the same applicationhelps hopefully minimizes confusion.

I'm also splitting `tolerance` to `widthTolerance` and `lengthTolerance` to be able to deal with situations where while the body of a part is easily detected, the metal pin/pad pixels may be changing under different lighting conditions, thus the stage might fail to detect. In such a situation `length` (the narrow side) should have a tighter tolerance than `width`, as `width` should take into consideration the detection or no detection of the pin/pad pixels. See the images below:

result1661389490739781924
_`tolerance` = -10, `width` = 42, `length` = 25 . Old procedure. The same `tolerance` applies equally to `width` and `length`, is wide enough to allow for differences in the `width` of the desired parts, but fails because it also detects the metal strip of an undesirable part._

result6900090335188959950
_`lengthTolerance` = -5, `widthTolerance` = -10, `width` = 42, `length` = 25. New procedure. A tighter tolerance is applied to `length`, so the narrow strip of the undesirable part is filtered out._

cri-s:
The filter stage was ok, why you change it now?
Only the debug output was wrong.

For loose feeder, you need a uniform background color that you can
easy filter out,
use colored paper or thicker paper using a color that is difficult to
find on components.
Filter it out with maskHSV.
Then do contour detection using External countours only.
convert the contours to RotatedRectContours.
Do filtering with this filter.


2017-04-02 17:52 GMT+02:00, Dimitrios <notifications@github.com>:

dzach:
@cri-s thank you for bearing with me.

> The filter stage was ok, why you change it now?

Two reasons: 
the first, I explained above. The second is that that code was bad. Even I could not easily read it. This version is a lot easier to maintain, I believe. 

> For loose feeder, you need a uniform background color that you can easy filter out, easy filter out, use colored paper or thicker paper using a color that is difficult to find on components.

Have you noticed the difference in the lighting? I doubled the separation of the 4 LEDs and things improved a lot. I have a number of colored cardboards next to the sanded gray ABS plate that you see, this one is an almost black gray, for the PLCC-6 LEDs:
result_8630079874682112492

but do you see any point in testing with such conditions? There will never be a missed detection to test.
So I prefer the gray ABS plate for testing. 

It's a pity I can't use the full 1280x1024px resolution of these cameras with OpenPnP, and stay with 640x480px. In one setup I had 6 of these working in a 360° configuration with _mjpg_streamer_. I'm now waiting to test the new work of @vonnieda with the camera drivers.

dzach:
Question:

Should I include code to read directly contours, i.e. copy 5 lines of code from MinAreaRectContoures, instead of reading RotatedRects? That would make FilterRects independent from MinAreaRectContoures and reduce the length of the stage list. It would make it also kind of a "mega" stage, but Jason's philosophy is to keep them simple so they can be reused easily. But is there any other pipeline stage  that outputs RotatedRects?

cri-s:
On old code you need to use -20% to have similar behavior of new code and limits.
Gray or black background are not good generic backgrounds. Green or pink backgrounds are excellent backgrounds.

dzach:
Pushed a new commit.
Spent some time trying to make Log data useful for post-processing, e.g. statistics abt how tuned detection is etc.

I also tried to improve the execution time, and I might have succeeded to save a few CPU cycles here and there. 

Please test and comment once more.
Thanks!

dzach:
This is all nice (apsect 2.0 any size):

result_5420731905792913451

... but the screw at the bottom says applying tolerance to aspect might ~~~be screwed~~~ need a little more attention:

result_5870436065425956927

cri-s:
The unrecognized resistor mean, you need additional (gaussian)blur and
threshold for the
contours in order to remove that noise. publish a photo with this
background without any
additional drawings in it..
The new code is buggier than the older code.
Please check the necessity for separate width/height limits and not
using the % for the same
thing.




2017-04-03 15:24 GMT+02:00, Dimitrios <notifications@github.com>:

dzach:
These images are only present because they helped identify a problem when applying tolerance to aspectRatio, the are not the problem themselves. In any case, if you want to try it as an excersize, here is one similar to the original.

result_5822148720993644498

cri-s:
That is a good image for detection, can you please make more image shoots like this not for filterRect but
for loose feeder, eventually with more components and similar components too,

dzach:
Sure. Here are a few:

loose_6475804881969579554
loose_2023604720924459451
loose_7579681608410872380
loose_3947585495643440989
loose_750681448779290302
loose_9028477986572849606
loose_6485248539885362866

Some differ only by one component being on its side, or upside down.

cri-s:
On 4th image from previous post pin1 orientation is not machine recognizable, on other image it could be detected.
It is possible that you recreate that situation, make a shoot and travel over soic8 center, make another shoot, move can that ic is under the spot glaze from the LEDs and make another shoot. As example the ldo if its under the spot glaze the molded hole on the body reflects.

vonnieda:
@dzach Is this PR ready to go?

cri-s:
Just to show, i have clustered the color to extract the background color, the first image is before lightning modification, the second after.
gray
color

dzach:
@vonnieda I just pushed the latest code, which I've been using for an couple of days to see how it works. I like it a lot, but had hoped for a clean UI, and I am not sure it is such. The main reason is that the concept of `tolerance` although common in engineering, doesn't help much with the pipeline stages, and the need for a calculator is still there, as @cri-s commented in the beginning.

I have second thoughts about requesting to include this stage to openpnp, at least as it stands now. I'd like to try changing the parameters to more common ones, similar to those used in the other pipelines e.g.:
```
aspectRationMax
aspectRatioMin
contoursStageName
enableLogging
lengthMax
lengthMin
widthMax
widthMin
```
The number of the parameters is the same, but I think their meaning is clearer and similar to the min-max parameters used throughout openpnp's pipelines. `Tolerance` is gone.

I'd also like to have some advise on the question I asked above:
>Question:
> Should I include code to read directly contours, i.e. copy 5 lines of code from MinAreaRectContours, instead of reading RotatedRects? That would make FilterRects independent from MinAreaRectContours and reduce the length of the stage list. It would make it also kind of a "mega" stage, but Jason's philosophy is to keep them simple so they can be reused easily. But is there any other pipeline stage that outputs RotatedRects?

Ii did, this FilterRects (or with a better name) stage would be an alternative to MinAreaRectContours.

So I think it's better to close this PR now and resubmit a new one, once the new code is ready.

dzach:
> Just to show, i have clustered the color to extract the background color, the first image is before lightning modification, the second after.

@cri-s, I am very interested in working with the loose feeder pipeline, as you can tell from the above, but maybe we should open another issue for that subject, and exchange observations there, that may eventually end up on some wiki page, as a reference for others too.

vonnieda:
@dzach Sounds good. If you'd like to submit later, that's fine with me. I think it might be worth it to have a very simple version of this without all the bells and whistles, but no one is clamoring for it so it can certainly wait until you are happy with it.

As to your question, as you noted, I prefer the stages to be very short, simple and focused. The purpose of the pipeline system is to allow people to build up complexity with simple to understand stages. Currently only MinAreaRect and MinAreaRectContours output RotatedRect, but RotatedRect is considered a core, basic, type so I do expect other stages to output them eventually.

So, my opinion is that if the only benefit of adding the contour code would be to decrease the number of stages in a pipeline, then no, I don't think you should. That reduces flexibility. We already have a stage that does that part, so I think we should keep it that way.

That being said, I am not following this thread as closely as you and @cri-s are, so if you feel it's important for the stage then I won't block it.

dzach:
@vonnieda I'm afraid I don't know what a very simple version of this might be. Having a min and max is the simplest form of parameters that can be. Multiply this by 3, because there are 3 dimensions we are manipulating, width, length and aspect ratio, makes the list of parameters I mentioned above, which also happen to be the ones that don't need a calculator to set.

I'll leave the input to the stage then as it is now, expecting RotatedRects and not contours.

vonnieda:
@dzach Apologies. I was under the impression that the scope of this feature had grown quite a bit over the course of the discussion. I'll leave it to you. When you are happy with it, let me know.

dzach:
@vonnieda No, you are right, the initial scope changed to include @cri-s 's suggestion for including aspect ratio checking.

I mistakingly just included some more commits to this PR, thinking that I could push separately some other minor commits. I don't know how to correct this, it was supposed to be a single file change, but now its more than one.

vonnieda:
@dzach Since the other changes are just descriptive, I think it's okay. Do you want me to go ahead and merge this, or are you still working? If you are still working, it may be best to close this PR and just re-submit when you are ready.

dzach:
I'm testing the new code and will commit it very soon.

dzach:
Ok, I think I'm done. I did enough tests to gain confidence that this thing works as intended. 
To do calculated (non real life) testing I developed a new pipeline stage, RotatedRectsRead, which reads rotated rectangles from a file. It's not a fully fledged testing suit, but helps testing by keeping the input data stable. The code (see also #511), is [FilterRects.java.zip](https://github.com/openpnp/openpnp/files/907857/FilterRects.java.zip).

A simple pipeline to test this RotatedRectsRead stage together with FilterRects would be:
```
<cv-pipeline>
   <stages>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRead" name="2" enabled="true" file="/home/<user>/.openpnp/vision/white.png"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.MaskCircle" name="3" enabled="true" diameter="0"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.RotatedRectsRead" name="1" enabled="true" file="/home/<user>/.openpnp/vision/rects.dat"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DrawRotatedRects" name="0" enabled="true" rotated-rects-stage-name="1" thickness="1"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.FilterRects" name="results" enabled="true" width-max="0.0" width-min="0.0" length-max="0.0" length-min="0.0" aspect-ratio-max="1.1" aspect-ratio-min="0.9" enable-logging="true" rotated-rects-stage-name="1"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DrawRotatedRects" name="10" enabled="true" rotated-rects-stage-name="results" thickness="2">
         <color r="255" g="0" b="255" a="255"/>
      </cv-stage>
   </stages>
</cv-pipeline>
```
and some rotated rects for simple testing in file `rects.dat`
```
# numeric fields are:
# x y width length angle
#
# aspect ratio 2
50 50 25 50 0
# Aspect ratio 2, +10% off length
130 90 100 55 15
# Aspect ratio 2, -10% off length\
300 100 100 45 15
# Aspect ratio 2, +10% off width
250 150 110 50 30
# Aspect ratio 2, +10% off width
400 150 90 50 30
# Zero rect dimensions
0 0 0 0 0
0 0 100 0 0
0 0 0 100 0
# square
100 200 100 100 45
# 4/3
250 250 80 60 60

```

@cri-s thank you for your suggestion about _aspect ratio_ and for the testing and comments you have done. Hopefully, this PR is now completed.

src/main/java/org/openpnp/vision/pipeline/stages/FilterRects.java
```diff
@@ -0,0 +1,253 @@
...
+            // Any size:
+            // Select rects based only on the aspect ratio of each input rectangle
+            lmin = rl;
+            lmax = rl;
+            wmax = rl *armax; 
+            wmin = rl * armin;
+          }
+
+          // check criteria
+          if (rw >= wmin && rw <= wmax) wok = true;
```
vonnieda: Please see https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style with regards to use of optional braces. There's several instances of that in this file, so I won't flag each one, but please fix them all.

vonnieda:
Thank you @dzach!

dzach:
Thank you @vonnieda and @cri-s  !

Issue #462, opened 2017-03-15
bsh is not working on windows build fc9736


cri-s:
I'm open this low priority issue for someone other.
bsh scripts don't appear on script menues, including after new scan, so i assume that bsh is not recognized
as valid scripting resource. As i have distributed various .bsh script to some users receiving feedback
, i know that in some cases it worked. I include a bsh script that center camera over fiducial using hough circles.
[centerCircle.zip](https://github.com/openpnp/openpnp/files/844252/centerCircle.zip)

Regards
Cri.

vonnieda:
Confirmed not working on Windows. Works okay on Mac. Choosing Menu -> Scripts -> Examples -> Print_Scripting_Info.js shows that the bsh interpreter is not being loaded. Not sure why. Could be a Windows specific bug in bsh.

```
2017-03-15 08:44:23 SystemLogger INFO: Scripting Environment
2017-03-15 08:44:23 SystemLogger INFO:   Global Variables
2017-03-15 08:44:23 SystemLogger INFO:     config: org.openpnp.model.Configuration@7a664378
2017-03-15 08:44:23 SystemLogger INFO:     machine: org.openpnp.machine.reference.ReferenceMachine@59eee832
2017-03-15 08:44:23 SystemLogger INFO:     gui: org.openpnp.gui.MainFrame[frame0,77,41,1523x914,invalid,layout=java.awt.BorderLayout,title=OpenPnP - Untitled.job.xml,resizable,normal,defaultCloseOperation=DO_NOTHING_ON_CLOSE,rootPane=javax.swing.JRootPane[,8,31,1507x875,invalid,layout=javax.swing.JRootPane$RootLayout,alignmentX=0.0,alignmentY=0.0,border=,flags=16777673,maximumSize=,minimumSize=,preferredSize=],rootPaneCheckingEnabled=true]
2017-03-15 08:44:23 SystemLogger INFO:     scripting: org.openpnp.Scripting@2c93e75f
2017-03-15 08:44:23 SystemLogger INFO:   Scripts Directory: C:\Users\Jason\.openpnp\scripts
2017-03-15 08:44:23 SystemLogger INFO:   Scripting Engines
2017-03-15 08:44:23 SystemLogger INFO:     Engine: Oracle Nashorn (1.8.0_66)
2017-03-15 08:44:23 SystemLogger INFO:       Language: ECMAScript (ECMA - 262 Edition 5.1)
[...]

PR #496, opened 2017-03-30
workaround for  #462

cri-s:
This is a workaround for  #462 until the bug is resolved .

### DIFF
--- a/src/main/java/org/openpnp/Scripting.java
+++ b/src/main/java/org/openpnp/Scripting.java
@@ -98,7 +98,7 @@ public class Scripting {
         try {
             watcher = FileSystems.getDefault().newWatchService();
             watchDirectory(getScriptsDirectory());
-            new Thread(() -> {
+       	Thread thread = new Thread(() -> {
                 for (;;) {
                     try {
                         // wait for an event
@@ -107,12 +107,13 @@ public class Scripting {
                         key.reset();
                         // rescan
                         synchronizeMenu(menu, getScriptsDirectory());
-                    }
-                    catch (Exception e) {
+                    } 	catch (Exception e) {
                         e.printStackTrace();
                     }
                 }
-            }).start();
+            });
+            thread.setDaemon(true);
+            thread.start();
         }
         catch (Exception e) {
             e.printStackTrace();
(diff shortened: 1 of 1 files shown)


vonnieda:
Hi @cri-s, I think this can be fixed much more simply like this:

[patch.diff.txt](https://github.com/openpnp/openpnp/files/882951/patch.diff.txt)

This small patch fixes it for me. Can you try it?

vonnieda:
@cri-s Thoughts on my comment above? I think we can fix this with a much smaller change. See the diff.

cri-s:
I test it out today, i wanted it test yesterday but it was not possible.
I want test it out, not that it uses stricter java rules and then
limits the scriptings a lot,
not allowing easy access to private variables.

2017-04-02 3:51 GMT+02:00, Jason von Nieda <notifications@github.com>:

cri-s:
Unfortunatly is not the same, i don't have tested private variable access.
running a simple script with you'r mods, the strict java mode is enforced.
It produced only the error message:
2017-04-02 23:04:00 MessageBoxes DEBUG: Error: javax.script.ScriptException: Sourced file: eval stream : Attempt to resolve method: size() on undefined variable or class name: locs : at Line: 24 : in file: eval stream : locs .size ( ) 
 in eval stream at line number 24

 I post the code, line 24 begins with i=

	locs = machine.getProperty("bsh-locations");
	if(locs==null) locs=new ArrayList();
	i=locs.size();


Running bsh in strict java mode is useless as there are other engines when using java mode that can
handle arrays and other java language features better.

cri-s:
PS: as scripting class is visible in script interpreter, it should be modified in order to allow access to 
the file item that are used to call the script in order to know the own filename, the base directory.
Example file as variable of type File inside class declared as public or if protected/private works as such.

vonnieda:
@cri-s I think you just have a script error. This works fine for me:

```
print("BSH is working");
locs = machine.getFeeders();
if(locs==null) locs=new ArrayList();
i=locs.size();
print(i);
```

Output is:
```
BSH is working
8
```

cri-s:
Yes, your patch works.

vonnieda:
Fixed in https://github.com/openpnp/openpnp/commit/06b337d282124db35689f7d7b668658ef98431a4

PR #497, opened 2017-03-31
added tray rotation and measurement of tray item positions/rotation

sp-apertus:


vonnieda:
@sp-apertus Can you provide a detailed description of how this all works? I'm finding it a bit confusing. What I'd like to see is:

1. What are the First Component and Row 1 Last Component fields used for?
2. How does the Calculate X-Offset and Tray Rotation button work? Or more specifically, how should it be used?

Also, I'm not sure if this is an error or if I did something wrong:

screen shot 2017-04-01 at 9 13 46 pm

I captured the position of two components in a strip (so, a 1D tray effectively) and hit the calculate button and got this result.

sp-apertus:
You need to set the Tray Count X to a value higher than 1.
I guess its a good idea to add a pop up message with this info if the value is 1, will take care of this.

sp-apertus:
pushed:https://github.com/sp-apertus/openpnp/commit/a37383b4c58460c0aeb122f9be4751c4ece670fc

cri-s:
Below is my code to rotate feeders based on two fiducials inside beanshell scripting:
For the Modified TrayFeeder, if the PR get merged, how to rotate it based on location changes ?

	map(Feeder feeder, boolean forward,Location pos1, Location pos2) {
		fdr.location = map(fdr.location,forward,pos1,pos2);
		if("org.openpnp.machine.reference.feeder.ReferenceStripFeeder"		.equals(feeder.getClass().getName())) {
				fdr.referenceHoleLocation=map(fdr.referenceHoleLocation,forward,pos1,pos2);
				fdr.lastHoleLocation=map(fdr.lastHoleLocation,forward,pos1,pos2);
				fdr.visionLocation=map(fdr.visionLocation,forward,pos1,pos2);
		} else
		if("org.openpnp.machine.reference.feeder.ReferenceDragFeeder"		.equals(feeder.getClass().getName())) {
				fdr.feedStartLocation=map(fdr.feedStartLocation,forward,pos1,pos2);
				fdr.feedEndLocation=map(fdr.feedEndLocation,forward,pos1,pos2);
				fdr.pickLocation=map(fdr.pickLocation,forward,pos1,pos2);
		}

cri-s:
Why you could not only update the wizard and leave the Tray feeder as is ?

sp-apertus:
> Why you could not only update the wizard and leave the Tray feeder as is ?

What do you mean?

Measuring the tray with two fiducials is a great idea!

BendRocks:
> Measuring the tray with two fiducials is a great idea!

Yes, agree. I make simple trays with a laser cutter using some SW I wrote and it generates fiducials automatically in the output DXF. I hope some day to take advantage of them. 


image

cri-s:
@bendrocks 
I can you give a stripped down version For the tray feeder with fiducials, it don't works with this PR and on unmodified spi any error is silently discarded, the user need to double check if enabled/disabled was sucessfull.

sp-apertus:
Requested Documentation:

Usage guide:
1. Move the camera over the center of your first component in the tray to be picked up and set "First Component X, Y and Z coordinates"
2. Move the camera over the last component in the first row of the tray and again set "Row 1 Last Component X, Y and Z coordinates"
3. Set "Count X" to the number of components in the first row of the tray.
4. Optionally set "Count Y" if your tray as multiple rows and "Offset Y" (note that it should be a negative number if the second row is below the first one) to the nominal Y axis tray grid distance
5. Click "Calculate X-Offsets and Tray Rotation" Button

Your tray should now be set up successfully.

Will record screencapture video shortly.

sp-apertus:
Video showing how it works: https://www.youtube.com/watch?v=q5tu596jk_A

vonnieda:
Thanks @sp-apertus. I think this is a nice improvement, but I find it somewhat confusing and I think other users will too. Is there any reason not to detect both X and Y? It seems odd that we're auto detecting X but the user has to fill in Y. The added complexity of "note that it should be a negative number if the second row is below the first one" makes it more confusing.

If we're going to add this kind of auto setup, I'd like it to be more complete. What about a process like:
1. Ask user to locate the top left component in the tray.
2. Ask user to locate the bottom right component in the tray.
3. Ask user to enter width of tray in parts (x count).
4. Ask user to enter height of tray in parts (y count).
5. Calculate offsets and rotation.

In addition, I'd prefer if it used the camera instruction system, similar to how the strip feeder auto setup works. Guide the user through the process rather than just giving them a bunch of fields to guess at.

This isn't a final decree or anything. I'm open to further discussion or convincing.

sp-apertus:
I agree that the usability could be improved.
But I am not a big fan of the strip feeder auto setup I have to admit

>    Ask user to locate the top left component in the tray.
>    Ask user to locate the bottom right component in the tray.
>    Ask user to enter width of tray in parts (x count).
>    Ask user to enter height of tray in parts (y count).
>    Calculate offsets and rotation.

That process would only work if we assume that X and Y offsets are the same value which might not be the case for all trays. 

So I would rather propose to define the tray through 3 positions:
- first component
- last component in first row
- last component in last row

BendRocks:
> That process would only work if we assume that X and Y offsets are the same value which might not be the case for all trays.

I don't think that's true. This is the same setup info required on the TVM920 software (mark first part in array, mark last part in array, enter number of rows and columns) and it works fine.

vonnieda:
> That process would only work if we assume that X and Y offsets are the same value which might not be the case for all trays.

Why is that? It's the same thing, just done twice. offsetX = deltaX / countX, offsetY = deltaY / countY.

cri-s:
For quadratic square part , only 28mm jedec trays have X and Y the same offset.
If you need to travel over 300mm with a narrow camera that don't have
the whole component in view and you see only you'r setup is not nice.
Trays are for wider sized components and
i see no reason to complicate the live because the pickup error is
minimal compared to the
component size in order to justify going to the last components.
Further for trays there exist tables that are used for setup, example:
Metric	JEDEC	VER	TYPE columns  Rows   Cells (+X)	(-Y)	-Y step  +X Step
TQFP	CS-007	AA	5mm Square	12	30	360	10.20	11.05	10.10	10.50
TQFP	CS-007	AB	7mm Square	10	25	250	11.25	11.10	12.20	12.60
 QFP	CS-004	AA	10mm Square	6	16	96	18.30	17.25	18.70	19.86

Personally i see no justification that someone should not align the
tray to x/y axis.
Small errors are not a issue similar to resistor strips.
Please use annother tray name for this feeder and don't remove the
ReferenceTrayFeeder.


2017-04-07 17:32 GMT+02:00, Jason von Nieda <notifications@github.com>:

sp-apertus:
> I don't think that's true. This is the same setup info required on the TVM920 software (mark first part in array, mark last part in array, enter number of rows and columns) and it works fine.
Are you sure this process takes tray rotation into account or does not assume the tray slots are square?


> Why is that? It's the same thing, just done twice. offsetX = deltaX / countX, offsetY = deltaY / countY.
I created an illustration to explain the issue:
Left and right trays have the same first/last locations, they also have the same Count X and Count Y:
rotated-tray-illustration

So either one of tray rotation, x-offset or y-offset values needs to be defined to calculate the others.

BendRocks:
OK I get it, you are correct, the assumption on the TVM920 is that things are square and if you add tray angle to the mix (which I *do* think is valuable) then you don't have enough info to solve using start, end, row count, col count.

vonnieda:
@sp-apertus I *think* this is possible, but I would need to sit down and really try to work it out. Here's my idea, roughly:

1. Call the first point A and the second point B.
2. Find a point C that lies at the intersection of lines drawn through A and B which form a right angle.
3. Determine the angle R of the line AC.
4. Rotate B by the negative of R, creating BR.
5. Calculate offsetX = (BRx - Ax) / countX, offsetY = (BRy - Ay) / countY.

My geometry is too rusty, and it's too early in the morning for me to prove this all out, but I drew it up a bit and it looks okay to me. Let me know if you see any obvious problems and I'll see if I can prove it later today when I have a bit more time.

sp-apertus:
> Find a point C that lies at the intersection of lines drawn through A and B which form a right angle.
Let the coffee sink in :)

There are infinite points that fulfill this these conditions:
pict46

vonnieda:
Right you are! :)

Okay, subject dropped. I will look closer at the PR later and see if I can
suggest some small UI changes that will make it (IMHO) less confusing
without having to change everything.

Jason

sp-apertus:
Right :)

@vonnieda what do you think of @cri-s idea to fork the old tray feeder, then we can be a bit more experimental and collect user feedback without affecting the old tray functionality.

vonnieda:
That would be fine with me.

vonnieda:
Closing this PR, will accept a PR for a new feeder class when it's ready as discussed above.

Issue #501, opened 2017-04-02
Auto Panelization Improvements


vonnieda:
This is a follow up to #456 with some suggestions for improvements to the new feature.

* [x] Move DlgPanelXOut and DlgAutoPanelize out into their own files. JobPanel is already too long and these can be cleanly broken out.
* [x] Bug: Add two unique boards. Click the panelize button. Get an error that only one board can be used but dialog opens anyway. After this the state of the add board and delete board buttons is broken.
* [ ] Enhancement: Would like to see the single board requirement removed. Everything is in place now for multiple panels, so even if it's not used, there's no reason to have all the special cases for a single board and it just makes it confusing.
* [x] Bug: Opening the panel dialog and then closing it without hitting Ok leaves the job in an unloadable state due to the part-id in the panel not being populated.
* [x] Documentation in https://github.com/openpnp/openpnp/wiki/Panelization

vonnieda:
FYI: I am currently working on moving the dialogs into their own files.

BendRocks:
For the "Opening the panel dialog and then closing it..." issue you note, is that as simple as adding the require=false attribute on partId field? Or is something bigger needed? I tried it and it seems to work but I want to make sure I'm not missing something larger

    @Element(required = false)
    private String partId;

tonyluken:
@vonnieda,

Can we close this issue as I believe it was address by [PR1507](https://github.com/openpnp/openpnp/pull/1578)?

vonnieda:
Yes, looks good to me. Thank you @tonyluken!

PR #503, opened 2017-04-02
Panelize icon state bug #501

BendRocks:
This should fix bug reported in #501. With this, it shouldn't be possible to autopanlize with 2 or more boards in the job list as the autopanelize icon is disabled. But in the event it does happen via unforeseen events (eg future changes to tricky icon logic), the user will see a dlg telling them you cannot panelize with 2 or more jobs present and the dlg will not longer show.  

There was also a bug related to normal board add functionality getting disabled when it shouldn't have been due to panelize logic. That was fixed too. 

There's a new issue that might be related to breaking out UI code. The drop combo box for global fiducial part no longer will drop down into a list when mouse clicked. It's just a combo box and you can only scroll via keys. Perhaps that is a UI switch that wasn't set?


vonnieda:
Hi @BendRocks, I can't reproduce the combo box error. Can you show a video or something? Seems to be working okay here. See https://www.youtube.com/watch?v=S0B__0m4Fts

BendRocks:
I see what it is--if the part-id in the job you open doesn't exist in the parts database, then the dropdown doesn't happen.  You can only scroll through the items with arrows.

What would you prefer to happen in this case? When the dlg is being set up, if the part doesn't exist in the database then reset it to the first item in the db?

vonnieda:
I think that'd be fine, or just leave the part set to null and handle the null case in the code. The null handling may be better since that could be the situation and the user may not open the dialog before starting a job.

As for not showing the dropdown, that sounds like a bug. I'd expect the same thing to happen in the placement parts dropdown, or any feeder part selection dropdown. Maybe there is a difference in how it's being initialized.

PR #504, opened 2017-04-04
DRO multistate pr

BendRocks:
The DRO today can be cycled from current nozzle -> relative with a mouse click. This adds a 3rd option, allowing you to cycle from currentnozzle -> relative -> defaultcamera. Tools tips were added so a hover over the DRO will remind you what you are looking at. 

I haven't found a good way to locate features on my machine. For example, if I navigate the over a hardware fiducial used for homing the machine, there is no way to read off the camera location. Yes, I can read the nozzle location and then go in and calculate the offset, but that needs paper and pencil.

Also, let's say you want to make a 100 mm move that might bring you close to a limit. Currently, it's not easy to know if a move of 100mm will hit an end stop or not by looking at the DRO.  And if you have several nozzles, it gets even more confusing as you need to understand which nozzle might be selected.

For me, it makes a lot of sense to think in camera/head locations instead of 1-of-4 nozzle locations. 

This uses the defaultcamera on the head. I double checked things still work if there's not a default camera. Ideally, without a default camera, it'd revert to the two swaps as before. 

Changes were to a single file. If approved, I'll also look through the docs when the time is right and update the text.

### DIFF
--- a/src/main/java/org/openpnp/gui/MachineControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/MachineControlsPanel.java
@@ -173,7 +173,7 @@ public class MachineControlsPanel extends JPanel {
 
     public void updateDros() {
         Location l = getCurrentLocation();
-        if (l == null) {
+        if (l == null) { 
             return;
         }
 
--- a/src/main/java/org/openpnp/gui/MachineControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/MachineControlsPanel.java
@@ -73,10 +73,20 @@ public class MachineControlsPanel extends JPanel {
 
     private JogControlsPanel jogControlsPanel;
 
-    private Location markLocation = null;
+    private enum DroDisplayEnum {NOZZLE, SAVED, DEFAULTCAM;
+    		private static DroDisplayEnum[] vals = values();
+    		public DroDisplayEnum next(){
+    			return vals[(this.ordinal()+1) % values().length];
+    		}
+    }
+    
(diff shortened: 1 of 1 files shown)


cri-s:
You do it wrong, because you suppose that camera is the same as machine location.
This is extreme dangerous, as if camera need to be touched (refocussed, ...) there are 2-3 hours to setup
the machien. If you have a hardware 0 point, hole , screw, .... it saves you hours of work in that case.
For Setup, just add a Nozzle with offset 0:0:0 and use that during setup.
I feel if you overload the dro with camera locations, it's too confusing.

vonnieda:
Hi @BendRocks, let me offer an alternate solution:

The nozzle selection dropdown was originally a "tool" selection dropdown, which could include basically anything mounted to the head. This would allow you to see the coordinates of any tool whether it was an actuator, camera, nozzle, dispenser, etc. This functionality was lost in a refactor a long time ago because I needed to get more important work out the door. Why don't we just add that functionality back? The bones for it are already in place with getSelectedTool() in MachineControlsPanel (IIRC, not looking at the code ATM). I've been intending to add this functionality back anyway because it's needed for dispense.

I think it would be a very simple change. The only real "catch" is that some parts of the code rely on getSelectedNozzle() and I think that we'd just make that function return getDefaultNozzle() if something other than a nozzle is selected.

BendRocks:
This makes perfect sense, I'll put something together tonight and try it on for size

BendRocks:
Can be closed. Resubmitted as #507. I would have done as a continuation of this, but my git skills are not quite there yet

Issue #279, opened 2016-05-26
Park Improvements


vonnieda:
Originally from #273, some improvements to the Park system are needed:
- Park should be a method on head rather than being a bunch of copypasted code everywhere. One thing to consider is that this would break the UI which allows parking of XY, C or Z. Maybe the method would take those as booleans.
- Park should always park at the same location by subtracting the offsets of the object it moves. Due to the driver API we can't just send an absolute coordinate, but we can pre-subtract the head offsets from the default nozzle, for instance, and send that.

cri-s:
Take care for machines having more then one head, and for that park can issue a exception if the move is not allowed because of position from other head.

vonnieda:
@cri-s Do you know of any machines currently using OpenPnP with more than one head?

PR #507, opened 2017-04-05
Add Camera to Nav Combobox and then show selected item coords in DRO

BendRocks:
This is continuation of #504....This is to check if on the right path (not a lot of changes were needed). There was CameraItem and NozzleItem already. Looks like these are used to give a nice output string. I created a headmountableitem and derived cameraitem and nozzleitem from that, and then used the headmountable item in the combo box. Combobox populates as expected (with nozzle and head camera) and DRO reads as expected based on which is selected in the listbox. 

One bug I might have found that is unrelated to these changes: If you have two nozzles, one offset of -10,-10 and another of -5, -5, and you XY park, your absolute park location seems to depend on which nozzle was selected when you parked. That seems not right to me as the docs say park should park the head at its park location. If you confirm that is incorrect, I'll fix that in this pr.

### DIFF
--- a/src/main/java/org/openpnp/gui/MachineControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/MachineControlsPanel.java
@@ -45,6 +45,7 @@ import org.openpnp.gui.support.HeadMountableItem;
 import org.openpnp.gui.support.Icons;
 import org.openpnp.gui.support.MessageBoxes;
 import org.openpnp.gui.support.NozzleItem;
+import org.openpnp.machine.reference.ReferenceCamera;
 import org.openpnp.model.Configuration;
 import org.openpnp.model.Location;
 import org.openpnp.spi.Camera;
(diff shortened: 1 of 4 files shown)


cri-s:
Is this then in camera coords or machine coords?
Park location is a location where the nozzles park, depending on last used and last head used.

vonnieda:
@BendRocks Looks perfect. As to the park problem, see https://github.com/openpnp/openpnp/issues/279. I fear that might actually be a big change, but I haven't fully thought it through. If it's more than a couple lines I'd prefer if it's done as two PRs.

vonnieda:
@BendRocks Is this ready to go?

BendRocks:
>Is this ready to go?

Almost! @vonnieda, could you take a look around line 382 in MachineControlsPanel and explain what is going on there? This looks listener code looks like code that will run when a nozzle gets added (the IF part) OR if a first nozzle is added (the ELSE part). 

Is that right?

The reason I'm confused is I'm not sure I need to extend to look for camera because if I add a camera to the head, it somehow ends up in the list box. But I don't understand how.  Any insight?

vonnieda:
The if adds a newly arrived nozzle, when a user adds one in Machine Setup, and the else removes one, when a user deletes it from Machine Setup. These are for handling dynamic updates during runtime. Everything gets loaded during startup at 359, so this is probably where you are seeing cameras added, after a restart. If cameras are showing up without a restart then I am not sure why.

I would expect you to need to update both parts for camera.

BendRocks:
OK, this should handle the runtime changes.  Unless you see something in review, I think done on my side.

I didn't try to fix anything mentioned above about #279.

vonnieda:
Great work @BendRocks, this is a nice improvement and a really nice PR!

PR #508, opened 2017-04-06
Added PPH metric to job time log item

BendRocks:
Added a var to count the number of parts placed, and then in addition to the already computed job time in the log, also print out the number of parts per hour the last placement achieved. Very helpful for machine tuning and comparing various strategies.

Verified no issue  with empty job (eg zero placements).

PS @pfried, the logging stuff is awesome


vonnieda:
Looks good, thanks @BendRocks!

cri-s:
why not include cph , components per hour, where the timer starts after aquiring the fiducials and stops
before park nozzle and eventually fire some events.
and add 10 seconds to pph for the time needed to change the pcb in order that value is near takt time cph.

PR #512, opened 2017-04-08
fixes #510

### LINKED ISSUE #510: Bottom pipeline crashes application on unchecking ImageCapture stage
# Problem Report
## Describe the Issue
Bottom pipeline crashes the application when the ImageCapture pipeline stage is unchecked.
## Steps to Reproduce
1. Start OpenPnP
2. Click the Parts panel.
3. Click on any existing part in the top panel and click Pipeline 'Edit` in the bottom panel.
4. In the pipeline stages list, click to uncheck ImageCapture
5. Program crashes immediately.

## Expected Result
Program should not crash. 

## Actual Result
Program crashed

## Notes, Log Files, Screen Captures, Videos, etc. to Show the Issue
The problem doesn't happen if an image is present through a preceding __successful__ ImageRead stage. If the ImageRead stage was __unsuccessful__, crash happens. It seems that the bottom pipeline cannot handle a null in place of the expected image, or that the exception thrown is not handled properly.

I tested the same on a loose feeder pipeline and the problem doesn't seem to happen there.

cri-s:
see #510

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
@@ -139,7 +139,7 @@ public abstract class ReferenceCamera extends AbstractCamera implements Referenc
         catch (Exception e) {
             Logger.warn(e);
         }
-        return image;
+        return last=image;
     }
     
     protected abstract BufferedImage internalCapture();
--- a/src/main/java/org/openpnp/spi/Camera.java
+++ b/src/main/java/org/openpnp/spi/Camera.java
@@ -59,6 +59,14 @@ public interface Camera extends HeadMountable, WizardConfigurable,
      * @return
      */
     public BufferedImage capture();
+               
+    /**
+     * Return the last captured image from the camera and returns it in it's native format. Fires
+     * the Camera.BeforeCapture and Camera.AfterCapture scripting events before and after.
+     * @return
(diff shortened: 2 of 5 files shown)


vonnieda:
Please use the template provided when submitting a pull request: https://github.com/openpnp/openpnp/blob/develop/PULL_REQUEST_TEMPLATE.md

cri-s:
Read This First

Be sure to review the [[Developers Guide]] and the Contributing Guidelines.

Make your Pull Request as small as possible. Only include one bug or feature. Large pull requests that change dozens of files or add multiple features are unlikely to be accepted.

Description

It fixes the #510 by providing workingImage with lastCaptured image or if that is not disponible actual capture() image. Hovewer if no camera is configured, it still returns null.

Fixes a bug.
Instructions for Use

for developer only
Implementation Details

    How did you test the change?
traying
    Did you add automated tests?
nog
    Did you follow the coding style?
yes, i think.
    If you made changes in the org.openpnp.spi or org.openpnp.model packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.

lastCaptured saves time and increase cph as it don't need wait the usual 130 ms for a new image and
further it don't call the two scriptings events that need additional time. 
Having access to last captured image can be helpful too on function that takes camera as argument and
capture image by itself.

PR #519, opened 2017-04-14
namedCsv importer change

cri-s:
# Description
code cleanup, removed trace logging entry.

# Justification
The logging entry was necessary at time of writing and later, but was hard to follow.
Now, after it have proven it works, this logging could be removed.

# Instructions for Use
This PR don't change anything for the user.

# Implementation Details
not tested, just diff checked in order it contains only Logging lines.


vonnieda:
Thank you @cri-s!

PR #533, opened 2017-04-26
Fixed an issue with the non-squareness factor

aauer1:
# Description
I used the non-squareness factor for my PNP machine. But when I moved the head only in Y-direction, I noticed that the X coordinate hasn't changed accordingly.  The original code doesn't change the x-coordinate when there is no x-movement. This is the reason why the non-squareness factor isn't used in this case.
My fix recalculates the X coordinate as soon as there is a change of the Y coordinate.

# Justification
Every user who is using the non-squareness factor will be effected by the problem. In some rare cases when the machine is only moving in the Y direction, the position will be wrong because of not using the non-squareness factor to correct the X coordinate.

# Implementation Details
- First test: I used some plotting paper and moved the head only in Y direction. With the help of the plotting paper, I can see that the camera position also changed in X.
- Second test: Inspected the debug output. If a movement in Y is started, I can see that also the X coordinate is corrected.

No

Yes, I think so.


vonnieda:
Thank you for the PR @aauer1. I'm sorry I have not been able to review it yet. I've been busy the last few nights. I will try to get it reviewed and merged by this weekend.

PR #539, opened 2017-05-07
New OpenCV pipeline stage: MaskMultiple

dzach:
# Description
A new OpenCV pileline stage is proposed, named `MaskMultiple`, which offers masking of multiple image areas based on shapes derived from previous stages, or from user provided data.

There are three OpenCV pipeline stages that provide a masking operation, namely: `MaskCircle`, `MaskRect` and `MaskHsv`. The two former apply the mask as a geometric shape at the center of the image,

screenshot-1

screenshot-3

while the latter operates on the whole image based on the color of the features.

The new `MaskMultiple` stage can apply multiple mask shapes anywhere on the image:

screenshot-2

The above image was produced by  defining shape coordinates in parameter `shapes`:
```
shapes: 640,400,50;440,532,540,532,540,632;700,532,800,532,800,632,700,632
```
### Using multiple mask shapes
Multiple shapes can be used to mask out unwanted features scattered in the image. For example, the following image contains bright spots that may interfere with the part detection process:

screenshot-7

By applying a multiple shape mask, the unwanted features can be eliminated (here, a colored mask has been used just for demonstration):

screenshot-6

The result (mask is black):

screenshot-9

### Isolating detected features

`MaskMultiple` can operate with shapes (Circles and RotatedRects) created by other stages, earlier in the pipeline. Here, the outline of a TQFP-32-8 part is used to completely isolate it from the rest of the image content:

screenshot

The result:

screenshot-5

### Using isolated features

Isolated features may be easier to work with, since most of the side effects of image processing on the rest of the image can be masked out. Here, circle detection has been used to isolate features on the previously isolated TQFP-32-8 part:

screenshot-10

# Justification
Isolating parts of an image can help in extracting useful features. Similarly, blanking unwanted areas on an image can help reduce noise and improve part detection. Users of OpenPnP's vision system will be able to utilize the new `MaskMultiple` stage to add flexibility and advanced masking capabilities to their part detection OpenCV pipelines.

# Instructions for Use

- Define the shape(s) to be used as a mask, by using results of previous stages:
screenshot-13
Strings or single numbers can be used as stage names.

- Define the shape by specifying shape coordinates:
1. Circle: `shapes: 731,942,80`
2. Triangle:`shapes: 440,532,540,532,540,632`
3. Polygon`shapes: 1090,370,985,475,985,635,1090,745,1280,680,1280,450`

- Define the shapes by combining all of the above, separated by semicolons `;`
screenshot-12

- Use the `inverted` parameter checkbox to change operation from masking to filtering.
- Use the `color` parameter to visualize the shape of the mask, or experiment with tinting the colors of the image. `Black` is the default color for masking.

# Implementation Details

Visually.

No.

I think so.

No such changes were made.


vonnieda:
Hi @dzach, this is a very interesting stage, and I think it will be very helpful.

I'd like to suggest two changes:
* Change name of stage to MaskShapes as this more clearly describes what it does. MaskMultiple is not clear. Multiple what?
* Change the format of the shapes property to more clearly define coordinate pairs. I suggest something like: "X1,Y1:X2,Y2:X3,Y3;X1,Y1:...". So, comma separates X and Y, colon separates coordinate pairs and semicolon separates shapes / stages.

Jason

cri-s:
maskShapes remember shapes from packages, why not DrawShapes ?

ps: the above image is a ideal example for the usage of maskRectangle.

vonnieda:
DrawShapes would be fine with me too. I don't understand this though "maskShapes remember shapes from packages, why not DrawShapes ?" What do you mean by this?

dzach:
Regarding coordinates, it would be easier to define the coordinates by clicking on the image, but I haven't been around that part and java GUI development is far from being easy or intuitive to me. I'll change the format of `shapes` to the suggested.

Regarding the name:

Its original name was MultiMask, but that would not put it along with the other masks in the stage list. Maybe it is better then MaskShapes, despite the sorting problem.

@cri-s I'm afraid DrawShapes doesn't even hint about the stage being a mask operation.

Thinking about it, with just a small addition this stage can replace both MaskCircle and MaskRect, as follows:
```
circle:100; 
rect: 100; 
100,100:100; 
100,100:300:200; 
100,100: 200,90: 230,180
```
The first item is MaskCircle, the second item is MaskRect, the third is a Circle located at 100,100, the fourth is a rect, specified in a similar with Circle format, located at 100,100 with width=300 and length=200, and the fifth is a triangle as it was before. This allows for additional formal or otherwise known shapes to be included at a later stage, if needed, without breaking the format of the `shapes` parameter.

Is this too complicated?

vonnieda:
@dzach In general, I prefer the stages to be as simple as possible, and to only do a single thing. I'd rather not add more complexity to this one. In fact, I was previously going to suggest you simplify this and break it up into different stages, but I figured it would do no harm to leave it.

Stages are intended to be similar to how UNIX chains text utilities together. You combine many simple things to create complex behavior. My concern with this is that as the stages become more complex, they become less flexible.

An alternate way to think of this is just changing MaskCircle to allow specification of a center point, and add MaskPolygon which takes a list of points. This would accomplish the same thing with just two very simple stages. The user can just stack as many as they need to mask out everything they want.

And to the point of Draw vs. Mask, the only difference is whether the object is filled. We eliminate the Mask* stages altogether and just have Draw* by adding a "fill" property.

cri-s:
some pnp software use the name shape instead of package,
as the packages 0603 could be named R0603 or C0603 because resistors and caps have different (package) shape for vision and using the name shape is then more intuitive as package, just because vision difference.

circle and rectangle mask have annother important function.
With negative values, it mask out the central part, for masking single or two hole nozzles for really small
component where otherwise the reflection from nozzle tip would confuse vision algorithm.

dzach:
@vonnieda Isn't stage overhead a thing to consider?

An important (for me) feature of the proposed stage is that it takes input from other stages, while MaskCircle and MaskRect to not. As a matter of fact, this is why I started writing it in the first place.

vonnieda:
I consider stage overhead to be a pretty small concern. It's only a couple
milliseconds, and in most cases these pipelines are only called once per
image.

In any case, I'm not requiring that you make the changes I suggested, aside
from the name change. I'd like something more descriptive than
MaskMultiple. If you prefer not to change the other stuff that's fine.

Jason

cri-s:
Do you know that specify size -1 and color black remove the circle,polygons, ....  from previous stages ?
If using circle, one must use drawCircle, if using contours, one must use drawContours .

dzach:
@vonnieda I  enjoy very much the discussion and appreciate the suggestions. I like the MaskPolygon name, but what about the missing functionality for Circles and RotatedRects, coming from models produced by other stages?

cri-s:
use drawRoatedRect with size -1 and color black (0,0,0)

dzach:
@cri-s Can you elaborate? Is this a comment for this proposal or a fact for something else?

dzach:
There is no `size` property in DrawRotatedRects. If I put -1 to rectCenterRadius I get:
```
CvException [org.opencv.core.CvException: cv::Exception: /home/jason/opencv/opencv-2.4.11/modules/core/src/drawing.cpp:1737: error: (-215) radius >= 0 && thickness <= 255 && 0 <= shift && shift <= XY_SHIFT in function circle
]
```

cri-s:
The parameter on DrawRotatedRects is thickness. 
If thickness is 2, the drawed border is 2 pixel width in case no AntiAlias flag is passed to the function.
If thickness is -1, the object get filled completly with the choosen color.

dzach:
@cri-s It seems that negative numbers are not allowed in `thickness` either:
```
CvException [org.opencv.core.CvException: cv::Exception: /home/jason/opencv/opencv-2.4.11/modules/core/src/drawing.cpp:1663: error: (-215) 0 <= thickness && thickness <= 255 in function line
]
```

cri-s:
tested. it works on DrawCircle and DrawContours, it was my assumption it would work on drawRotatedRect too.  Set -1 on thickness and in this two stages and it works. btw, in this case it is more probable you use contours as rotatedrect .

dzach:
@vonnieda I may be wrong, but I see similar stage times in MaskCircle, MaskRect and MaskMultiple, all around 20ms on this machine.

In any case, the change in coordinate format is done according to the suggestion, and it's much better than it was previously, a sequence of numbers. I added white space trimming when parsing the parameter, so it can be easier to read with additional spaces.

The only remaining issue seems to be the name. I propose to use MultiMask, as it sounds more like a noun than a verb, and hopefully is not as unclear as MaskMultiple. 

If that's ok, I'm ready to commit the changes.

dzach:
Committed the changes. Please have a look and comment.

dzach:
If there are still concerns about the complexity of this new stage, then here is another proposal that takes into consideration the `simple stage` principle:

1. MutliMask, or whatever it is called, splits into simpler stages. 
2. The polygon mask functionality will become `MaskPolygon`.
3. The functionality for masking model results from previous stages will become `MaskModel`, or something similar. This will make it easier to include a model `dilate/erode` functionality that I have  tested but did not include in here, trying to keep it simple. Dilate/erode is sometimes necessary in order to mask image content around the outline of shapes still remaining after `MaskHsv` operations.

cri-s:
Off topic

If you make illumination and/or centering optics from camera better,
then you could
detect bend pins and filter + count valid pins and further use masks
like this for checking
aligment using shapes reticles. Without proper illumination setup,
this never works.

dzach:
Nice! Actually this is a used ATMega328P chip that I de-soldered from a broken Arduino Nano.

atmega328p

What I am getting to, is an easy way to achieve the following result, without using SIFT or other advance feature detection methods of OpenCV:

screenshot

The above is achieved quite easily by isolating the `RotatedRect` of the chip, and then using `DetectCirclesHough` to find the dot in one of the corners. The result will be the true position and orientation of the chip, and hopefully a prescription of how to achieve such results.

dzach:
Now, the above is from the down looking camera, where the dot for pin1 is more or less obvious. Can you tel me how I can do the same with the up looking camera?

chip

This is how far I've got, at the moment.

features

dzach:
More on the above, it seems that a way to vary the illumination and/or the height of the part when trying to extract pin1 may be necessary. Fortunately this can be done with a ScriptRun with arguments.

vonnieda:
@dzach I'm sorry, the name MultiMask is no more clear than MaskMultiple. It still does not tell the user *what* it masks. In addition, the name is worse because it no longer starts with Mask, as the others do. Why not MaskShapes? This clearly tells the user what it does and keeps with the Mask* naming scheme.

dzach:
Well, it actually does all masks,  and also can do all of them at once, thus `MultiMask`.
However,  yes, why not `MaskShapes`, I am not insisting on the name. @cri-s had a concern about using the word `shapes` but it's ok with me.

Having said that, I have now proceeded with the split of this stage on the basis of making it simpler. Have already a working new `MaskModel` stage, which I've  used for one of the images above, and am working on a new `ExpandModel` that expands/contracts (dilates/erodes) a model by a given number of pixels. I prefer the `ExpandModel` name versus `DilateModel` because `erode` is usually used for pixel data while here we are dealing with a geometric model, but will gladly take suggestions for changing it. `MaskPolygon` is also ready and needs to be tested. 

So, what do you think we should do? Keep the `whatever its name` one-does-it-all stage, or go for the simpler ones?

dzach:
Of course `expand` has also other meanings that can create confusion.

cri-s:
@dzach.
Threshold to Zero, Inverted . If you don't have that option, use
findContours and
black the contours out, that is the same. Now normalize. Find Contours.
Then get COG (center of gravity of the blobs)
COG = Point(mu.m10 / mu.m00, mu.m01 / mu.m00); //find the COG of ROI
This gives you the quadrant where the marking is, after you have
translated/rotated the COG.


2017-05-08 16:30 GMT+02:00, Jason von Nieda <notifications@github.com>:

dzach:
@cri-s I guess you are replying to my  question `Can you tel me how I can do the same with the up looking camera?` but I can't follow you description. Do you have a pipeline that works with any of the images above? Or is that a suggestion for some java code (or beanshell), because `SimpleBlobDetector` does not expose any parameters. I'm keen on exploring the possibilities for detecting the markings on the bottom side of a chip, and was planning to have a look into the Blob detection code and see if I can do something there. Maybe we should open an issue here in github about this subject and post some ideas/pipelines there for reference.

cri-s:
Do you need the simoleblob pipeline?

Il lunedì 8 maggio 2017, Dimitrios <notifications@github.com> ha scritto:

dzach:
I don't know what it is, but yes.

vonnieda:
@dzach It's your call. If you are willing to rename to MaskShapes then I'll be happy to merge this as it is. If you'd prefer to hold off for the new stages, that's fine too.  I'd also accept MaskMultipleShapes, MaskMultiShapes, or other variations on the theme. I just want there to be some indication of *what* it masks. Basically: VerbNoun :)

In addition, @dzach and @cri-s, please don't use PRs for off topic discussion. This makes it much harder for me to follow and eventually merge the PR. There's plenty of other places for discussion. Making these threads dozens of comments long uses a lot of my limited time for merging PRs.

dzach:
@vonnieda Right. My apologies, that was not my intension here. I'll open a new issue for this very  subject with a goal to create a workable pipeline + whatever it takes to detect polarized part orientation, if it can be done at all with the current status of the pipeline.

I've already proceeded with breaking the stage to simpler ones and will be ready to commit separate ones soon. It is really more flexible this way. So I'm closing this one and will open separate ones, as soon as they are ready.

PR #547, opened 2017-05-09
New OpenCV pipeline stage: DilateModel

dzach:
# Description
When building an OpenCV pipeline, there are cases where the detected models of parts contain unwanted content, e.g. portions of pins, shadows etc, which may be difficult to remove without affecting other parts of the image. While processing the image content directly may be difficult in such cases, there exists an alternative  path to take: **process the geometry of the model** instead.

A new OpenCV pipeline stage is thus proposed, named `DilateModel`, that performs dilation or contraction on the shape of a model. At present, it operates on Circles and RotatedRect models. The following series of images show an example of a dilate operation.

Original image:

screenshot-3

Model masked (makes use of a `MaskModel` stage to isolate the original image content inside the model) :

screenshot-4

The result of applying the new `DilateModel` pipeline stage, with a dilation, or rather, in this case, a contraction width of -10px:

screenshot-5

_Before_ (red) and _after_ (Cyan) models, showing the result of applying a `DilateModel` stage with a contraction width of -10px:

screenshot-7

# Justification
Users of OpenPnP vision system may benefit from the additional flexibility offered by the new pipeline stage by using it together with other stages, like MaskHsv, to improve the performance of a pipeline.

# Instructions for Use
screenshot-6

- Set the `modelStageName` parameter to the pipeline stage to use as an input for the model.
- Set the `dilate` parameter to the number of pixels to dilate (negative number to contract) the model.

# Implementation Details

Visually.

No.

If memory serves me right, yes.  
___________________________________________
**NOTE:** 
_4.2 Block indentation: +2 spaces_ of the Google document asks for 2 spaces of indentation. OpenPnP seems to prefer 4 spaces instead, and keeps correcting the code. **This creates conflicts** when merging back the code from OpenPnP github to the local repository, which need to be resolved **every time**.
___________________________________________

No such changes were made.

### DIFF
--- a/src/main/java/org/openpnp/vision/pipeline/stages/DilateModel.java
+++ b/src/main/java/org/openpnp/vision/pipeline/stages/DilateModel.java
@@ -1,0 +1,109 @@
+package org.openpnp.vision.pipeline.stages;
+
+import java.awt.Color;
+import java.util.List;
+import java.util.ArrayList;
+
+import org.opencv.core.Core;
+import org.opencv.core.Mat;
+import org.opencv.core.RotatedRect;
+import org.opencv.core.Point;
+import org.opencv.core.Size;
+import org.opencv.core.Scalar;
+
+import org.openpnp.vision.pipeline.CvPipeline;
+import org.openpnp.vision.pipeline.CvStage;
+import org.openpnp.vision.pipeline.stages.convert.ColorConverter;
+import org.openpnp.vision.pipeline.Stage;
+import org.openpnp.vision.pipeline.Property;
+import org.openpnp.vision.FluentCv;
+
+import org.simpleframework.xml.Attribute;
+import org.simpleframework.xml.Element;
+import org.simpleframework.xml.convert.Convert;
+import org.pmw.tinylog.Logger;
+
+@Stage(
+  category="Image Processing", 
+  description="Dilate or contract a model by a given number of pixels.")
+  
+public class DilateModel extends CvStage {
+
+    @Attribute(required = false)
+    @Property(description="Name of stage to input model data from.")
+    private String modelStageName = null;
+    
+    @Attribute(required = false)
+    @Property(description="Dilate the model by given pixels. Negative values contract the model.")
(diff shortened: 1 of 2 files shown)


cri-s:
Why you name the stage dilate where you model get smaller, erode ?

@dzach 
As this stages are similar , why you don't make one single Stage incorporating all.
Ask @vonnieda what hi think about this.

I search out example code from my stages:

	public enum Mode {
		Test, Quantize, Add, Range, ERODE_DILATE
	}


	@Attribute
    private Mode mode=Mode.Quantize;
	
	public void setMode (Mode mode) {
		this.mode = mode;
	}

	public Mode getMode() {
		return mode;
	}

and inside code take different action based on mode
				if(mode==Mode.Quantize) {
				} else 
				if(mode==Mode.Add) {
				} else 


Further you should overload description() and description(String) , using that instead of annotation,
and depending on Mode return the appropriate String.

vonnieda:
I agree with @cri-s. I suspect you may also want to support, in the future, erosion at least in addition to dilation. It might make sense to change this to MorphModel and then add modes as you find you need them. For now it could just have the single Dilate mode.

dzach:
Erosion is already supported, I call it `contraction` since `erosion` is a word used with pixel data, I believe, but not with geometrical. Here the operation is done on the geometry. Just put negative values in parameter `dilate`.

I don't understand which stages are similar. Also, can you give me a hint about what else could be added? For Circles and RotatedRects it seems quite easy to implement.

Please don't forget that this is an `Infinite monkey` hitting keys and programming in Java. Don't expect me to understand every little hidden detail, as if I were a real Java programmer ;-)

vonnieda:
@dzach Ah, I see. I missed the negative value parameter. I think it's fine as is now. I'll merge it.

dzach:
Thanks! 
If there are any other operations that can be included here, or in a similar but more inclusive stage, please tell me. I'd like to bring this idea of manipulating the geometry of a model a little forward, if I can.
All 3 of the last stages do exactly this.

PR #551, opened 2017-05-10
Feature multiple part aligner

tracklogic:
# Read This First

Be sure to review the [[Developers Guide]] and the [Contributing Guidelines](https://github.com/openpnp/openpnp/blob/develop/CONTRIBUTING.md).

# Description
Mechanism for allowing multiple part alignment stages (i.e. bottom vision camera and mechanical centering jaws). Each alignment stage has a list of parts it can handle, which are editable via the gui. Each alignment stage has its own wizard page added to the parts settings.

# Justification
Flexibility

# Instructions for Use

# Implementation Details

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -52,13 +52,6 @@ public class ReferenceBottomVision implements PartAlignment {
     @ElementMap(required = false)
     protected Map<String, PartSettings> partSettingsByPartId = new HashMap<>();
 
-    @Attribute(required = false)
-    private boolean allowIncompatiblePackages;
-
-    private Set<org.openpnp.model.Package> compatiblePackages = new HashSet<>();
-
-    @ElementList(required = false, entry = "id")
-    private Set<String> compatiblePackageIds = new HashSet<>();
 
     @Override
     public PartAlignmentOffset findOffsets(Part part, BoardLocation boardLocation, Location placementLocation, Nozzle nozzle) throws Exception {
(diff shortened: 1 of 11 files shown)


src/main/java/org/openpnp/machine/reference/vision/wizards/ReferenceBottomVisionConfigurationWizard.java
```diff
@@ -110,6 +110,11 @@ private void editPipeline() throws Exception {
     }
 
     @Override
+    public String getWizardName() {
```
vonnieda: If this is shown in the tab it should be "ReferenceBottomVision". If another "BottomVision" type class is added we want to be able to tell them apart. Same with the one below in the wizard.

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -114,6 +119,14 @@ public PartAlignmentOffset findOffsets(Part part, BoardLocation boardLocation, L
         return new PartAlignmentOffset(offsets,false);
     }
 
+    @Override
+    public boolean canHandle(Part part) {
```
vonnieda: There may be a small issue here. canHandle is checking the part settings map directly, whereas findOffsets is calling getPartSettings(). getPartSettings creates a default part settings when called, whereas checking the map directly will fail if the part has not been previously configured. Is this what you intended, or would it be better to use getPartSettings()?

The main issue I see here is that this changes default behavior a bit. Previously, if the user has bottom vision enabled it will be enabled by default for all parts, and the user can disable it on a part by part basis, as needed. With your changes it will default to being disabled for all parts unless the user has previously configured the part settings for that part.

Thoughts?

vonnieda:
@tracklogic The change looks good, but this is failing Maven tests. It looks like the change is somehow turning bottom vision on for the test suite, which fails since the test suite does not include a bottom vision camera. Can you run `mvn test` and see what is going on?

vonnieda:
Hi @tracklogic, I merged this with very minor changes. I reverted the changes to machine.xml, as I didn't want the pipeline duplicated and I changed ReferenceMachine to add a ReferenceBottomVision to the partAligners if none exist. Let me know if these changes cause you any problems. I don't think they should, as it effectively replicates what you were doing, but in a slightly different way that duplicates less code.

PR #553, opened 2017-05-11
Added center on paste-dispenser button to the GUI.

tracklogic:
# Read This First

Be sure to review the [[Developers Guide]] and the [Contributing Guidelines](https://github.com/openpnp/openpnp/blob/develop/CONTRIBUTING.md).

# Description
Describe the change in detail. Explain how it works. The better you describe the change the more likely it is to be accepted.

Adds a button to the paste-dispenser tab to centre the paste-dispenser tool at the camera position.

# Justification

# Instructions for Use

# Implementation Details

### DIFF
--- a/src/main/java/org/openpnp/gui/JogControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/JogControlsPanel.java
@@ -374,6 +374,11 @@ public class JogControlsPanel extends JPanel {
         tabbedPane_1.addTab("Dispense", null, panelDispensers, null);
         FlowLayout flowLayout = (FlowLayout) panelDispensers.getLayout();
         flowLayout.setAlignment(FlowLayout.LEFT);
+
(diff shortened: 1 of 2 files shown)


vonnieda:
This functionality already exists. Just select the paste dispenser from the dropdown and click the position at camera button in the jog controls.

vonnieda:
Actually, that may be a lie. Let me double check. In any case, that's how this *will* work. We're not going to add a new button for it.

vonnieda:
Okay,  yea, I was mistaken. Never added dispensers to the dropdown. But a recent PR made it possible to do this, so instead of this change all that needs to happen is that paste dispensers get added to the dropdown for them to become a first class citizen. I'm making that change now since it's like a 10 liner and will commit it in a moment. Closing this PR.

vonnieda:
Okay, that's added now, so you can select a paste dispenser from the dropdown and use all the normal things like jog buttons, position at camera, position camera at tool, etc. Should do the same as you were looking for here without adding another button.

PR #557, opened 2017-05-12
Added contours to MaskModel, corrected parameter description.

dzach:
Added ability to filter or mask a `contours` model. Corrected parameter description to make it clearer when the stage is in masking or filtering mode.


vonnieda:
Nice addition, thank you @dzach!

PR #558, opened 2017-05-12
Added getResult("..") to CvPipeline

dzach:
Cascading of pipeline stages that operate on the `working image` is the default mode for the user of the opencv pipeline, and it is easy for the developer: method `getWorkingImage()` brings in the last image put in the pipeline by a previous stage.

The same is not true for _a model_ produced by a previous stage. To get such model, the user/developer has to explicitly ask for it by naming the specific stage or by using `getResult(stage)` in code.

This PR adds code that stores the result produced by the previous stage and makes it available through the method `getResult()`, the same method used before to get the result from a specific stage, but now without arguments.

This change will make it easier to cascade stages that work on models, in the same way that they work for images: a stage can be moved around without the need to change its reference to a previous model, by using the input of the stage that precedes.

### DIFF
--- a/src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
+++ b/src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
@@ -246,7 +246,11 @@ public class CvPipeline {
             results.put(stage, new Result(image, model, processingTimeNs));
             // if there is an error, then there is no point in storing the error condition, just skip it
             if (!err) {
-              previousResult = new Result(image, model, processingTimeNs);
+              Mat pmat = null;
+              if (image != null) {
+                pmat = image.clone();
+              }
+              previousResult = new Result(pmat, model, processingTimeNs);
             }
         }
     }
--- a/src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
+++ b/src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
@@ -134,8 +134,9 @@ public class CvPipeline {
     }
 
     /**
-     * Get the Result returned by the CvStage with the given name. May return null if the stage did
-     * not return a result.
+     * Get the Result returned by the CvStage with the given name. 
+     * If name is '..', then return the previous result. 
+     * May return null if the stage did not return a result.
      * 
      * @param name
      * @return
@@ -143,6 +144,8 @@ public class CvPipeline {
     public Result getResult(String name) {
         if (name == null) {
             return null;
+        } else if (name.matches("..")) {
+          return previousResult;
         }
         return getResult(getStage(name));
     }
@@ -159,19 +162,6 @@ public class CvPipeline {
             return null;
         }
         return results.get(stage);
-    }
-
-    /**
-     * Get the Result returned by the previous CvStage. May return null if the stage did not return a
-     * result.
-     * 
-     * @return
-     */
-    public Result getResult() {
-        if (previousResult == null) {
-            return null;
-        }
-        return previousResult;
     }
 
     /**
--- a/src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
+++ b/src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
@@ -203,6 +203,7 @@ public class CvPipeline {
             // Process and time the stage and get the result.
             long processingTimeNs = System.nanoTime();
             Result result = null;
+            boolean err = false;
             try {
                 if (!stage.isEnabled()) {
                     throw new Exception("Stage not enabled.");
@@ -211,6 +212,7 @@ public class CvPipeline {
             }
             catch (Exception e) {
                 result = new Result(null, e);
+                err = true;
             }
             processingTimeNs = System.nanoTime() - processingTimeNs;
             totalProcessingTimeNs += processingTimeNs;
@@ -242,7 +244,12 @@ public class CvPipeline {
             }
 
             results.put(stage, new Result(image, model, processingTimeNs));
-            previousResult = new Result(image, model, processingTimeNs);
+
+            if (model == null && previousResult != null && previousResult.model != null && !err) {
+              previousResult = new Result(image, previousResult.model, processingTimeNs);
+            } else {
+              previousResult = new Result(image, model, processingTimeNs);
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


vonnieda:
@dzach What are your plans for how to use this? Do you want to start changing stages that currently get models by name to use the previous model? Or is this for new stages? Or other?

cri-s:
I think insert the name is fine, otherwise the risk is that stage get coded where no input of stage is possible
because on 90% the case the previous stage is used, but for the 10% remain cases that are more important
the trouble begins.

dzach:
I may change stages that get models by name by allowing for a name e.g. `..` or similar to mean `previous stage`. All stages requiring a name will work exactly the same, and will also be able to use `..` to mean `previous stage`. The users will see no change in their pipelines.

dzach:
No, default operation will remain for the image. Getting the previous model will require a decision by the user and will not be automatic.

vonnieda:
How about if we just take no stage name to mean previous? Using ".." makes sense to me as a programmer, but won't to a regular user. But I think if we just just leave the name blank and that means grab the previous that would be easy to understand. I will add that I think we need to make sure that stages continue to use a name to get the model data rather than new stages assuming the data is in the previous. 

If we do this, I think we can actually just modify getResult(String name) to return the previous stage's model if name is null. What do you think?

dzach:
E.g the ImageWrite stage, will get a `modelStageName`, empty by default. If something like `..` is entered there, then the model in the pipeline, if any, will be used as ROI for the write operation. This will allow to move this stage around and write ROIs as they are produced by each stage, without needing to change the stage name every time.

ImageWrite is not currently capable of doing this, but this is needed for getting template images fast and automatically out of pipeline stages, if part orientation with templates is to be pursued.

vonnieda:
That's not a good example. Adding this to ImageWrite would be too limiting. We'd just add a Crop or ROI stage and then you can use it anywhere.

dzach:
> If we do this, I think we can actually just modify getResult(String name) to return the previous stage's model if name is null. What do you think?

Yea, that may be easier. But I am afraid that having no name as a default for get previous result may lead to conflicts and confusion. I think empty name should mean exactly that: `no model should be used`. We can think of a better name to mean 'previous stage model', if `..` is not a good idea.

cri-s:
Why do you not simple crop the image  or pass the resulting roi image as image to the next stage.
Whitout rotating image the result is not usable as example for getting the shape from result or using
it as reticle.

cri-s:
I want repeat the danger. Allowing using model from previous stage fool developer to use that without
giving user the possibility to configure the name of model and this create then inflexibility.

dzach:
Whatever. ImageWrite is almost an one-liner anyway. But been able to move a stage around and not change its model input makes things a lot easier, I think.

vonnieda:
@cri-s We're definitely not going to support stages that *only* take the previous model. As I said, we'd stick with having a model name and we're working on deciding how that can be used to specify either a name or "the previous result".

dzach:
@cri-s "allowing" doesn't do this, "forcing" may do it. Allowing it is supposed to give flexibility, as it is something that is not possible today for the user.

vonnieda:
@dzach I agree, I think it's a valuable idea. I also agree that there is some risk in defaulting to the previous if left null. I can't really think of a place where this would cause a major problem, but it could cause some confusion. With that in mind, I'd be fine with changing getResult(String name) to accept ".." as "previous" if you'd like to make that change. I can't really think of any better token. I considered "previous", "-1", "<", etc. but they all have their problems.

dzach:
Will do

dzach:
@cri-s `Rotate` is also a candidate for the same think. All these need to be able to take model input and produce a result for part of the image, and possibly "in place", if that is required or useful.

dzach:
Done.

dzach:
There is an issue when the previous stage is not enabled. This seems to be more involved, so please hold the merge until it's fixed.

dzach:
Ok. This is now fixed.

dzach:
@cri-s are you referring to the latest fix 3bc76ed or the previous commit? The fix is supposed to do exactly that, avoid propagating the error.

cri-s:
The latest fix.
Example:
Detecting contours, filtering contours that are less then 26 pixel in size, draw that contours filled with black.
This for cleanup the noise. Now finding the part, part is missing from nozzle, black center was painted to hide the nozzle.
Result with you modifications: the noise that are unwanted get  detected as valid part. Without modification the missing part get noticed and retry performed.

dzach:
Ok, the logic was bad, the code was bad, it didn't actually catch the exception.
This is fixed. Please, test and comment, I much appreciate it.

I really like this capability. I already have a number of old and new stages that make use of it and is a breeze of user friendliness to move stages around and immediately see the results.

dzach:
@cri-s if you hit a problem, could you please provide a pipeline that demonstrates it? This is a really simple change, it shouldn't affect any old code or pipeline. If it does, then there is still an error in the logic that needs to be corrected.

cri-s:
The error still exist, but i'm out from discussion here, merge it, and
probalby if you use it with
you'r new stage that is a alias of some pipeline commands apparently
the error for you don't exist or is not relevant. For ther other,
there are free to not use this  feature.

2017-05-13 15:52 GMT+02:00, Dimitrios <notifications@github.com>:

dzach:
Turns out that the bug was not fixed. I've also got some unexpected results upstream in the pipeline that probably indicate that result.image has to be cloned for storage. I did just that.

@cri-s you were probably right, but your comments, now removed (!), are so cryptic that do not practically help in resolving the problems.

So, as a true `infinite monkey` Java coder, I've committed the next fix.

Jokes aside, please give this some thought, I am at my limits with Java, and need some help to get this through. If of course this feature is of any value, which I think it is.

dzach:
See  #560 for an example of use.

dzach:
I've encountered more problems with this. Closing this PR until it is ready.

cri-s:
Hi @dzac, do you have a image of sot23 where the rotatedRect is wrong ?
I would need this case for the reference image, as i cannot use
pipeline and a pic would help
for the test case. Unfortunatly i don't find it now such pic.

2017-05-18 13:18 GMT+02:00, Dimitrios <notifications@github.com>:

PR #563, opened 2017-05-15
Rotary Tray Feeder now with maven test clean ups :)

sp-apertus:


vonnieda:
Looks great, thanks @sp-apertus!

PR #566, opened 2017-05-20
added nozzle variable to pipeline

cri-s:
# Description
Added Nozzle variable to pipeline including getter/setter 

# Justification
This allows accessing the nozzle, from scripts.
This is useful for stages that need to stitch images in order to get a wider field of view,
access to all classes behind nozzle, as part, nozzleTip, and differentiate to bottom or top vision.
If nozzle.getPart() is not null, the pipeline is uplooking because nozzle have part loaded.
Otherwise it is downlooking for the normal usage.
Special pipelines like NozzleTip Calibration pipeline are exceptions to this rules and as special
pipeline that rule don't need to be respected.

# Instructions for Use
moveTo(pipeline.getNozzle().getLocation().add(new Location(LengthUnit.Millimeters,10,10,0,0)):
As note, inside StandaloneEditor, getNozzle return always null.
For Wizards that don't have access to the nozzle parameter, selected Nozzle from gui selector is used.

# Implementation Details
added variable to pipeline and updated code that use pipeline to include nozzle settings.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTip.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTip.java
@@ -193,7 +193,7 @@ public class ReferenceNozzleTip extends AbstractNozzleTip {
     private Nozzle getParentNozzle() {
         for (Head head : Configuration.get().getMachine().getHeads()) {
             for (Nozzle nozzle : head.getNozzles()) {
-                for (NozzleTip nozzleTip : nozzle.getNozzleTips()) {
+                for (NozzleTip nozzleTip : nozzle.Tips()) {
                     if (nozzleTip == this) {
                         return nozzle;
                     }
@@ -473,6 +473,7 @@ public class ReferenceNozzleTip extends AbstractNozzleTip {
 
         public CvPipeline getPipeline() throws Exception {
             pipeline.setCamera(VisionUtils.getBottomVisionCamera());
+			pipeline.setNozzle(getParentNozzle());
             return pipeline;
         }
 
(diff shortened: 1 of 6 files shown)


vonnieda:
Hi @cri-s, looks mostly good. I'm a little concerned about the getPickLocation hack for nozzle. Would it be better if we just added Nozzle to getPickLocation like getPickLocation(Nozzle)? That's an easy API change that might be worth making if you think it would be valuable.

cri-s:
You refer to
private getPickLocation(Camera camera) {}
It could be changed to getPickLocation(Camera,Nozzle) 
Hovewer on Wizard a button calls getPickLocation(Camera) in order to check if it finds the reference part.
It is ok that nozzle is null, that case must be handled because on StandalonePipelineEditor actually
nozzle is null. Do you see this problematic, i see this solution cleaner as calling Nozzle with zero as this is
unusual and usually not expected on first look over code.

vonnieda:
@cri-s Now that I look closer, I am not sure why you added the Nozzle nozzle; and this.nozzle = nozzle; getPickLocation(Camera) is called from feed() which already has nozzle, so we can just pass it, right? No need to store it?

vonnieda:
@cri-s Ignore my previous comment. I see you already considered this. I will look at this closer when I have a few minutes later today and see what to do. Should be close to merge.

cri-s:
It's a cooding preference.
Loosefeeder will have a refernce Mat and the wizard can overimpose it
on camera view.
Now  a button can run the pipeline and check for part in order to verify
that all is ok. This code don't have access to nozzle, but if you want
can call it
as getPickLocation(Configuration.get().getMachine.getDefaultHead().getDefaultCamera(),
null);
.


2017-05-20 23:07 GMT+02:00, Jason von Nieda <notifications@github.com>:

vonnieda:
I'm going to go ahead and merge this as is, and we can revisit the nozzle variable later if it turns out to be an issue.

vonnieda:
Thanks @cri-s!

vonnieda:
@cri-s Okay, I checked the idea I had and it seems okay. Do you see any problem with this? https://github.com/openpnp/openpnp/commit/7f2e6ddbea34c654119e35ab5652ecceef0f7f6e

cri-s:
no.
Further i think the function name is inappropiate as this is the name
of feeder spi api and it
is better named as example findPart(...) instead.

2017-05-21 1:32 GMT+02:00, Jason von Nieda <notifications@github.com>:

Issue #567, opened 2017-05-20
PR single merge issue


cri-s:
# Read This First
To submit an issue for OpenPnP you must use this template or it will be deleted. Please fill out either the Problem Report or Feature Request section below and delete the other section.

-----------------------------------------------------------------------

# Problem Report
## Describe the Issue
I want create a PR for accessing Part inside pipeline.
changes shoud be adding variable part to pipeline,
and on 
public Part getPart() {
  if(part==null) if(feeder!=null) part=feeder.getPart(); else if(nozzle!=null) part=nozzle.getFeeder();
 return part;
}

How to commit this, if  for this #565 and #566 is needed in order to access to this variables ?
AS i have experienced, committing  file where thing is missing and adding this missing later,
block travis checking.

## Expected Result
Response how to submit such patch having the reuqirement to reduce patch to such small and detailed
patch level.

## Notes, Log Files, Screen Captures, Videos, etc. to Show the Issue
The more information you can provide about the issue the more likely it is that a developer will be able to respond.

Whenever possible, please use the Menu -> Help -> Submit Diagnostics feature and include the URL here. Submit Diagnostics makes it easy to include log files, screen capture, debug images, configuration files, etc.

-----------------------------------------------------------------------

# Feature Request
## Describe the Feature
What do you want it to do? Be as descriptive as possible.

## How Will It Look?
If the feature will include changes to the user interface, how would they look? Include drawings or mockups if possible.

## Why Do We Need It?
1. What problem does it solve?
2. Is it useful for everyone, or does it only solve a problem for your machine?

## Provide Examples
Is this something that exists in other pick and place software? Provide examples and references if possible.

vonnieda:
These are merged now, so it should be okay to add in your next change. If you mean to add getPart() to Pipeline, though, I'd prefer if you do it the same as the others and set it where needed, rather than try to derive it from Nozzle or Feeder. setPart and getPart will be more flexible.

PR #570, opened 2017-05-22
changed CvPipeline.java

cri-s:
# Read This First

will not be accepted.

# Description
from memory, 
Changed S1 instead of 1  for stages default names in order stages can accept numbers or stages.
As example rotation now can accept fixed rotation value or stage name where it parses rotatedRect.
Changed Map<CvStage,Result> to <String,Result> that speedup the execution of stages.
Addes some Options and Variables for Feeder/Part/.... in order that stages can access this variables.
Changed the cross for blank image to blank with red border, because if giving a mask, the red cross
always is detected as valid component. 
Improved image separation from stages, it is now not possible that the next stage change the image from previous stage that is displayed on editor or used on other stages later.
It is unclear if hook for changing the previous working image is eventually needed, i think the change
was necessary.
Updated image releasing and fixed memory leaks for images not released.
Added setWorkingImage(Mat) for allowing setting WorkingImage from outside the stages.
Increased processing timing measurements precision.

# Justification
Update of Pipeline for LooseFeeder and Top Vision, all users benefits.

# Instructions for User
Stages names  should begin like variables , first char must be letter or underscore, 
reserved are identifier beginning with underscore and identifiers beginning with result
Now it is possible to have multiple result stages, 
Example:  result1 result2 result3 result
Now it is possible to loop from stage result_ to result stage.

# Implementation Details
Please wait for merging, this PR.
Not currently implemented but desiderable is the variable for the final placement location of part 
for allowing closed loop adjustment or checking the placement error.

### DIFF
--- a/src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
+++ b/src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
@@ -13,7 +13,10 @@ import org.opencv.core.CvType;
 import org.opencv.core.Mat;
 import org.opencv.core.Point;
 import org.opencv.core.Scalar;
+import org.opencv.core.Size;
+import org.opencv.imgproc.Imgproc;
 import org.openpnp.model.Configuration;
+import org.openpnp.model.Part;
 import org.openpnp.spi.Camera;
 import org.openpnp.spi.Nozzle;
 import org.openpnp.spi.Feeder;
@@ -21,6 +24,7 @@ import org.openpnp.vision.pipeline.CvStage.Result;
 import org.simpleframework.xml.ElementList;
 import org.simpleframework.xml.Root;
 import org.simpleframework.xml.Serializer;
+
 
 /**
  * A CvPipeline performs computer vision operations on a working image by processing in series a
@@ -51,16 +55,38 @@ public class CvPipeline {
     @ElementList
     private ArrayList<CvStage> stages = new ArrayList<>();
 
-    private Map<CvStage, Result> results = new HashMap<CvStage, Result>();
+    private Map<String, Result> results = new HashMap<String, Result>();
 
     private Mat workingImage;
 
     private Camera camera;
     private Nozzle nozzle;
     private Feeder feeder;
+    private Part	part;
     
     private long totalProcessingTimeNs;
-    
+
+	
+    /**************************************************************************
+     * Setter for internal flags
+     * Different Naming is used, because internal use only || checkstyle don't allow that
+     * Every flag have a prepending comment line that describes there use
+	 * Default value for flag must be false.
+     */
+	 
+	// indicate if camera is downlooking
+	private boolean downLooking;
+	// indicate if fiducial is searched
+	private boolean fiducial=true;
+	// if inside editor, different path is used, no optimisations
+	private boolean editor=true; // TODO need to be false, actually for legacy code.
+    public void setOptionEditor(boolean val) { 	editor=val; }
+	
+	// return single result, looping disabled
+	private boolean single;
+    public void setOptionSingle(boolean val) {  single=val; }
+    /**************************************************************************/
+
     public CvPipeline() {
         
     }
@@ -116,6 +142,7 @@ public class CvPipeline {
     }
 
     public void remove(CvStage stage) {
+		results.remove(stage.getName()); 
         stages.remove(stage);
     }
 
@@ -146,7 +173,7 @@ public class CvPipeline {
         if (name == null) {
             return null;
         }
-        return getResult(getStage(name));
+        return results.get(name);
     }
 
     /**
@@ -160,25 +187,37 @@ public class CvPipeline {
         if (stage == null) {
             return null;
         }
-        return results.get(stage);
+        return results.get(stage.getName());
     }
 
+	List <Mat> workingImages = new ArrayList<>();
     /**
      * Get the current working image. Primarily intended to be called from CvStage implementations.
      * 
      * @return
      */
     public Mat getWorkingImage() {
+		Mat img;
         if (workingImage == null || (workingImage.cols() == 0 && workingImage.rows() == 0)) {
-            workingImage = new Mat(480, 640, CvType.CV_8UC3, new Scalar(0, 0, 0));
-            Core.line(workingImage, new Point(0, 0), new Point(640, 480), new Scalar(0, 0, 255));
-            Core.line(workingImage, new Point(640, 0), new Point(0, 480), new Scalar(0, 0, 255));
+            workingImage = new Mat(480, 640, CvType.CV_8UC3, new Scalar(0, 0, 255));
+			Core.rectangle(workingImage, new Point(10,10), new Point(470,630), new Scalar(0), -1);
+			workingImages.add(workingImage);
         }
-        return workingImage;
+		img = workingImage.clone();
+		workingImages.add(img);
+        return img;
+    }
+
+    public void setWorkingImage(Mat img) { // img get not released !!!
+        workingImage = img;
     }
 
     public void setCamera(Camera camera) {
         this.camera = camera;
+		downLooking = camera.getHead()!=null;
+		if(!downLooking) {
+			fiducial=false;
+		}
     }
 
     public Camera getCamera() {
@@ -194,27 +233,47 @@ public class CvPipeline {
     }
   
     public void setFeeder(Feeder feeder) {
+		fiducial = false;
         this.feeder = feeder;
     }
 
     public Feeder getFeeder() {
         return feeder;
     }
+  
+    public void setPart(Part part) {
+        this.part = part;
+    }
+
+    public Part getPart() {
+        return part;
+    }
 
     public long getTotalProcessingTimeNs() {
       return totalProcessingTimeNs;
     }
 
-    public void setTotalProcessingTimeNs(long totalProcessingTimeNs) {
-      this.totalProcessingTimeNs = totalProcessingTimeNs;
-    }
-
+	
+	private boolean isNull(Object obj) {
+							if(obj!=null) {
+								if(obj instanceof List<?>) {
+									if(((List)obj).size()==0) {
+										return true;
+									}
+									return false;
+								}
+								return false;	
+							}
+							return true;	
+	}
+	
     public void process() {
-
-        totalProcessingTimeNs = 0;
+        long totalProcessingTimeNs = System.nanoTime();
         release();
-        for (CvStage stage : stages) {
-            // Process and time the stage and get the result.
+		int mark=-1,last=stages.size()-1,pos=0;
+		for(int limit=1000;pos>=0&&pos<=last;pos++,limit--) {
+			CvStage stage = stages.get(pos);
+		    // Process and time the stage and get the result.
             long processingTimeNs = System.nanoTime();
             Result result = null;
             try {
@@ -226,8 +285,6 @@ public class CvPipeline {
             catch (Exception e) {
                 result = new Result(null, e);
             }
-            processingTimeNs = System.nanoTime() - processingTimeNs;
-            totalProcessingTimeNs += processingTimeNs;
 
             Mat image = null;
             Object model = null;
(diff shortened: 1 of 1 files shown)


vonnieda:
Hi @cri-s, please don't submit the PR until it's ready for merge. This code contains all kinds of hardcoded functionality, and pipeline specific variables and functionality. When you are ready for it to be reviewed please resubmit it.

cri-s:
apart the "result_" function and the "resultN" aliases for result, what do you see problematic ?

vonnieda:
You said "Please wait for merging, this PR." If you don't want it merged yet, I don't want to review it. It is a waste of time to review code that isn't finished as I will just have to do it again when it's finished.

cri-s:
The code is ok, several stages need to be updated in order to work,
basically instead of return null  it must be  return new Result(mat)


2017-05-22 15:55 GMT+02:00, Jason von Nieda <notifications@github.com>:

src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
```diff
@@ -51,16 +55,38 @@
...
+	
+    /**************************************************************************
+     * Setter for internal flags
+     * Different Naming is used, because internal use only || checkstyle don't allow that
+     * Every flag have a prepending comment line that describes there use
+	 * Default value for flag must be false.
+     */
+	 
+	// indicate if camera is downlooking
+	private boolean downLooking;
```
vonnieda: Why? Camera.Looking already tells you this.

src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
```diff
@@ -51,16 +55,38 @@
...
+	// indicate if fiducial is searched
+	private boolean fiducial=true;
```
vonnieda: Why? This is specific to one single algorithm and does not belong in CvPipeline.

src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
```diff
@@ -51,16 +55,38 @@
...
+	// if inside editor, different path is used, no optimisations
+	private boolean editor=true; // TODO need to be false, actually for legacy code.
```
vonnieda: This does not appear to be set anywhere.

src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
```diff
@@ -51,16 +55,38 @@
...
+    public void setOptionEditor(boolean val) { 	editor=val; }
+	
+	// return single result, looping disabled
+	private boolean single;
+    public void setOptionSingle(boolean val) {  single=val; }
```
vonnieda: This does not appear to be set anywhere. If it is intended for future functionality, submit it when you submit that functionality.

src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
```diff
@@ -51,16 +55,38 @@
     @ElementList
     private ArrayList<CvStage> stages = new ArrayList<>();
 
-    private Map<CvStage, Result> results = new HashMap<CvStage, Result>();
+    private Map<String, Result> results = new HashMap<String, Result>();
```
vonnieda: Why? How does this improve performance?

src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
```diff
@@ -194,27 +233,47 @@ public Nozzle getNozzle() {
...
+	
     public void process() {
-
-        totalProcessingTimeNs = 0;
+        long totalProcessingTimeNs = System.nanoTime();
         release();
-        for (CvStage stage : stages) {
-            // Process and time the stage and get the result.
+		int mark=-1,last=stages.size()-1,pos=0;
+		for(int limit=1000;pos>=0&&pos<=last;pos++,limit--) {
```
vonnieda: What is all of this supposed to do? Why does it need to be changed? Why is there a hardcoded value of 1000?

src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
```diff
@@ -238,25 +295,43 @@ public void process() {
...
-                image = image.clone();
+				if(!workingImages.contains(image)) {
+					workingImages.add(image);
+				}
             }
-
-            results.put(stage, new Result(image, model, processingTimeNs));
+			String id = stage.getName();
+			result= new Result(workingImage, model, processingTimeNs = System.nanoTime() - processingTimeNs);
+			if(id.startsWith("result")&&id.length()==7) {
```
vonnieda: This is sloppy. Hardcoding a format for stage names and having special functionality does not improve the system, it makes it more fragile and increases the code complexity. If you want to be able to handle multiple result stages it needs to be thought through better, and should probably just be handled in the calling code.

src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
```diff
@@ -308,7 +380,7 @@ public void fromXmlString(String s) throws Exception {
 
     private String generateUniqueName() {
         for (int i = 0;; i++) {
-            String name = "" + i;
+            String name = "S" + i;
```
vonnieda: Why?

cri-s:
2017-05-22 16:14 GMT+02:00, Jason von Nieda <notifications@github.com>:
> vonnieda commented on this pull request.
>
> I have highlighted the issues in this code. I'm not going to merge it. It
> adds all kinds of complicated logic to a very simple system and you didn't
> explain the value of any of it. I am extremely frustrated by constantly
> having to do this over and over. If you aren't willing to follow the
> guidelines in the template please just don't submit PRs.
>
This is the big change, if i broke every change down with explaining
the bug and why,
i need a week only for this and doing it sequencially as you could not
give answer to
#567 i must wait befor you approve any change,and then doing the next
that week then become several and during that time the build is
broken.
For this i have prefered to doing this change at once and discuss
the/every issue with you.
It don't make sense if i change the other part and you prefer doing it
differently or don't merge it. Surely this code need some polish and
small changes, overall functionality is here.
You could say, you prefer two pipelines, one that check the possible
candidates, and the
other that verify it. This remove the need for the loop. and some
control variables like
single or editor. Instead of one single file then there are two files
for the pipeline.
It has advantage and disadvantages that code don't need to be that
complicated, the disadvantage is for the user that it have to manage
and syncronize and debug two pipelines.
Having the possibility of more then one return values i think is necessary.
On my code i have used a break() call from stage. This have the
disadvantage that it is not
debuggable visually and not suitable for users without experience with
low level vision.
Because this i have choosen to have 10 stages that are full visible
inside editor instead of the
clean api.

>
>> @@ -308,7 +380,7 @@ public void fromXmlString(String s) throws Exception
>> {
>
>      private String generateUniqueName() {
>          for (int i = 0;; i++) {
> -            String name = "" + i;
> +            String name = "S" + i;
>
> Why?
>
As described in description with the example given for rotate, it is a
enhanchment.
This allows on stages to have a field that accept either numeric or
stiring input.
If input begins with letter, it is a stage name, otheriwse numeric value.
Example on rotate, it could rotate a fixed arbitrary value, or it
could rotate based on rotatedRectangle from that stage.
If default stages name begin with numbers, this is not possible,
Further this better adress ScriptRun  where there is one string argument.
If beginning default stage name with S0 for Stage nr 0 allows to check
scripts if argument
is stage or one or multiple values.


>>              }
> -
> -            results.put(stage, new Result(image, model,
> processingTimeNs));
> +			String id = stage.getName();
> +			result= new Result(workingImage, model, processingTimeNs =
> System.nanoTime() - processingTimeNs);
> +			if(id.startsWith("result")&&id.length()==7) {
>
> This is sloppy. Hardcoding a format for stage names and having special
> functionality does not improve the system, it makes it more fragile and
> increases the code complexity. If you want to be able to handle multiple
> result stages it needs to be thought through better, and should probably
> just be handled in the calling code.
Actually the system use hardcoded name result and results .
Using that convention, i have added the result[0-9] and result_
result[0-9] are alternative result and result_ loop trought result_
and result[0-9]?
The api interface is still getResult(), there is no special handling
outside this with
exception to the result_ stage and results where probably a api call
get inserted
for getResults().
The editor and single flag specifiy if the stage can be exited previously.
On editor, one needs to know the results and timing of all stages.
If only a single component is searched, there is no need to do feature
extraction on all
possible candidates, after the first component is found, the pipeline
is terminated.

As this hovewer change possible evaluation of legacy code and is a
breaking feature,
actually there is no code that use this and the editor flag is set to
true in order to not
break any code. Code that retain functionality, but sequencing must be changed.
As result1 as example terminate the pipeline processing, any code for
user display need to
be before result stage, because no more getWorkingImage() is used for
display overlay image but the resulting image from result1 stage in
this case.
In order to easily migrate that code a new stage that can copy model
data from other stage
is needed. Hovewer it can be decided that the abort action should only
trigger for stages
named result[0-9] and not for result stage in order to silently still
support legacy code but
introducing special cases and code handling.

Hypotetical example of code:
  [color] capture
  blur
  convertColor
  normalize
  [gray] blur
  crop
  templatematch
  recall image color
  Draw template center and bounding box
  [result1] If found, return result using workingImage and
templatematch data converted to rotatedRect
  recall image crop
  invert image
  templatematch
  recall image color
  Draw template center and bounding box
  [result2] If found, return result using workingImage and
templatematch data converted to rotatedRect
 recall image gray1
 [crop1] crop image with bigger FOV as crop
  templatematch
  recall image color
  Draw template center and bounding box
  [result3] If found, return result using workingImage and
templatematch data converted to rotatedRect
  recall image crop1
  invert image
  templatematch
  recall image color
  Draw template center and bounding box
  [result4] If found, return result using workingImage and
templatematch data converted to  rotatedRect
 recall image gray1
 try hough circle ... with and without canny
>
>>          release();
> -        for (CvStage stage : stages) {
> -            // Process and time the stage and get the result.
> +		int mark=-1,last=stages.size()-1,pos=0;
> +		for(int limit=1000;pos>=0&&pos<=last;pos++,limit--) {
>
> What is all of this supposed to do? Why does it need to be changed? Why is
> there a hardcoded value of 1000?
mark is the mark pointer for loops, last is the last stage number, pos
is the actual stage nr.
The pipeline need to search for possible targets, identify that, sort
/ filter that target.
This rotatedRect stage is the result_ stage that have as source a list
of rotatedRects and
gives out one rotatedRect removing it (delayed) from the source list.
The rest of the pipeline with several sequencial branches (first hit
wins) validate that possible
candidate. If there is a new getResult() the result get added to
results and rotatedRect removed, otherwise the rotatedRect get
relisted at the end of the queue for to be evaluated again after all
others are evaluated and maybe multiple possibility are removed.

Because the pipeline have now the ability to loop and jump (pos
variable is the actual stage)
This is a resource limit that after approx 1000 stages it terminates
when looping,
this are 40 to 100 components depending on pipeline or maybe 10 if
false positive are
present that later can be removed. .It is a failsafe in case one
construct a pipeline that
loops forever. It don't hit immediatly, only after reaching result
stage or processing the
last stage inhibiting the possibility to loop again.




>>
>      private long totalProcessingTimeNs;
> -
> +
> +	
> +
> /**************************************************************************
> +     * Setter for internal flags
> +     * Different Naming is used, because internal use only || checkstyle
> don't allow that
> +     * Every flag have a prepending comment line that describes there use
> +	 * Default value for flag must be false.
> +     */
> +	
> +	// indicate if camera is downlooking
> +	private boolean downLooking;
>
> Why? Camera.Looking already tells you this.
Camera looking can be set by user,
this is intended to make it clear what it is it and because looking
direction get extracted
from camera.getHead() because that identifies looking correctly. If it
is a head camera it
looks down, if it is a camera without mounted on head it don't do
fiducial or top vision.
The intended usage is on Wizard and on templatematch state


>
>>
>      private long totalProcessingTimeNs;
> -
> +
> +	
> +
> /**************************************************************************
> +     * Setter for internal flags
> +     * Different Naming is used, because internal use only || checkstyle
> don't allow that
> +     * Every flag have a prepending comment line that describes there use
> +	 * Default value for flag must be false.
> +     */
> +	
> +	// indicate if camera is downlooking
> +	private boolean downLooking;
> +	// indicate if fiducial is searched
> +	private boolean fiducial=true;
>
> Why? This is specific to one single algorithm and does not belong in
> CvPipeline.
It is for internal use (Wizard and statistical data).

>
>> -
> +
> +	
> +
> /**************************************************************************
> +     * Setter for internal flags
> +     * Different Naming is used, because internal use only || checkstyle
> don't allow that
> +     * Every flag have a prepending comment line that describes there use
> +	 * Default value for flag must be false.
> +     */
> +	
> +	// indicate if camera is downlooking
> +	private boolean downLooking;
> +	// indicate if fiducial is searched
> +	private boolean fiducial=true;
> +	// if inside editor, different path is used, no optimisations
> +	private boolean editor=true; // TODO need to be false, actually for legacy
> code.
>
> This does not appear to be set anywhere.
>
>> +     * Different Naming is used, because internal use only || checkstyle
>> don't allow that
> +     * Every flag have a prepending comment line that describes there use
> +	 * Default value for flag must be false.
> +     */
> +	
> +	// indicate if camera is downlooking
> +	private boolean downLooking;
> +	// indicate if fiducial is searched
> +	private boolean fiducial=true;
> +	// if inside editor, different path is used, no optimisations
> +	private boolean editor=true; // TODO need to be false, actually for legacy
> code.
> +    public void setOptionEditor(boolean val) { 	editor=val; }
> +	
> +	// return single result, looping disabled
> +	private boolean single;
> +    public void setOptionSingle(boolean val) {  single=val; }
>
> This does not appear to be set anywhere. If it is intended for future
> functionality, submit it when you submit that functionality.
It is used when checked for result_ , it don't loop but find only one
single component.

>
>> @@ -51,16 +55,38 @@
>      @ElementList
>      private ArrayList<CvStage> stages = new ArrayList<>();
>
> -    private Map<CvStage, Result> results = new HashMap<CvStage, Result>();
> +    private Map<String, Result> results = new HashMap<String, Result>();
>
> Why? How does this improve performance?
because results are retrived frequently inside pipeline and using
Strings as key.
looking up results using string mean looping trough all stages and
comparing all names
until the searched name is found and only than doing Hash table lookup.
I

PR #572, opened 2017-05-26
Add parameter `normalize` to the `MatchTemplate` CvPipeline stage.

dzach:
# Description
This PR adds one parameter `normalize` defaulting to `true` to the `MatchTemplate` CvPipeline stage. This allows getting output results that are not normalized to maxVal, which is necessary for comparing results from multiple MatchTemplate stages in the same pipeline. Parameter descriptions are also added.

# Justification
This change is necessary in order to be able to compare results from multiple `MatchTemplate` stages. Multiple `MatchTemplate` stages are used in an upcoming solution to properly detect part orientation in a loose part feeder.

The newly added parameter `normalize` defaults to `true`, thus its addition does not affect existing pipelines. When unchecked, it prevents normalization of the winning match to 1.0, thus allowing similarly non-normalized results from other `MatchTemplate` stages to be compared and the best match to be selected. Otherwise all such stages output `1.0` and comparisons may   give false results.

# Instructions for Use
Just uncheck the new parameter `normalize` to have the stage output non-normalized results.

# Implementation Details
1. Multiple modified stages were tested in practice with an upcoming pipeline solution for detecting orientation of parts in the loose part feeder. A simple comparison of the non-normalized results can now give the best match between them.
2. Coding style was followed and automatically checked.
3. No changes to the `org.openpnp.spi` or `org.openpnp.model` packages were made
4. `mvn test` was run successfully.


dzach:
Most of the commits that appear in the list here are remnants of a previous, now withdrawn PR, and are irrelevant to the present PR. Only one file was actually changed, and that is 'MatchTemplate`

vonnieda:
Nice work @dzach, thank you for taking the extra time to add attribute documentation!

dzach:
Thank you @vonnieda !

PR #574, opened 2017-05-28
Added currentStage, previousStage to CvPipeline. Added special stage …

dzach:
…name `..` to mean `previousStage` for convenience.

# Description
This PR is a clean resubmission in place of #558 .

This PR adds the `previousStage` and `currentStage` properties to the CvPipeline class. The properties are read-only and return existing pipeline stages. No extra storage is required other than for the properties themselves and their getter methods.

This PR also adds the convenience stage name `..` to mean `previousStage` and return the contents of the `previousStage` property.

# Justification

Cascading of pipeline stages that operate on the working image is the default mode for the user of the opencv pipeline, and it is easy for the developer: method `getWorkingImage()` brings in the last image put in the pipeline by a previous stage.

The same is not true for a model produced by a previous stage. There is no method such as `getWorkingModel()` available to developers of pipeline stages, nor is there a way for a pipeline user to get the model output by the last stage, other than naming that stage explicitly. The pipeline is image-centric, but significant work is done too on models created by processing these images. These additions aim to make working with models similarly easy.

For pipeline users a convenience stage name, `..`,  is added as an alias to  `previousStage`. This makes it possible for users to cascade stages operating on models without the need to name these stages explicitly.

# Instructions for Use
The `previousStage` and `currentStage` are intended to be used by developers. `currentStage` can be used to get the name of a stage currently being processed, and will also help getting a stage's relative position in a pipeline, without resorting to looping through all stages of the pipeline and comparing. `previousStage` can help getting the results from the previous stage, both image and model, similarly without looping through all the pipeline stages.

Users will be able to use the name `..` for a stage to mean _previous stage_, which will allow them to cascade stages dealing with models.

# Implementation Details
1. The changes have been in continuous use for about 10 days in developing pipeline stages. No problems have been observed from such use. It should be noticed that contrary to the ill fated #558 PR, the pressent changes do not create or use new storage for images or models; the new class properties merely point to existing stage objects.
2. The coding style required by this project has been followed, and automatically verified.
3. No changes to `org.openpnp.spi` or `org.openpnp.model` packages were made.
4. The `mvn test` was run successfully before submitting the Pull Request.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -129,7 +129,7 @@ public class ReferenceBottomVision implements PartAlignment {
                 1500);
 
 
-        return new PartAlignmentOffset(offsets.derive(0d, 0d, 0d, preRotateAngle), preRotate);
+        return new PartAlignmentOffset(offsets.derive(null, null, null, offsets.getRotation() + preRotateAngle),false);
     }
 
     @Override
--- a/src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
+++ b/src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
@@ -56,9 +56,6 @@ public class CvPipeline {
     private Mat workingImage;
     private Object workingModel;
     
-    private CvStage currentStage;
-    private CvStage previousStage;
-
     private Camera camera;
     private Nozzle nozzle;
     private Feeder feeder;
@@ -150,9 +147,6 @@ public class CvPipeline {
         if (name == null) {
             return null;
         }
-        if (name.trim().equals("..")) {
-          return getResult(getPreviousStage());
-        }
         return getResult(getStage(name));
     }
 
@@ -188,14 +182,6 @@ public class CvPipeline {
       return workingModel;
     }
 
-    public CvStage getCurrentStage() {
-      return this.currentStage;
-    }
-    
-    public CvStage getPreviousStage() {
-      return this.previousStage;
-    }
-    
     public void setCamera(Camera camera) {
         this.camera = camera;
     }
@@ -235,7 +221,6 @@ public class CvPipeline {
         for (CvStage stage : stages) {
             // Process and time the stage and get the result.
             long processingTimeNs = System.nanoTime();
-            currentStage = stage;
             Result result = null;
             try {
                 if (!stage.isEnabled()) {
@@ -248,7 +233,6 @@ public class CvPipeline {
             }
             processingTimeNs = System.nanoTime() - processingTimeNs;
             totalProcessingTimeNs += processingTimeNs;
-            previousStage = stage;
 
             Mat image = null;
             Object model = null;
--- a/src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
+++ b/src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
@@ -54,6 +54,7 @@ public class CvPipeline {
     private Map<CvStage, Result> results = new HashMap<CvStage, Result>();
 
     private Mat workingImage;
+    private Object workingModel;
     
     private CvStage currentStage;
     private CvStage previousStage;
@@ -183,6 +184,10 @@ public class CvPipeline {
         return workingImage;
     }
 
+    public Object getWorkingModel() {
+      return workingModel;
+    }
+
     public CvStage getCurrentStage() {
       return this.currentStage;
     }
@@ -251,7 +256,9 @@ public class CvPipeline {
                 image = result.image;
                 model = result.model;
             }
-
+            if(stage.isEnabled() && model != null) {
+              workingModel=model;
+            }
             // If the result image is null and there is a working image, replace the result image
             // replace the result image with a clone of the working image.
             if (image == null) {
@@ -289,6 +296,7 @@ public class CvPipeline {
                 result.image.release();
             }
         }
+        workingModel=null;
         results.clear();
     }
 
--- a/src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
+++ b/src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
@@ -54,6 +54,9 @@ public class CvPipeline {
     private Map<CvStage, Result> results = new HashMap<CvStage, Result>();
 
     private Mat workingImage;
+    
+    private CvStage currentStage;
+    private CvStage previousStage;
 
     private Camera camera;
     private Nozzle nozzle;
@@ -146,6 +149,9 @@ public class CvPipeline {
         if (name == null) {
             return null;
         }
+        if (name.trim().equals("..")) {
+          return getResult(getPreviousStage());
+        }
         return getResult(getStage(name));
     }
 
@@ -177,6 +183,14 @@ public class CvPipeline {
         return workingImage;
     }
 
+    public CvStage getCurrentStage() {
+      return this.currentStage;
+    }
+    
+    public CvStage getPreviousStage() {
+      return this.previousStage;
+    }
+    
     public void setCamera(Camera camera) {
         this.camera = camera;
     }
@@ -216,6 +230,7 @@ public class CvPipeline {
         for (CvStage stage : stages) {
             // Process and time the stage and get the result.
             long processingTimeNs = System.nanoTime();
+            currentStage = stage;
             Result result = null;
             try {
                 if (!stage.isEnabled()) {
@@ -228,6 +243,7 @@ public class CvPipeline {
             }
             processingTimeNs = System.nanoTime() - processingTimeNs;
             totalProcessingTimeNs += processingTimeNs;
+            previousStage = stage;
 
             Mat image = null;
             Object model = null;


cri-s:
current stage is simple this from current stage.
It probably don't work if there is not enabled stage prevoiusly.

dzach:
That seems not to be the case with ScriptRun stages:
Run the script in a pipeline:
```
print(pipeline.getCurrentStage());
print(pipeline.getCurrentStage().getName());
print(this);
print(this.getName());
```
Results:
```
> org.openpnp.vision.pipeline.stages.ScriptRun@26b6fc5d
> thisStageName
> 'this' reference (XThis) to Bsh object: NameSpace: javax_script_context (bsh.ExternalNameSpace@216d86b2)
> javax.script.ScriptException: Sourced file: eval stream : Method getName() not found in bsh scripted object: javax_script_context : at Line: 4 : in file: eval stream : this .getName ( ) 
 in eval stream at line number 4
```

dzach:
> It probably don't work if there is not enabled stage prevoiusly.

This script:
```
print(pipeline.getPreviousStage());
print(pipeline.getPreviousStage().getName());
```
with the previous pipeline disabled, 

screenshot-12

seems to work:
```
org.openpnp.vision.pipeline.stages.ImageRead@5a59ca5e
previousStageName
```

cri-s:
Inside ScriptRun, there is a variable stage that is the actual stage,
because this is the interpreter, not the stage in this case, RunScript
is a special case.
Test it, one stage that gives result (model).
next stage is disabled.
next stage (whatever stage) access last stage using ".."
My understanding or fast view is, that ".." get the exception from
previous disabled stage and not the last real result.



2017-05-28 21:08 GMT+02:00, Dimitrios <notifications@github.com>:

dzach:
'previousStage` is expected to return whatever explicitly naming the previous stage would return. It doesn't do anything more than that.
I think they both return the exception.

dzach:
@cri-s if you know a way to do it differently I'll be happy to remove this PR and adopt whatever is better. It's just not helpful to have the image propagate through the pipeline and not the model. Getting the exception instead of a result from upstream the pipeline is not so bad for me, since I can just slide the stage to the place I need it and it'll work right away. Changing stage names every time a stage is moved is a PITA.

dzach:
> Inside ScriptRun, there is a variable stage that is the actual stage, because this is the interpreter, not the stage in this case, RunScript is a special case. 

This is good to know. Then the `currentStage` can be omitted, although it doesn't hurt. I'd prefer having a common method to access a stage than handling it as a special case though.

cri-s:
Insert only the laststage if it is enabled.
If (stage.isEnabled()) lastStage = stage; // at example, i don't
remember exactly how do you
insert it.

previousStage need return only stages that are enabled, otherwise it
is not usable inside
pipeline as every not enabled stage break the pipeline processing flow.

Further this type of coding is extremly dangerous, as you shoud implement only
    public Object getWorkingModel()
if you really need it and not getPreviousStage() as this open the door
to a lot of bugs.


2017-05-28 23:01 GMT+02:00, Dimitrios <notifications@github.com>:

dzach:
@cri-s it's hard to decipher your comments. They seem not to be pertinent to this PR but more general, about the pipeline. I can't do something about that. For this PR though:
* Access to the stage objects already exists, it is not something introduced in this PR. `pipeline.getStage()` does that already.
* There is no "insertion" into the pipeline anywhere. Just two getters and one alias.
* In your snippet code, do you mean not to return the stage if it is disabled? If so, there is always `stage.isEnabled()` that a stage developer can use to check it. Why should we do it for him?; We are not writing stage code here.
* Can you elaborate on why it is "extremely dangerous" to use `..` as a stage name, or `previousStage`, and why this is only when someone writes `..` as a stage name and not when this same 'someone' writes `10` instead? 

I fail to see why I must loop through all the stages of the pipeline to get to the one I'm after, instead of getting it straight away.

cri-s:
The response was in relation to this PR, i copy the justification and use you have written.
To the last point you mentioned, i never said that in this context.
From you'r usage instructions:
>  previousStage can help getting the results from the previous stage, both image and model, similarly without looping through all the pipeline stages.
Exactly that is dangerous and instead just getWorkginkModel() should be implemented as when developer get images from getResult() it must double check or follow some rules in order to not
create memory leaks.. getWorkingImage() is totally safe , getResult() for accessing image not.

I'm not comment further, good luck with you'r code.

> Justification
> 
> Cascading of pipeline stages that operate on the working image is the default mode for the user of the opencv pipeline, and it is easy for the developer: method getWorkingImage() brings in the last image put in the pipeline by a previous stage.
> 
> The same is not true for a model produced by a previous stage. There is no method such as getWorkingModel() available to developers of pipeline stages, nor is there a way for a pipeline user to get the model output by the last stage, other than naming that stage explicitly. The pipeline is image-centric, but significant work is done too on models created by processing these images. These additions aim to make working with models similarly easy.
> 
> For pipeline users a convenience stage name, .., is added as an alias to previousStage. This makes it possible for users to cascade stages operating on models without the need to name these stages explicitly.
> Instructions for Use
> 
> The previousStage and currentStage are intended to be used by developers. currentStage can be used to get the name of a stage currently being processed, and will also help getting a stage's relative position in a pipeline, without resorting to looping through all stages of the pipeline and comparing. previousStage can help getting the results from the previous stage, both image and model, similarly without looping through all the pipeline stages.
> 
> Users will be able to use the name .. for a stage to mean previous stage, which will allow them to cascade stages dealing with models.

dzach:
@cri-s What you are criticizing is the way the pipeline stands today and some imaginative future developers, not the current PR. The pipeline offers numerous ways for "bad programmers" to do "dangerous programming". You say `getResult()` is one of them and so is `previousStage`, but you haven't shown code on how to do this "safely". Do you have some code to share for the proper `getWorkginkModel()` you are proposing? Cause I tried something about that, and got stuck. 

I understand one being cautious, but being negative does not advance any issue.

cri-s:
// for getWorkingModel() you need this code change:
// below  private Mat workingImage;
  private Object workingModel;
// near getWorkingImage function:
  public Object getWorkingModel() { return workingModel; }
// inside release()
   workingModel=null;
// inside pipeline.process after this existing code:
            if (result != null) {
                image = result.image;
                model = result.model;
            }

           if(stage.isEnabled()&&model!=null) {workingModel=model;}


2017-05-29 14:52 GMT+02:00, Dimitrios <notifications@github.com>:

dzach:
Thank you @cri-s I'll put that code to use and see how it goes. Have you tested this code?

dzach:
I am using @cri-s 's `getWorkingModel()` in a couple of my new stages, and it works really nicely. It does fix the problem of getting the Exception instead of the model when a previous stage is disabled. A question that needs to be answered is if modifying old stages to propagate a model in their results, where they didn't before, can potentially break some existing user pipelines.

I have modified a `ConvertColor` stage to use like this in tests, and it feels very natural to have this functionality. The gif below shows rotatedRects in red color from stage 9, `MinAreaRectContours`, passing though to the last stage, where the display comes from, while a `MaskModel` stage's enable property is toggled on/off . Without the modification, `ConvertColor` blocked the model coming from upstream the pipeline : 

getworkingmodel

cri-s:
I don't understand.
Do you have modified convertColor stage? If yes why ?

2017-05-31 21:46 GMT+02:00, Dimitrios <notifications@github.com>:

dzach:
Yes, that was an early test to show that the `java.lang.Exception: Stage not enabled.` problem is fixed with your solution and that using `result.model` from a previous stage may not yield the desired results,  as you were saying. My wording above is not very clear indeed.

MaskModel, stage 11 takes its input from ConvertColor, stage 10, by naming that state explicitly, well, with a `..`, i.e. `getPreviousStage()` which is the same with an explicit name, and not by using `getWorkingModel()`. And the problem with the Exception is there, as you can see in the picture below. 

The unmodified `ConvertColor` is asked explicitly for its model, which is null, and this causes the null exception in the next stage. The modified `ConvertColor` passes the result.model from upstream to its result. `MaskModel` is not modified to use `getWorkingModel()` but reads `result.model` which is null and throws the exception.  `DrawModel` however uses `getWorkingModel()` from the pipeline, and shows both the model and the image that are in the pipeline, which model should have been masked by `MaskModel` if it could read the model, as it did in the previous animation.

Sounds confusing but it does prove you point. An animated picture may tell more:

getworkingmodel1

Compare this animation with the previous.

cri-s:
to respond to you'r question about modify stages for picking data without knowing what data is and
returning this data as result.
This is dangerous, as when you need stages that have image data as model, there is a need to release all
result.model inside release() and if you release data two times, because someone had the ill idea to pick random data
and store it inside result bad things or core dump occurs when already after ram is deallocated, it is free again for data with native pointers, as with Mat and other data from opencv.
Actually the amount of data that is not deallocated because of return.model is small compared to other
never reclaimed memory.

dzach:
@cri-s I see the reasons for your cautiousness about using result.model and the need for safe retrieval of results.

With your nice `getWorkingModel()` code I don't really have any actual case where `previousStage` etc is needed. So to get to the end of this, I believe I should withdraw this PR so that you submit a new one with the `getWorkingModel()` additions. This way credit goes where credit is due.

I this ok with you?

dzach:
Went ahead and removed currentStage, previousStage and convenience stage name `..`. Now only .getWorkingModel() remains as a change in this PR

vonnieda:

dzach:
No problem @vonnieda , I've been happy using @cri-s little jewel about everywhere locally without a single problem.

vonnieda:
The simplified change looks good to me, thanks @dzach!

dzach:
Thanks @vonnieda and @cri-s !

PR #577, opened 2017-06-01
prerotate patch

cri-s:
# Justification
correct xy offset handling for prerotate and adding preliminary suport for per part prerotate.
What was previously wrong.
XY offsets:
Using preRotate boolean returned from partAligner there was not considered.
Returning false, there was calculated wrong.
As example if part is prerotated to 266 degree and part is wrong -16 degree, the X/Y offsets need to be
rotated only 16 degree instead of 272 degree as the previous code does.
The reason is that the rotation difference from reference picture to target position is only 16 degree and
not 272 degree because part was pictured with nozzle at 266 degree.
Further offset calc was presumibly wrong as the vision was checked for +-45 degree from horizontal 
and not related to final part rotation.

# Instructions for Use
same as before, 

# Implementation Detail
added generic double value support for pipeline, setValue and getValue 
that can be used to pass a double between pipeline and calling code.
On Bottomvision it passes down the final rotation angle, and if pipeline set it to NaN
bottomvision return the final offset preRotated.
corrected some offsets calcs, but all unproben as i actually cannot test it on real machine.
It should be verified before merging, user smdude ???

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -673,8 +673,9 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
                  */
                 if(plannedPlacement.alignmentOffsets.getPreRotated())
                 {
-                    Location location = plannedPlacement.alignmentOffsets.getLocation().multiply(-1.,-1.,0.,1);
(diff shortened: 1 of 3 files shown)


vonnieda:

cri-s:
At the moment it needs a bit more testing.

vonnieda:
Closing until fully tested, please resubmit at that time.

Issue #515, opened 2017-04-14
Jython (Python wrapper) adds a 2nd newline to the print-statement in the Log-TAB. 


ldpgh:
Jython (Python wrapper) adds a 2nd newline to the print-statement in the Log-TAB. 
- Use of sys.stdout.write() is not a real option. It adds effort to convert everything to a string.
- In case of a Javax/ScriptEngine topic ... close this issue.
- Btw. using `from __future__ import print_function` seems to hide the output completely. But it smells like a Jython issue.

```
print "line 1"
print "line 2"
print "line 3"
print 1,
print 2,
print 3
```
will be in the  Log-TAB
```
2017-04-14 09:08:17 SystemLogger INFO: line 1

2017-04-14 09:08:17 SystemLogger INFO: line 2

2017-04-14 09:08:17 SystemLogger INFO: line 3

2017-04-14 09:08:17 SystemLogger INFO: 1
2017-04-14 09:08:17 SystemLogger INFO:  2
2017-04-14 09:08:17 SystemLogger INFO:  3
```

-------------------------
Desired result in the Log-Tab
```
2017-04-14 09:08:17 SystemLogger INFO: line 1
2017-04-14 09:08:17 SystemLogger INFO: line 2
2017-04-14 09:08:17 SystemLogger INFO: line 3
2017-04-14 09:08:17 SystemLogger INFO: 123
```
The script wrapper should not issue a newline.

vonnieda:
I believe this is actually a bug in the logging code if someone wants to look into it: #288

pfried:
I pretty much figured it out, and the necessary changes will be a bit ugly. The core problem is that the PrintStream class has private methods which i cannot override which add characters like the newline to the output. To avoid this i will have to "reimplement" / override a lot of its methods. Overriding only the `write()` method is not a good solution as it results in the issue described above.

@ldpgh Which OS are you running?

cri-s:
I have a mod that send output data to separate tab as is without adding anything if script panel is present and script for adding the tab. Would this resolve you issue?

ldpgh:
OS = Win7

pfried:
> Desired result in the Log-Tab
>     2017-04-14 09:08:17 SystemLogger INFO: line 1
>     2017-04-14 09:08:17 SystemLogger INFO: line 2
>     2017-04-14 09:08:17 SystemLogger INFO: line 3
>     2017-04-14 09:08:17 SystemLogger INFO: 123

[...]

PR #579, opened 2017-06-04
Fix the double line breaks from the script runners

pfried:
# Description

As promised I took a look at this. Had not much time recently, but today I found some hours.

1. Fix the StackOverflow on shutdown where Logger gets called but is no longer available (exception loop) #288 
2. Fix the double new line issues with the Script Runners #515

The Logger will attach a listener to the VM Shutdown. Otherwise I cannot know if the tinylog logger is still available. This could be replaced by a teardown function, but right now this stuff is split up between the `Main` class and the `MainFrame`.

The code for fixing the newline issue is not the best, but using the `StringBuilder` was my favorite option altough maybe not the best in this case (some bytearray might be better or something similar).

Best, Friedrich

### DIFF
--- a/src/main/java/org/openpnp/logging/SystemLogger.java
+++ b/src/main/java/org/openpnp/logging/SystemLogger.java
@@ -13,12 +13,10 @@ public class SystemLogger extends PrintStream {
     private StringBuilder logMessage = new StringBuilder();
 
     private Level logLevel;
-    private OutputStream out;
     private boolean logToLogger = true;
 
     public SystemLogger(OutputStream out, Level logLevel) {
         super(out);
-        this.out = out;
         this.logLevel = logLevel;
 
         // In order to stop logging on shutdown, we need to make sure that we do not call the Logger anymore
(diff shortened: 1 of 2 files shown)


vonnieda:

src/main/java/org/openpnp/logging/SystemLogger.java
```diff
@@ -42,5 +38,34 @@ public void write(byte[] buf, int off, int len) {
...
+    }
+
+    @Override
+    public void write(byte[] buf, int off, int len) {
+        /*
+         * Log to tinylog as long as the Logger is available (before shutdown)
+         * As tinylog will forward all logs to the console, we do not call the super method
+         */
+        if (logToLogger) {
+            byte[] pb = new byte[len];
```
vonnieda: I'm having some trouble understanding the complexity of this function. Couldn't the same thing be accomplished with code like:
```
        if (logToLogger) {
            String str = new String(buf, off, len);
            if (str.endsWith("\n")) {
                logMessage.append(str, 0, str.length() - 1);
                flushLogMessage();
            }
            else {
                logMessage.append(str);
            }
        }
        else {
            super.write(buf, off, len);
        }
```

Specifically, I don't see the reason for the new byte array, the array copy or the char by char loop of the string.

Other than this, the change looks good. I tested it with and without system output and it seems to work well, whether with your code as specified  of my change listed above.

pfried: you are right, i did not think it over after refactoring (where it came from), will change it

vonnieda:
Looks great, thanks @pfried!

Issue #580, opened 2017-06-08
Add 'tip diameter' to nozzle tip wizard


dzach:
# Feature Request
## Describe the Feature
Add an entry for setting the tip diameter for each nozzle and have it available to the pipeline and the rest of the system. It can be used in many places to represent the size of the tip nozzle, e.g. as a reticle, as a center mark in the pipeline image, as the COG in models, e.g. rotated rects, helping diagnose incompatibility of nozzle tip with part size:

![screenshot](https://user-images.githubusercontent.com/1109829/26934719-da4ea336-4c72-11e7-98c0-873cf4e97f99.png)

## How Will It Look?
If the feature will include changes to the user interface, how would they look? Include drawings or mockups if possible.

![tipdiameter](https://user-images.githubusercontent.com/1109829/26934657-b2030502-4c72-11e7-99d1-1b732600353c.png)

## Why Do We Need It?
1. What problem does it solve?
Today, we are not able to directly compare nozzle tip size with part size. The proposed feature will provide visual clues for such comparison.
2. Is it useful for everyone, or does it only solve a problem for your machine?
This is for the benefit of everyone using OpenPnP.

vonnieda:
Great idea!

dzach:
Ok, I've got my `infinite monkey coder (TM)`to go as far as getting the label and the text box to appear in the wizard. 

![screenshot-1](https://user-images.githubusercontent.com/1109829/26940139-43fbdfe2-4c82-11e7-9aa8-b1e69fb5f81e.png)

There is a chance I have also succeeded to properly include the new property `tipDiameter` to the `AbstractNozzleTip` class.

How do I get the value of the tip diameter, say from a pipeline script stage? `pipeline.getNozzle()` returns just `null`. Is this correct?

I can reach my `tipDiameter` with this:
```
pipeline.getCamera().getHead().getDefaultNozzle().getNozzleTip().getTipDiameter()
```
but this doesn't feel right. Or does it?

vonnieda:
Nozzle has to be set on the pipeline before it's processed. So, in whatever class you want to use this, you need to call pipeline.setNozzle(the nozzle) before calling pipeline.process().

dzach:
Yes, now that I think of it, it is logical. In my example with the pipeline: a pipeline does not have a direct connection to a nozzle, but it does have one with the camera that is grabbing the image it is processing. So, I guess, it does feel right to get the tipDiameter through the camera, after all.

[...]

PR #581, opened 2017-06-08
Feature nozzle tip diameter

dzach:
# Description
Add an entry for setting the tip diameter for each nozzle and have it available to the pipeline and the rest of the system. It can be used in many places to represent the size of the tip nozzle, e.g. as a reticle, as a center mark in the pipeline image, as the COG in models, e.g. rotated rects, helping diagnose incompatibility of nozzle tip with part size:
screenshot-2
See discussion in https://github.com/openpnp/openpnp/issues/580

# Instructions for Use
As is, this concerns mostly programmers.
Users can use a text field to fill in each nozzle tip diameter. Eventually this tip diameter may be used to size the reticle in the down camera view, or similar markers in a pipeline stage.

# Implementation Details
Used the following script as a SriptRun pipeline stage:
```
print(pipeline.getNozzle().getNozzleTip().getTipDiameter());
```
Result:
```
1.000mm
```

Yes.

Changed were made to package `org.openpnp.spi.base`, file `AbstractNozzleTip.java`

`mvn test` run successfully.


dzach:
Need to do some corrections. It turns out that the string from the text field needs to be properly casted for a Length. Was ok with the double before.

Will push a fix as soon as I have it working.

dzach:
My apologies for the noise.
It's fixed now, meaning the field appears properly in the NozzleTip configuration wizard and the methods  `getTipDiameter()` and `setTipDIameter()` work properly, tested from a ScriptRun pipeline stage with:
```
import org.openpnp.model.Configuration;
import org.openpnp.model.Length;

print(pipeline.getNozzle().getNozzleTip().getTipDiameter());

ndia = new Length(1.5,Configuration.get().getSystemUnits());
pipeline.getNozzle().getNozzleTip().setTipDiameter(ndia);

print(pipeline.getNozzle().getNozzleTip().getTipDiameter());
```
Output:
```
-> 1.000mm
-> 1.500mm
```

I have put a default value for the tip diameter, `1.0mm`, which will show enough of a circle if rendered as reticle. Can't think of anything that could make more sense. Please correct it to anything suitable.

dzach:
Added a screenshot (in the first post) of an actual implementation of the feature in a ScriptRun pipeline stage which draws a RotatedRect, its orientation and a circle at the COG of the rect with size equal to the nozzle tip diameter, here `3.6mm` for a `Juki-505`.

cri-s:
I advise to not do that.
If you need mask out the nozzle on uplooking vision, just use circle diameter of  Math.min() of bodyWidth and bodyHeight from Part->Package->Footprint , that works without problems.
Small nozzles have two holes instead of ones or cross and this then need rectangle and not circles.
Led nozzles as example need negative circles mask value as the nozzle is visible outside of part and reflect.
Further JUKI NOZZLE 505 ASSEMBLY Ø 3.5 / Ø 1.7  you need to draw the 1.7mm inner nozzle hole and
not the outer dimension of pickup nozzle. This, it maybe is needed to mask of from vision pipeline,
but what matter to suction is the inner diameter. Further as example juki 501 nozzle have two 0.2mm holes
(robotdigg sells a different nozzle with single hole having 0.3mm drilling -- official 0.25mm as seen from several  users).
There are other odd shape nozzles with gripper or connectors that really don't fit on this description,
or special melf/led nozzles that have molded or structured rubber hold by metal on outside, same for
connector nozzles.

vonnieda:
That's a good point @cri-s - many nozzles are not round. Do you and @dzach think there is value in storing this information anyway? It seems to me like it would still be a good bit of info to have available even if it doesn't model the nozzle exactly.

dzach:
@cri-s Good point indeed. However it's all a matter of choice. Having a choice is a lot better than not having one. I believe that the majority of OpenPnP users could benefit from having an indication for the size of the tip, even if it is not an exact drawing of it. Not having anything is worst, in my view.

Speaking of an exact drawing, one could possibly use something like **the facility to draw packages** and draw a better outline for a given nozzle, and not just a circle, and then use that as a reticle, or whatever, to indicate the size. 

Can you provide some workable code with an implementation of the above, or an even better solution?

dzach:
Also, using it as a reticle or as an educated indication of the COG of a rotated rect would be **optional**. One may decide to have an 1.5mm circle as a reticle instead of an 1.5mm circle indicating the approximate size of the tip.

dzach:
This is what I mean with **the facility to draw packages**:

screenshot

This is a Juki-505 with its internal and external tip diameters, and an extra square to show what I mean. How about that @cri-s ?

cri-s:
I don't see any value, as i have used nozzleTip calibration and even
that the circle diameter
for nozzleTip never matched the ideal nozzleTip diameter.
For masking out nozzle, Math.min of body width and height works better
as having nozzle
parameter.
Having a representation of nozzle diameter is useless in real life.
On setup, you want component footprint, not nozzle footprint.
And as you have the table of nozzles and diameters if you need that
info, it is faster to
setup the reticle with the diameter you want, knowing then what it is,
inner/outer diameter or
average of that two values.
The problem is not this value, the problem is when this is used on
other part, as example
nozzleTip calibration, and then the other nozzles don't work, where
when this useless part
was not implemented, then the other part was implemented differently.
This is similar as the eagle import code limits the vision part or
the NavView JavaFX code limit  UI friendness.

It was only a advice, if you want having it i suggest you to setup
nozzleTip id as package
and possibly part. Part can be anything, as example Nozzle-505 .
package need to be the same id or name of NozzleTip. This gives you
the outline/shape drawing for free, same for the editor and the
reticle.




2017-06-09 17:47 GMT+02:00, Dimitrios <notifications@github.com>:

dzach:
I've already done this @cri-s from a pipeline stage:

screenshot-3
>And as you have the table of nozzles and diameters

Where is the "table of nozzles and **diameters**"?

dzach:
@cri-s thank you for the comments. They are really helpful and full of information, although frequently  negative.

If there is a "table of nozzles and **diameters**" I can certainly use that.

I too find that displaying a complex nozzle tip as a reticle not very useful, however I think it would be  useful to have a choice of `nozzle` as a reticle in the camera view right click menu, that would be similar to the fiducial -> circle one, only automatically sized to nozzle diameter. Either that, or a choice of `nozzle` in fiducial -> size.

I also think it would be useful to (optionally) draw a correctly sized but simple outline of the nozzle tip on oriented rotated rects (I've added that to `DrawRotatedRects` stage some time ago), like the one shown in the first post, because it is a no cost element of HMI that offers visual cues connecting the actual world (image) with the model that the machine considers as "world", as long as these cues do not clutter the image.

So, @vonnieda , answering your question above, yes, I think there is value in storing this information, both for the reticle and for the pipeline. I will certainly make use of it in `DrawRotatedRects` where a choice of `drawRectCenter` and a `rectCenterRadius`=0 (or -1) will draw the nozzle diameter set for the current nozzle.

dzach:
@cri-s 
> If you want having it i suggest you to setup nozzleTip id as package and possibly part. Part can be anything, as example Nozzle-505 . package need to be the same id or name of NozzleTip. This gives you the outline/shape drawing for free, same for the editor and the reticle. 

I've already tested this, and is a viable solution. I use it in a stage that is called `ShowTipReticle` at the top of a pipeline with the intend to exist while the loose part feeder feeds parts, and then, at the bottom of the pipeline, the original reticle is restored with a `RestoreReticle`. At present it was using the tipDiameter, but I've modified the code and it now uses any package with the same name as the nozzle tip.

This can potentially lead to other interesting applications, like match part template using the package shape, for all parts that have a package with a shape defined et.c.

I'll pursue that venue for now, which offers more possibilities and does  not require any change.
So, I'm closing this PR.

@cri-s thank you for your helpful comments.
@vonnieda , even if the PR is withdrawn, the Java lesson was great. Thank you!

PR #587, opened 2017-06-22
Pre rotate test

cri-s:
Test prerotate feature with ClosedLoop Stage removed as wanted by  @vonnieda  comments.
It now work globally without the possibility to have it per part or to do better alignment using closed loop
vision process.

Superset of this feature was tested, it should work and just one picture was tested on this branch.
If first stage is ImageRead and second stage is Rotate it works on stored images.
I have leaved this feature for testing purposes.


cri-s:
The branch used was one before removing the prerotate configurations, you probably need to revert back first.

vonnieda:
Merged with additional commit re-enabling the pre-rotate checkbox. Should be available for download from http://openpnp.org/test-downloads/ once the build finishes: https://travis-ci.org/openpnp/openpnp/builds/245876689

Issue #401, opened 2017-01-03
"Final Rotation" option for bottom vision


vonnieda:
Add a "Final Rotation" option to ReferenceBottomVision which will cause the findOffsets function to rotate the part to it's final rotation before performing bottom vision. By doing this before performing bottom vision we can eliminate the majority of any nozzle runout that might exist. 

More information about this idea can be found at:
* https://github.com/openpnp/openpnp/issues/264#issuecomment-214624198
* https://github.com/openpnp/openpnp/issues/286
* https://github.com/openpnp/openpnp/pull/311

vonnieda:
Note, there is significant discussion about this in https://groups.google.com/d/msgid/openpnp/cb9112c2-5d71-467f-abc1-c798417f920c%40googlegroups.com?utm_medium=email&utm_source=footer

As of right now there is code checked in that is not working. I am disabling that code for now and it will be reenabled when fixed.

vonnieda:
Disabled in https://github.com/openpnp/openpnp/commit/6534644a1636b158bccd8ba6cb47e2f76208999f

vonnieda:
Fixed in #590.

PR #588, opened 2017-06-24
Pre rotate fix

cri-s:
# Description
It does prerote removing nozzle runout offset.

setValue/getValue is needed for several stages that need to know the intended rotation on bottom vision, one example is ImageCapture if it need takin 2 snapshoot and stitch the image because camera
FOV is too small.
It require the "Temporarily disable the Bottom Vision pre-rotate feature until ..." commit to be rolled back.

# Justification
#401 was mentioned , but was not the origin of this PR:

# Instructions for Use
Just set the checkbox and prerotation get enabled.

# Implementation Details
SMdude have done the tests , a big thank to him.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -69,6 +69,8 @@ public class ReferenceBottomVision implements PartAlignment {
         // Pre-rotate to minimize runout
         double preRotateAngle = 0;
         if (preRotate) {
+		// Prerotate just prerotate nozzle, take picture, adjust rotation and take second picture before getting xy offsets.
+		// is is complicated by the fact that it need some workaround for rotation angle that is not normalized.
 			if(part == null || nozzle.getPart() == null) { throw new Exception("no part on nozzle"); }			
 			if((!part.toString().equals(nozzle.getPart().toString()))) { throw new Exception("Part mismatch with part on nozzle"); }
 			part=nozzle.getPart();
@@ -88,6 +90,7 @@ public class ReferenceBottomVision implements PartAlignment {
 				throw new Exception("Bottom vision alignment failed for part " + part.getId()
 						+ " on nozzle " + nozzle.getName() + ". No result found.");
 			}
+		// first picture was ok, now calc error
 
 
 			RotatedRect rect = ((RotatedRect)(pipeline.getResult("result")).model);
@@ -96,13 +99,14 @@ public class ReferenceBottomVision implements PartAlignment {
 			while (Math.abs(rect.angle) > 45.0) { rect.angle += (rect.angle < 0)? 90 : -90; } 
 			angle += rect.angle;
 			while (Math.abs(angle) > 45.0) { angle += (angle < 0.)? 90 : -90; } 
-
+		// error is -angle
 			nozzle.moveTo(new Location(LengthUnit.Millimeters,Double.NaN, Double.NaN, Double.NaN, placementAngle+angle), part.getSpeed());
 			pipeline.process();
 			if (!((pipeline.getResult("result")).model instanceof RotatedRect)) {
 				throw new Exception("Bottom vision alignment failed for part " + part.getId()
-					 " on nozzle " + nozzle.getName() + ". No result found.");
+					+ " on nozzle " + nozzle.getName() + ". No result found.");
 			}
+		// second picture taken
 
 			rect = (RotatedRect) pipeline.getResult("result").model;
                         Logger.debug("Result rect {}", rect);
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -100,6 +100,8 @@ public class ReferenceBottomVision implements PartAlignment {
 			angle += rect.angle;
 			while (Math.abs(angle) > 45.0) { angle += (angle < 0.)? 90 : -90; } 
 		// error is -angle
+			if(Math.abs(angle)>0.0765) { angle+=0.0567 * Math.signum(angle); } // rounding 
+		// error 
 			nozzle.moveTo(new Location(LengthUnit.Millimeters,Double.NaN, Double.NaN, Double.NaN, placementAngle+angle), part.getSpeed());
 			pipeline.process();
 			if (!((pipeline.getResult("result")).model instanceof RotatedRect)) {
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -19,6 +19,7 @@ import org.openpnp.model.Length;
 import org.openpnp.model.LengthUnit;
 import org.openpnp.model.Location;
 import org.openpnp.model.Part;
+import org.openpnp.gui.MainFrame;
 import org.openpnp.spi.Camera;
 import org.openpnp.spi.Nozzle;
 import org.openpnp.spi.PartAlignment;
@@ -28,6 +29,7 @@ import org.openpnp.util.OpenCvUtils;
 import org.openpnp.util.Utils2D;
 import org.openpnp.util.VisionUtils;
 import org.openpnp.vision.pipeline.CvPipeline;
+import org.openpnp.vision.pipeline.stages.Rotate;
 import org.openpnp.vision.pipeline.CvStage.Result;
 import org.pmw.tinylog.Logger;
 import org.simpleframework.xml.Attribute;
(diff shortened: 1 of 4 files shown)


src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -65,10 +68,77 @@ public PartAlignmentOffset findOffsets(Part part, BoardLocation boardLocation,
...
+			while (Math.abs(angle) > 180.0) { angle += (angle < 0.)? 360. : -360.; } 
+			double nozzleAngle=angle;
+			double placementAngle=angle;
+            MovableUtils.moveToLocationAtSafeZ(nozzle, VisionUtils.getBottomVisionCamera().getLocation().add(new Location(part.getHeight().getUnits(), 0.0, 0.0,part.getHeight().getValue(), 0.0)).derive(null, null, null, angle));
+            CvPipeline pipeline = partSettings.getPipeline();
+            pipeline.setCamera(VisionUtils.getBottomVisionCamera());
+            pipeline.setNozzle(nozzle);
+			pipeline.setValue(angle);
+			if(Boolean.getBoolean("develop")) {
+            if ("Rotate".equals(pipeline.getStages().get(1).getClass().getSimpleName())) {
```
vonnieda: What is the purpose of this code? Is it intended to go into production, or was it just used for development?

cri-s:
It is intended to go into production.

The part compare is for the wizard when user press the button "Align"
Same the check for boardLocation == null.
The Rotate stage is for running without real capture and for eventual
automatic testing,
To not conflict with user and create confusion it require starting openPnP with
-Ddevelop=true  argument in order to enable it, hovewer if there could
be a problem that
code should be present inside production code in order to eventually
trace down the problem.


2017-06-25 22:05 GMT+02:00, Jason von Nieda <notifications@github.com>:

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
vonnieda: Please add comments documenting how this is to be used, and under what conditions, or just remove it. I'd rather it was just removed.

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -65,10 +68,77 @@ public PartAlignmentOffset findOffsets(Part part, BoardLocation boardLocation,
...
+							+ " on nozzle " + nozzle.getName() + ". No result found.");
+				}
+
+				rect = (RotatedRect) pipeline.getResult("result").model;
+		                Logger.debug("Result rect {} inside vision loop", rect);
+				offsets = VisionUtils.getPixelCenterOffsets(pipeline.getCamera(), rect.center.x, rect.center.y).derive(null,null,null,-angle);
+				s = "Align offset for "+part.getName()+": " + offsets.toString() + "     " ;
+				MainFrame.get().setStatus(s);
+				offsets = offsets.derive(null,null,null,Double.NaN);
+				if(cnt>Integer.getInteger("contours",40)||Double.isNaN(pipeline.getValue())) {
```
vonnieda: Please add comments documenting the Integer.getInteger("contours", 40) part, and explain what it is for.

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -65,10 +68,77 @@ public PartAlignmentOffset findOffsets(Part part, BoardLocation boardLocation,
...
+				angle = Utils2D.calculateBoardPlacementLocation(boardLocation, placementLocation).getRotation();              
+			}
+			while (Math.abs(angle) > 180.0) { angle += (angle < 0.)? 360. : -360.; } 
+			double nozzleAngle=angle;
+			double placementAngle=angle;
+            MovableUtils.moveToLocationAtSafeZ(nozzle, VisionUtils.getBottomVisionCamera().getLocation().add(new Location(part.getHeight().getUnits(), 0.0, 0.0,part.getHeight().getValue(), 0.0)).derive(null, null, null, angle));
+            CvPipeline pipeline = partSettings.getPipeline();
+            pipeline.setCamera(VisionUtils.getBottomVisionCamera());
+            pipeline.setNozzle(nozzle);
+			pipeline.setValue(angle);
```
vonnieda: How is setValue used and why is it needed? Is it being used in a custom stage? None of the existing stages or any of the changes you checked in use it, aside from one use below which can be accomplished with a local variable.

src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
```diff
@@ -197,6 +198,14 @@ public void setNozzle(Nozzle nozzle) {
     public Nozzle getNozzle() {
         return nozzle;
     }
+	
+    public void setValue(double value) {
```
vonnieda: This does not appear to be used in any stages. It should be removed. If it is required, please commit code that actually uses it and rename it to something more descriptive. "value" is too generic for a property of the pipeline.

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -65,10 +68,77 @@ public PartAlignmentOffset findOffsets(Part part, BoardLocation boardLocation,
...
+
+			int cnt=1;
+			if (pipeline.getResult("contours")!=null&&pipeline.getResult("contours").model instanceof List<?>) {
+				cnt=((List<?>)(pipeline.getResult("contours")).model).size();
+			}
+
+			String s="";
+			Location offsets = null;
+			int old=999;
+			int rot=999; do {
```
vonnieda: Please add some comments describing how this algorithm works.

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -65,10 +68,77 @@ public PartAlignmentOffset findOffsets(Part part, BoardLocation boardLocation,
...
+				offsets = VisionUtils.getPixelCenterOffsets(pipeline.getCamera(), rect.center.x, rect.center.y).derive(null,null,null,-angle);
+				s = "Align offset for "+part.getName()+": " + offsets.toString() + "     " ;
+				MainFrame.get().setStatus(s);
+				offsets = offsets.derive(null,null,null,Double.NaN);
+				if(cnt>Integer.getInteger("contours",40)||Double.isNaN(pipeline.getValue())) {
+					if (rect.size.width < rect.size.height) { rect.angle = 90 + rect.angle; }
+					while (Math.abs(rect.angle) > 45.0) { rect.angle += (rect.angle < 0)? 90 : -90; } 
+					angle=nozzleAngle + rect.angle;
+					while (Math.abs(angle) > 45.0) { angle += (angle < 0.)? 90 : -90; }
+					double step=(360./200./16.);
```
vonnieda: Why 360/200/16? That looks like it's assuming 16 microsteps on a direct drive nozzle. Is that important? Will it work with other types of nozzles?

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -65,10 +68,77 @@ public PartAlignmentOffset findOffsets(Part part, BoardLocation boardLocation,
...
+			while (Math.abs(angle) > 45.0) { angle += (angle < 0.)? 90 : -90; } 
+
+			int cnt=1;
+			if (pipeline.getResult("contours")!=null&&pipeline.getResult("contours").model instanceof List<?>) {
+				cnt=((List<?>)(pipeline.getResult("contours")).model).size();
+			}
+
+			String s="";
+			Location offsets = null;
+			int old=999;
```
vonnieda: What is the meaning of 999 here? Can it be replaced with a more meaningful value?

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
cri-s: Yes, it is something like that and it is important for closed loop. It works with other types of nozzles
and different microstepping without updating this numbers.

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
cri-s: it can be replaced with 99 or any other value bigger as that. Ideally it should be greater as 360 and
for security greater as 720, so 999 fit nicely. Max_Integer fit too. The other requirement is that variable rot need to be initialized with the same value. As maxInt when converted to double is Infinite, i prefer safer value 999.  It is used inside old=Math.min(old,rot); and no , it is not a special
case  code, 765 or 123 works equals.

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
cri-s: setValue is double type variable that just set a value.
In this case the value variable inside pipeline holds the target part angle value, that differ from
nozzle rotation value.
It is used as example inside ImageCapture when part size is bigger as camera FOV or inside MaskRectangle when value is less then 3 (double) where it set the rectangle to part size,
Further when having prerotate without having access to this variable it is not possible to extract
footprints like this:
https://cloud.githubusercontent.com/assets/9506454/25804378/d56d8682-33fb-11e7-8c2a-92b63ef54c6d.png
and as consequence drop the ic because of coplanarity error .

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
cri-s: This is a feature for better usage.
If instead of MinAreaRect  the follow stages get used:
Threshold with value of min from MinAreaRect
findContours  external
MinAreaRectContours with findContours stage
The result is the same, but findContours result hold the number of discovered pins.
It has to be noted that each pin on package like soic/tssop/qfp/... each pin get detected with
two contours, and 32 pins get 64 contours on correct illumination. Because not all have correct illumination, maybe 80 or 70% of pins get correctly detected, and the number is then near 50
for 32pin tqfp package.  Now if vision detect that part have more then 40 pins, it uses automatically closed loop instead of prerotate. To enable this behaviour, the findContour stage
mentioned above must be named "contours" and with -Dcontours=20 that value could be lowered or increased. For tqfp32 it is never needed, but for ram as bga on wson package as example yes.  It was a user request during the extended beta test.

src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
cri-s: ok, i see, i need to remove closed loop and other functionality for having more simplier code

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
cri-s: The algorithm implement the old BottomVision-Cri.s stage rotating the real nozzle instead of
the image.
The result is a big workaround and required a lot of testing and debugging.
Basically code inside Location need to be changed , but as this have sideffects when loading position from camera or nozzle as example, it was choosen to do workaround.
Further as @vonnieda don't have liked the ClosedLoop stage it was implemented using the prerotate flag and some code was just copied because it was tested as is . I simplifiy it.

vonnieda: @cri-s I don't understand why you keep claiming I don't like the ClosedLoop stage. The ClosedLoop stage you committed was empty. It was a class with an empty process() body. I was just asking you to remove the empty class.

src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
vonnieda: @cri-s I didn't ask you to remove the "closed loop and other functionality". I'm asking you to remove setValue() which does not appear to be used. If it's not used, then what is the value of adding it? And if it *is* used, then commit the code that uses it. Right now none of the code you submitted uses it.

cri-s:
The closedLoop stage was empty, as i have used code inside BottomVision to
check the presence of ClosedLoop stage. As from other submitting i
know you don't like such coding and when you have asked to remove that
stage  i have implemented the prerotate setting with the same
functionality with the bottomVisionWizard.
It had the same code and functionality, just different method to enable it.
below extract code form how the empty class was used originally inside
bottomVision,
i

            CvPipeline pipeline = partSettings.getPipeline();
            pipeline.setCamera(camera);
            pipeline.setNozzle(nozzle);
            boolean preRotate = false;
            boolean postRotate = false;
                for (CvStage stage : pipeline.getStages()) {
                    if
(!"ClosedLoop".equals(stage.getClass().getSimpleName())) { continue; }
                    preRotate = true;
                    postRotate = stage.isEnabled();
                }
			Logger.warn("preRotate = {} - postRotate = {}",preRotate,postRotate);
            preRotateAngle = preRotate ? preRotateAngle : 0.0;
			double placementAngle = preRotateAngle;
            startLocation =
startLocation.add(partHeightLocation).derive(null, null, null,
preRotateAngle) ;
            MovableUtils.moveToLocationAtSafeZ(nozzle, startLocation);


2017-06-26 1:45 GMT+02:00, Jason von Nieda <notifications@github.com>:

cri-s:
> @cri-s I didn't ask you to remove the "closed loop and other functionality"
That is true,
.
> I'm asking you to remove setValue() which does not appear to be used. If
> it's not used, then what is the value of adding it? And if it *is* used,
> then commit the code that uses it. Right now none of the code you submitted
> uses it.
>
At the end this PR is for prerotate.
Removing this variable removes closed loop.feature, unfortunatly i
don't have code
to show the usage of this variable or what i have is totally untested
and too premature
for inclusion into this PR. If closed loop is requested in future, it
can be added later.

Cheers Cri.s

cri-s:
HI Jason, probably i have a problem viewing this PR on website, as i see 1 review request but cannot locate it or respond it.
I'm done with this PR, maybe check the last ReferenceBottomVision and let me know if i should
change something or you need further clearification about something because i have little difficulty with
the git web interface, don't know why.

src/main/java/org/openpnp/vision/pipeline/CvPipeline.java
cri-s: > @cri-s I didn't ask you to remove the "closed loop and other functionality"
That is true,
.
> I'm asking you to remove setValue() which does not appear to be used. If
> it's not used, then what is the value of adding it? And if it *is* used,
> then commit the code that uses it. Right now none of the code you submitted
> uses it.
>
At the end this PR is for prerotate.
Removing this variable removes closed loop.feature, unfortunatly i
don't have code
to show the usage of this variable or what i have is totally untested
and too premature
for inclusion into this PR. If closed loop is requested in future, it
can be added later.
 
Cheers Cri.s

cri-s:
I'm close this and resubmit it later.

PR #598, opened 2017-07-10
Protect board against nozzle

machinekoder:
# Description
This patch adds a feature that checks the nozzle position against the board position and Z distance. If the nozzle would move into the board either from the side or from the top it throws an exception.

# Justification
Crashing the nozzle into board can damage the board as well as the nozzle. Therefore, it makes sense to prevent jog moves from accidentally crashing into the board. The code is generic and not machine specific.

# Instructions for Use
1. Load a board and set the Z position.
2. Move your head above the board e.g. by using the camera.
3. Move the nozzle to the head position.
4. Try to move the nozzle into the board (be careful!).
5. When reaching the Z tolerance position an exception message will pop up.
6. To override the Z limit you can use the button in the jog panel.

image

# Implementation Details
The code has been tested on a real machine with the board rotated in different angles. Changes only affect the JogControlPanel.

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -147,8 +147,7 @@ public class JobPanel extends JPanel {
 
     private FiniteStateMachine<State, Message> fsm = new FiniteStateMachine<>(State.Stopped);
 
-    public JobPanel(Configuration configuration, MainFrame frame,
-            MachineControlsPanel machineControlsPanel) {
+    public JobPanel(Configuration configuration, MainFrame frame) {
         this.configuration = configuration;
         this.frame = frame;
 
--- a/src/main/java/org/openpnp/gui/JogControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/JogControlsPanel.java
@@ -221,7 +221,8 @@ public class JogControlsPanel extends JPanel {
                             targetLocation, new Length(1.0, l.getUnits()));
                     if (!safe) {
                         throw new Exception(
-                                "Z Axis would move against " + boardLocation.toString());
+                                "Nozzle would crash into board: " + boardLocation.toString() + "\n" +
+                                "To disable the board protection go to the \"Safety\" tab in the \"Machine Controls\" panel.");
                     }
                 }
             }
@@ -325,9 +326,6 @@ public class JogControlsPanel extends JPanel {
         homeButton.setIcon(Icons.home);
         homeButton.setHideActionText(true);
         panelControls.add(homeButton, "2, 2");
-
-        boardProtectionOverrideCheck = new JCheckBox("Override Board Protection");
-        panelControls.add(boardProtectionOverrideCheck, "24, 2");
 
         JLabel lblXy = new JLabel("X/Y");
         lblXy.setFont(new Font("Lucida Grande", Font.PLAIN, 22));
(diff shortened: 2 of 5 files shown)


vonnieda:
Thanks for the PR @machinekoder, this looks very handy! I'm a bit swamped at the moment, so it may take a day or two for me to be able to review and merge this. Please stand by.

machinekoder:
@vonnieda No need to hurry. I work with OpenPNP on Monday evenings.

vonnieda:
@machinekoder This looks good. My only issue with it is the placement of the checkbox as that will force that pane to be much larger than default. How would you feel about moving it into the top level menu? Do you think that would be too many clicks to get to it?

machinekoder:
@vonnieda You usually need this option when calibrating the head to nozzle offset and the board Z distance. I think moving the option to the top menu should work. However, the function closely relates to the jog controls, so I have the feeling that it feels more natural there. What do you think about moving it below the run controls maybe?

vonnieda:
@machinekoder Here is the issue:

screen shot 2017-07-12 at 9 18 37 am

This is the default size for the components in the UI. The current UI is designed to fit on a common laptop screen size of 1440x900. This was chosen based on a lot of feedback from users who were having trouble fitting everything on the screen. 

So, there is very little room to add anything to the jog controls unless it's an icon. So, here are a few more suggestions:
* Add a new Machine Controls tab (like Jog, Special, Actuators, etc.) called Safety and put it there. This would be good because there has recently been interest in adding other safety functions and this would be a good place for them to live.
* Add it as an icon button to the jog controls, maybe by moving the two positioning buttons on the right up to the top and put the new one at the bottom. We would make it a two state button with two icons, one being "safe" and one being "unsafe" and it would switch between the two. I can help with the icons if needed.
* And as I mentioned before, I'd be fine with having it in the top level menu, too. I still think this might be the best choice as it would let users try it out without taking up valuable UI space. If it turns out that people really like it and want it to be more visible we could move it then.

machinekoder:
@vonnieda Thank you for your feedback. I updated the PR by moving the Override CheckBox to a new "Safety" tab.

vonnieda:
Hi @machinekoder, I am doing some testing on this and it's not working the way I'd expect. Here is what seems odd to me:

1. Load a job where board is set to Z = 0.
2. Jog nozzle away from board.
3. Jog nozzle to Z = -3.
4. Jog nozzle over board.

I expected this to throw an error since I am dragging the nozzle through the board, but it didn't. It does seem to work if I try to move over the board and then lower the nozzle further, but at that point it's already below the surface of the board.

machinekoder:
@vonnieda The board protection does not work for movements made via the camera window (dragging) yet. Can you point me to the code where the movement command is executed?

vonnieda:
Oh, of course. I didn't even consider that even though I looked at the
code. Sorry about that :)

The camera jog happens in CameraView.java:moveToClick, but putting stuff
about the Job there is not going to fly.

I'm open to suggestions. If you think it will be beneficial to merge as is,
I'm fine with that, or if you'd like to improve it by adding camera
support, that's good too. If you do the latter, I'd like to centralize the
code, rather than copying it around. Maybe a public function in
MachineControlsPanel like jogToLocation(HeadMountable, Location), and do
the checks there, then modify both jog and moveToClick to use it?

Jason

vonnieda:
Ping @machinekoder - can you see my above comment and respond please?

machinekoder:
@vonnieda From what I understand the nozzle moves to a safe z distance before any move done with the camera so it should not be possible to crash into a board using the camera window. Your test case with the board at Z0 or at any location higher than the safe z height is a special case. However, this would indicate an error in the configuration. Why would you set your safe z height in a way that will guarantee you to crash into the board?

vonnieda:
@machinekoder Ah, great point that I hadn't considered. Sorry about that. I'll take another look at this tonight after work and will likely be able to go ahead and merge it. Thanks for your patience!

vonnieda:
Thanks for your work on this @machinekoder!

machinekoder:
@vonnieda It was a pleasure.

Issue #601, opened 2017-07-22
Panelization Dialog Fiducial Dropdown Failure


vonnieda:
On Windows, at least, you can't drop the global fiducial dropdown in the panelization panel with the mouse, you must use the keyboard. This is likely due to a bad value in the dropdown but I haven't had time to look yet.

See further discussion in https://groups.google.com/d/msgid/openpnp/b6f2e5e7-dfe6-4426-8690-4b37ecee5469%40googlegroups.com?utm_medium=email&utm_source=footer

vonnieda:
@BendRocks Tagging you in hopes you will take a look at this :)

vonnieda:
This may be related to http://bugs.java.com/bugdatabase/view_bug.do;jsessionid=d238b89a70e0051d5edca359b731?bug_id=4106334

BendRocks:
OK, I'm on it. The dropdown not happening was noted in #501. I'll re-visit and report back

vonnieda:
@BendRocks If you'll just focus on the dropdown and ignore #602 I'd appreciate it. I have a few hours to blow so I am fixing the stuff in #602 now, and since this one (the dropdown) doesn't happen on Mac it's harder for me to fix.

BendRocks:
OK, got it.

vonnieda:
Potential fix pushed in https://github.com/openpnp/openpnp/commit/92465bf4e4df7c61201fd299a6b1d1c55699eafd, waiting for user feedback.

BendRocks:
Confirmed this fixes the dropdown issue on windows. But why? There are other places with dropdowns that  don't have this issue and don't need this code.

vonnieda:
@BendRocks See http://bugs.java.com/bugdatabase/view_bug.do;jsessionid=d238b89a70e0051d5edca359b731?bug_id=4106334. The fix contains two fixes, really. The first makes the dropdown menu a heavy weight component, which means it's a Windows native component. This doesn't have the limitation of not being able to extend outside the dialog box. The second is that it limits the length of the dropdown to 7 items, which further ensures it won't try to extend outside the dialog box. The root problem is that dropdowns can't extend outside dialogs, and since this is the only (or one of the only) place we use a dropdown in a dialog, we only see it here.

BendRocks:
[...]

PR #603, opened 2017-07-22
Update DlgAutoPanelize.java

cri-s:
drop down fix.

# Read This First

will not be accepted.

# Description
Describe the change in detail. Explain how it works. The better you describe the change the more likely it is to be accepted.

# Justification

# Instructions for Use

# Implementation Details

### DIFF
--- a/src/main/java/org/openpnp/gui/panelization/DlgAutoPanelize.java
+++ b/src/main/java/org/openpnp/gui/panelization/DlgAutoPanelize.java
@@ -113,8 +113,21 @@ public class DlgAutoPanelize extends JDialog {
         jPanel.add(textFieldboardPanelFid2Y, "4, 16, fill, default");
 
         jPanel.add(new JLabel("Panel Fiducial Part", JLabel.RIGHT), "2, 16, right, default");
-        partsComboBox = new JComboBox(new PartsComboBoxModel());
-        partsComboBox.setRenderer(new IdentifiableListCellRenderer<Part>());
+        partsComboBox = new JComboBox();
+        partsComboBox.setEditable(false);
+        partsComboBox.setLightWeightPopupEnabled(false);
+        partsComboBox.setMaximumRowCount(7);
+
+        for (Part p: Configuration.get().getParts()) {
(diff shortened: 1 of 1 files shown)


vonnieda:
Thanks @cri-s! I've incorporated the primary changes in https://github.com/openpnp/openpnp/commit/92465bf4e4df7c61201fd299a6b1d1c55699eafd and we'll see if this fixes the problem. I have not tested on Windows yet. I did not include some of the other code that was less important since it changes the storage model of the dialog.

vonnieda:
@BendRocks Looks like we may have a fix here. It's been pushed and will wait for feedback.

BendRocks:
@vonnieda , confirmed this fixes on windows. 

For future reference:

There are 3 lines in the commit, but the 'setlightweightpopupenabled' line does the magic.. 

Most curious is there are jcombobox throughout the code and they all work on windows. This is the only place the setlightweightpopupenabled is needed.

I wonder why that is?

In any case, @cri-s, I'm glad you were able to fix this. I'd not have hit on the fix in hours and hours of trying.

vonnieda:
I commented as the reason in #601. In general, it's not a good idea to use the heavy weight components in a Swing app since it can cause graphics issues, but this is a good solution for now. If it turns out to cause problems later we'll face that when it happens. Thanks again @cri-s!

vonnieda:
@cri-s To your comment in the code regarding 1000 parts: Of course, I don't want there to be 1000+ parts in the dropdown, but adding in a hardcoded filter that no one knows about is not the answer, either. We have no guarantee that every user uses "fid" in their fiducials, and this changes the behavior of the fid dropdown in one place but nowhere else, which makes the code inconsistent.

A good solution to this, and to part selection in general, would be to make the part selection dropdown a searchable combobox instead. Then the user can easily start to type "FID" and get just the parts containing that name. This would improve part selection throughout the system.

I appreciate your help, but please stop with the rudeness and acting like I'm out to get you. I have been completely respectful to you in our interactions over the years, and I would appreciate the same from you.

Issue #330, opened 2016-08-19
Strip feeder should use the CvPipeline for vision processing


issus:
The stripfeeder currently does not use the CvPipeline for recognition. As converting this completely would be a large task, I propose this is split into two stages.

**Stage 1**
Add pipeline for pre-processing the image, allowing it to be cleaned up for improved recognition by the current algorithm.

**Stage2**
Completely convert the vision system to use the CvPipeline.

Stage 1 would allow the user to configure steps to clean up the image in their Machine.xml file, but still use the current recognition code on the cleaned up image. Stage 2 would be a complete conversion to using the CvPipeline for recognition.

dzach:
What does the 'ready' label mean? Is this already implemented? Where can I find more info on how to tweak the strip feeder pipeline?

vonnieda:
Ready just means it's ready to be worked on. It's not done yet.

dzach:
Oh, ok! Is there any other way to tweak some parameters of the pipeline to suit different tapes/feeders? E.g. threshold per feeder?

vonnieda:
No, not currently. That's the purpose of this ticket - to add that functionality.

richard-sim:
If anyone interested in this feature could test out my pull request #604 and report back, it'd be appreciated.

richard-sim:
#604 has been updated with Stage 2 now completed also.

vonnieda:
Will check in the next few days, currently on vacation.

vonnieda:
Fixed in #604

Issue #597, opened 2017-07-08
CvPipelineEditor often sets stage values to null unexpectedly


vonnieda:
I've noticed, and others have commented, that the pipeline editor will often change a property value to null seemingly randomly. I've noticed this mostly with the ConvertColor stage, for the conversion property and any stage that uses a Color property. 

There are two issues. The first is that this happens at all, which delete user data, and the second is that once it happens it's impossible to recover from. The stage must be deleted and re-added, often losing other settings.

So ideally, try to figure out why it's happening and fix it, and a a fallback, try to make it so that it's recoverable at least.

dzach:
Hi @vonnieda 
This happens to me all the time. I think it can be reproduced easily if one tries to change a parameter fast enough while the pipeline is still processing. If that's indeed the case, then disabling the pipeline stages from getting user input while processing might be a way to fix it.

vonnieda:
I figured out a way to reproduce this on demand:
1. Add at least two ConvertColor stages to a pipeline.
2. Select the first, select the conversion dropdown and then click away from the dropdown to have it fold back up.
3. Now press space bar, which will highlight the dropdown but not drop it down.
4. Click on the second ConvertColor stage and it's conversion value will get set to null.

vonnieda:
Note, the above seem to work with any stage that has a nullable first property.

vonnieda:
Fixed in https://github.com/openpnp/openpnp/pull/611

PR #604, opened 2017-07-25
ReferenceStripFeeder converted to use CvPipeline

richard-sim:
# Description
Stage 1 of adding CvPipeline to ReferenceStripFeeder, as per #330: Add pipeline for pre-processing the image, allowing it to be cleaned up for improved recognition by the current algorithm. Stage 1 would allow the user to configure steps to clean up the image in their Machine.xml file, but still use the current recognition code on the cleaned up image.
Stage 2 of adding CvPipeline to ReferenceStripFeeder, as per #330: Completely convert the vision system to use the CvPipeline. Stage 2 would be a complete conversion to using the CvPipeline for recognition.

Compiles and runs, however I cannot test the functionality on real hardware at this time so further work may be required.

# Justification
Addresses bug report.

# Instructions for Use
A CvPipeline can be customized to clean up the cameras' image for improved recognition. The pipeline can be configured by clicking Edit Pipeline in the feeders configuration. The final pipeline stage (with detected circles) should have the name 'results'.

During the feeder setup wizard, the strip feeder hole debug colors are as follows:
  * Red are any holes that are found.
  * Orange are holes that passed the distance check but failed the line check. **
  * Blue are holes that passed the line check but are considered superfluous. **
  * Green passed all checks and are considered good.

** - only displayed when Alt/Option is held down when the setup wizard is started.

# Implementation Details
  Has only been tested with a webcam at this stage, which is not configured properly. If proper testing can be done to confirm the functionality it would be appreciated. Unfortunately I have no access to working hardware at this point.
  pnp-test.job.xml runs successfully (in addition to using -DoverrideUserConfig=true), and appears to use vision with its ReferenceStripFeeders, so I'm hopeful that everything is working correctly. The strip feeder setup wizard runs successfully with the test job loaded, and all circles are displayed as expected in the debug camera image.
  I believe so.
  No
  Done.

One possibly dubious change has been the addition of a BufferedImageCamera, which is only to be used internally (not as a selectable camera type by the user). It is used to use a BufferedImage as the source for CvPipeline rather than an actual Camera (used in ReferenceStripFeederConfigurationWizard.showHoles()).

A second change that potentially impacts some users is changing the DetectCirclesHough CvStage to have unit-based (Length) for its sizes and distances, rather than using pixels. This change was made because:
1. It matches the old behaviour from FluentCv
2. It allows the stage to be used across different hardware configurations (no longer tied to the camera resolution, distance, etc)
3. It matches what a user expects (detect a 1mm fid, 1.5mm tape hole, etc)
4. I don't expect that DetectCirclesHough is actually being used by any users presently due to all parameters being marked as required, despite most being undocumented and unintuitive. These parameters are now marked as not-required.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/camera/BufferedImageCamera.java
+++ b/src/main/java/org/openpnp/machine/reference/camera/BufferedImageCamera.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2011 Jason von Nieda <jason@vonnieda.org>
+ *
+ * This file is part of OpenPnP.
+ *
+ * OpenPnP is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * OpenPnP is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with OpenPnP. If not, see
+ * <http://www.gnu.org/licenses/>.
+ *
+ * For more information about OpenPnP visit http://openpnp.org
+ */
+
+package org.openpnp.machine.reference.camera;
+
+import java.awt.image.BufferedImage;
+import java.beans.PropertyChangeSupport;
+
+import org.openpnp.gui.support.Wizard;
+import org.openpnp.machine.reference.ReferenceCamera;
+import org.openpnp.spi.Camera;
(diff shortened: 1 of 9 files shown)


vonnieda:
I am on vacation. Will test this next week. Thanks for the PR!

richard-sim:
As of c7d16678085dc31b680c7bad72d6ca8fa3224e38 this should completely address #330.

richard-sim:
erk, please ignore those last 3 commits - they weren't supposed to get merged into this pull request.  :/ They're for #597.

richard-sim:
Ok for the life of me I can't figure out how to get GitHub to remove those commits from this PR, and can't create a new PR without them. Fail.

vonnieda:
Create a new branch from develop, merge only the commits you want (cherry
pick) and then create a PR from that branch.

richard-sim:
Thanks - done. Superseded by #610.

PR #610, opened 2017-07-29
ReferenceStripFeeder converted to use CvPipeline

richard-sim:
# Description
Supersedes #604.

Stage 1 of adding CvPipeline to ReferenceStripFeeder, as per #330: Add pipeline for pre-processing the image, allowing it to be cleaned up for improved recognition by the current algorithm. Stage 1 would allow the user to configure steps to clean up the image in their Machine.xml file, but still use the current recognition code on the cleaned up image.
Stage 2 of adding CvPipeline to ReferenceStripFeeder, as per #330: Completely convert the vision system to use the CvPipeline. Stage 2 would be a complete conversion to using the CvPipeline for recognition.

Compiles and runs, however I cannot test the functionality on real hardware at this time so further work may be required.

# Justification
Addresses bug report.

# Instructions for Use
A CvPipeline can be customized to clean up the cameras' image for improved recognition. The pipeline can be configured by clicking Edit Pipeline in the feeders configuration. The final pipeline stage (with detected circles) should have the name 'results'.

During the feeder setup wizard, the strip feeder hole debug colors are as follows:
  * Red are any holes that are found.
  * Orange are holes that passed the distance check but failed the line check. **
  * Blue are holes that passed the line check but are considered superfluous. **
  * Green passed all checks and are considered good.

** - only displayed when Alt/Option is held down when the setup wizard is started.

# Implementation Details
  Has only been tested with a webcam at this stage, which is not configured properly. If proper testing can be done to confirm the functionality it would be appreciated. Unfortunately I have no access to working hardware at this point.
  pnp-test.job.xml runs successfully (in addition to using -DoverrideUserConfig=true), and appears to use vision with its ReferenceStripFeeders, so I'm hopeful that everything is working correctly. The strip feeder setup wizard runs successfully with the test job loaded, and all circles are displayed as expected in the debug camera image.
  I believe so.
  No
  Done.

One possibly dubious change has been the addition of a BufferedImageCamera, which is only to be used internally (not as a selectable camera type by the user). It is used to use a BufferedImage as the source for CvPipeline rather than an actual Camera (used in ReferenceStripFeederConfigurationWizard.showHoles()).

A second change that potentially impacts some users is changing the DetectCirclesHough CvStage to have unit-based (Length) for its sizes and distances, rather than using pixels. This change was made because:
1. It matches the old behaviour from FluentCv
2. It allows the stage to be used across different hardware configurations (no longer tied to the camera resolution, distance, etc)
3. It matches what a user expects (detect a 1mm fid, 1.5mm tape hole, etc)
4. I don't expect that DetectCirclesHough is actually being used by any users presently due to all parameters being marked as required, despite most being undocumented and unintuitive. These parameters are now marked as not-required.


vonnieda:
Thanks! Should be able to look at this tomorrow. Almost home :)

vonnieda:
Hi @richard-sim,

First, thanks for taking this on. It's a big task and overall the PR looks pretty good.

I am concerned about the changes to DetectCirclesHough. There are two problems, as I see it:
1. I know that there are users using it as it is now and this change will break any pipelines that contain it. I try very hard to make all changes to serialized classes backwards compatible and this will definitely not be.
2. This change locks the stage to only working inside OpenPnP, since it now requires a camera to work. I have been trying to keep the CvPipeline basically independent from OpenPnP so that it can be used outside of it and in other projects. I've allowed certain stages that are OpenPnP specific to use things like cameras, but hough circles is pretty fundamental and I don't like the idea of restricting it only to OpenPnP.

With that being said, I really like the idea of being able to use real world units in stages, and I think this is an improvement over specifying pixel dimensions. But I think we need to consider a better way to do it. I'd prefer a method that uses the pixel conversions available from a Camera when there is a camera property set, or uses pixel dimensions when it's not.

I'm going to mull this over for a day or two and see if I come up with anything that might work. Please let me know if you have any ideas, too.

richard-sim:
Yep, all good points - and good to know about the design decision to keep stages decoupled from cameras so they can be used outside OpenPnP. I'll think over this some more too - and we probably should bounce some ideas around on IRC, as I'm back from vacation now too. It seems there should be a really elegant way to achieve this, so let's find it! :)

richard-sim:
This is a first-pass at addressing the issues - while it does address them (as you suggested), I feel we can still do better somehow.

KarlZeilhofer:
Very interesting!
How about the progress of this PR? We Tried today our first job on out Lite-Placer, and the StripFeeder had sometimes problems with neighbour strips with the same pattern of holes. A modifyable CVpipeline would help very much. 
Greets, Karl from Team14

richard-sim:
Hi Karl,

Thanks for the interest! Glad to know there are people wanting this feature. Good motivation for me to get on it.

Currently we're mulling over some details of the implementation - we've got some ideas of how we want to handle it, but I don't think we're totally satisfied with the solution at this time. I'll try to get on it later this week and will see how it goes.

vonnieda:
My apologies for not getting to this sooner. @richard-sim, would you be willing to drop the length stuff entirely for now and just merge this using pixel dimensions? i.e. remove the changes to DetectCirclesHough.java? This way we can get the major feature merged and work on making it more user friendly in the future?

richard-sim:
Good call - will do, then we can tackle the units-in-CvPipeline in a separate issue rather than muddy a potentially big change with this somewhat unrelated issue.

Only issue will be how to handle backwards compatability, given ReferenceStripFeeder currently does use units. Is there a mechanism for translating old config versions, so I can do the conversion for the user (hopefully using whatever camera they have set up)?

vonnieda:
@richard-sim Ah, good point. I had forgotten that we need this data to move forward in one way or another.

The current values are not stored in config anywhere - they are hardcoded. We could add a configuration listener to ReferenceStripFeeder which would calculate the pixel dimensions and manipulate the pipeline to set them. The configuration listener is run after the configuration has been loaded. That's pretty hacky though.

Okay, different suggestion: How about just encapsulating this functionality in a completely new stage for now? It could be TapeHoleFinder to keep it specific and easy to replace later, or more generic like DetectCirclesHoughUnits or something. 

I kind of like the idea of making it specific, e.g. TapeHoleFinder. There isn't a whole lot of values in making the values used editable. Holes are always 1.5mm, they are always Nmm from the edge of the tape, etc. but at least this way we could make them configurable if you think it's helpful, and we can do it without affecting existing pipelines or locking ourselves into a unit conversion system we haven't fully considered.

Thoughts?

richard-sim:
Yep, I like the idea of adding a TapeHoleFinder stage that can be replaced when a proper system is in place. There's definitely other uses for a DetectCirclesHoughUnits (fids), but that can be held off for the proper system, or a similar approach can be taken on a case-by-case basis if required.

Thinking forward to when it comes time to replace TapeHoleFinder, having a version-aware upgrade system would be very helpful. That's a task for another day, but something I'm going to start thinking about. In the past when I've written similar systems, it's made it things a whole lot cleaner in the long-run, as you don't have to support legacy versions beyond the conversion stage, and can remove and rename things easily - keeping things lean and simple rather than cluttered with "the old stuff" that can't be removed.

richard-sim:
I went with a slightly different approach - I reverted DetectCirclesHough (mostly; some minor changes remain), and added DetectFixedCirclesHough. The new stage operates in pixels, except it is "fixed" in configuration by the current pipeline; it's min/max diameter and min distance are set on the pipeline (i.e. provided by ReferenceStripFeeder). This both completely avoids the Units issue, and allows for a generic pipeline stage that I think can be reused safely by other pipelines that need similar functionality.

As it has stage-specific properties set on the pipeline, I decided to go with a naming convention of <stage>.<property> (e.g. DetectFixedCirclesHough.minDistance) to avoid future conflicts.

vonnieda:
Merged, thank you @richard-sim for all your hard work on this and for your patience while working through the issues. This is a really huge improvement!

KarlZeilhofer:
great to here! Thank you guys!

PR #611, opened 2017-07-29
CvPipelineEditor issue #597

richard-sim:
# Description
PropertyChangeListener is fired asynchronously, so when it is received the PropertySheetPanel may no longer reflect the items the change was fired for (as is the case when the user is editing a cell of a stage’s properties and clicks to another stage, cancelling the edit while the properties change to reflect the newly selected stage).

# Justification
Addresses #597.

# Instructions for Use
No change in existing behaviours.

# Implementation Details
1. How did you test the change? Tested the pipeline editor according to the repro steps described in the issue.
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? I believe so.
No changes.
Done.


vonnieda:
Thanks for finding this. This was a huge issue and this fix is greatly appreciated. In addition, thank you for the clean, easy to merge PR!

PR #620, opened 2017-08-15
Add Check Box "Use only Value as PartId", ...

KarlZeilhofer:
TODO: fix GUI container row, see screenshot:

# Read This First

will not be accepted.

# Description
1. Adds a Checkbox "Use only Value as PartId" to the board importd dialog for KiCad. 

2. Unchecked both checkboxes by default. 

3. Both checkboxes got a tooltip. 

# Justification
ad 1) PartId could be a primary key from a company internal data base.  
Therefore mixing up value and package isn't of any help and Value in the .pos file is exactly that primary key in this case. 

ad 2) The Parts.xml is a sensitive table, which should not be polluted by default. The user has to do it with care!

ad 3) Tooltips are always a good idea. 

# Instructions for Use
If you want to create parts on import, you can either give them an ID named 'Package'-'Value' or, if both check boxes are enabled, the ID is only the 'Value' from the .pos-file. Use this, if your Value is a unique value, e.g. from a database. 

# Implementation Details
I imported a .pos file, and the created parts had the PartId only from the value. 
Yes
```Tests run: 13, Failures: 0, Errors: 0, Skipped: 0```

### DIFF
--- a/src/main/java/org/openpnp/gui/importer/KicadPosImporter.java
+++ b/src/main/java/org/openpnp/gui/importer/KicadPosImporter.java
@@ -89,7 +89,8 @@ public class KicadPosImporter implements BoardImporter {
         return board;
     }
 
-    private static List<Placement> parseFile(File file, Side side, boolean createMissingParts)
+    private static List<Placement> parseFile(File file, Side side, boolean createMissingParts, 
+    		boolean useOnlyValueAsPartId)
             throws Exception {
         BufferedReader reader =
                 new BufferedReader(new InputStreamReader(new FileInputStream(file)));
@@ -134,7 +135,12 @@ public class KicadPosImporter implements BoardImporter {
                     placementRotation));
             Configuration cfg = Configuration.get();
             if (cfg != null && createMissingParts) {
-                String partId = pkgName + "-" + partValue;
+                String partId;
+                if(useOnlyValueAsPartId == true) {
+                	partId = partValue;
+                }else {
+                	partId = pkgName + "-" + partValue;
+                }
                 Part part = cfg.getPart(partId);
                 if (part == null) {
                     part = new Part(partId);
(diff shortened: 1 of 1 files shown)


vonnieda:
This looks good @KarlZeilhofer. I saw you have a TODO up there for a screen layout issue. Do you want to fix that before this gets merged? Do you need some help?

KarlZeilhofer:
I need some help here, please. I'm used to Qt in C++, but not to Java. When I set the row to 3, I got a runtime error...

vonnieda:
@KarlZeilhofer Okay, I will have a look tonight. For your reference, all GUIs in OpenPnP are developed with https://eclipse.org/windowbuilder/. I highly recommend using this rather than trying to edit the GUI code manually.

vonnieda:
Merged with minor updates. I moved the dialog into it's own file so it can be opened more easily with WindowBuilder, and I fixed up the checkbox that was in the wrong position. Thanks for the PR @KarlZeilhofer!

KarlZeilhofer:
@vonnieda Thank you very much for your help and for the fast merge!

I will have a look into the WindowBuilder.

KarlZeilhofer:
@vonnieda Another question related to this topic: is there anywhere an overview of the dependencies of the different objects in OpenPnP? 

Something like: A Part has an PartId, a hight, a Package and a bottom vision pipeline. 
A Package has optional pads and body dimensions, and is referred by the nozzle compatibility settings. 
...

vonnieda:
@KarlZeilhofer No, there is no documentation on that. It changes quite often, but always in a backwards compatible manner. I feel the best documentation for this is either the code itself or the API documentation at http://openpnp.github.io/openpnp/develop/org/openpnp/model/package-frame.html.

In addition, the machine.xml, parts.xml and packages.xml effectively describe the entire object model for the application. These are used to load the models exactly as listed, so they are a good reference.

PR #622, opened 2017-08-18
Rotated Tray Feeder update pick location when changing Feed Count now also works without executing a feed operation

sp-apertus:
minor addition so you can preview pick location by positioning the camera over the next to be picked component. Before this was only possible after executing a feed operation, now the picklocation updates also when just moving the camera to the pick location


vonnieda:
Looks good to me, thanks @sp-apertus!

PR #626, opened 2017-08-24
Added "placed" checkbox in pick and place job panel

sp-apertus:
# Description
Added a new checkbox to each component to be placed in the "Pick and Place" Tab of a board. 

# Justification
If a component is miss picked or skipped there is currently no way to see these is the component list. Now with the "placed" checkbox its easy to see which parts have been placed and which not. Also components that have successfully been placed will not be placed anymore in the next run.

I know the job processor is currently undergoing a major rewrite so things might already be diffferent behind the scenes - still I consider this is a small addition that would be extremely useful in operation.

# Implementation Details
Tested with nulldriver simulated machine.

mvn test completed successfully

Real world test would be appreciated, I will get to our machine on Monday to test it.

openpnp-placed

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
@@ -301,6 +301,7 @@ public class JobPlacementsPanel extends JPanel {
 
             boardLocation.getBoard().addPlacement(placement);
             tableModel.fireTableDataChanged();
+            placement.setPlaced(false);
             Helpers.selectLastTableRow(table);
         }
     };
--- a/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
@@ -222,6 +222,11 @@ public class JobPlacementsPanel extends JPanel {
             setSideMenu.add(new SetSideAction(side));
         }
         popupMenu.add(setSideMenu);
+        
+        JMenu setPlacedMenu = new JMenu(setPlacedAction);
+        setPlacedMenu.add(new SetPlacedAction(true));
+        setPlacedMenu.add(new SetPlacedAction(false));
+        popupMenu.add(setPlacedMenu);
 
         table.setComponentPopupMenu(popupMenu);
 
@@ -464,6 +469,7 @@ public class JobPlacementsPanel extends JPanel {
         public void actionPerformed(ActionEvent arg0) {
             for (Placement placement : getSelections()) {
                 placement.setType(type);
+                tableModel.fireTableDataChanged();
             }
         }
     };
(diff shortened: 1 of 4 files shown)


KarlZeilhofer:
Thank you for your contribution, seems to be a very useful feature. 
Is the checkbox automatically set on a placement?
Is the state saved in the job xml?
How would you normally use this checkbox?

sp-apertus:
> Is the checkbox automatically set on a placement?
yes:  line 731 in src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java

> Is the state saved in the job xml?
yes

> How would you normally use this checkbox?
Please elaborate further.... 

I will add being able to reset all "placed" checkboxes through the right click menu when you start populating a new board tomorrow.

KarlZeilhofer:
this seems to be very, very useful! Thumbs up for a merge!

sp-apertus:
Thanks :)

sp-apertus:
Right click menu to change "placed" status of multiple lines done.

cri-s:
As I have similar thing for years using drop down instead of checkboxes ,  and it seems    to me that you don't store it inside job.XML but inside board.XML, same as me, to tell you that this is incompatible with Panelization. The reason is that for Panelization, it refers to the same board file and dataset and doing panelisation, I need to copy the board file for each sub-pcb in order to decouple that status field from different PCB inside one panel.
There exist other possibility to do it, of course

vonnieda:
I won't be able to look at this till Monday, but if it is being set on
Board it will not work in a job where multiple copies of the same board are
used. The object is shared, so setting it on one will set it for all.

sp-apertus:
> As I have similar thing for years using drop down instead of checkboxes

Interesting, why didn't you commit it?

> store it inside job.XML but inside board.XML
Right, indeed.

Any suggestions how to fix that?

vonnieda:
Sebastian, I'd ask that you don't spend anymore time on this till I am back
on Monday. This is a big part of the job processor rewrite I've been
working on for months and I've put a lot of thought into how I want to
store placement status. I'm more than happy to collaborate on it, but I am
on vacation right now and can't do that till I get back.

sp-apertus:
Great, lets chat on Monday!

cri-s:
It have send to @vonnieda and discussed with him using email about
more then one time.
The orginal  implementation was from @Ami .and @vonnieda helped to
update the status of placing during job inside table and not only
after have paused/stopped the job.
There was a table update/display issue where i was not able to
understand  the reason
at that time.








2017-08-25 14:42 GMT+02:00, Sebastian Pichelhofer <notifications@github.com>:

vonnieda:
Closing this as it will not be merged in the form it's in now. Discussed with @sp-apertus on IRC.

PR #633, opened 2017-08-31
ReferenceStripFeeder CvPipeline improvements

richard-sim:
# Description
- Changes to the ReferenceStripFeeder default CvPipeline to improve detection accuracy with real-world cameras (works with the simulated camera too). Tested and proven quite reliable on white tape, black tape, and even transparent(!) tape
- Added descriptions for some Hough circle parameters
- Exposed the threshold/detection related parameters from DetectFixedCirclesHough, as only the feature size ought to be fixed
- Added odd-numbers-only check to BlurGaussian and BlurMedian, to avoid an assertion in OpenCV. Matches the assertion (and description) in ThresholdAdaptive for blockSize.
- Auto Setup: The best-fit line of holes for Auto Setup is now taken as the longest line that is within our desired distance ranges rather than the closest, as that would often get influenced by falsely detected circles, and there should only ever be one (long) line of holes anyway. Both works better and makes more sense.
- Auto Setup: The detected holes are now fitted to what we expect their perfect locations/spacing to be along the best-fit line. This stops them being heavily affected by not-so-consistent circle detection in the pipeline, improves the stability of the results, and as a side benefit really highlights when the tape isn't at the same height that the camera's units-per-pixel were configured for (as the resulting holes are spaced at exactly 4mm, and you can see them offset from the detected holes)

# Justification
Michael on the Google Group has been testing the new behaviour, and these changes are necessary to work with his setup (which is pretty typical of the hardware in use by users).

# Instructions for Use
To get the improved ReferenceStripFeeder CvPipeline changes, you will need to reset the pipelines of each ReferenceStripFeeder in your machine. If necessary, changes for tape hole detection for specific hardware and lighting setups can then be made to the pipeline.

# Implementation Details
 Tested on white tape, black tape, and even transparent(!) tape, and the simulated camera.
Yes
N/A
Done.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
@@ -868,12 +868,17 @@ public class ReferenceStripFeederConfigurationWizard extends AbstractConfigurati
         Integer pxMinDiameter = (int) VisionUtils.toPixels(feeder.getHoleDiameterMin(), camera);
         Integer pxMaxDiameter = (int) VisionUtils.toPixels(feeder.getHoleDiameterMax(), camera);
 
-        CvPipeline pipeline = feeder.getPipeline();
-        pipeline.setProperty("camera", camera);
-        pipeline.setProperty("feeder", feeder);
-        pipeline.setProperty("DetectFixedCirclesHough.minDistance", pxMinDistance);
-        pipeline.setProperty("DetectFixedCirclesHough.minDiameter", pxMinDiameter);
-        pipeline.setProperty("DetectFixedCirclesHough.maxDiameter", pxMaxDiameter);
-        return pipeline;
+        try {
+            CvPipeline pipeline = feeder.getPipeline().clone();
+            pipeline.setProperty("camera", camera);
+            pipeline.setProperty("feeder", feeder);
+            pipeline.setProperty("DetectFixedCirclesHough.minDistance", pxMinDistance);
+            pipeline.setProperty("DetectFixedCirclesHough.minDiameter", pxMinDiameter);
+            pipeline.setProperty("DetectFixedCirclesHough.maxDiameter", pxMaxDiameter);
+            return pipeline;
+        }
+        catch (CloneNotSupportedException e) {
+            throw new Error(e);
+        }
     }
 }
--- a/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
@@ -19,8 +19,6 @@
 
 package org.openpnp.machine.reference.feeder.wizards;
 
-import static org.openpnp.vision.FluentCv.pointToLineDistance;
-
 import java.awt.BorderLayout;
 import java.awt.Color;
 import java.awt.event.ActionEvent;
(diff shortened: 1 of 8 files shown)


richard-sim:
I want to wait for Michael to do some more testing before we merge this, but for now wanted others to be able to get a hold of the changes for wider testing/more feedback.

cri-s:
Why not remove FixedCircleHough and integrated it into CircleHough using this code:

		int dist,min,max;
		dist=minDistance==-1?(Integer) pipeline.getProperty("DetectFixedCirclesHough.minDistance"):minDistance;
		min=minDiameter==-1?(Integer) pipeline.getProperty("DetectFixedCirclesHough.minDiameter"):minDiameter;
		max=maxDiameter==-1?(Integer) pipeline.getProperty("DetectFixedCirclesHough.maxDiameter"):maxDiameter;
		 Imgproc.HoughCircles(mat, output, Imgproc.CV_HOUGH_GRADIENT, dp, dist, param1, param2, min / 2, max/ 2);


Inside the pipeline the values must be -1 for dist, min, max in order to use the values supplied by pipeline.

richard-sim:
@vonnieda: OK I'm pretty happy with how this is working now; it appears to be very reliable with what I'm throwing at it, and as such I feel it's ready for prime-time.

@cri-s: There are a couple of reasons for putting the fixed-behaviour logic in its own stage.
* Magic-numbers and special-case behaviour is kept out of the DetectCirclesHough stage. I'd originally implemented this inside DetectCirclesHough, with support for Units, however it was a very wrong decision as far as usability goes.
* It keeps any future changes localized to the one stage to reduce impact elsewhere. This is especially important as I do want to revisit the Units support in CvPipeline.

vonnieda:
Looks good, thanks @richard-sim!

mgrl:
Hi Jason, Hi Richard,

while testing openpnp in eclipse I fell into an issue: I can edit a pipeline in the editor, but when I quit it won't be applied and not be saved. This is exclusive to the pipeline-editor in the feeder tab. Modify a pipeline for a fiducial works as expected.

I'm on branch devel and have the current pull request #639 applied.

vonnieda:
@mgrl @richard-sim Ah, shoot. That's my fault. When I changed the code to clone the pipeline to avoid concurrency issues I forgot that that would cause it not to save. I'll see if I can come up with a decent fix today.

vonnieda:
Okay @mgrl, that bug is fixed and pushed to develop. Try pulling develop down and see if it's working better.

cri-s:
Can you please tell why concurrent pipeline/camera don't work.
Both pipeline and camera at first look seem ok .

2017-09-04 20:43 GMT+02:00, Jason von Nieda <notifications@github.com>:

mgrl:
OK, I'm now on commit 628d0f4e8b404213d6a0f30cdd5eb673da88415e in eclipse. Pipeline-Editor works well again.
On that commit I tried autosetup a couple of times and it worked very well, nearly 100%, ONLY if I add ImageCapture a second time at the beginning of the pipeline to have more settle time. The default pipeline works fine then. Have to say that this is not needed for other vision tasks e.g. fiducials, so its still related to the camera handling in autosetup I would guess.

Now going on to #639 ...

vonnieda:
@cri-s When two threads access the pipeline one can read results from the other since they are both referencing the same results map.

cri-s:
the results map is not defined static, so two class instances should
have two distinct
hashmaps, i'm missing something ?

2017-09-05 0:44 GMT+02:00, Jason von Nieda <notifications@github.com>:

vonnieda:
Using two instances is fine, but the code we are referring to was sharing
one instance.

mgrl:
playing around with the autosetup again I can state, that the error messages from the past went away for me after switchting to the new OpenPnP camera system. I don't need to capture 2 images from the camera any more and autosetup worked in 100% so far. even settle time could be reduced to 150ms.

on the other side I noticed a quite long time, the autosetup routine hangs right after clicking a part, most of the times clicking the second part. it's not always and I can't reproduce... anybody noticed that, too?

PR #639, opened 2017-09-04
Ensure there are no dangling references to BufferedImageCamera

richard-sim:
# Description
Clean up dangling references to the (transient) BufferedImageCamera in ReferenceStripFeederConfigurationWizard:
- Clear references to cameras and feeders in pipeline properties when done
- Added ability to remove the ConfigurationListener from a Camera when we're done with it, so the Configuration doesn't hold onto a reference to the Camera

# Justification
BufferedImageCamera holds onto a BufferedImage, so it's a significant memory leak if anything is still referencing the BufferedImageCamera so it can't be cleaned up by the GC.

# Instructions for Use
N/A

# Implementation Details
Tested by leaving the ReferenceStripFeeder Auto Setup feature running for an extended time and verifying that an OOM situation doesn't occur.
Yes.
Yes. AbstractCamera.release() was added to remove the ConfigurationListener.Adapter from the Configuration, as otherwise it holds onto a reference to the Camera.
Done.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
@@ -587,7 +587,7 @@ public class ReferenceStripFeederConfigurationWizard extends AbstractConfigurati
             return holeLocations;
         }
         finally {
-            pipeline.release();
+            releaseCvPipeline(pipeline);
         }
     }
 
@@ -628,7 +628,8 @@ public class ReferenceStripFeederConfigurationWizard extends AbstractConfigurati
             return showResult;
         }
         finally {
-            pipeline.release();
+            releaseCvPipeline(pipeline);
+            bufferedImageCamera.release();
         }
     }
 
@@ -880,5 +881,12 @@ public class ReferenceStripFeederConfigurationWizard extends AbstractConfigurati
         catch (CloneNotSupportedException e) {
             throw new Error(e);
         }
+    }
+
+    private void releaseCvPipeline(CvPipeline pipeline) {
+        pipeline.setProperty("camera", null);
+        pipeline.setProperty("feeder", null);
+
+        pipeline.release();
     }
 }
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java
@@ -306,6 +306,8 @@ public class ReferenceStripFeeder extends ReferenceFeeder {
             return holeLocation;
         }
         finally {
+            pipeline.setProperty("camera", null);
+            pipeline.setProperty("feeder", null);
             pipeline.release();
         }
     }
--- a/src/main/java/org/openpnp/spi/base/AbstractCamera.java
+++ b/src/main/java/org/openpnp/spi/base/AbstractCamera.java
@@ -40,6 +40,7 @@ public abstract class AbstractCamera extends AbstractModelObject implements Came
     @Attribute(required = false)
     protected long settleTimeMs = 250;
 
+    protected ConfigurationListenerAdapter configurationListener = new ConfigurationListenerAdapter();
     protected Set<ListenerEntry> listeners = Collections.synchronizedSet(new HashSet<>());
 
     protected Head head;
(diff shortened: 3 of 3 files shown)


src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
```diff
@@ -587,7 +587,7 @@ public void run() {
             return holeLocations;
         }
         finally {
-            pipeline.release();
+            releaseCvPipeline(pipeline);
```
vonnieda: I don't think any of this is needed, but I'm open to being convinced I'm wrong. 

Since we're cloning the pipeline, there shouldn't be any references to it held past the time it's used. Once those references are orphaned the entire pipeline, including the properties map and the references to the feeder, camera, etc. will be collected by the GC.

I do agree that the configuration listener stuff below is needed for now, since the configuration listener will keep the camera around. I'm willing to merge that bit temporarily, but long term I want to see two other changes:

1. VisionProvider is on it's way out. It's only used by ReferenceDragFeeder now so if we convert ReferenceDragFeeder either to a pipeline to just use FluentCv we can drop the entire VisionProvider interface and everything that goes along with it.
2. I'm not crazy about the creation of a Camera every time we call the pipeline. I realize it's just a wrapper for an image, but it feels hacky and I think we'd be better off just injecting the image somehow.

richard-sim:
Good points - the releaseCvPipeline was done before the cloning. Now that the pipeline is sometimes not cloned (editing), it may be useful again.

Yeah the BufferedImageCamera is a hack. I like the idea of injecting the image instead - I'm thinking something along the lines of adding support for overriding stages. Pipeline.addStageOverride<T, OverrideType extends CvStage>(OverrideType override) and Pipeline.removeStageOverride<T, OverrideType extends CvStage>(). Then when processing the stages, calls to any stages of type T are replaced by calls to overrideMapping[T].process().

mgrl:
...coming from #633 I then applied the changes proposed in this pull request.
After then nothing worked any more with the same pipeline as used before (err: "no holes detected", "wrong direction"). Reverting the patch, everything is fine again.
Sorry that I can't give further information than that. If you provide me information on how to help further I would do.

```
> <cv-pipeline>
>    <stages>
>       <cv-stage class="org.openpnp.vision.pipeline.stages.ImageCapture" name="original" enabled="true" settle-first="true"/>
>       <cv-stage class="org.openpnp.vision.pipeline.stages.ImageCapture" name="0" enabled="true" settle-first="true"/>
>       <cv-stage class="org.openpnp.vision.pipeline.stages.BlurMedian" name="cleanup-original" enabled="true" kernel-size="5"/>
>       <cv-stage class="org.openpnp.vision.pipeline.stages.ConvertColor" name="gray" enabled="true" conversion="Bgr2Gray"/>
>       <cv-stage class="org.openpnp.vision.pipeline.stages.DetectEdgesCanny" name="find-edges" enabled="true" threshold-1="5.0" threshold-2="15.0"/>
>       <cv-stage class="org.openpnp.vision.pipeline.stages.BlurGaussian" name="predetect-1" enabled="true" kernel-size="5"/>
>       <cv-stage class="org.openpnp.vision.pipeline.stages.BlurMedian" name="predetect-2" enabled="false" kernel-size="7"/>
>       <cv-stage class="org.openpnp.vision.pipeline.stages.DetectFixedCirclesHough" name="results" enabled="true" dp="1.0" param-1="25.0" param-2="20.0"/>
>       <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRecall" name="recalled" enabled="true" image-stage-name="original"/>
>       <cv-stage class="org.openpnp.vision.pipeline.stages.DrawCircles" name="display" enabled="true" circles-stage-name="results" thickness="1">
>          <color r="255" g="0" b="0" a="255"/>
>       </cv-stage>
>    </stages>
> </cv-pipeline>
```

vonnieda:
@richard-sim I'm suggest we make the following changes:
1. Leave in the setting camera and feeder to null.
2. Instead of all the configuration listener changes, we can just change the configuration class to use weak references to the listeners. This way when they go out of scope the reference will break and the configuration won't hold on to the cameras. I can't think of any reason we'd ever want the configuration to hang on to an object once it's otherwise ready for GC.

vonnieda:
Ping @richard-sim, what do you think of my suggested changes above?

vonnieda:
Closing until this can be revisited. The PR's code no longer applies to the current situation as discussed above.

Issue #402, opened 2017-01-04
Warn when starting a job and board Z is 0


vonnieda:
Peter Betz suggest that we warn the user when starting a job if there are board locations with Z = 0. Sounds like a good idea!

PR #640, opened 2017-09-04
throw error if board z height is still set to 0

sp-apertus:
deals with: https://github.com/openpnp/openpnp/issues/402

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -279,6 +279,12 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
                 continue;
             }
             
+            // Check if board z-height is still 0 
+            if ((Double.compare(boardLocation.getLocation().getZ(), 0.0)) == 0) {
+            	throw new Exception(String.format("Board z-height is set to 0"));
+            }
+            
+            
             // Check for ID duplicates - throw error if any are found
             HashSet<String> idlist = new HashSet<String>();
             for (Placement placement : boardLocation.getBoard().getPlacements()) {


vonnieda:
The original ticket indicates that this should be a warning, and I think that should remain true. 0 is a valid value for a particular type of setup and I don't want to stop users from being able to use it. I think it would be better to make this a status value + color, similar to how we handle missing feeder, missing part, etc. Alternately, when the user presses the start button, we could pop a dialog and let them continue if they choose.

sp-apertus:
Turned the error into a warning popup now.

For adding a job status column deeper going changes will be required. Will investigate.

sp-apertus:
I added a status column now to the job panel.

We should discuss though what information could be shown there beside the warning about Z-Height being zero.

It could summarize placement status errors, like if this board has a placement that has status "missing feeder" the board could also list any errors/warnings related to this boards placements. But then again its not possible to list multiple errors/warning at the same time with this single line entry so we would need to create a priority list....

Thoughts?

PS: I need a little pointer how to update the BoardStatus display when the z value gets changed.

src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
```diff
@@ -279,6 +281,12 @@ protected void doPreFlight() throws Exception {
                 continue;
             }
             
+            // Check if board z-height is still 0
```
vonnieda: The reference classes cannot refer to the GUI classes as this breaks tests and locks the job processor to this specific UI. Suggest just removing this block entirely. If you really feel a popup warning is necessary it can be done in the JobPanel on job start.

src/main/java/org/openpnp/gui/tablemodel/BoardLocationsTableModel.java
```diff
@@ -33,13 +33,19 @@
 
 public class BoardLocationsTableModel extends AbstractTableModel {
     private final Configuration configuration;
-
+    public enum BoardStatus {
+        Ready,
+        MissingPart,
```
vonnieda: MissingPart, MissingFeeder and ZeroPartHeight aren't used here as far as I can tell. They should be removed from the enum if they aren't used.

src/main/java/org/openpnp/gui/tablemodel/BoardLocationsTableModel.java
```diff
@@ -51,11 +57,18 @@ public void setJob(Job job) {
...
         fireTableDataChanged();
     }
-
+    
     public Job getJob() {
         return job;
     }
 
+    public BoardStatus getBoardStatus(int index) {
+    	if (Double.compare(job.getBoardLocations().get(index).getLocation().getLengthZ().getValue(), 0.0) == 0) {
```
vonnieda: This can be simplified to (job.getBoardLocations().get(index).getZ() == 0), no need for the complex compare. Not a critical change though.

src/main/java/org/openpnp/gui/tablemodel/BoardLocationsTableModel.java
sp-apertus: Right, will add just "Error" instead indicating that there are unresolved issues to deal with in the future.

src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
sp-apertus: done

src/main/java/org/openpnp/gui/tablemodel/BoardLocationsTableModel.java
sp-apertus: done

sp-apertus:
Suggestions incorporated.

One thing I still need help with is how to update a table rows status when any column was changed...

sp-apertus:
I think I found a solution to the table row updating, will push shortly

sp-apertus:
Last remaining thing: how to update/invalidate BoardLocations Table when Placements Table has been altered?

sp-apertus:
Good progress was made, thanks for the tip with the PropertyChangeListener Jason.

The only thing missing now is to make this work with the autopanelization feature.

sp-apertus:
another small feature added:
display yes/no dialog when starting a job and if all placements are marked "placed" to reset placements and start job

src/main/java/org/openpnp/model/Placement.java
```diff
@@ -99,7 +109,40 @@ public void setPart(Part part) {
...
+            }
+        }
+        setPlacementStatus(Placement.Status.Ready);
+        return Placement.Status.Ready;
+    }
 
+    public void setPlacementStatus(Status PlacementStatus) {
+    	Status oldValue = this.placementStatus;
+        this.placementStatus = PlacementStatus;
+        firePropertyChange("PlacementStatus", oldValue, PlacementStatus);
```
vonnieda: Property name should be placementStatus, not PlacementStatus, to adhere to Java naming standards. Likewise the incoming parameter should be named the same.

src/main/java/org/openpnp/model/BoardLocation.java
```diff
@@ -98,6 +106,39 @@ public void setSide(Side side) {
...
     }
 
+    public void setBoardStatus(BoardStatus boardstatus) {
+        Object oldValue = this.boardStatus;
+        this.boardStatus = boardstatus;
+        firePropertyChange("boardStatus", oldValue, boardStatus);
+    }
+
+    public void initBoardStatus() {
+    	for (Placement placement : getBoard().getPlacements()) {
```
vonnieda: Suggest adding a break for each case. No need to iterate through all the placements once you've found an error.

src/main/java/org/openpnp/gui/JobPanel.java
```diff
@@ -768,6 +810,10 @@ public boolean accept(File dir, String name) {
                 Job job = configuration.loadJob(file);
                 setJob(job);
                 addRecentJob(file);
+                
+                for (BoardLocation location : job.getBoardLocations()) {
```
vonnieda: I believe this will cause the listener to be added every time refresh is called, and it will result in it being called once for every time it has been added, eventually slowing things down significantly. See https://docs.oracle.com/javase/7/docs/api/java/beans/PropertyChangeSupport.html#addPropertyChangeListener(java.beans.PropertyChangeListener).

src/main/java/org/openpnp/model/Placement.java
sp-apertus: done

src/main/java/org/openpnp/model/BoardLocation.java
sp-apertus: done

src/main/java/org/openpnp/gui/JobPanel.java
sp-apertus: I make sure now to remove any listeners before adding the new ones. First thoughtthat would be a dirty way to do it but https://docs.oracle.com/javase/7/docs/api/java/beans/PropertyChangeSupport.html#addPropertyChangeListener(java.beans.PropertyChangeListener) says:
"If listener is null, or was never added, no exception is thrown and no action is taken." so that should not be a problem

sp-apertus:
made a quick screencapture to show the changes: https://www.youtube.com/watch?v=Mxih6XtEeR0

cri-s:
It is possible to change the code in order to not hardcode 0 but instead use getSafeZ from primary Nozzle
for displaying the error.

sp-apertus:
> It is possible to change the code in order to not hardcode 0 but instead use getSafeZ from primary Nozzle for displaying the error.

Good idea! Will look into it.

sp-apertus:
> It is possible to change the code in order to not hardcode 0 but instead use getSafeZ from primary Nozzle for displaying the error.

interesting idea, I will look into it!

sp-apertus:
it was my mistake that this PR has grown beyond the original title, I just saw things that I deemed worth improving as well...

I will break the improvements up into one per PR if I get the time...

PR #641, opened 2017-09-06
Editable feeder names

mgrl:
# Description
Added a textfield in the general settings-panel in the feeder wizards to edit the name of feeder via GUI instead of machine.xml.

# Justification
Users can easily distinct between enumerated feeder if they are able to edit the default name.

# Instructions for Use
It's easy: Just choose a name, put it into the textfield and click apply.

# Implementation Details
Started OpenPnP in Eclipse and tried to modify every feeders name (except SlotFeeder, didn't want to mess with the internal names that are already present).
Yes.
Have not.
Don't know how to do in Eclipse, tested manually.


vonnieda:
@mgrl You can already change the name of a feeder by editing the column (double click) in the Feeders tab. This seems to duplicate that functionality. Is this what you intended, or just not aware of the column edit capability?

mgrl:
Haha, not expected the lists to be that interactive. Maybe its worth to mention in the feeder setup or anywhere else where applicable? So consider this to be needless.
The only line you could take over would be that in https://github.com/openpnp/openpnp/pull/641/commits/70a4d60b08f38add360a38cef50a75371c42e07d to have an error message with the name of the feeder instead "null".

PR #642, opened 2017-09-07
Feature - Add peel and stuck drag pin sensor to reference drag feeder

johngrabner:
# Read This First

will not be accepted.

# Description
It is common for the drag pin to get stuck in the tape. Machines like the TVM802 incorporate a sensor to determine if the drag pin has retracted successfully.  An optional actuator can now be specified for this functionality.  

Commercial runs for stuffing boards require a lot of components.  Manually peeling the tape from component reels is not practical.  A stepper for friction wheels allows for a very low cost implementation of automatic peeling.  This update add the functionality of being able to specify an actuator that will peel the tape.

Machines with multiple nozzles may have more than one pump.  TVM802 is such a machine.  The update allows Pump on / off actuator to be specified on a per nozzle basis rather than turning both pumps on when only one is needed.

A minor bug was also corrected.  Pin actuator may have their name changed in the UI.  In machines like TVM802B where there are 48 feeders, changing the names makes a lot of sense.  A bug required the name to be same as the id.  This bug was fixed.

# Justification
OpenPnP has a much cleaner UI than QiHe native software.  If companies like QiHe are ever to convert to OpenPnP, OpenPnP needs to have parity on critical features.

# Instructions for Use
I will update the Wiki page with a full description and images of the drag feeder.

In short, 

1) specify an optional pin sensor.  If pin sensor detects a stuck pin, it will trigger minor corrective, if corrective action is not successful, it will raise an error.
2) specify an actuator for the peel action (extrude) and a multiplier action, and this action will be triggered anytime a tap is dragged.

# Implementation Details

a) configured a drag feeder before my change, upgraded code to this version and validate drag feeder does what it did before (ie: no pin sensor and no peel stepper)
b) configured a drag feeder with pin sensor.  Cause pin to be stuck down and validated detection.  Caused pin to be stuck up and validated detection.  Ran lots of components through feeder and noticed on occasion recovery action triggering.
c) configured an peel actuator and validated the peel tape function operates reliably.
d) deleted configuration and ran the pnp-test.job.xml

To the best of my ability.  Please let me know if I deviated.


vonnieda:
One more comment, regarding the pump changes. The pump commands are only for pumps shared by multiple nozzles. If you have separate pumps you can just use PICK_COMMAND and PLACE_COMMAND to turn the pumps on and off. Or do you mean you have multiple shared pumps? e.g. 4 nozzles, 2 on each pump?

johngrabner:
Hi Jason,

Current drag has a multiplier presumably to keep parts from falling of the tape while dragging.  So if tape advances at rate x mm/sec, peeling can't be any speed different from this x since a slower speed will allow tape cover to stay on while tape advancing and getting stuck on tape retainer. Similarly, a faster speed of the peel will result cover either stretching or peel would slip.

Small components will likely be dragged at rate different from large component. So speed is essential.

The peel stepper is connected to collection wheels that are friction  connected to a shaft. So it needs to start moving before drag function and continue for a short time after the drag. So either a fix time, fix distance or a multiplier if distance. Without, the cover tape under peeling will accumute with every feed. So also essential.

Extruder double is easily done.  I just thought it obfuscates the functionality. But no big deal changing.

I had not understand the mvn test, i thougth the was the test job.

We can talk tomorrow.



Sent via the Samsung Galaxy S7 edge, an AT&T 4G LTE smartphone


-------- Original message --------

johngrabner:
I supposed overloading pick / place functionality would work if every time the pump needs to be ON a pick occurs and likewise a place occurs every time the pump would be turned OFf.

If that is a good approach, why even have a pump on/off?

Pump takes time to build pressure/vacumm. Do you not envision turning pump on x ms before pick and not turning off between picks if pause is not activated  ?

Discard does a place ?

We can talk tomorrow.

Sent via the Samsung Galaxy S7 edge, an AT&T 4G LTE smartphone

-------- Original message --------

vonnieda:
@johngrabner The PUMP_ON and PUMP_OFF commands are specifically for systems where one pump runs more than one nozzle with solenoids deciding where the air goes. It includes logic such that if both nozzles are carrying parts, placing one will not turn off the pump. So, a two nozzle sequence looks like this:

PUMP_ON
N1 PICK_COMMAND
N2 PICK_COMMAND
N1 PLACE_COMMAND
N2 PLACE_COMMAND
PUMP_OFF

Yes, discard does a place. You can think of pick as "vacuum on" and place as "vacuum off".

If you describe your machine's plumbing to me I can tell you how to best use the current functionality, or if changes are needed.

johngrabner:
here is my plumbing.



________________________________

johngrabner:
slide 5 - controls

vonnieda:
@johngrabner Based on the feedback in the mailing list thread, what are your thoughts? Do you want to make any changes, or maybe make this it's own feeder type?

Here is where I stand on the major changes:
1. The pump change is fine. It doesn't harm anything and it will address your issue. 
2. I'm not comfortable merging the extrude() change. I think it is far too specific to this use case to be in the interface.
3. Based on user feedback, I feel the speed is overkill. By removing it we also remove the need for 
the extrude() change above. We can just use distance. I feel this is supported by the responses in the thread.

Let me know what you think.

Reference thread: https://groups.google.com/d/msgid/openpnp/26f5eef9-af90-43db-9649-f8ca5833370b%40googlegroups.com?utm_medium=email&utm_source=footer

johngrabner:
Feedback convinced me that my changes were an overkill for the openpnp.

I'll keep my changes in my branch.



Sent via the Samsung Galaxy S7 edge, an AT&T 4G LTE smartphone


-------- Original message --------

vonnieda:
Thanks @johngrabner, will close this then.

PR #656, opened 2017-10-05
show warning pop up to reset "placed" status if starting job with all placed

sp-apertus:
show warning pop up to reset "placed" status if starting job with all components marked placed already (fiducials and ignored placements are ignored)

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -1001,22 +1001,8 @@ public class JobPanel extends JPanel {
         }
 
         @Override
-        public void actionPerformed(ActionEvent arg0) {
-        	Boolean allPlaced = true;
-        	
-        	boardloop:
-        	for (BoardLocation boardLocation : job.getBoardLocations()) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/openpnp/gui/JobPanel.java
```diff
@@ -1001,9 +1002,38 @@ public void populatePanelSettingsIntoBoardLocations() {
 
         @Override
         public void actionPerformed(ActionEvent arg0) {
-            UiUtils.messageBoxOnException(() -> {
-                fsm.send(Message.StartOrPause);
-            });
+        	Boolean allPlaced = true;
+        	
+        	boardloop:
```
vonnieda: Please break this out into a function - something like `boolean isAllPlaced()`, or you could make it checkAllPlaced and have it perform the prompt as well. As this code stands now it buries the actual intent of the action (start the job) with the check.

sp-apertus:
good idea, done!

vonnieda:
Thanks @sp-apertus, merged with a few minor changes. Removed a bit of duplicated code and improved the message a bit.

PR #657, opened 2017-10-05
Update row immediately when changing part or placement type and not only when clicking on another row

sp-apertus:


vonnieda:
Look good, thanks @sp-apertus!

PR #661, opened 2017-10-11
Offset for fiducial recognition configurable per part

mgrl:
# Description
Watching videos of other PnP software I noticed there was a configuration for offset when doing vision tasks on fiducials. Using an offset would be helpful if the fids to be detected are very glossy or mirroring. The camera is mirrored in the fiducials if centered over them. See images attached. In this case, the fids would be recognized better if moved about 4mm off center.

fiducial_centered
fiducial_offcenter

# Justification
Everyone having cheap glossy pcbs could profit by the feature.

# Instructions for Use
In current implementation the configured offset is added to the actual fiducial's x-location. Configure a value in the part's referenceFiducialLocator tab and click apply. the next fiducial recognition is made with respect to the offset.
screenshot

# Implementation Details
with simulated machine and eagle demo-board.
no changes there.
JUnit tests passed.

# Notes
This patch works, but needs cleanup for sure. So this pull request is just for discussion as there are some decisions to be made / questions left open for me.
-Would you think it's valuable for OpenPnP?
-Is it needed to be configured per part? Or per machine?
-The offset-value should be validated

Would like to hear feedback.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
@@ -180,10 +180,11 @@ public class ReferenceFiducialLocator implements FiducialLocator {
                     pkg.getId()));
         }
 
-        //add offset to real fiducial's location
-        Location offsetLocation = location.add(part.getOffsetVision());
+        //add offset to fiducial's real location
+        Location offsetVision = part.getOffsetVision();
+        Location offsetLocation = location.add(offsetVision);
         
-        Logger.debug("Looking for {} at {} including {} offset", part.getId(), offsetLocation, part.getOffsetVision());
+        Logger.debug("Looking for {} at {} plus {} vision offset", part.getId(), location, offsetVision);
         MovableUtils.moveToLocationAtSafeZ(camera, offsetLocation);
 
         PartSettings partSettings = getPartSettings(part);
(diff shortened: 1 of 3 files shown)


vonnieda:
Hi @mgrl, this sounds like a nice improvement. Here are my thoughts:

* The offset should work like the pipeline works, in that there is a default on the fiducial locator object itself, in Machine Setup -> Vision -> Fiducial Locator and then the per-part one can override it.
* The offset should be a Location, or at least an X,Y since people may have different lighting conditions that require the offset to be in Y.
* I would suggest testing this with a real machine and camera before going much further. At the least it will require lens calibration to not introduce additional error. I'd like to hear that it actually improves things in the real world before we move forward.

mgrl:
hi @vonnieda, okay will try to implement as you suggest, but test it with my machine next.
- X and Y makes sense of course.
- A default value, active for all parts might lead to problems if the part is near as big as the FOV. Would prefer opt-in instead of opt-out.
- another variant could be to only apply an offset if the fiducial to be detected lays within the area around the center marked as "bad". the size of the area is defined by the offset.

vonnieda:
@mgrl Good point about the default value. Forget I said anything :) 

Let's keep it simple for now. Just add a field for Y and test on a real machine and call it good.

mgrl:
In real it works, see video. Unfortunately on my white PCB (see above) the fiducial has no clearance around it so there is no chance to recognize it any better. the contrast to the white silkscreen is too low. but that is another problem and will not happen again. learning by doing...

What I observed in tryout:
- the camera units have to be quite exact to have best results.
- dialing in 3 times is sometimes not enough in my setup (this is not related to the offset fiducials, but now that I know how to, I think about making it configurable and additionally calculating an average over the dial-in which could give a better resolution)
- the parts.xml is now a little overloaded with stuff... maybe two attributes would be sufficient instead of a location?

here is a short video with using an offset:
[fiducial-offset.zip](https://github.com/openpnp/openpnp/files/1377023/fiducial-offset.zip)

cri-s:
I think this is the wrong method and i suggest to do it inside vision stage and not here globally.
Globally it could work if no visiual pipeline was present, but with visual pipeline, things changes.
As example for fiducials, it is important to do maskCircle as example with 100 pixel  , approx 3,5mm
diameter. If offsetting the image, this is impossible to do. Further there is code for sorting starting
at image center for selecting the correct match.
Inside stage  it is possible to move nozzle the offset value, call ImageCapture stage,
travel back and crop image in order that the fiducial is at center of result image as it was when no
offset was applied.
I understand that if one don't know how to work with visual pipelines it is difficult to work with
this fiducials that are out of specs.
res1
res2
As you see, with a specific pipeline it can be detected.
It works better with dedicated stage but that is not really necessary.
res3

Unfortunatly that bad lightning gives false readings on other visual code and it is better to adress the
root of cause instead of adding code to fix errors of bad lightning, or averaging random results like #662 
hoping it gives enought confidence.

vonnieda:
I agree with @cri-s. I think this is a fix for a setup problem and should be addressed with improved lighting and tuning of the pipeline.

mgrl:
okay, so no vision-offset for everyone ;) Just wanted to add where I one day found and now found again the information, how others do it with offsetting also on commercial machines: http://www.ladyada.net/wiki/mdcpickandplace/fiducials#offsetting

KarlZeilhofer:
What I want to throw in: We still have no adoption of the pixel resolution with different z-levels. So this feature would most likly add systematic errors.
Oherwise it could be helpful, I think.

cri-s:
@micheal
That is really specific to Luna PnP, as it utilized the diffuses
lights from annother camera.
On lightning used by openpnp, the light pattern at camera center is
good and outside that
center is really bad. You can see it from videos from @ the only exception
is when the camera optical axis is really bad. In this case, it is
better to resolve the root of problem adjusting the camera mounting.
As example https://www.youtube.com/watch?v=aeW66c9oN5Y at 4:55
As you see, because it is not diffused light , but more direct, it
create more problems as solves  something.



2017-10-29 5:12 GMT+01:00, Karl Zeilhofer <notifications@github.com>:

cri-s:
If you look further the images, only the center of image is sharp,
outside the center it is
unsharp and optically blurred. Check as example the resitor marking.
Further the recognized fiducial have errors as the blurring of
fiducial is only partial and as result it don't recognize the correct
fiducials center.

2017-10-29 5:49 GMT+01:00, Cri S <phone.cri@gmail.com>:

PR #662, opened 2017-10-11
Configurable repetition of fiducial recognition and averaging result

mgrl:
# Description
While fiddling with the machine I found it would be nice if the number of "dial ins" while recognizing fiducials would be configurable. Additionally some type of averaging as option would be great. With just 3 repetitions and without averaging it happens sometimes, the machine goes back and forth in a very small amount but never hitting the exact center... For my setup this improved the accuracy of detection (used 6 matches instead default 3). 

Maybe how the average is calculated could be thought over: Is it better to calculate a moving average while dialing in or just apply the average in the end, as it is implemented now.

# Justification
Giving the user the ability to configure the number of repetitions and averaging enables him with little effort to check whether the accuracy can be enhanced further or not. Anyway it defaults to the detection algorithm used until today, that is 3 repetitions and no averaging.
Having the idea of the planned feature "affine transformation" in my mind, a very accurate detection would be needed.

# Instructions for Use
See screenshot where settings were added:
fidusetup
after configuring start a fiducial scan on an board.

# Implementation Details
with simulated machine and eagle demo-board as well as my machine.
no changes there.
JUnit tests passed.

Any chance to have that within OpenPnP? Otherwise it would be no problem to cancel the pull request. 
kind regards.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
@@ -241,8 +241,7 @@ public class ReferenceFiducialLocator implements FiducialLocator {
             location = locations.get(0);
             
             if (i > 0) {
-            	//hold all matches to calculate the average later
-            	//keep a list, but don't keep the first match, since its probably most off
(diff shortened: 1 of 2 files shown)


vonnieda:
Merged, thank you @mgrl!

PR #666, opened 2017-10-18
fixed reference loose part feeder

KarlZeilhofer:
* don't rotate the nozzle between the 3 retries. 
* constant 90° rotation (width >= length)

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceLoosePartFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceLoosePartFeeder.java
@@ -62,7 +62,7 @@ public class ReferenceLoosePartFeeder extends ReferenceFeeder {
         MovableUtils.moveToLocationAtSafeZ(camera, location);
         for (int i = 0; i < 3; i++) {
             pickLocation = getPickLocation(camera, nozzle);
-            camera.moveTo(pickLocation);
+            camera.moveTo(pickLocation.derive(null, null, null, 0.0));
         }
     }
 
@@ -89,7 +89,12 @@ public class ReferenceLoosePartFeeder extends ReferenceFeeder {
         Location location = VisionUtils.getPixelLocation(camera, result.center.x, result.center.y);
         // Get the result's Location
         // Update the location with the result's rotation
-        location = location.derive(null, null, null, result.angle);
+    	double angleCorrection=0;
+    	if(result.size.width < result.size.height) {
+    		angleCorrection = 90;
+    	}
+
+        location = location.derive(null, null, null, -(result.angle+angleCorrection + location.getRotation()));
         // Update the location with the correct Z, which is the configured Location's Z
         // plus the part height.
         location =


vonnieda:
Hi @KarlZeilhofer, a few comments:

* **don't rotate the nozzle between the 3 retries:** This should be handled in the driver. If your nozzle is rotating when the camera is commanded to rotate then you have your axes setup incorrectly in GcodeDriver. See https://github.com/openpnp/openpnp/wiki/FAQ#my-nozzle-moves-in-z-when-the-camera-is-moved - the situation is the same for both the Z and Rotation axes.
* **constant 90° rotation (width >= length):** Can you explain what this does please?

KarlZeilhofer:
## Skip rotation
Thanks @vonnieda , I've added this sentence to the wiki page:
```This solve also a limited speed of moves in the x,y,z axis, if the camera is commanded to turn and the nozzle rotates as well (rather slowly). ``` and changed the title a bit. 
I've read this title, and didn't thought, this is a problem, which I have :)

But nevertheless it would help others with this sub-optimal setup I have. Does it help anyone, if the camera rotates on the retries?

## Constant rotation
The need for this code was already in a discussion end of August in #624 . It rotates the rotated rect always into landscape orientation, instead of randomly mixed landscape/portrait. This is very, very useful for resistors and capacitors.

File: src/main/java/org/openpnp/machine/reference/feeder/ReferenceLoosePartFeeder.java [hunk 1]
```diff
@@ -89,7 +89,12 @@ private Location getPickLocation(Camera camera, Nozzle nozzle) throws Exception
         Location location = VisionUtils.getPixelLocation(camera, result.center.x, result.center.y);
         // Get the result's Location
         // Update the location with the result's rotation
-        location = location.derive(null, null, null, result.angle);
+    	double angleCorrection=0;
+    	if(result.size.width < result.size.height) {
+    		angleCorrection = 90;
+    	}
+
+        location = location.derive(null, null, null, -(result.angle+angleCorrection + location.getRotation()));
```
vonnieda: Hi @KarlZeilhofer, can you explain why you are negating the angle here? I refer to the `-` before `(result.angle...`

KarlZeilhofer: Good question, which seems I cannot explain. I tried it without the negation before, but it didn't work.
I'll try to come back with a proper explanation...

vonnieda: I figured it out. It actually fixes a bug that wasn't previously know. The angle sign is inverted between OpenCV and OpenPnP, so it needs to be inverted. Most of the other vision operations already do this.

vonnieda:
This is merged in https://github.com/openpnp/openpnp/commit/9c33ac8ff2c25945c129c7a6408c85fe0e497125

KarlZeilhofer:
great to hear!

src/main/java/org/openpnp/machine/reference/feeder/ReferenceLoosePartFeeder.java
KarlZeilhofer: Thank you, @vonnieda !
Still I wonder why this hasn't noticed anyone before. Wasn't the ReferenceLoosePartFeeder not properly tested on a real machine?

vonnieda: @KarlZeilhofer This code was released unfinished. People wanted to use it and modify it, so I merged it. That's why there are big red letters that tell you to use at your own risk in the wizard :)

vonnieda:
Okay, here is how things are looking now:

screen shot 2017-10-25 at 1 15 36 pm

This is with the addition of two new Stages. They are Add and OrientRotatedRects. Add just adds two images together, and OrientRotatedRects ensures that RotatedRects are laying flat. 

Here's the pipeline:

```
<cv-pipeline>
   <stages>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRead" name="4" enabled="true" file="/Users/jason/Desktop/Head_2017-10-25_10.49.18.838.png"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageCapture" name="9" enabled="true" settle-first="true"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ConvertColor" name="2" enabled="true" conversion="Bgr2Gray"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.Threshold" name="1" enabled="true" threshold="200" auto="false" invert="false"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRecall" name="7" enabled="true" image-stage-name="2"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.Threshold" name="11" enabled="true" threshold="120" auto="false" invert="true"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.Add" name="12" enabled="true" first-stage-name="1" second-stage-name="11"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.BlurMedian" name="5" enabled="true" kernel-size="9"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.FindContours" name="3" enabled="true" retrieval-mode="External" approximation-method="Simple"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.FilterContours" name="6" enabled="true" contours-stage-name="3" min-area="1000.0" max-area="2500.0"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.MinAreaRectContours" name="13" enabled="true" contours-stage-name="6"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.OrientRotatedRects" name="results" enabled="true" rotated-rects-stage-name="13"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRecall" name="8" enabled="true" image-stage-name="9"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DrawContours" name="0" enabled="true" contours-stage-name="3" thickness="2" index="-1">
      </cv-stage>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DrawRotatedRects" name="10" enabled="true" rotated-rects-stage-name="results" thickness="2" draw-rect-center="false" rect-center-radius="20" show-orientation="true">
         <color r="51" g="255" b="0" a="255"/>
      </cv-stage>
   </stages>
</cv-pipeline>
```

I was playing with this last night and it came together nicely. The pipeline uses two thresholds. The first finds highlights, which are electrodes, and the second finds lowlights, which are part bodies. Adding the two images gets us a good segmentation of the parts. By doing it this way we are able to filter out parts that are upside down.

I'll do some more testing on this tonight and then release it. It should cause no changes to any existing functionality.

KarlZeilhofer:
Sounds and looks perfect! 

I wonder why this is called add? Shouldn't there be simple logic operations like and, or, exor, negate? They are very basic, universally usable, intuitive and should do this job too.

KarlZeilhofer:
What about parts that should be placed in portrait orientation?

I see 2 ways for that.
1. Set the feeder rotation to 90°
2. Add a checkbox to the OrientRotatedRects stage named portrait, which is default false.

About the 1st one I'm nit quite shure, if it really works.
For now I have no demand for portrait part orientations, but limiting to landscape only seems to be too constrained.

vonnieda:
The function is called Add in OpenCV, so it's just a thin wrapper around
that.
https://docs.opencv.org/2.4/modules/core/doc/operations_on_arrays.html?highlight=addweighted#add

In this case it's effectively the same as an OR since the images are binary.

I've already added a selection of Landscape or Portrait to
OrientRotatedRects.

This code should be released tonight. Last thing I need to do is figure out
whether or not we should be negating angles. Need to do some further
testing.

Jason

vonnieda:
@KarlZeilhofer @cri-s The updated code is now released. See https://github.com/openpnp/openpnp/commit/0d482c113a1e720d9c4aa410aa8d8791cd44488d for details. In brief:

1. Removed the angle negation.
2. Removed the hard coded angle correction.
3. Added new OrientRotatedRects stage which does the two above tasks.
4. Updated the default pipeline to a new one I've been working on. It includes the above changes and an improvement to detecting rectangular components.

PR #670, opened 2017-10-22
Board Z height error

sp-apertus:
Set Z height of newly added board to NaN.

Throw error pop up message if the value is not changed before running a job.


vonnieda:
Tested this for merge, but ran into a big annoyance. This will cause every existing job in the wild to warn users next time they run it, even though they have surely set their Z heights.

In addition, it's popping the dialog multiple times with the same error message for jobs with multiple boards.

I don't think this is a feature we need in OpenPnP. I suggest you just implement it as a scripting event (Job.Starting) for your machine. This way you get the functionality you want without adding a nag for all existing users. If you want to post the resulting script, maybe others will find it useful.

KarlZeilhofer:
Perhaps this could be improved? I have not looked into the code, but that feature would have helped me multible times in the past.

What about not using NaN but 0.000? If one wants to use this value, he could easyly set it to 0.001 without heaving real issues.

vonnieda:
@KarlZeilhofer As I mentioned above, I think this should be done with a script event. It's just a couple liner. This would be a good way to proof this feature for a bit before merging it into the main codebase.

sp-apertus:
Sorry, very busy currently so I cant work on this immediately. 

The NaN approach was implemented (https://github.com/openpnp/openpnp/pull/670/commits/2aaf8f557d3a1022070a11dc9ef7b37a708c1c6f) but dropped after vonnieda advised against it: vonnieda: "NaN is off limits, sorry. It's currently used to do some weird driver trickery, and it has caused a lot of problems. One of my tasks is to remove every reference of it from the app. I'm not gonna add a new one that will just need to be removed. Maybe just a boolean property on BoardLocation? Like BoardLocation.zSet?"

Issue #235, opened 2016-02-27
Nozzle Tip Runout Calibration


vonnieda:
I'm noticing a lot of runout on my nozzles and others are too. This severely degrades the performance bottom vision ( #104 ) along with general placement. We need to be able to compensate for runout by checking the runout with the up looking camera and interpolating X / Y movements.

Code for this was submitted a long time ago in #57 but it was not merged due to the complexities of the change. That code can be used as a reference, along with the good circle finding we now have throughout the system.

My initial thought on this is:
- Whenever a Nozzle is commanded to move, if it has not been calibrated it should calibrate first.
- Nozzle tip change should reset the calibration status.
- It should be possible to enable / disable calibration through the UI.
- I think most of the changes can be done by branching off of ReferenceNozzle.moveTo.

cri-s:
>Whenever a Nozzle is commanded to move, if it has not been calibrated it should calibrate first.
It should be calibrated using component, checking nozzle only have some issues.
>Nozzle tip change should reset the calibration status
This is not true, if automaitc changer is used, as in this case it is unload and load using rotation 0.
It's thinkable to use red dots/color, and filter that out in order to check if nozzle position is ok or not,
in the event nozzle was removed manually.

vonnieda:
This code is now in and working, but it's not yet documented at all. Leaving this ticket open until that's done.

vonnieda:
Note: Previous comment said this is working, but that may not be the case. More testing is needed, along with documentation and any fixes required.

vonnieda:
See comments from @cri-s on #337 for potential issues in the current implementation.

vonnieda:
Also see comments on #336 for additional issues related to angle and rotation.

vonnieda:
This feature has temporarily been removed from the configuration UI since it is not working correctly and was causing confusion. See https://github.com/openpnp/openpnp/commit/5f586166bb74a81b11a78c44a15909ebc81b2bf2

vonnieda:
@markmaker The issue was related to how OpenPnP was reading coordinates from the nozzles when the compensation was applied with regards to bottom vision. Basically, there is a stack of transforms that were being applied and it seems like some were either incorrect or backwards. The comments were not particularly clear on the error, so I think it's best that someone interested in this feature dig it out and do some testing.

ldpgh:
ReferenceNozzleTip.java (develop-branac @ line 456 ... it seems to be divison-by-0
1. develop-branch
[...]

PR #671, opened 2017-10-23
ReferenceCamera: adding support for auto centering to nozzle

machinekoder:
# Description
Adds automatic object detection via a OpenCV pipeline to the camera wizard.

# Justification
This feature is intended to be used for calibrating the bottom vision position. Calibrating the bottom vision by hand is a cumbersome task that may be repeated at each machine startup (if your end stops aren't precise). This step is easy to automate.

In theory, this feature can be useful for other things as well.

# Instructions for Use
* Select a camera in the machine configuration.
* Navigate to "Camera Specific" > "Center Detected Object to Camera"
* Edit the Pipeline for your needs.
image
* Press "Detect and Center" to auto center on the detected object.
image

# Implementation Details
The feature was detected on a real machine with the bottom vision to detect the nozzle position.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
@@ -25,12 +25,15 @@ import java.awt.event.ActionEvent;
 import java.awt.image.BufferedImage;
 import java.io.IOException;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import javax.swing.AbstractAction;
 import javax.swing.Action;
 import javax.swing.JOptionPane;
 
+import org.apache.commons.io.IOUtils;
 import org.opencv.core.Core;
 import org.opencv.core.CvType;
 import org.opencv.core.Mat;
@@ -38,6 +41,7 @@ import org.opencv.core.Point;
 import org.opencv.core.Rect;
 import org.opencv.core.RotatedRect;
 import org.opencv.core.Size;
+import org.opencv.features2d.KeyPoint;
 import org.opencv.imgproc.Imgproc;
 import org.openpnp.gui.MainFrame;
 import org.openpnp.gui.support.Icons;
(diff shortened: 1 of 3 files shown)


cri-s:
I'm not commenting on this PR, it is just a hint.
Openpnp allow to increase homing accuracy using vision and fiducial location.
https://github.com/openpnp/openpnp/wiki/GcodeDriver#visual-homing

vonnieda:
I've decided not to merge this. I think this is a fix for a machine specific setup problem and it doesn't add a lot of value for most users. If you are having to reset your camera position every startup it would be better to address the core issue of bad endstops than to have to run through this process every time.

KarlZeilhofer:
Isn't this already very close to an automated nozzle runout compensation?
The main reason for doing bottom vision in our case is to compensate for nozzle runout. This could be skipped, if the runout is known before pick.

cri-s:
For nozzle runout check, you must do vision with component, why you do
it without ?
The problem with detecting the hole is , that usually this hole is not
the same as the
center of rotation. Practically, measuring the hole, you just adding
random error to the
error compensation.



2017-10-27 19:03 GMT+02:00, Karl Zeilhofer <notifications@github.com>:

vonnieda:
@KarlZeilhofer What you are referring to is Nozzle Calibration. We have code for that in place currently, but it doesn't work right, so it's been removed until it can be finished. There's more info about that in #235

sp-apertus:
This PR uses part of the code from #235 and intends to fix the issues and usability problems step by step. We figured we split the feature into smaller pieces and contribute them one after the other rather than making one huge PR. This first PR is the first puzzle piece to allow using a pipeline to detect the nozzle center of bottom vision camera and automatically move it to camera center. 

Once that works we would attempt to:
* automatically validate bottom vision camera axis orientation (rotation, flipping).
* automatically measure units/pixel
* automatically measure nozzle runout

cri-s:
I had the "nozzle tip calibration working", there are two bigger bugs,
but discardet it
as it create a lot of side effect that are bigger issues.

2017-11-08 13:50 GMT+01:00, Sebastian Pichelhofer <notifications@github.com>:

sp-apertus:
> I had the "nozzle tip calibration working", there are two bigger bugs

please elaborate/share

cri-s:
It don't take center of rotation (average of 4x 90 degree) and the
interpolation code is
totally wrong if angle is over certain degrees. I don't remember if 0
degree is correct.

The big side effect is that "this" code introduces random offsets to
camera during setup
of parts and the big problem is , if feeder or some parts are setup at
different times,
then it immediatly gives position errors on real results.

2017-11-08 15:45 GMT+01:00, Sebastian Pichelhofer <notifications@github.com>:

Issue #668, opened 2017-10-22
Play Sound, when job is interrupted


KarlZeilhofer:
# Feature Request
## Describe the Feature
When a job is interrupted and asks for user input, it should play some sound to acoustically notify the operator. 

## How Will It Look?
To keep it simple, i would provide a default sound deployed with OpenPnP. If the user provides an other file in his .openpnp folder, this file would be preferred.

## Why Do We Need It?
1. What problem does it solve?
In our lab we also have 3D printer, which makes similar noise as our LitePlacer. If PnP stops, one cannot immediately nitice that. 
2. Is it useful for everyone, or does it only solve a problem for your machine?
Notifications should be helpful for any user.

vonnieda:
We do have the Signaler and SoundSignaler system, so it could be used for
this very easily.

KarlZeilhofer:
Oh, I wasn't aware of that. I think it already does what I want. 
Though, it would be nice, if we can place files under .openpnp/sounds
which override the default sounds.

pfried:
Good idea, I will setup a pull request where you can override the error.wav and success.wav files

PR #683, opened 2017-12-06
Allowing users to override the default sounds of the sound signaler

pfried:
# Description
The sound signaler is shipped with two default free sounds for indicating success or error. This PR allows users to override those files.

Solves #668

# Instructions for Use
There are two files which can be replaced and need to be wav files: 'error.wav' and 'success.wav'.

Place them in the OpenPnP configuration directory under sounds e.g. '~/.openpnp/sounds/success.wav' or 'C:\Users\pfried\.openpnp\sounds\success.wav' 

@vonnieda If you accept this PR I will update the Wiki accordingly


vonnieda:
Looks good to me. Thanks @pfried!

PR #688, opened 2017-12-23
Added option to ignore error and continue assembly

mgrl:
# Description
Gives the user one more option on how to handle errors while assembly. The new option ignore and continue allows the user to ignore errors raised from alignment per bottom vision as if everything is right.

# Justification
if user checked for pick being ok, it should be allowed to continue.

# Instructions for Use
ignore_continue
1. pick a part with insufficient vac (vac sensor needed) and click continue -> it goes to alignment
2. picked part will be checked on bottom camera -> if it fails, one could continue anyway.

# Implementation Details
unit test and on real machine (vac insufficient and bottom vision failed if removed part).
yes
maybe @netzmark can test further?

OK

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -884,12 +884,14 @@ public class JobPanel extends JPanel {
             List<String> options = new ArrayList<>();
             String retryOption = "Try Again";
             String skipOption = "Skip";
+            String ignoreContinueOption = "Ignore and Continue";
             String pauseOption = "Pause Job";
 
             options.add(retryOption);
             if (jobProcessor.canSkip()) {
                 options.add(skipOption);
             }
+            options.add(ignoreContinueOption);
             options.add(pauseOption);
             int result = JOptionPane.showOptionDialog(getTopLevelAncestor(), t.getMessage(),
                     "Job Error", JOptionPane.YES_NO_CANCEL_OPTION, JOptionPane.ERROR_MESSAGE, null,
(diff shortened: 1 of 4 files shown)


netzmark:
Great Michael,
I'll check test it with pleasure of course but a bit later - must mix your changes with my modified jobProcessor first and it'll take some time. Then I'll test it somehow on the simulator, on the real machine after Christmas only.
brgds
Marek

vonnieda:
Looks good, thanks @mgrl!

netzmark:
Hi Michael, 
I have tested it today and it's not working good for me.
- test conditions: 3 nozzles head
- picking the part up by N1 properly
- missing the part on N2 because of empty feeder
after that last operation i got the message window with "Ignore and Continue" option. So placed the part on N2 manualy and pressed "I&C":
- system even didn't try to pick up the part with N3 - it's first mistake: should pick up the part with N3;
- head moved with N2 to the camera - it's second mistake: should go to camera with N1 first;
- alligned N2 it, and at once after thrown exception window that missing part on N3 for bottom vision - rather normal because didn't pick it up at all...

mgrl:
@netzmark sorry for that. I was missing some things obviously.

Made another pull request #697 . That should now work as intended and fixes your issues.

netzmark:
Hi Michael,
I have found still one bug in the modification.
Situation: we start procedure to pick parts using three nozzles. A window containing "Ignore/Continue" is thrown properly at the first misspick only. If another misspick appears on another nozzle - above windows doesn't come but we get old window with message "vacuum error" or so so and there is only one option to choose - Ok or Cancel, don't rememver exactly, I'm not at the machine now. Then we have to go to feeders, take this part "manually" and press Play - like before modification.
Of course, when all 3 parts are placed, then if misspick appears again at next 3-nozzles feeding procedure - proper "Ignore/Continue" windows comes as should do.

mgrl:
hi!

I think this is not related to these modifications, as I noticed that before already sometimes. I filed an issue some time ago to keep that in mind: #695  don't know where that comes from, maybe some event/error handler listeners are not right in place or something like that...

netzmark:
Understand. Pity :-(. It would be better if related to the modifications - would give a chance that you find it quick :-).
Anyways, I bless you everyday for your mods! Can't count how many times have pressed Continue" today only :-).

PR #689, opened 2017-12-23
renamed pipeline editor title for stripfeeder to prevent nullpointerexception

mgrl:
# Description
had issues editing the pipeline for a feeder but didn't know what the problem was. the error message was unclear to me:
err_feeder_no-part

so I changed the title of the pipeline editor, since it is feeder specific, not part specific...

# Justification
if a part is deleted and a feeder had it assigned, the feeder has no part any more. editing the pipeline gives the error, which is unexpected and doesn't indicate the problem.

# Instructions for Use
no instructions...

# Implementation Details
started openpnp and edited a strip feeder without part assigned. -> works now.
yes
no
OK


vonnieda:
Looks good, thanks @mgrl!

PR #691, opened 2017-12-23
Mark placed again

### LINKED ISSUE #663: Placed Flag isn't set any more
# Problem Report
## Describe the Issue
The placed-flag was working properly, until a special point in time. But now it doesn't any more.
Actual behaveour: the left column (References) gets colored from yellow over blue to green - this is as expected.
But the checkboxes all stay unchecked when processing the job. 
When I do check them manually, they get saved and restored as expected,
when closing and reloading the job. Also the job processor skips that parts which were loaded with a checked placed-flag - again as expected.

What I am aware of what is different to before:
* I've merged Jason's patch regarding the affine transform for the board area
* added fiducials to the board (which I added manually to the board's parts list)
* marked several parts of the board to be ignored

## Steps to Reproduce
How can a developer reproduce the issue you are experiencing? Use a step by step format like this:
1. Start OpenPnP
2. Load a job
3. Start the Job

## Expected Result
The checkboxes should be checked one after the other (according to the green state of the referenc-column)

## Actual Result
the checkboxes don't get checked

## Notes, Log Files, Screen Captures, Videos, etc. to Show the Issue
If it helps, I could provide some additional files - please tell me. 

Discussion: https://groups.google.com/forum/#!topic/openpnp/4MKg7JaUTAk

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/50346363-placed-flag-isn-t-set-any-more?utm_campaign=plugin&utm_content=tracker%2F429103&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F429103&utm_medium=issues&utm_source=github).
</bountysource-plugin>

mgrl:
# Description
fixes #663: if having fiducials checked, placed flag is set correctly.
don't know whether it's an naive approach or it breaks something else? but at least it fixes the bug ;)

# Justification
it's annoying to have the placement flags not set properly.

# Instructions for Use
use as expected, no matter if with or without checking.

# Implementation Details
on my machine in a testrun
think so
no changes
done.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -897,7 +897,7 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
     BoardLocation getFiducialCompensatedBoardLocation(BoardLocation boardLocation) {
         // Check if there is a fiducial override for the board location and if so, use it.
         if (boardLocationFiducialOverrides.containsKey(boardLocation)) {
-        	BoardLocation boardLocation2 = new BoardLocation(boardLocation);
+            BoardLocation boardLocation2 = new BoardLocation(boardLocation);
             boardLocation2.setLocation(boardLocationFiducialOverrides.get(boardLocation));
             return boardLocation2;
         }
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -26,7 +26,6 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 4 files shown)


vonnieda:
Okay, I knew there was a problem with this fix, but it took me all day to find it. Here's my comment from when I tried this myself several weeks ago:

```
// TODO STOPSHIP: This fixes the placed problem, but it fails to send notifications to the
// job panel so it doesn't know the job is dirty. Probably need to move the location
// stuff into the boardlocation.
BoardLocation boardLocation2 = new BoardLocation(boardLocation);
```

To expand on that comment, the problem is that when changes are made to BoardLocation it fires a property change notification. The Job that the BoardLocation belongs to listens for these notifications and updates it's dirty flag. This is what tells the UI that the job has changed when the user tries to exit.

So, the end result is that if you use this fix, the Placed flag is set correctly, but when you go to exit OpenPnP it doesn't know the job is dirty, so it lets you exit without saving, and you lose the Placed status. 

For "Probably need to move the location stuff into the boardlocation.", what I meant is that to fix this we should move the fidicual override into the BoardLocation as a transient property. This way we won't need to clone the BoardLocation for fidicuals, and thus we won't lose the Placed status as we run the job.

mgrl:
okay, tried my best to address the issues you mentioned. 
hopefully it's okay this way and I understood right.
did a quick check on my machine: the job is marked dirty by placements (* in the title bar) and placed is set likewise.

KarlZeilhofer:
@mgrl thank you for fixing this issue! we will try it soon.

PR #693, opened 2017-12-26
Bugfix: On Skip discard right nozzle

mgrl:
# Description
until today if skipping a part due to an error (vac/aligning failed) always nozzle 1 is cleared per discarding a part into the discard box. this patch fixes this an takes the right nozzle into account.

# Justification
it's very annoying to have the wrong nozzle cleared and investigate which part was assembled, which not and remember which parts to fix in next run...

# Instructions for Use
click "skip" if an error occured. now the correct nozzle is cleared and the assembly for the rest goes on.

# Implementation Details
tested on my machine with 2 nozzles on one head.
I tested 2 cases:
1: removed nozzle tip from nozzle 1. pickup from feeder failed of course, vacuum sense check failed and it gave me the option to skip/retry, etc. -> clicked skip -> nozzle 1 went to discard location, discarded and went with nozzle 2 to the feeder and continues as usual.
2:  removed nozzle tip from nozzle 2. pickup from feeder failed of course, vacuum sense check failed and it gave me the option to skip/retry, etc. -> clicked skip -> nozzle 2 went to discard location, discarded and went with nozzle 1 to the feeder and continiues as normal.

yes
no changes
done.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -804,14 +804,29 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
      * @throws Exception
      */
     protected void doSkip() throws Exception {
-        if (plannedPlacements.size() > 0) {
-            PlannedPlacement plannedPlacement = plannedPlacements.remove(0);
-            JobPlacement jobPlacement = plannedPlacement.jobPlacement;
-            Nozzle nozzle = plannedPlacement.nozzle;
-            discard(nozzle);
-            jobPlacement.status = Status.Skipped;
-            Logger.debug("Skipped {}", jobPlacement.placement);
-        }
+    	if (plannedPlacements.size() > 0) {
+    		
+    		//iterate through nozzles to determine which one to clear
+    		for (int i = 0; i < head.getNozzles().size(); i++) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
```diff
@@ -816,12 +816,35 @@ protected void doReset() throws Exception {
...
+            for (PlannedPlacement plannedPlacement : plannedPlacements) {
+                JobPlacement jobPlacement = plannedPlacement.jobPlacement;
+
+                if (plannedPlacement.stepComplete) {
+                    // go over placements having the current step already completed
+                    continue;
+                }
+
+                // remove the placement to be skipped from list
+                plannedPlacements.remove(plannedPlacements.indexOf(plannedPlacement));
```
vonnieda: This is likely to cause a ConcurrentModificationException. You can't modify a collection that you are iterating unless you use an Iterator and the Iterator.remove() method. See https://stackoverflow.com/questions/18448671/how-to-avoid-concurrentmodificationexception-while-removing-elements-from-arr. Other that that, this change looks solid!

mgrl:
updated according to review. indeed the ConcurrentModificationException was not raised since only one element was called and loop left afterwards. but this way it's better obviously.

netzmark:
Hi Michael, Jason
Tested it today. Seems be working just properly!

mgrl:
yay! :)

PR #697, opened 2018-01-02
Correctly handle overridden errors

mgrl:
Sorry for another bugfix-bugfix ;)

Fixes the unintended behaviour @netzmark describes in #688.
I was missing to set stepComplete for the error to be handled at the moment the user clicks a message box and additionally the FSM-states added were wrong.

This fix is tested on my 2 nozzle machine again, this time with a failing vacuum on nozzle 1 and afterwards with failing vac. on nozzle 2. in both cases the assembly went correctly on.


netzmark:
Hi Michael,
Great that you do what you do, bugs don't appear only when we do nothing ;).
Going to compile it and tomorrow will give a test result feedback, thx!!!

netzmark:
So tested it today and as for me it's working great now :-).
Making some proposal of another improvement on the group in "skip function" thread (few minutes).

PR #700, opened 2018-01-29
Make scripts Python 3 compatible by using __future__ imports

johnthagen:
# Description
Add Python 3 compatibility to example scripts using the built in `__future__` package.

# Justification
Since examples are used as places for users to start, this is a good practice for new Python 2 code as it will make porting to Python 3 in the future easier.

# Instructions for Use
N/A

# Implementation Details
N/A


johnthagen:
@vonnieda I tested these updates and they work as they did on Python 2. It's not possible to test yet on Python 3 because OpenPnP/Jython do not yet support Python 3, but this PR will make the scripts more compatible with Python 3.

johnthagen:
Upon further testing, it looks like this is not working. See http://bugs.jython.org/issue2007

PR #702, opened 2018-01-31
Make hallo openpnp script Python 3 compatible

johnthagen:
# Description
Use parenthesis when calling `print` to make Python 3 compatible.

# Justification
This will be used as an example by others, so helping them write code that is Python 3 compatible will be useful.

According to the Python PEP8 style guide, modules should be `lower_snake_case` so this makes this module adhere to that standard.

# Instructions for Use
N/A

# Implementation Details
This code was tested using the virtual device/driver and behaves the same as it did.

### DIFF
--- a/src/main/resources/scripts/Examples/Python/Print_Hallo_OpenPnP.py
+++ b/src/main/resources/scripts/Examples/Python/Print_Hallo_OpenPnP.py
@@ -1,2 +1,0 @@
-print "Hallo OpenPnP"
-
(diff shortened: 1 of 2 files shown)


johnthagen:
Just curious, as I'm a native English speaker, does "Hallo" mean "Hello" in another language? 😄

vonnieda:
@johnthagen Yep, it does. Whoever originally contributed that script uses their native language instead of English :)

PR #709, opened 2018-02-01
Copy new Python script examples to OpenPnP working directory

johnthagen:
# Description
Copy new Python files into OpenPnP scripts directory.

# Justification
Requested in #708


cri-s:
What is the reason to upgrade to python 3x where version 2.7 is used and where opencv python code examples from web works with python 2x (after adding numpy... Jar file), and to my knowledge the 3x notation don't work.

johnthagen:
This PR only adds examples already merged, but to answer your question, all of the changes recently merged work in both Python 2 and Python 3.

So Python 2 compatibility is preserved, and it helps make them ready for Python 3 whenever support for that lands in Jython.

cri-s:
Do you know that python 3x don't support opencv 2.x , only 3.x with a
lot of code change for the existing scripts and further updating
(opencv) vision code to 3.x is a lot of work where the result is just
code that works slower.
As there is a intend from project creator to migrate to opencv 3.x
(from open issues),
so yes, it is possible that it go that way, my personal opinion is
that is just a lot of work for
nothing. Further on linux Lubuntu LTE python3 don't really work.

rgds Cri.S

vonnieda:
We're not upgrading to Python 3. The changes are just to make the scripts
compatible with Python 3 and they are intended to be backwards compatible
with Python 2. @johnthagen has tested the scripts in OpenPnP to make sure
they still work as modified. This shouldn't cause any issues.

cri-s:
I have just asked the reason, as running opencv python examples custom
compilation with
machine depend JyNI.jar is needed, as otherwise it don't work or works
only for simple things.

2018-02-02 21:10 GMT+01:00, Jason von Nieda <notifications@github.com>:
> We're not upgrading to Python 3. The changes are just to make the scripts
> compatible with Python 3 and they are intended to be backwards compatible
> with Python 2. @johnthagen has tested the scripts in OpenPnP to make sure
> they still work as modified. This shouldn't cause any issues.
>

vonnieda:
Honestly, I don't know the reason :) Maybe @johnthagen can answer that. It
didn't seem like it would do any harm, so I merged it :)

Jason

johnthagen:
The reason was that since these are examples, the examples now show how to use Python 2.7 in a way that is also compatible with Python 3. It has absolutely no bearing on forcing people to use Python 3 in any circumstances, it's just a good practice in general.

I don't really understand how making these scripts compatible with both Python 2.7 and 3 have anything to do with opencv. They are just executed in OpenPnP in the context of Jython and call OpenPnP **Java** objects.

Sorry, also I'm having a difficult time understanding you, so that might be some of the miscommunication.


___________
This is completely off topic to this PR, but for what it's worth I have used opencv bindings in a Python application that supports both Python 2.7 and Python 3 and it worked fine for me: https://pypi.python.org/pypi/opencv-python. If you scroll to the bottom of that PyPI page you'll see they support Python 2.7, 3.4, 3.5, & 3.6 simultaneously.

cri-s:
Jython can be used to perform imaging processing stages callable by
vision pipeline or
that could be used inside automatic or manually triggered script.
Most of this require numpy and similar aux libs that are included
inside jython using Cpython
and Jni interface.
This allows to use opencv 2.4 python code directly on openpnp and is
great for the persons
that knows python 2.x and opencv and only .
The prerequestite is to not use internal opencv lib, but external
linked, and to add the
architecture depend and locally compiled required external python
dependencys. jar to targets/lib directory.
--------------------------------------------------------------------------------------------------------
The page https://pypi.python.org/pypi/opencv-python say:
Supported Python versions
Python 2.7 is the only supported version in 2.x series
The meaning should be:
Python 2.7 is the only supported version in OpenCv 2.x series, OpenCv 3.x could
be installed over both, python 2.7 and 3.x  if numpy supports that
subversion.. .


2018-02-02 22:00 GMT+01:00, johnthagen <notifications@github.com>:

johnthagen:
@cri-s Feel free to try out the modified scripts and see if they don't work for your case. They should be fully compatible with Jython 2.7, so there's no reason they shouldn't work the same when using opencv 2.4.

If you do find issues, it's probably best to open a new issue about them as I think this discussion (while interesting) is off topic for this PR.

vonnieda:
Looks good, thanks @johnthagen!

PR #711, opened 2018-02-19
Upgrade to final release of Jython 2.7.1

### LINKED ISSUE #710: Upgrade to Jython 2.7.1 final
Jython 2.7..1 final has been released: https://fwierzbicki.blogspot.com/2017/07/jython-271-final-released.html

OpenPnP is still currently still using Jython 2.7.1b3

Link to the updated `jython-standalone` Maven artifact: https://search.maven.org/#artifactdetails%7Corg.python%7Cjython-standalone%7C2.7.1%7Cjar

The new version is about 1.5 years newer than what is currently being used by OpenPnP: https://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.python%22%20AND%20a%3A%22jython-standalone%22

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/55239290-upgrade-to-jython-2-7-1-final?utm_campaign=plugin&utm_content=tracker%2F429103&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F429103&utm_medium=issues&utm_source=github).
</bountysource-plugin>

johnthagen:
Fixes #710 

All of the current Python example scripts were tested and they all function as before.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -118,7 +118,7 @@
 		<dependency>
 			<groupId>org.python</groupId>
 			<artifactId>jython-standalone</artifactId>
-			<version>2.7.1b3</version>
+			<version>2.7.1</version>
 		</dependency>
 		<dependency>
 			<groupId>org.beanshell</groupId>


vonnieda:
Looks good to me. Thanks @johnthagen!

cri-s:
can you make a example with python, numpy and opencv.
I normally remove jython and replace it with jep/pyjnius for real python use,
with jython i was never able to do it or it was really, really slow.

johnthagen:
@cri-s In my limited experience using OpenPnP, the way we solved needing external Python dependencies from PyPI, we would create a Python virtual environment, install what we needed into it, and "shell out" to it using `subprocess.call`. This of course isn't ideal, but I don't know of any way to use `pip` install packages into the Jython environment itself.

What complicates this further is that a lot packages will check the `platform` before doing something system related. In Jython the platform is `Java`, and I haven't seen any packages that support Jython this way. It's just not used enough.

So my only suggestion is to shell out from Jython to another Python process.

cri-s:
What is then the reason to have such a heavy (MB) package as jython if
you need shell out
to a real python interpreter to perform the real job. Any other
scripting language can that too,
or modify openpnp script code to call a script that shell it out.
The reason for the script is that the location or name of binary is
installation dependent.


2018-02-21 17:18 GMT+01:00, johnthagen <notifications@github.com>:

johnthagen:
OpenPnP and Jython are both very new to me. That caveat aside, I'd imagine that the reason that the scripting interface exists is to allow you to easily modify or reuse the **existing** OpenPnP Java code, without having to write Java and recompile yourself. So you can take advantage of existing OpenPnP libraries, but working in a language that might be more familiar and/or ergonomic.

Extending OpenPnP with new libraries (like Python packages from PyPI such as numpy or opencv) was probably not in the scope of what they were originally intending for the scripting interface.

Again, @vonnieda probably knows way more about the **why** than me.

vonnieda:
That is correct. The scripting system is intended for doing lightweight,
simple tasks. Large pieces of new functionality should be written in Java,
and that includes extensive CV operations. It was never intended to be used
with external libraries. People can jump through whatever hoops they want,
of course, and make it do whatever they want, but that's not what it's
designed for.

Jason

PR #738, opened 2018-06-15
active placement number indicator

sp-apertus:
# Description
Added indicator label how many active and how many completed placements are listed in this board.

active-placements

# Justification
It provides the user with better overview of the current placement progress (like a progress bar - maybe a good future addition) and before starting a job provides valuable information about the number of placements planned for this PCB.

# Instructions for Use
The label automatically updated whenever a placement is set to "fiducial", "ignore", "place", when a placed flag is set or when a placement side is changed. It also automatically updates during job runs and when a job placement status is reset.

# Implementation Details
I used the nulldriver and with multiple boards and also tried the autopanel feature.
In autopanel mode the active placements are "per" board.

yes
I did: BUILD SUCCESS

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
@@ -18,6 +18,7 @@ import javax.swing.Action;
 import javax.swing.DefaultCellEditor;
 import javax.swing.JButton;
 import javax.swing.JComboBox;
+import javax.swing.JLabel;
 import javax.swing.JMenu;
 import javax.swing.JOptionPane;
 import javax.swing.JPanel;
@@ -69,6 +70,7 @@ public class JobPlacementsPanel extends JPanel {
     private ActionGroup multiSelectionActionGroup;
     private ActionGroup captureAndPositionActionGroup;
     private BoardLocation boardLocation;
+    private JLabel labelActivePlacements;
 
     private static Color typeColorIgnore = new Color(252, 255, 157);
     private static Color typeColorFiducial = new Color(157, 188, 255);
@@ -141,7 +143,12 @@ public class JobPlacementsPanel extends JPanel {
         JButton btnEditFeeder = new JButton(editPlacementFeederAction);
         btnEditFeeder.setHideActionText(true);
         toolBarPlacements.add(btnEditFeeder);
-
+        
+        toolBarPlacements.addSeparator();
+        
+        labelActivePlacements = new JLabel("");
+        toolBarPlacements.add(labelActivePlacements);
+        
         tableModel = new PlacementsTableModel(configuration);
 
         table = new AutoSelectTextTable(tableModel);
@@ -155,6 +162,7 @@ public class JobPlacementsPanel extends JPanel {
         table.setDefaultRenderer(PlacementsTableModel.Status.class, new StatusRenderer());
         table.setDefaultRenderer(Placement.Type.class, new TypeRenderer());
         table.setDefaultRenderer(PartCellValue.class, new IdRenderer());
+        tableModel.setJobPlacementsPanel(this);
         table.getSelectionModel().addListSelectionListener(new ListSelectionListener() {
             @Override
             public void valueChanged(ListSelectionEvent e) {
@@ -202,6 +210,7 @@ public class JobPlacementsPanel extends JPanel {
                     Placement placement = getSelection();
                     placement.setType(placement.getType() == Type.Place ? Type.Ignore : Type.Place);
                     tableModel.fireTableRowsUpdated(table.getSelectedRow(), table.getSelectedRow());
+                    UpdateActivePlacements();
                 }
                 else {
                     super.keyTyped(e);
@@ -236,6 +245,7 @@ public class JobPlacementsPanel extends JPanel {
     
     public void refresh() {
         tableModel.fireTableDataChanged();
+        UpdateActivePlacements();
     }
 
     public void selectPlacement(Placement placement) {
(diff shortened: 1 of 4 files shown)


cri-s:
It is possible you include a boolean variable on machine.xml that as default enable this code, but
it allows disable it editing machine.xml file manually..
The resons for me:
I have scripts that add icons/actions to the job processor, as example check all polarized components just
one time, as example if red led is present 10 times, it get checked only one time with indication of pin1
overlayed. This label give me problems . Other things are if using dispenser/solder iron other icons
are added for that specific things.
Further having jobs with 4000 (top+bot) parts and 24 panelisations, i don't want that after every placement it traverses 96000 list and calls isPlaced that count of time, slowing down the assembly
a lot.

sp-apertus:
Adding a checkbox in machine settings GUI (and machine.xml) to enable/disable this feature is fine for me. What are other peoples thoughts?

vonnieda:
Hold off on that till I’ve had a chance to look at this please.

cri-s:
To clearify, I add icons using startup.bsh to this tab. Clearly this text
label disturb the GUI.
Il 16/giu/2018 15:08 "Jason von Nieda" <notifications@github.com> ha
scritto:

> Hold off on that till I’ve had a chance to look at this please.
> On Sat, Jun 16, 2018 at 2:06 AM Sebastian Pichelhofer <

vonnieda:
I agree with @cri-s. I don't think this belongs in the toolbar. In addition, it adds another status item to a system I think is already too confusing. We have text in the ID column, colors in the ID column and status info in the status bar already. 

I'd like to see all these systems combined, or one or more removed. I think an appropriate place for this kind of data is in the status bar, broken out like the DROs are.

sp-apertus:
Moving it to the status bar sounds like a good idea. Either merging it with the status info or removing the status info and leaving that information for the logs.

I don't have time to work on it immediately but hopefully in a few weeks. If anybody else feels like starting already please go ahead and I will join later on.

cri-s:
The status info is really needed for scripting usage, please don't remove that.

2018-06-26 10:15 GMT+02:00, Sebastian Pichelhofer <notifications@github.com>:

sp-apertus:
New proposal to add the active placements count to the status bar before the current status bar text:

pnp---status-text

vonnieda:
Hi @sp-apertus, this is an improvement, but if we're going to add more status right now I'd really like to consolidate and improve what we've got. I think we can do that very easily. How about something like this?

screen shot 2018-06-29 at 8 39 39 pm

The numbers aren't correct, this is just a mockup. But here is what I'm thinking:

1. A new little status bar item between the main status text and the DROs which shows the progress of the job. It shows Job: Finished / Total and Board: Finished / Total. The Job one reflects all the active placements in the job, the board one reflects all the active placements in the selected board.
2. Get rid of the "(2/4)" stuff in the ID column. This is just different way to show the same thing and I think it's confusing.
3. Keep the colors in the Placements panel which shows the status of each ID. As a reminder: Yellow is pending, Blue is currently placing and Green is placed.
4. Add colors to the Board panel, too, following the same rules as #3. This will make it very easy for you to tell which board is being placed, and by clicking on that board you can easily see which placements are done. And using the new stats below, how far there is to go.

We could make the new status item a little smaller by using a couple small icons in place of "Job" and "Board" and we could color the numbers yellow and green to more easily visually indicate what they represent.

The main benefit of this, I think, is that we finally have something telling us how far along in the whole job we are. Right now you have to click through each board to see and kind of add it up in your head.

What do you think?

sp-apertus:
Looks good!

1. I would reword it as currently it reads like there are 2 of 12 jobs done. Maybe "Placements: 2 / 12 Total | 1 / 4 Selected Board"

Do you opt for keeping the "3 of 4 active placements" in the Pick and Place panel (as its still in your mockup)?

vonnieda:
@sp-apertus Yep, I like that wording, and no, no need for the "3 of 4". Just forgot to remove it from the mockup. Do you agree that the new status item would take care of this, or am I misunderstanding what that text represents?

vonnieda:
Yes indeed, I agree it should update each time a row is edited. I think
this ultimately is very similar to your code - just moving where it's
displayed. Do you want to work on this?

Jason

sp-apertus:
I am happy to work on this tomorrow.

Would you just be so kind and give me a pointer in which file/part of the code the status line/DRO is being handled so I have a reference for this new element?

vonnieda:
It's in MainFrame.java. Recommend looking at the DRO item and copying it's
attributes - in particular, the border. If you have trouble let me know and
I will add it.

Jason

sp-apertus:
Thanks!

sp-apertus:
I went ahead and implemented the discussed changes, even had a go at the progress bar and to my surprise everything already seems to work flawlessly. Tested with single Board, autopanelized boards and multiple boards.

pick-progress-gui

vonnieda:
Great work @sp-apertus, this looks really nice! Thank you for your work on this!

I'm going to go ahead and merge it, and then will start a discussion with the group about removing the "(2/4)" stuff in the ID column. I find it to be confusing, but others might be using it, so it's worth discussing.

vonnieda:
Hey @sp-apertus, made a few small changes in https://github.com/openpnp/openpnp/commit/39a929b4517487ab3d0647f178d3bee994d8bb17 to make is so the status updates even when no board is selected, and it updates immediately upon job load. Let me know what you think.

sp-apertus:
Glad I could help.

Great!

sp-apertus:
I will add a new line to change.md about this in a new PR.

Issue #715, opened 2018-04-01
Strip feeder option "Rotation in Tape" is inconsistency with another feeders


aneox:
# Problem Report
## Describe the Issue
In strip feeder option "Rotation in Tape" is inconsistency with another feeders. For example in drag or auto feeders we set pick rotation and it works as expected. In strip feeder, if we set 90 degreed Rotation in tape, we expect 90 degree pick rotation, but if strip is rotated 270 on table, vision detect it and pick rotation will be 270+90 = 360 = 0, fully wrong. So we need set Rotation in tape = 0, but in another feeder, with same strip rotation, 90 pick rotation works as expected, inconsistency.

I suggest to change concept of "Rotation in tape" for strip feeder, make it just "Pick rotation", as we were used to with drag and auto feeders. Code is easy to fix, we need just made vision correction rotate to nearly 90 or 0 or 180 or 270.. i going to test it and share patch. But i do not know how to continue to support machines, which is already using wrong concept of "Rotation in tape". Or maybe iam wrong?
I think if part on board has rotation 90 and part in strip looks in camera the same rotation, it should have same 90 rotation. It will be easy to understand and setup correct way
## Steps to Reproduce
set strip rotated 270 on table and set Rotation in tape 90

## Expected Result
pick with 90 degree rotation nozzle

## Actual Result
rotation will be 270+90 = 360 = 0, fully wrong

-----------------------------------------------------------------------

# Feature Request
## Describe the Feature
I suggest to change concept of "Rotation in tape" for strip feeder, make it just "Pick rotation", as we used to, with drag and auto feeders. 

## How Will It Look?
https://yadi.sk/d/4RpHnBWV3TxYuS
But need to add compatibility for already working machines.

## Why Do We Need It?
It will be easy to understand and setup correct way

mgrl:
hi!
From my understanding you should come out like this with the current implementation, don't you?
![rotationintape](https://user-images.githubusercontent.com/3868450/40044502-d1b0fd02-5827-11e8-9149-c3be74e16160.png)

At first I found it confusing, too. But how it is currently implemented, it is independent from the feeders orientation on the table. Maybe it's inconsistent because the stripe feeders' rotation could change very likely compared to autofeeder, so it's taken into account only for them, not for the fixed autofeeder?

aneox:
i have added my approach in pull request. By default its off, and program will work in old way. To active, you need to set checkbox in machine settings.

vonnieda:
[...]

PR #740, opened 2018-07-02
My changes and impromevents

aneox:
- Filter PlacementsTableModel, show only active board side.
- Added option to AutoHome after machine enabled. To activate, need to set checkbox in machine settings.
- Windows saves sizes and position in Multiple Windows Mode
- Save configureation menu button
- Camera window can be split in vertical or horizontal style
- Added new approach to pick rotation for Strip Feeder. To activate, need to set checkbox in machine settings. https://github.com/openpnp/openpnp/issues/715
- Some changes in job order. Now user can change job order by sorting table.
- Job autosave after each placement.
- Added peel off actuator option for Drag Feeder.
- Drag Feeder improve accurance of feed, now drag distance can be adaptive with vision enabled.
- Drag Feeder can work with 0402.

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -95,6 +95,7 @@ import org.openpnp.spi.MachineListener;
 import org.openpnp.util.FiniteStateMachine;
 import org.openpnp.util.MovableUtils;
 import org.openpnp.util.UiUtils;
+import org.pmw.tinylog.Logger;
 
 import com.google.common.eventbus.Subscribe;
 
@@ -245,7 +246,12 @@ public class JobPanel extends JPanel {
                 	updatePanelizationIconState();
                 	
                 }
-
+                if (job.getBoardLocations().size() > 0) {
+					BoardLocation boardLocation = getSelectedBoardLocation();
+					jobPlacementsPanel.setBoardLocation(boardLocation);
+                }
+                else
+                    jobPlacementsPanel.setBoardLocation(null);
             }
         });
 
--- a/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
@@ -12,6 +12,7 @@ import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.PatternSyntaxException;
 
 import javax.swing.AbstractAction;
 import javax.swing.Action;
@@ -26,11 +27,13 @@ import javax.swing.JScrollPane;
 import javax.swing.JTable;
 import javax.swing.JToolBar;
 import javax.swing.ListSelectionModel;
+import javax.swing.RowFilter;
 import javax.swing.UIManager;
 import javax.swing.border.LineBorder;
 import javax.swing.event.ListSelectionEvent;
 import javax.swing.event.ListSelectionListener;
 import javax.swing.table.DefaultTableCellRenderer;
+import javax.swing.table.TableRowSorter;
 
 import org.openpnp.events.PlacementSelectedEvent;
 import org.openpnp.gui.components.AutoSelectTextTable;
@@ -60,10 +63,12 @@ import org.openpnp.spi.PnpJobProcessor.JobPlacement;
 import org.openpnp.util.MovableUtils;
 import org.openpnp.util.UiUtils;
 import org.openpnp.util.Utils2D;
+import org.pmw.tinylog.Logger;
 
 public class JobPlacementsPanel extends JPanel {
     private JTable table;
     private PlacementsTableModel tableModel;
+    private TableRowSorter<PlacementsTableModel> tableSorter;
     private ActionGroup boardLocationSelectionActionGroup;
     private ActionGroup singleSelectionActionGroup;
     private ActionGroup multiSelectionActionGroup;
@@ -143,9 +148,10 @@ public class JobPlacementsPanel extends JPanel {
         toolBarPlacements.add(btnEditFeeder);
 
         tableModel = new PlacementsTableModel(configuration);
+        tableSorter = new TableRowSorter<>(tableModel);
 
         table = new AutoSelectTextTable(tableModel);
-        table.setAutoCreateRowSorter(true);
+        table.setRowSorter(tableSorter);
         table.getTableHeader().setDefaultRenderer(new MultisortTableHeaderCellRenderer());
         table.setSelectionMode(ListSelectionModel.MULTIPLE_INTERVAL_SELECTION);
         table.setDefaultEditor(Side.class, new DefaultCellEditor(sidesComboBox));
@@ -258,6 +264,17 @@ public class JobPlacementsPanel extends JPanel {
         else {
             tableModel.setBoardLocation(boardLocation);
             boardLocationSelectionActionGroup.setEnabled(true);
+
+            RowFilter<PlacementsTableModel, Object> rf = null;
+            // If current expression doesn't parse, don't update.
+			try {
+				rf = RowFilter.regexFilter("(?i)" + boardLocation.getSide().toString());
+			}
+            catch (PatternSyntaxException e) {
+                Logger.warn("Side sort failed", e);
+                return;
+            }
+            tableSorter.setRowFilter(rf);
         }
     }
 
--- a/src/main/java/org/openpnp/machine/reference/ReferenceMachine.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceMachine.java
@@ -86,6 +86,12 @@ public class ReferenceMachine extends AbstractMachine {
     @Element(required = false)
     protected FiducialLocator fiducialLocator = new ReferenceFiducialLocator();
 
+    @Element(required = false)
+    private boolean homeAfterEnabled = false;
+
+    @Element(required = false)
+    private boolean usePickRotationInsteadOfRotationInTapeForStripFeeders = false;
+
     private boolean enabled;
 
     private List<Class<? extends Feeder>> registeredFeederClasses = new ArrayList<>();
@@ -297,5 +303,21 @@ public class ReferenceMachine extends AbstractMachine {
     @Override
     public PasteDispenseJobProcessor getPasteDispenseJobProcessor() {
         return pasteDispenseJobProcessor;
+    }
+
+    public boolean getHomeAfterEnabled() {
+        return homeAfterEnabled;
+    }
+
+    public boolean getUsePickRotationInsteadOfRotationInTapeForStripFeeders() {
+        return usePickRotationInsteadOfRotationInTapeForStripFeeders;
+    }
+
+    public void setHomeAfterEnabled(boolean newValue) {
+        this.homeAfterEnabled = newValue;
+    }
+
+    public void setUsePickRotationInsteadOfRotationInTapeForStripFeeders(boolean newValue) {
+        this.usePickRotationInsteadOfRotationInTapeForStripFeeders = newValue;
     }
 }
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -19,6 +19,7 @@
 
 package org.openpnp.machine.reference;
 
+import java.io.File;
 import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -426,8 +427,9 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
         fireTextStatus("Planning placements.");
 
         // Get the list of unfinished placements and sort them by part height.
+
         List<JobPlacement> jobPlacements = getPendingJobPlacements().stream()
-                .sorted(Comparator.comparing(JobPlacement::getPartHeight))
+                .sorted(Comparator.comparing(JobPlacement::getPartId))
                 .collect(Collectors.toList());
 
         if (jobPlacements.isEmpty()) {
@@ -788,6 +790,10 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
 	        }
             
             Logger.debug("Place {} with {}", part, nozzle.getName());
+
+            File file = job.getFile();
+            Configuration.get().saveJob(job, file);
+            Configuration.get().save();
         }
 
         clearStepComplete();
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
@@ -79,12 +79,18 @@ public class ReferenceDragFeeder extends ReferenceFeeder {
     protected double feedSpeed = 1.0;
     @Attribute(required = false)
     protected String actuatorName;
+    @Attribute(required = false)
+    protected String peelOffActuatorName;
     @Element(required = false)
     protected Vision vision = new Vision();
     @Element(required = false)
     protected Length backoffDistance = new Length(0, LengthUnit.Millimeters);    
 
     protected Location pickLocation;
+
+    private double feededCount = 0;
+    private double partsPitchX = -2; //-2mm for 0402
+    private double partsPitchY = 0;
 
     /*
      * visionOffset contains the difference between where the part was expected to be and where it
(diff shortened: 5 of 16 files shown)


mgrl:
Too bad it's an all in one commit and the automatic checks fail... Nevertheless some features are useful for having it in the openpnp main branch, I think. So maybe it's worth it to separate each feature to separate pull requests after jason states what would probably be merged if proper, tested pull requests are submitted?

vonnieda:
Hi @aneox, thank you for the PR! There are a lot of great features in this and I'm very interested in merging them. As @mgrl said, we will want to split this up into some smaller PRs, and the PRs will need to pass the tests. 

You can see which tests failed by clicking the Details link above. It looks like most of it is just style problems, like this:

```
[ERROR] /home/travis/build/openpnp/openpnp/src/main/java/org/openpnp/gui/components/CameraPanel.java:203: 'if' construct must use '{}'s. [NeedBraces]
[ERROR] /home/travis/build/openpnp/openpnp/src/main/java/org/openpnp/gui/components/CameraPanel.java:205: 'else' construct must use '{}'s. [NeedBraces]
[ERROR] /home/travis/build/openpnp/openpnp/src/main/java/org/openpnp/gui/JobPanel.java:253: 'else' construct must use '{}'s. [NeedBraces]
[ERROR] /home/travis/build/openpnp/openpnp/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java:111: 'if' construct must use '{}'s. [NeedBraces]
[ERROR] /home/travis/build/openpnp/openpnp/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java:113: 'else' construct must use '{}'s. [NeedBraces]
[ERROR] /home/travis/build/openpnp/openpnp/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java:220: 'else' construct must use '{}'s. [NeedBraces]
[ERROR] /home/travis/build/openpnp/openpnp/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java:234: 'else' construct must use '{}'s. [NeedBraces]
[ERROR] /home/travis/build/openpnp/openpnp/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java:347:24: Name 'isPart_0402' must match pattern '^[a-z][a-zA-Z0-9]*$'. [MethodName]
[ERROR] /home/travis/build/openpnp/openpnp/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java:190: 'if' construct must use '{}'s. [NeedBraces]
[ERROR] /home/travis/build/openpnp/openpnp/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java:192: 'else' construct must use '{}'s. [NeedBraces]
[ERROR] /home/travis/build/openpnp/openpnp/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java:195: 'if' construct must use '{}'s. [NeedBraces]
[ERROR] /home/travis/build/openpnp/openpnp/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java:197: 'else' construct must use '{}'s. [NeedBraces]
```

As for splitting the features, please see https://github.com/openpnp/openpnp/wiki/Developers-Guide#granularity. The goal is to have PRs for each feature, or at least a few related features, so that we can discuss and decide on the features, rather than a huge set of changes.

Since this is your first PR, I can understand if splitting it up might be a bit overwhelming. Let me know if you need some help and I'd be happy to provide some guidance.

aneox:
Hi! 
@vonnieda 
I understand, may be next week i will find some free time to split. Should i delete this PR?

vonnieda:
Lets leave it open for now. I’ll spend more time looking at the features so
we have a head start on the new PRs.

KarlZeilhofer:
I'm excited about this PR! Great work @aneox! This would solve many feature requests.

Phara0h:
Nice PR features hope he takes the time to confirm with the standards.

vonnieda:
Hi @aneox, I have some free time this weekend so I'm going to try get this merged and will just leave out the things that need further work. I'll follow up here and let you know.

src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
```diff
@@ -426,8 +427,9 @@ protected void doPlan() throws Exception {
         fireTextStatus("Planning placements.");
 
         // Get the list of unfinished placements and sort them by part height.
+
         List<JobPlacement> jobPlacements = getPendingJobPlacements().stream()
-                .sorted(Comparator.comparing(JobPlacement::getPartHeight))
+                .sorted(Comparator.comparing(JobPlacement::getPartId))
```
vonnieda: This is a major change to how the order of placements works and I think it needs to be discussed before adding this. The reason we use height is so that the machine doesn't run into tall parts after they are placed

src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
```diff
@@ -149,45 +176,69 @@ public void feed(Nozzle nozzle) throws Exception {
...
         pickLocation = this.location;
-        if (visionOffset != null) {
-            feedStartLocation = feedStartLocation.subtract(visionOffset);
-            feedEndLocation = feedEndLocation.subtract(visionOffset);
-            pickLocation = pickLocation.subtract(visionOffset);
-        }
 
-        // Move the actuator to the feed start location.
-        actuator.moveTo(feedStartLocation.derive(null, null, Double.NaN, Double.NaN));
+        if (feededCount == 0) {
```
vonnieda: This is a big change, and I don't really understand what it does. Can you explain the changes to me please?

src/main/java/org/openpnp/gui/JobPanel.java
```diff
@@ -245,7 +246,12 @@ else if (e.getFirstRow() == 0 && e.getLastRow() == Integer.MAX_VALUE) {
                 	updatePanelizationIconState();
                 	
                 }
-
+                if (job.getBoardLocations().size() > 0) {
```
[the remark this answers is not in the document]
aneox: To refresh placement table, if we change board side, as i remember.

src/main/java/org/openpnp/gui/JobPlacementsPanel.java
```diff
@@ -258,6 +264,17 @@ public void setBoardLocation(BoardLocation boardLocation) {
         else {
             tableModel.setBoardLocation(boardLocation);
             boardLocationSelectionActionGroup.setEnabled(true);
+
```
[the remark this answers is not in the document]
aneox: I think to edit both sides, user can switch sides using combobox. no need to add board twice.

src/main/java/org/openpnp/machine/reference/ReferenceMachine.java
```diff
@@ -298,4 +304,20 @@ public PnpJobProcessor getPnpJobProcessor() {
...
+
+    public boolean getUsePickRotationInsteadOfRotationInTapeForStripFeeders() {
+        return usePickRotationInsteadOfRotationInTapeForStripFeeders;
+    }
+
+    public void setHomeAfterEnabled(boolean newValue) {
+        this.homeAfterEnabled = newValue;
+    }
+
+    public void setUsePickRotationInsteadOfRotationInTapeForStripFeeders(boolean newValue) {
```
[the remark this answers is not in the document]
aneox: i place there, to change this setting for ALL strip feeders. If we put it in feeder wizard, it will apply only for 1 feeder, i think.

src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
```diff
@@ -788,6 +790,10 @@ protected void doPlace() throws Exception {
 	        }
             
             Logger.debug("Place {} with {}", part, nozzle.getName());
+
+            File file = job.getFile();
```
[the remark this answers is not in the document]
aneox: yeah, it should be some checkbox in machine settings, to disable this feature. For me, as i already said in chat, mac crashes sometimes if openpnp run, so this feature must be checking on all time)

src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
aneox: oh, i didnt see any problems with. already done a lot of board with it. order are more predictable and can be set by table sorting. very good feature.

with PartHeight order, absolutely not predicable and not user friendly order.

KarlZeilhofer: it's a good idea, to take the part height into account. Perhaps this value should be listed, so the user can sort the list by the parts height. Or at least he has it in focus, before starting the job. 

Another idea could be, that the user gets warned when starting a job with "mixed" heights. A configurable threshold should help to give it some tolerance (e.g. 0.2mm). 

Third Idea: I would like to group e.g. all resistors, then all ceramic caps, then diodes, then ICs, ... Similar to a manual pick and place. I would like to select multiple lines and define a group (floating point number). Then one can assign that number (normally integer) as he wants, and then sort the whole list by the grouping number. Special parts can then be handled by using further digits of that floating point number. 

@aneox: what is you typical column you are sorting the table?

vonnieda: @aneox @KarlZeilhofer One thing I want to clarify: The note in the PR said "Some changes in job order. Now user can change job order by sorting table." but as far as I can tell this is not the case. It was just sorting by placement ID. The old behavior is by height, and the new behavior was by placement ID. I stuck with the old behavior.

aneox: @KarlZeilhofer 
for me, best Part column. Job goes from one feeder to another and never return back to previous feeders. easy to control proccess.

@vonnieda 
sorry, i didnt test another columns. I used only part column and this order works well.
Maybe you misunderstand "sorting by placement ID", job should go thought placement ID column. However you can sort table by any column. I cant test now, please play with table column sorting with my code, it should work, as i remember.

src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
KarlZeilhofer: I like it!

src/main/java/org/openpnp/gui/JobPanel.java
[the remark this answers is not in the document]
vonnieda: Ah yes, makes sense. Thanks! It can be simplified to `jobPlacementsPanel.setBoardLocation(getSelectedBoardLocation());`. I'll make that change.

src/main/java/org/openpnp/gui/JobPlacementsPanel.java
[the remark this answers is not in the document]
vonnieda: Good point, I didn't notice that. Works for me!

src/main/java/org/openpnp/machine/reference/ReferenceMachine.java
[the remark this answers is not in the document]
vonnieda: Yes, it will only apply to one. We could add it as a machine property, and set it as a default from within the strip feeder, but it needs to live there one way or another. I won't put feeder implementation specific settings in the machine UI. It is a recipe for clutter.

src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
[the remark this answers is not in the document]
vonnieda: Yep, understandable. We'll see how it goes.

src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
vonnieda: Okay, I see now that it uses partId, rather than placementId. That is better, but I still can't allow this change in until users can change it themselves. This completely changes the order that jobs progress in, and I can't make that decision for people who have been using OpenPnP for a long time and who depend on it being ordered by height.

My goal is to eventually allow users to set order themselves, in which case the point is moot.

If you feel strongly about this, you could add a setting to the ReferencePnpJobProcessor wizard so that users can select how the ordering happens.

aneox: @vonnieda 
As KarlZeilhofer said, good idea to add partHeight column and use it to sort by default. Then user can choose themself.

src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
aneox: I didnt made main correction to backoff movement. It should be done without backslash, but i didnt find any move function without backslash. It will be nice to do it. Now iam trying to make mgrl autoFeeders, maybe later i will return to drag feeder improvements.

About is0402, it depends on package name, not part name. If user want use drag feeder with 0402, he can rename package). 0201 easy to add same way.
Instead of adding this checkbox, we can add "parts per drag" input and it add more features. For example drag 4 sop8 ICs, then pick it one by one, then drag again.

src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
vonnieda: Hi @aneox and @KarlZeilhofer: I've tested the code twice and it does not follow the table sorting, it only follows the partId sorting. I don't see how it could. The only change you made was to change the sort in the JobProcessor from partHeight to partId. 

The table is a Swing view, and the sorting is only reflected in the view. It does not reorder anything in the model. The only way to get the sort order from the table is to query it directly, and the JobProcessor can't do that. 

If I've missed something, please let me know. Perhaps there was a change you didn't submit?

Here is a video showing the code from this PR with three different sorts. As you can see, the only sort applied is the partId.

https://www.youtube.com/watch?v=vOBkw_ljq44&feature=youtu.be

aneox: @vonnieda 
Yop, thank you for testing and i am sorry for confusion. I was using only part column sorting and it seemed like it works by table sorting.

src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
KarlZeilhofer: As Jason said, doing this based on a package name is a bad idea. We for example use metrical packages, e.g. 1005 for your 0402. A "parts per drag" seems to be a good option, defined in each feeder instance. 

@aneox For our "Heapfeeder", https://www.youtube.com/watch?v=rYFK_RLNU-g we also had to move the head without backlash compensation. 

This is what we use in our code, it's a hack, but it works: 
```
    // skips the backlash compensation
    // TODO 4: we assume millimeters here
    // TODO 4: only GcodeDriver is supported
    private void directMoveTo(HeadMountable generalHm, Location location) throws Exception{
    	directMoveTo(generalHm, location, getMachine().getSpeed());
    }
    
    private void directMoveToAtSafeZ(HeadMountable generalHm, Location location) throws Exception{
    	generalHm.moveToSafeZ();
    	double safeZ = generalHm.getLocation().getZ();
    	directMoveTo(generalHm, location.derive(null,  null,  safeZ, null));
    	directMoveTo(generalHm, location);
    }
    
    private void directMoveTo(HeadMountable generalHm, Location location, double speed) throws Exception{
    	ReferenceHeadMountable hm = (ReferenceHeadMountable) generalHm;
    	
    	ReferenceDriver driver =  getMachine().getDriver();
    	Location lHead = location.subtract(hm.getHeadOffsets()).convertToUnits(LengthUnit.Millimeters);

    	if(driver instanceof GcodeDriver) {
    		GcodeDriver d = (GcodeDriver) driver;

    		// TODO 4: fix this hardcoded axis handling
    		if(hm instanceof ReferenceCamera) {
        		d.moveTo(hm, lHead.derive(null,  null,  0.0,  null), speed, true); // TODO 3: fix save-z constant
    		}else {
    			d.moveTo(hm, lHead, speed, true);
    		}
    		
    	} else {
    		hm.moveTo(location);
    	}
    }
    
    // wrapper function, using backlash compensation, if configured
    private void preciseMoveTo(ReferenceHeadMountable hm, Location location) throws Exception{
    	hm.moveTo(location);
    }
```

We extended the interface in the GCodeDriver by a `skipBacklashCompensation`: 
```
	public void moveTo(ReferenceHeadMountable hm, Location location, double speed, boolean skipBacklashCompensation)
			throws Exception {
```

You can find the current code there:
https://github.com/KarlZeilhofer/openpnp/tree/feature/heapfeeder

aneox: @KarlZeilhofer thank you moveTo with skipBacklashCompensation good idea!

PR #746, opened 2018-07-10
Backoff movement without backslash compensation.

aneox:
To avoid small parts and parts in plastic tapes from bouncing.

Done in Gcode driver. Exclude Gcode lines with Backslash variables.

### DIFF
--- a/src/main/java/org/openpnp/machine/openbuilds/OpenBuildsDriver.java
+++ b/src/main/java/org/openpnp/machine/openbuilds/OpenBuildsDriver.java
@@ -230,6 +230,12 @@ public class OpenBuildsDriver extends AbstractSerialPortDriver implements Runnab
         }
     }
 
+    public void moveTo(ReferenceHeadMountable hm, Location location, double speed, boolean skipBacklashCompensation)
+            throws Exception {
+
+        moveTo(hm, location, speed, false);
+    }
+
     /**
      * Returns 0 or 1 for either the first or second Nozzle.
      * 
--- a/src/main/java/org/openpnp/machine/reference/ReferenceDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceDriver.java
@@ -59,6 +59,7 @@ public interface ReferenceDriver extends WizardConfigurable, PropertySheetHolder
      * @throws Exception
      */
     public void moveTo(ReferenceHeadMountable hm, Location location, double speed) throws Exception;
+    public void moveTo(ReferenceHeadMountable hm, Location location, double speed, boolean skipBacklashCompensation) throws Exception;
 
     /**
      * Returns a clone of the HeadMountable's current location. It's important that the returned
--- a/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
@@ -544,8 +544,9 @@ public class GcodeDriver extends AbstractSerialPortDriver implements Named, Runn
                 String newCommand = "";
 
                 for (int i = 0; i < commandsArray.length; i ++) {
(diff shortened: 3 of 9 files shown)


KarlZeilhofer:
I think, this should be fine for a quick fix. But perhaps there should be something like an "options object". Then the function prototypes must not be changed in future. If a driver doesn't support an option, it is just ignored. But if it shoud, it has to be implemented for that specific driver. 

something like
```
class DriverOption
{
  enum OptionType{SkipBacklashCompensation, ForceSaveZ, PlanAhead, ...} optionType;
  enum ValueType{bool, int, float, string} valueType;
  boolean boolValue;
  int intValue;
  double floatValue;
  String stringValue;

  void setValue(OptionType type, int value)¬
    this.optionType = type;
    this.intValue = value;
  }
  ...
}
```

vonnieda:
Hi @aneox,

Thank you for the PR, but unfortunately I will not merge this as it is. The modification to the driver interface is too specific. As @KarlZeilhofer said, it would be better to make this more extensible so that future changes do not require additional updates to every driver.

To give some background: There are many people who use a custom driver with OpenPnP. This change would mean that their code will break the next time they upgrade OpenPnP. I understand that we may need to do that from time to time, but I don't feel that this is a good enough reason to release a breaking change.

If you feel that this is critical to the operation of your machine, let's explore some ways we can do it without changing the driver interface.

Thanks,
Jason

aneox:
@vonnieda Yeah, alright. But this bug should be fixed. May be you can advise me how to do it in correct way?

vonnieda:
@aneox I think @KarlZeilhofer's idea of an options object would work, but I think we can simplify it a bit. How about something like this:

Add:
```
public interface MovementOption {
    // this is just a marker interface https://en.wikipedia.org/wiki/Marker_interface_pattern
}

public class SkipBacklashCompensationMovementOption implements MovementOption {
}

public class SkipSquarenessCompensationMovementOption implements MovementOption {
}
```

Add:
```
Movable.public void moveTo(Location location, double speed, MovementOption...) throws Exception;
```

Then you could do something like:
`nozzle.moveTo(location, 1.0, SkipBacklashCompensationMovementOption);`

Or even multiple options in a call:
`nozzle.moveTo(location, 1.0, SkipBacklashCompensationMovementOption, SkipSquarenessCompensationMovementOption);`

This is a pretty simple change, and it's extensible. We can easily add additional options in the future without changing the interface again.

Additionally, since the interface is just a marker, we can easily implement more complex options for people who develop their own drivers. For instance, you could have:

```
public class ChangeAccelerationMovementOption implements MovementOption {
    public ChangeAccelerationMovementOption(double acceleration) {
    }
}

nozzle.moveTo(location, 1.0, new ChangeAccelerationMovementOption(.3));
```


What do you think? If you think this is reasonable, and will work for you, let me know and I'll implement it.

KarlZeilhofer:
Sounds great with variable arguments and an interface class. Didn't use the former feature in Java. Here is some short example: 
https://stackoverflow.com/questions/2330942/java-variable-number-or-arguments-for-a-method#2330958

Who is implementing it?

vonnieda:
If @aneox thinks it will work for their use case, I'll go ahead and implement it.

aneox:
@vonnieda 
i am embedded c programmer and not familiar with java, especially variable arguments count functions.
But if it safe, it is the best way to fix backoff movement!

vonnieda:
Sounds good - I'll get this implemented in the next week or so.

aneox:
@vonnieda i have arrange several board with this fix, drag feeder is working more stable.
When you will have time to implement this?

vonnieda:
Hi @aneox, as with everything, unfortunately, the answer is "when I have time". This is one of my three top priority tickets so I'll get to it as soon as I can.

vonnieda:
It took a long time, but we now have this option available in #998

PR #747, opened 2018-07-10
continue of PR #740

aneox:
- Added Job order combobox to job processor wizard
- Added checkbox to strip feeders, to use different approach of rotation https://github.com/openpnp/openpnp/issues/715

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -86,6 +86,9 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
         Reset
     }
 
+    public static final String ORDER_JOB_PARTHEIGHT = "Part Height";
+    public static final String ORDER_JOB_PART = "Part";
+
     public static class PlannedPlacement {
         public final JobPlacement jobPlacement;
         public final Nozzle nozzle;
@@ -109,6 +112,9 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
 
     @Attribute(required = false)
     protected boolean parkWhenComplete = false;
+
+    @Attribute(required = false)
+    protected String jobOrder = "Part Height";
 
     private FiniteStateMachine<State, Message> fsm = new FiniteStateMachine<>(State.Uninitialized);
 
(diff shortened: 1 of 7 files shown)


src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
```diff
@@ -426,10 +432,26 @@ protected void doPlan() throws Exception {
...
+            jobPlacements = getPendingJobPlacements().stream()
+                    .sorted(Comparator.comparing(JobPlacement::getPartHeight))
+                    .collect(Collectors.toList());
+        } else {
+            if (this.jobOrder == ORDER_JOB_PART) {
+                // Get the list of unfinished placements and sort them by part.
+                jobPlacements = getPendingJobPlacements().stream()
+                        .sorted(Comparator.comparing(JobPlacement::getPartId))
+                        .collect(Collectors.toList());
+            } else {
```
vonnieda: This seems to be a duplicate of the above code?

src/main/java/org/openpnp/spi/Machine.java
```diff
@@ -163,6 +163,10 @@
 
     public boolean getHomeAfterEnabled();
 
+    public boolean getUsePickRotationInsteadOfRotationInTapeForStripFeeders();
```
vonnieda: As I said before, this does not belong on Machine. The property should be on the strip feeder it belongs to. If you want a global default, you can use machine.setProperty() and machine.getProperty() from the ReferenceStripFeederWizard.

src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
aneox: it is like default value if string is wrong

aneox: anyone can easy expand list of sort directions

src/main/java/org/openpnp/spi/Machine.java
aneox: ok, i didnt know about machine.setProperty() and machine.getProperty(), will fix it

Phara0h: Can this be renamed to something that is not a sentence? like "isPickRotation()"

aneox:
Hi @vonnieda !
Is this fix correct?

vonnieda:
* Merged the job order checkbox.
* Rejected strip feeder rotation change. See https://github.com/openpnp/openpnp/issues/715#issuecomment-441534350.

PR #748, opened 2018-07-11
Check that machine is started before sending gcode commands from console

PeeJay:
Signed-off-by: Paul Jones <paul@pauljones.id.au>


vonnieda:
Nice improvement, thanks @PeeJay!

PR #750, opened 2018-07-12
Nozzle Tip: Dwell Time

sp-apertus:
# Description
The Dwell Pick and Dwell Place times have been added to Nozzle Tip class and the Nozzle Tip Wizard GUI.
I also added a note that the total dwell time is the sum of both nozzle dwell time and nozzle tip dwell time.

I have found no reference of where/how getPlaceDwellMilliseconds() is actually used (in job planner or machine driver?) so I need some guidance before I can finish this feature.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,17 @@
 This file lists major or notable changes to OpenPnP in chronological order. This is not
 a complete change list, only those that may directly interest or affect users.
 
+
+# 2018-07-13
+
+* Dwell Times per Nozzle Tip
+
+	Pick dwell time and place dwell time has been added to nozzle tip. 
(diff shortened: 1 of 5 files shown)


vonnieda:
They are used in ReferenceDriver.pick and .place. Easy way to find this
kind of stuff (in Eclipse) is to right click the property and choose
References -> In Project. In this case, it uses the variable directly
(which is a little sloppy on my part) rather than the getter, so you have
to search for both the getter and the variable.

Jason

sp-apertus:
Thanks!

All done and tested with nulldriver simulation.

vonnieda:
Great work as usual @sp-apertus, thank you for making this easy to merge :)

sp-apertus:
as always: my pleasure :)

Issue #677, opened 2017-11-20
slash / backslash in partname


sp-apertus:
# Problem Report
## Describe the Issue
When the board contains a part with a slash or backslash in the name the template creation will fail, namely WritePartTemplateImage will not report an error but simple not create any template images.

vonnieda:
Makes sense, should be a simple fix to fix up the incoming name. We should probably just remove all punctuation aside from period, dash and underscore.

PR #761, opened 2018-08-18
Import package names #677

pfried:
# Description
If a user imports a boardfile from eagle openpnp takes the name. As the name is later used to create a path some characters cause trouble. This code removes those characters on import.

# Justification
Bug #677 

# Merge commits
Was the development branch rebased at some point? I have 7 commits here of which I do not know why they are there. But I can rebase to the latest development branch of course. Just curious why they are there


vonnieda:
Hi @pfried, thanks for the PR, but we need to find a way to do this in the actual importer code, and not in the Eagle XML code. The reason for that is that the Eagle XML code is generated from the XML and will be re-generated for new versions. When it's re-generated, your changes will be lost.

So, the changes will need to be made in https://github.com/openpnp/openpnp/blob/develop/src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java so that we don't lose them.

pfried:
I see, did not know that this was auto generated, but makes sense. I will redo it in a different location then.

PR #762, opened 2018-08-19
Clear search field on new feeder created, fixes #714

### LINKED ISSUE #714: Error Message: "Row Index out of Range", on creating new Feeder with Search List
# Problem Report
## Describe the Issue
Error message appears: "Row index out of range", when in the feeders tab a search term is active (list shows only search results) and the user adds a new feeder. 

## Steps to Reproduce
How can a developer reproduce the issue you are experiencing? Use a step by step format like this:
1. Start OpenPnP
2. Open Feeders Tab
3. Enter some search term, showing only a sub set of all feeders
4. Click on "Add new feeder"

## Expected Result
Good Question. Either no message, and the feeder ist most probably hidden due to the sear term, or clear the search term, and select/highlight the newly created feeder. 

## Actual Result
Error message apperas. New feeder ist still hidden. The user perhaps cannot interpret the error message, and creates another feeder, with same result...

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56411493-error-message-row-index-out-of-range-on-creating-new-feeder-with-search-list?utm_campaign=plugin&utm_content=tracker%2F429103&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F429103&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pfried:
Also Correct wrong javadoc param

# Description
Clears the feeder search textfield when a new feeder is created so it can be selected in the list, avoids the index error filed in #714

# Justification
Its best to show the new feeder in the list. If it is not visible the user might think it has not been created

# Implementation Details
Clears the textfield and triggers search

### DIFF
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -223,7 +223,7 @@ public class FeedersPanel extends JPanel implements WizardContainer {
      * Activate the Feeders tab and show the Feeder for the specified Part. If none exists, prompt
      * the user to create a new one.
      * 
-     * @param feeder
+     * @param part
      */
     public void showFeederForPart(Part part) {
         mainFrame.showTab("Feeders");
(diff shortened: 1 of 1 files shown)


vonnieda:
Look good, thanks @pfried!

PR #767, opened 2018-09-18
OpenBuildsDriver, GcodeDriver and LinuxCNC disconnect when disabled.

markmaker:
# Description
With this Pull Request the OpenPNP drivers OpenBuildsDriver, GcodeDriver and LinuxCNC disconnect when the Machine is disabled.

# Justification
The disconnect allows for a serial port to be reused in another program (i.e. Pronterface) or unplugged, or the controller reset without closing OpenPNP. If the serial port becomes available again, the machine can be enabled again and the connection is successfully re-established.

Before this change this failed at least under Windows where a COM port name is blocked until explicitly closed by the process. Even resetting/re-plugging the controller will not free the port and also not re-establish the connection as long as a process sits on it. OpenPNP has to be closed and the controller reset/re-plugged one more time, before a freshly started OpenPNP can then reconnect.

When closing OpenPNP the user loses the GUI navigation which is annoying when editing Gcode fragments. Along with Gcode editing, users may want to test Gcode on Pronterface (or similar) or need to change config.txt on Smoothieware controllers, which requires a reset of the controller. As described above, a reset was only possible after exiting OpenPNP (at least on Windows).

# Instructions for Use
The disconnect can be used through the GUI Start/Stop button. The colour of that button will now also show the state of the connection. 

# Implementation Details
1. The code changes are small and symmetric to the enable/connect case. I added the disconnect at the end of the method after everything has been taken care of, including sub-drivers recursion in GcodeDriver. I tested the feature with the GcodeDriver plus serial port on Windows. I was able to connect/disconnect using the Start/Stop button. I alternated between using Pronterface and OpenPNP without any problems. I tested resetting, unplugging and unpowering the controller with no problems while the machine was disabled. _I wish I had this when I setup my machine :)_
3. I did not change code inside the `org.openpnp.spi` or `org.openpnp.model` packages.

I assume the justification is equally valid for OpenBuildsDriver, LinuxCNC and TCP/IP connections. However it would have to be tested by someone who has such a setup.

### DIFF
--- a/src/main/java/org/openpnp/machine/openbuilds/OpenBuildsDriver.java
+++ b/src/main/java/org/openpnp/machine/openbuilds/OpenBuildsDriver.java
@@ -72,6 +72,9 @@ public class OpenBuildsDriver extends AbstractReferenceDriver implements Runnabl
                 pump(false);
             }
         }
+        if (connected && !enabled) {
+            disconnect();
+        }
     }
 
     @Override
--- a/src/main/java/org/openpnp/machine/openbuilds/OpenBuildsDriver.java
+++ b/src/main/java/org/openpnp/machine/openbuilds/OpenBuildsDriver.java
@@ -73,7 +73,9 @@ public class OpenBuildsDriver extends AbstractReferenceDriver implements Runnabl
             }
         }
         if (connected && !enabled) {
-            disconnect();
+            if (!connectionKeepAlive) {
+            	disconnect();
+            }
         }
     }
 
--- a/src/main/java/org/openpnp/machine/reference/driver/AbstractReferenceDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/AbstractReferenceDriver.java
@@ -32,6 +32,9 @@ public abstract class AbstractReferenceDriver extends AbstractModelObject implem
 
     @Attribute(required = false, name = "communications")
     protected String communicationsType = "serial";
+    
+    @Attribute(required = false)
+    protected boolean connectionKeepAlive = false;
 
     /**
      * TODO The following properties are for backwards compatibility and can be removed after 2019-07-15. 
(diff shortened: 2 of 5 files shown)


vonnieda:
This causes problems on Mac (at least) with certain controllers. The issue is that when you "open" the port, DTR is toggled, which for many Arduino style controllers resets the controller. This causes it to lose it's position where a normal enable/disable does not.

I found this to be a problem during early development, but I don't know if it's actually something that would bother users now. So, with that in mind, how about adding it as a default on checkbox option?

markmaker:
> So, with that in mind, how about adding it as a default on checkbox option?

OK, I'll probably need some coaching on this. 

I've tried in the last merge.  

Done it in AbstractReferenceDriver, so only the OpenBuildsDriver and GcodeDriver are covered, not LinuxCNC. If I understand the problem correctly, it is not affecting LinuxCNC anyway.

I directly coded the GUI in Java, hoping the [promise of WindowBuilder to be bi-directional](https://www.eclipse.org/windowbuilder/) holds true.

It was a bit trial and error for the JPanel.add() constraints. I haven't found any documentation for that in reasonable time. I have trouble browsing in Eclipse and finding proper documentation for implementations/overrides and polymorphism. While I can browse class hierarchies I have yet to find a "find all overrides of this method" browse command, had to search for those with text search like in the stone age. Hints are welcome. 

# Tests

- Tested Checkbox GUI. 
  
grafik

- Tested Apply, quitting, seeing the option restored.
- Located it in machine.xml
- Tested with open "Keep Alive" enabled, disabled Machine, reset controller, tried to enable the Machine, confirmed COM port became broken again:
grafik

- Ran `mvn test`.

Please review and advise :), Thanks.

_Mark

vonnieda:
This looks great @markmaker, thank you for making the requested changes. I'm swamped at the moment, so give me until the weekend to test both this and #766 and get them merged.

vonnieda:
Thanks @markmaker!

vonnieda:
Quick note: In https://github.com/openpnp/openpnp/commit/b3ae1ae3feb0c56bb0ab5d39ca153611cc5e7a10 I changed the keepalive option to default to on. Without this change every user that updates would get the new behavior and I don't want to surprise people. They can turn it off if they want the new behavior.

markmaker:
Agree on the default. Thanks!

Issue #769, opened 2018-09-27
NozzleOffsetWizard


pfried:
# Problem Report
## Describe the Issue
The nozzle offset wizard has an issue where a previous present offset is not consided correctly.

## Steps to Reproduce
NozzleOffsetWizard with present offset configuration

## Expected Result
Offset must be taken into account for the calculation

## Solution
subtract offset when storing the position

I will do a pull request soon

PR #770, opened 2018-09-27
Nozzle offset wizard #769

pfried:
# Description
Fixes an issue where the nozzle to head offset is not calculated correctly if an offset is already entered

# Justification
Calculation is wrong if using already corrected value

# Implementation Details
I did test on the machine as this happened to me and I figured out what the issue was since getLocation on the nozzle returns the corrected position already


vonnieda:
Hi @pfried, it looks like PR contains a whole new OpenPnP.install4j file for some reason. Can you resubmit without this please? See https://github.com/openpnp/openpnp/pull/770/files

pfried:
hey @vonnieda Actually those files were change by you / travis. I am not sure how this stuff always ends up in my PRs, since I usually branch from openpnp/develop directly

vonnieda:
@pfried Make sure you do a fetch and then pull before committing. This will sync your branch with the latest in the main branch and should avoid this issue. I think that if you do a pull from upstream now, it will merge that change in and then you can just commit and push and the diff should go away.

PR #771, opened 2018-09-29
Fixing an issue where an already present head offset will lead to wro…

pfried:
# Description
Fixes an issue where the nozzle to head offset is not calculated correctly if an offset is already entered

# Justification
Calculation is wrong if using already corrected value

# Implementation Details
I did test on the machine as this happened to me and I figured out what the issue was since getLocation on the nozzle returns the corrected position already


vonnieda:
Thanks @pfried!

PR #785, opened 2018-10-15
Improve display of scripting exceptions 

Crazor:
# Description
Jython exceptions with messages like `NameError: name 'prit' is not defined in <script> at line number 2` got truncated in the GUI message box because of the angle brackets. Also, since the script's filename was not populated prior to execution, the error message unhelpfully stated that the error occured in `<script>`.

# Justification
The change improves script error handling for developers and users who did not set the loglevel to debug prior to execution.

# Instructions for Use
Just execute any script that throws an exception.

# Implementation Details
I tested with a bunch of broken and exception-raising Python scripts. I could not find other causes for truncation of the messages.


vonnieda:
Nice work, thanks @Crazor!

PR #786, opened 2018-10-15
serial fix

aneox:
My system based on Mac has 10+ serial devices connected, but openpnp with SerialPortList.getPortNames() can see only 6 of them.
This patch worked for me, now i see all devices and can easily work with them, including /dev/tty.Bluetooth...


vonnieda:
@Crazor @aneox It seems we have two competing PRs for a similar feature. Can you two work out one PR that works for both cases and resubmit please?

Crazor:
I vote for this one to be merged, as it works for my case as well. I didn't like seeing the bt etc. ports, but since @aneox is actually using bluetooth to connect, I now think my taste is less important than having usable code upstream.

aneox:
@vonnieda 
https://github.com/openpnp/openpnp/pull/786
works for both cases, no reason to resubmit, @Crazor already agreed in comments day ago

vonnieda:
Great, thank you both!

PR #787, opened 2018-10-16
subDriverError

aneox:
If some gcode subdriver is disconnected, machine blocks. With missing subdriver i cant simply jog to home and any other job. If we had several subdrivers and remove it temporally, missing of them is not reason to block machine. Error message on start is enought i think. Catch of exception fix this.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
@@ -20,6 +20,7 @@ import javax.swing.Action;
 import javax.swing.JOptionPane;
 
 import org.openpnp.gui.MainFrame;
+import org.openpnp.gui.support.MessageBoxes;
 import org.openpnp.gui.support.Icons;
 import org.openpnp.gui.support.PropertySheetWizardAdapter;
 import org.openpnp.machine.reference.ReferenceActuator;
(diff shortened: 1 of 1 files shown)


vonnieda:
@aneox I disagree with this patch. The sub-driver could implement some critical part of the system, like vacuum checks, and then the user may harm their machine. What is the reason that you would run the machine without all the controllers attached?

aneox:
@vonnieda In my case i use changeable feeders block with own controller. May be you are right and in some case problems may occur. 

If replace error message with error dialog and ask user to continue or stop? Is it safe enough?

vonnieda:
@aneox So when you change out a feeder block the controller goes away? And there are other controllers that then become active? Do these all use different USB ports?

aneox:
@vonnieda machine has 1 USB port for these feeders. I change these feeder rarely, some jobs need this, some not. Sometimes i use only strip feeders and disconnect another controllers to have more free space. There is no option to disable driver and machine blocks.

vonnieda:
Hi @aneox - no, we can't have a dialog from a driver. Putting UI code in the reference or spi packages is forbidden because it breaks the testing code, which does not run within the GUI.

In addition, if you allow the driver to fail to connect, but then OpenPnP sends it commands, it may throw other exceptions since OpenPnP always expects commands sent to all drivers to complete.

I think this change is too specific to your machine and not generally applicable. You should maintain this change in your own branch, or come up with another way to make it more generally applicable.

PR #788, opened 2018-10-16
Add support for serial ports named tty.wchusbserial* on macOS.

Crazor:
# Description
WCH's CH340 chips show up on macOS as `/dev/tty.wchusberial*`. I added a search pattern to the macOS defaults from jssc.

# Justification
WCH chips are common cheap USB serial converters. Users should be able to select those on macOS.

# Instructions for Use
Connect a WCH based USB serial converter and pick it from the list of serial ports in the machine setup.

# Implementation Details
I tested the code with a bunch of USB serial adapters from different manufacturers on macOS. I changed the if/else to a switch statement for cleanliness, which should not affect the Linux case at all. Though I have technically touched but not tested the Linux case ;)

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/SerialPortCommunications.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/SerialPortCommunications.java
@@ -131,21 +131,23 @@ public class SerialPortCommunications extends ReferenceDriverCommunications {
      * @return array of Strings of serial port names
      */
     public static String[] getPortNames() {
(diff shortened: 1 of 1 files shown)


aneox:
Ahah)) https://github.com/openpnp/openpnp/pull/786

vonnieda:
@Crazor @aneox It seems we have two competing PRs for a similar feature. Can you two work out one PR that works for both cases and resubmit please?

PR #792, opened 2018-12-02
Update the Move Camera to Next Placement button.

smdood:
# Description
Change the functionality of the Move camera to next placement button.

# Justification
Old behavior: The camera would move to the current location and the highlighted part would move to the next part. Then when you want to adjust the placement location you must select the part you are actually looking at as the next part is already selected in the placements panel.

New behavior: The highlighted part is moved and then the camera is moved to its location.
This way, the part in the camera view and the part highlighted in the parts list are the same. Then if you adjust the placement, when you press the "Set placement to current camera location" etc the correct part will be edited instead of the next part in the list.

# Instructions for Use
To quickly move the camera to the next placement in the placement table just press the "Move camera to next placement location".

This way you can quickly move around the job to check all placement locations just using one button.

# Implementation Details
To test, I compiled and ran openpnp and tested the function of the button I modified.
I think so..
No Changes here
Ok

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
@@ -414,9 +414,9 @@ public class JobPlacementsPanel extends JPanel {
     public final Action moveCameraToPlacementLocationNext = new AbstractAction() {
         {
             putValue(SMALL_ICON, Icons.centerCameraMoveNext);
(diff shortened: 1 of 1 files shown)


smdood:
See what you think of this change to the function of the "Move camera to next" function.
I intend to update the other instances of this button to behave the same.

I hope I did everything correctly!

smdood:
Oops, that's what that button does!! 
Reopening!

vonnieda:
Makes sense to me, thanks for the PR!

Issue #795, opened 2018-12-21
Manual for automatic nozzle offset wizzard


mgrl:
Just a question related to the nozzle offset helper. Great addition, it helps to get the math right and makes the routine very simple.

My question is to step 4:
I use flour to get a mark on the table. Since we have some runout on the nozzle tips, I turn the nozzle in the flour at least by 360° to get the real center of the runout tainted nozzle tip. 
Am I right in doing it that way and it would be an improvement to add that advice to the wizzard?

vonnieda:
Sounds like a good idea to me. Maybe you could make a short video or GIF showing it and we could link to that as well?

pfried:
We could integrate the 360 turn thing (maybe a 360degree spin button), sounds like a good idea to me. I did the wizard, so maybe I can do a tutorial at some point (could be some weeks)

mgrl:
okay, updated the wiki according to the new steps: https://github.com/openpnp/openpnp/wiki/Setup-and-Calibration:-Nozzle-Setup#head-offsets

for me it's okay like it is. maybe I make a photo and add it to the wiki next days (flour is still on the pcb ;)

regards, michael

mgrl:
finally just for reference. I just found that the 360° degree advice makes sense to get the machine better calibrated for accuracy than without. Its in conjunction to the findings here:

- https://groups.google.com/d/msg/openpnp/16YV9sPXPFM/WIaTbpIkBgAJ
- https://groups.google.com/d/msg/openpnp/16YV9sPXPFM/8kiTyXxGBgAJ

vonnieda:
Nice work @mgrl!

PR #796, opened 2019-01-04
Added an uri to the nozzle wizard + 360° nozzle advice

mgrl:
# Description
as discussed in #795 I added an advice to turn the nozzle while measuring offset + added an link to the wiki.

# Justification
users may find the wiki link helpful and are directly pointed to further information. a video/gif could then be added to the wiki

# Instructions for Use
nothing changed in functionality.

# Implementation Details
compiled in eclipse and started openpnp, opened the wizard, looking good.

modded offset tool


mgrl:
grafik
for the wiki

PR #805, opened 2019-01-17
add isHomed

mgrl:
# Description
as mentioned in #804 I added an ishomed flag

# Implementation Details
live on machine and mvn
unfortunately I had to reformat the file due to some whitespace issues... I will fix the curly brakets later.

I felt it would be better to have the isHomed flag at the head, right? shall I change that, oder is it okay as it is?


mgrl:
doing this in #806

PR #806, opened 2019-01-17
Updated isHomed [OpenPnP formatted]

mgrl:
# Description
as mentioned in #804 I added an ishomed flag

# Implementation Details
live on machine and mvn
unfortunately I had to reformat the file due to some whitespace issues... I will fix the curly brakets later.

I felt it would be better to have the isHomed flag at the machine, right? shall I change that, oder is it okay as it is?

-due to the application of the automatic formatter some old stuff is shown as diff... its just another format, nothing more...

### DIFF
--- a/src/main/java/org/openpnp/gui/MachineControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/MachineControlsPanel.java
@@ -272,7 +272,7 @@ public class MachineControlsPanel extends JPanel {
 					Configuration.get().getMachine().setEnabled(enable);
 					setEnabled(true);
 					if (machine.getHomeAfterEnabled() && machine.isEnabled()) {
-						selectedTool.getHead().home();
+						machine.home();
 		                homeAction.putValue(Action.SMALL_ICON, Icons.home);
 					}
                 }
@@ -297,7 +297,8 @@ public class MachineControlsPanel extends JPanel {
         @Override
         public void actionPerformed(ActionEvent arg0) {
             UiUtils.submitUiMachineTask(() -> {
-                selectedTool.getHead().home();
+                Machine machine = Configuration.get().getMachine();
+                machine.home();
                 homeAction.putValue(Action.SMALL_ICON, Icons.home);
             });
         }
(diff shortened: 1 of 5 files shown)


vonnieda:
Hi @mgrl, I'm curious why you implemented this in ReferenceHead instead of ReferenceMachine? As this code is now, if one head succeeds but another fails isHomed will still be set true. I think this should be done in ReferenceMachine.home(). If that's okay with you I'll change and merge.

Thanks,
Jason

mgrl:
Or maybe ishomed has to be stored to every head and ishomed in machine iterates over all heads?

mgrl:
@vonnieda okay, just tested with the nulldriver:

the home() function in reference machine and
```
    @Override
    public void home() throws Exception {
        Logger.debug("home reference machine");
        super.home();
    }
```

its pendant in abstract machine never get called in the homing cycle.
```
    @Override
    public void home() throws Exception {
        Logger.debug("test");
        for (Head head : heads) {
            head.home();
        }
    }
```

is this a bug?

src/main/java/org/openpnp/gui/MachineControlsPanel.java
```diff
@@ -272,7 +272,7 @@ public void actionPerformed(ActionEvent arg0) {
 					Configuration.get().getMachine().setEnabled(enable);
 					setEnabled(true);
 					if (machine.getHomeAfterEnabled() && machine.isEnabled()) {
-						selectedTool.getHead().home();
```
mgrl: okay, changed the gui elements to perform a homing cycle on the machine instead directly on the head. with the nulldriver it works as expected now.

mgrl:
well, I found in the calling history that the gui directly calls .home on the head. I changed that in the latest commits. 
should be good then

mgrl:
@vonnieda tested now on machine, works as intended (at least for 1 head - has anybody have two heads? he must have homed both heads separate by choosing a nozzle attached to that head...)

vonnieda:
Great work, thank you @mgrl!

PR #809, opened 2019-01-24
Forced-output coordinate variables "XF", "YF", "RotationF" are omitted in moveToSafeZ()

markmaker:
# Description
If the command has forced-output coordinate variables "XF", "YF", "ZF" and "RotationF", the corresponding axes are always includes in the G-code command (see #766). This may be employed for shared physical axes, Z-probing, relative moves in custom G-code etc. where OpenPNP cannot keep track of when an axis has physically moved behind its back. 
        
There is an exception to the above, if a moveToSafeZ() is executed. The safe Z move does not count as a _real_ move in the sense that the headmountable is selected for _real_ work. Consequently coordinates (other than Z) should not be restored/forced in this case. 

# Justification
Tests of #766 with a shared C axis machine (Marek T's) have shown that unnecessary rotations are executed. This was tracked down to moveToSafeZ() moves on all head-mountables prior to X, Y moves with the actual nozzle. 

See also:
https://groups.google.com/d/msg/openpnp/zCVlqSgxzYk/RwoexxEhDQAJ

# Instructions for Use
When employing forced-output coordinate variables "XF", "YF", "ZF" and "RotationF" for the MOVE_TO_COMMAND these will automatically be omited if a moveToSafeZ() signature is detected. 

# Implementation Details
1. Tests of the change are **pending**, I hope Marek T. can test this with his machine. **This PR is still subject to review and testing**.
3. I had to make a change to the `org.openpnp.spi` package to add a getSafeZ() and setSafeZ() getter/setter to the HeadMountable interface in order to recognize a moveToSafeZ() method call signature. These changes are benign insofar as the existing identical methods in the implementation reference classes only had be annotated as overrides. 

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceActuator.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceActuator.java
@@ -170,10 +170,12 @@ public class ReferenceActuator extends AbstractActuator implements ReferenceHead
         return getName();
     }
 
+    @Override
     public Length getSafeZ() {
         return safeZ;
     }
 
+    @Override
     public void setSafeZ(Length safeZ) {
         this.safeZ = safeZ;
     }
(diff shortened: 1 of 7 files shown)


markmaker:
Closed due to:

> Sorry, I won't merge this one. I firmly believe it's the wrong approach to fixing the issue. I'm not going to add more technical debt for the sake of a quick fix.

https://groups.google.com/d/msg/openpnp/zCVlqSgxzYk/lj1o2QSQDQAJ

PR #812, opened 2019-02-10
More Robust Nozzle Tip Calibration

mgrl:
# Description
Continue to #804. This PR adds
1. automatic removal of nozzle tip offset locations that are above offsetThreshold and
2. sorting the pipeline's result list by linear distance to the bottom camera if multiple results are returned.
3. some minor changes + debug output

# Justification
to 1) wrong configured pipelines may return large offsets -> these are eliminated by this change
to 2) if there are multiple results (though it is recommended to configure the pipeline to have only ONE result), the list is sorted by distance to bottom camera. the location with the smallest distance is taken as valid for further processing. this may improve results, but doesn't have to since there can't be made any educated guesses which is a valid offset location.

# Instructions for Use
There will be added a offsetThreshold value to the GUI later. For now it's a hardcoded variable in the calibration class to prevent cluttering machine.xml.
This PR is tested on the nulldriver only and I would like to use this PR to test (@netzmark) and discuss assumptions and implementation as we did in #804.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTip.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTip.java
@@ -3,6 +3,8 @@ package org.openpnp.machine.reference;
 import java.awt.event.ActionEvent;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -438,7 +440,7 @@ public class ReferenceNozzleTip extends AbstractNozzleTip {
 
             @Override
             public String toString() {
-                return String.format(Locale.US, "0° Offset %f, %f", nozzleTipMeasuredLocations.get(0).getX(), nozzleTipMeasuredLocations.get(0).getY());
+                return String.format(Locale.US, "%i° Offset %f, %f", nozzleTipMeasuredLocations.get(0).getRotation(), nozzleTipMeasuredLocations.get(0).getX(), nozzleTipMeasuredLocations.get(0).getY());
             }
 
             @Override
@@ -787,37 +789,65 @@ public class ReferenceNozzleTip extends AbstractNozzleTip {
             try (CvPipeline pipeline = getPipeline()) {
                 pipeline.setProperty("camera", camera);
                 pipeline.process();
-                Location location;
-                
-                Object result = pipeline.getResult(VisionUtils.PIPELINE_RESULTS_NAME).model;
-                
-                // are there any results from the pipeline?
-                if (0==((List) result).size()) {
-                    throw new Exception("No results from vision. Check pipeline.");                    
-                }
+                List<Location> locations = new ArrayList<>();
 
-                if (result instanceof List) {
-                    if (((List) result).get(0) instanceof Result.Circle) {
-                        Result.Circle circle = ((List<Result.Circle>) result).get(0);
-                        location = VisionUtils.getPixelCenterOffsets(camera, circle.x, circle.y);
-                    }
-                    else if (((List) result).get(0) instanceof KeyPoint) {
-                        KeyPoint keyPoint = ((List<KeyPoint>) result).get(0);
-                        location = VisionUtils.getPixelCenterOffsets(camera, keyPoint.pt.x, keyPoint.pt.y);
-                    }
-                    else {
-                        throw new Exception("Unrecognized result " + result);
-                    }
-                }
-                else {
-                    throw new Exception("Unrecognized result " + result);
-                }
+                Object results = pipeline.getResult(VisionUtils.PIPELINE_RESULTS_NAME).model;
 
                 //show result from pipeline in camera view
                 MainFrame.get().get().getCameraViews().getCameraView(camera).showFilteredImage(
                         OpenCvUtils.toBufferedImage(pipeline.getWorkingImage()), 1000);
                 
-                return location;
+                // are there any results from the pipeline?
+                if (0==((List) results).size()) {
+                    throw new Exception("No results from vision. Check pipeline.");                    
+                }
+
+                // add all results from pipeline to a Location-list post processing
+                if (results instanceof List) {
+                    for (Object result : (List) results) {
+                        if ((result) instanceof Result.Circle) {
+                            Result.Circle circle = ((Result.Circle) result);
+                            locations.add(VisionUtils.getPixelCenterOffsets(camera, circle.x, circle.y));
+                        }
+                        else if ((result) instanceof KeyPoint) {
+                            KeyPoint keyPoint = ((KeyPoint) result);
+                            locations.add(VisionUtils.getPixelCenterOffsets(camera, keyPoint.pt.x, keyPoint.pt.y));
+                        }
+                        else {
+                            throw new Exception("Unrecognized result " + result);
+                        }
+                      }
+                }
+                
+                // remove all results that are above threshold
+                Iterator<Location> locationsIterator = locations.iterator();
+                while (locationsIterator.hasNext()) {
+                    Location location = locationsIterator.next();
+                    if (location.getLinearDistanceTo(0., 0.) > 1) {
+                        locations.remove(location);
+                    }
+                }
+                
+                // sort locations ascending. locations.get(0) is smallest distance to 0/0
+                Collections.sort(locations, new Comparator<Location>(){
+                    public int compare(Location s1, Location s2) {
+                        return Double.compare(s1.getLinearDistanceTo(0., 0.), s2.getLinearDistanceTo(0., 0.));
+                    }
+                });
+                
+                // check for a valid resultset
+                if (locations.size() == 0) {
+                    throw new Exception("No valid results from pipeline within threshold");
+                } else if (locations.size() > 1) {
+                    // one could throw an exception here, but we just log an info for now since
+                    // - invalid measurements above threshold are removed from results already and
+                    // - locations is sorted by distance to the camera. 
+                    // Assumption is that the smallest offset is the valid one.
+                    Logger.info("[nozzleTipCalibration]Got more than one result from pipeline. For best performance tweak pipeline to return exactly one result only.");
+                }
+                
+                // finally return the location at index (0) which is either a) the only one or b) the one with the smallest offset to bottom cam if locations.size>1 
+                return locations.get(0);
             }
         }
 
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTip.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTip.java
@@ -542,7 +542,7 @@ public class ReferenceNozzleTip extends AbstractNozzleTip {
 	    	    kasaB = (kasaD1 - kasaG12*kasaC)/kasaG11/2.0;
 	    	    
 	    	    // assembling the output
-	    	    Double centerX = kasaB + kasaMeanX;
+                Double centerX = kasaB + kasaMeanX;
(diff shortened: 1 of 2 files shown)


vonnieda:
Thanks @mgrl - just so I'm clear, you don't want this merged right away, right? Want to do some testing first?

mgrl:
right, don't merge. I'll give a sign.

@vonnieda a question:
The template matching stage returns a score in its results. From the [code comment](https://github.com/openpnp/openpnp/blob/develop/src/main/java/org/openpnp/vision/pipeline/stages/DetectCirclesHough.java#L51) for detecthoughcircles I guess that the best matches returned from the internal canny are returned first. so the score is implied by the order of the results. is this true for the output of detecthoughcircles then, too? in this case I would remove sorting the list by distance and rely on the higher score as being more apropriate.

netzmark:
Hi
I've tested it today and seems be working properly, and better than without it.
The only suggestion is to move this threshold variable to the machine.xml, what will be needed anyway at the moment of adding this to gui (if happens some day). But even without this I think it's worth to merge it into the develop.

netzmark:
Michael, thx for this!

mgrl:
okay, will make it configurable for sure - still within this PR.

but I think sorting the results by distance is wrong, I will remove this.

do you have your pipeline avail? would like to see if we can improve the default pipeline by combination.

mgrl:
okay, added the value to the gui now:

calibration-panel

mgrl:
@vonnieda: consider this to be ready to merge. if further changes (another default pipeline, etc.) are needed, these would come separate

netzmark:
Hi, I have sent you some email, se it later and answer in free time pls.

About the pipeline still not final success. I can't understand "impact" of param1 and para2 onto the houghCircle process. Don't understand how possible that have circle min=12 and circle max=18 but sometimes get detected circles D=10, some magic for me... still fight.

mgrl:
These are magic for me, too. I try to recreate the pipeline with imageplay which supports realtime processing. Maybe that gives more insight and helps finding pipelines...

mgrl:
So I successfully placed parts with that nozzle you see the calibration in the last post.
runout-nozzle

made a short youtube video: 

[IMAGE ALT TEXT HERE](https://www.youtube.com/watch?v=vTZsT_rND8A)
To test the nozzle runout compensation in OpenPnP I 3dprinted a nozzle tip with 1.5mm runout. The video starts from the park position, calibration run was done before already. It picks up a 0805 resistor from strip feeder and places it on the demo board.

netzmark:
Nozzle calibration is divided into two two offsets finding: finding xy-centerOffsets (distances between phisical center of Z axis to the head declared offsets) and finding the angle runout offsets. Then you sumarize it somehow and apply to the declared ofsettes which are used while sending the nozzle to the bottom vision and while placement.

And here is probably the problem.
The xy-centerOffsets apllying is not needed for those who have ideal machines and every nozzles' axes centers are visible perfectly in center of bottom vision. In this case this offset found as 0/0 is doing nothing.
But for those like me, you, Thomas, if it is found not 0/0 for nozzles other than first (which used used to catch bottom vision xy position will be always perfect obviously), because of different hardware limitations, but like ie 1/1 and added while runout calibration to head offsets - it causes that while alignment the part is brought ideal to centre of vision and vision compensates perfecttly only the part. This is good. But while the placement this xy-centerOffset is also applied and makes effect that part is moved xy from required placement position about this offset. Is placed in position "alignment +1/1". And this is bad of course.

So I think it should work as it is but xy-centerOffset should be removed from offset calculated for the placement. It will be neutral for ideal machines but the only solution for non-ideal machines.

How do you think, does this not sounds sensual?

mgrl:
> Nozzle calibration is divided into two two offsets finding: finding xy-centerOffsets (distances between phisical center of Z axis to the head declared offsets) and finding the angle runout offsets. Then you sumarize it somehow and apply to the declared ofsettes which are used while sending the nozzle to the bottom vision and while placement.

In principle yes. Let's find out how application of the offset works with nozzle runout compensation:

- Nozzle1 has NT1 loaded and calibrated with centerX=1, centerY=2 and runout=0.5 (this would be an offset at rot 0° x=1.5mm y=2)
- Nozzle1 is commanded to bottom cam pos (x10/y10/rot0)
- Nozzle1 goes physically to (x=10-1.5=8.5mm, y= 8mm, rot0)
- OpenPnP shows in it's DRO for Nozzle1 the commanded position, not the physical pos (x10/y10/rot0)
- -> runout compensation should be transparent to the application this way
- the bottom vision while placing parts finds an offset to the (x10/y10)-value, not the physical.
- while placement the runoutcompensation is reapplied to the (x10/y10) value that is now corrected by bottom vision (so not really 10/10 any more).

So there should not be a problem or?

> And here is probably the problem.
> The xy-centerOffsets apllying is not needed for those who have ideal machines and every nozzles' axes centers are visible perfectly in center of bottom vision. In this case this offset found as 0/0 is doing nothing.
Exactly, if the machine would be perfectly calibrated, it should be 0/0. This should be the goal to adjust the machine for if one sees large numbers for the center-values.
> But for those like me, you, Thomas, if it is found like ie 1/1 and added (while runout calibration) to head offset it causes that while alignment the part is brought ideal to centre of vision and vision compensates perfecttly only the part. This is good. But while the placement this xy-centerOffset is also applied and makes effect that part is moved xy from required placement position about this offset. Is placed in position "alignment +1/1". And this is bad of course.

As written above - it should make sense like it is. how could we test that bulletproof, to be sure? is a valid test to pick up a part, decenter it with the finger and to alignment + placement? do you have a better idea?  but from theorie it should not be wrong...
>
> So I think it should work as it is but xy-centerOffset should be removed from offset calculated for the placement.
>
> How do you think, does this not sounds sensual?

We need to provoke high offsets for picked up part, so that the offsets are clearly visible and way higher than any other inaccuracies from machine, etc. Additionally we could add a checkbox to remove the centerXY from calibration if someone wants to only compensate for that...


anyway all this has nothing to do with this PR... we should create an issue if you still think the math is wrong.

netzmark:
I see that first nozzle which is used to catch bottomVison position and this way is 0/0 with bottom vision - it does pnp perfectly. But another nozzle which is moved in accordance to the center of the bottom vision - it places the parts on the board just moved for the xy-centre offset.

CenterXY offset should be found while runout calibration, used while bottmVision alignment but removed from final offset used while placement only. This way it is, as said, neutral for ideal machines and good for non-ideal machines. So don't think it requires to be parametrized (I mean checkbox).

netzmark:
Maybe this is another PR, but note that without nozzle calibration non-ideal machine can be used even if head offsets are not idealy perfect. It can be somehow compensated using deslared head ofsetts. But with full nozzle calibration it is impossible because this calibration eliminates this kind of artificial offsets. So looking on this in this way - it is like the one complete PR :-)

mgrl:
hmmm, so many cases and pitfalls... I'm very willing to remove the centerX and centerY values from the total offset value. It's just changing these two lines:
https://github.com/openpnp/openpnp/blob/020f517c3d445c957e297a4d119d01e85120ad93/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTip.java#L470

have you tested whether it works better without this?  does the liteplacer have this feature in it? maybe we could take a look how it's implemented there...

netzmark:
Yes I have tested it.  
double offsetX = this.radius * Math.cos(angle);
But listen again.
- If calibration is off and N2 is not ideal, not xy-centerOffset 0/0. Then parts is brought to the bottom vision always little moved and then alligned (this center offset and wrong picking). And then wrong placed. BUT I can set artificial not real head offset, part is alligned also artificial bad and then placed good :-). It's stupid but works.
- Now calibration is ON and xyCenter offset is OFF. I can do the same as above having advantage of eliminated runout. So soft can be used cheated with this offset as was before. It is MUCH better than before but not final solution.
- IF calibration is ON, centerXY offset ON for alignment (to have part brought to vision with priorly compensated xyCenter offset) then we get only wrong picking compensation. After allignment this xyCenter offset should be removed and part should be placed properly - in every cases, by ideal and not-ideal machine. Bad news it it is probably needed to be done while bottomVision not in RefNozzleTip.java, isn't it?

netzmark:
it's probably hard for me to explain what I exactly mean due to my language limitation, but I hope you got it.

netzmark:
do you have issue that your N2 has xyCenter offset found while calibration? It seems to me you said you have some serious. So you should also have perfect pnp with N1 but moved with N2.  So maybe we can somehow hardcode this offset subtraction to test the solution...? And later think how to do it finally?

netzmark:
1. I can't compensate changing bottom vision position for each nozzle set other. This position is caught usin N1 and perfect for N1. For others I must cheat system setting little false N2 and N3 offsets.
2. I say about options with "double offsetX = this.radius * Math.cos(angle);", so calibration finds only runout not centerXY offset.
3. Let's not afraid to speak it out loud: I'm too limited with code understanding to be sure how totally your calibration code works :-). But I imagine you collect centerXY offsets and runout ofsetts, at least you display this in Status. So we have this information. So if we could these values (centerX offset and centerY offset) substract from final placement offset after bottomVision - it seems to me it could work how I mean and be universal for every machines. But as said - I can understand it all not too good :-).

vonnieda:
Hi @mgrl, since there has been some more discussion, I just want to confirm that you still want to merge this as is?

mgrl:
Yes, please merge. the discussion is on another topic. Will create an 
separate issue or discussion for this if we need further discussion.

Am 15. Februar 2019 05:33:42 schrieb Jason von Nieda 
<notifications@github.com>:

netzmark:
Ok, I made little code change or script. It allows me to send specific nozzle to specific bottom.camera position, other than declared as head offsets. So no matter what's going on with mechanic the nozzles always are landing in centre of the bottom vision but the placement is basing on the head offsets.
So about the bottom vision I will always have effect like the ideal machine, and no matter whether centerXY pulling is active or not - it must be great with nozzle.calibration now.
From my point of view the improvement we discussed above is no further required. We can wait the feedback when others will start to use it then worry if needed :-).
You should put to the wiki this information to users are warned with this centerXY information displayed in status. It can be frustrating - perfect calibration, perfect allignment and wrong placement in this situation is strange.

mgrl:
I will keep an eye on this topic when using the machine next time...

vonnieda:
Thank you @mgrl!

markmaker:
> So I successfully placed parts with that nozzle you see the calibration in the last post.
> runout-nozzle

Hi @mgrl 

Would it be possible for you to send me this 3D printing nozzle tip model but with zero runout or better yet the underlying geometry? 

I'm developing a feeder that uses the nozzle tip to push the feeder's cover open to prep it for use. Myself I have CP40 nozzle tips that are sturdy enough to do that. With Jukis I'm not so sure, as the spring loaded tip looks fragile on smaller tips. So for people that are reluctant to damage their tips I'd like to have an alternative. On job start the feeder would load the 3D printed "push" tip and then prep all feeders. 

The "push" tip would have to be sturdy enough for lateral forces, but for the lower-most 3mm at the point it needs to narrow down to 4mm in diameter (ideally a conical frustrum).  

Or do you have it in OpenSCAD? Then just send me the file you have :-)

Thanks a lot,
_Mark

CircuitSetup:
@markmaker I actually use Juki nozzles to move tape in my drag feeders currently. (Setup as an actuator) without any issues. My drag feeder is actually the same part used in the Charmhigh machines, so in sure the setup is a bit different, but the same idea.

markmaker:
@CircuitSetup that's good to know, makes things much easier! In fact the 3D printed nozzle tip only works with one sub-type of my new feeder, not with the other, so I'm very glad pushing is also viable with Juki nozzle tips.

I have now already implemented the software part, so I would still be grateful for @mgrl to provide the model. Thanks!

Issue #599, opened 2017-07-19
BottomVision Pre-Rotate Fails on Shared C Heads


vonnieda:
PK reports that the new bottom vision pre-rotate feature fails on their machine which uses a shared motor for the C axes. This is confirmed by other users in the below thread:

https://groups.google.com/d/msgid/openpnp/392748bb-206d-47b8-bea9-351692d78ad6%40googlegroups.com?utm_medium=email&utm_source=footer

Before this is fixed we need to add unit tests for the post-rotate and pre-rotate features to ensure that we're not breaking functionality when this code gets modified. It's too important for it not to be tested.

cri-s:
Marek have confirmed that the fix works for shared Axis. here the diff:
This fix should be applied only if machine have shared C axis

--- openpnp-develop/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ ReferencePnpJobProcessor.java
@@ -628,6 +628,10 @@
                         part,
                         boardLocation,
                         placement.getLocation(), nozzle);
+				// store the actual corrected rotation instead of the offset for shared C axis
+				if (plannedPlacement.alignmentOffsets.getPreRotated()) {
+					plannedPlacement.alignmentOffsets = new PartAlignment.PartAlignmentOffset(plannedPlacement.alignmentOffsets.getLocation().derive(null,null,null,nozzle.getLocation().getRotation()),true);
+				}
                 Logger.debug("Align {} with {}", part, nozzle);
             }
             else
@@ -670,14 +674,15 @@
                         preRotated means during alignment we have already rotated the component
                         - this is useful for say an external rotating stage that the component is
                         placed on, rotated to correct placement angle, and then picked up again.
-                 */
-                if (plannedPlacement.alignmentOffsets.getPreRotated()) {
-                    placementLocation = placementLocation.subtractWithRotation(
-                            plannedPlacement.alignmentOffsets.getLocation());
-                }
-                else {
-                    Location alignmentOffsets = plannedPlacement.alignmentOffsets.getLocation();
-                    // Rotate the point 0,0 using the alignment offsets as a center point by the angle
+			*/
+		    if (plannedPlacement.alignmentOffsets.getPreRotated()) {
+			    placementLocation = placementLocation.subtractWithRotation(plannedPlacement.alignmentOffsets.getLocation());
+				// restore the corrected rotation and override the offset calc for shared C axis
+                             if(Double.isNaN(placementLocation.getRotation()) {
[...]

Issue #255, opened 2016-04-06
Remove Double.NaN from Driver workflows


vonnieda:
Currently the drivers are supposed to accept Double.NaN on a moveTo() to mean "Don't move this axis" but not all drivers handle it correctly, it's not documented very well and it's confusing for driver authors. 

It's only really used in code that hasn't been updated since the immutable Location refactor and since that refactor it's not really very useful since the majority of operations are happening on all four axes anyway.

I want to remove the Double.NaN behavior (after reviewing everywhere it's used) and add a check to Location to make sure that none of the values are NaN.

vonnieda:
Started this work in https://github.com/openpnp/openpnp/tree/feature/remove-nan

markmaker:
Handled by Advanced Motion Control.

PR #815, opened 2019-03-13
BottomVision should store rotation offset in the PartAlignmentOffset

markmaker:
# Description
The PR contains the following extensions and cleanup:
* Bottom Vision now stores the rotation offset in the `PartAlignmentOffset`. 
* The angle calculation is simplified with regard to [distributive modulo equivalences](https://en.wikipedia.org/wiki/Modulo_operation#Equivalences).
* pre-rotate BottomVision is done in a multi-pass loop until a good fix is obtained.
* X, Y offsets are obtained together with the rotation offset, so with a good fix only one vision pass is needed.
* X, Y offsets are therefore compensated for rotation.
* Maximum allowed linear and angular offsets can be configured (GUI).
* `angleNorm()` is used instead of duplicated code in `findOffsetsPostRotate()`
* a redundant `Location.derive()` is removed in `findOffsetsPostRotate()`

# Justification
By storing the rotation offset in the `PartAlignmentOffset` the rotation can be restored when the part is actually placed later. Today the proper rotation is only "stored" in the current C axis rotation. However this information is lost with shared C axis heads when another nozzle picks a second part before the first one has been placed. The PR also helps removing `Double.NaN` from driver workflows. In addition the "smart" multi-pass vision process speeds up bottom vision in most cases. It is now as fast as post-rotate bottom vision in all but the most challenging situations and more robust in the rest. 

See also:
BottomVision Pre-Rotate Fails on Shared C Heads
https://github.com/openpnp/openpnp/issues/599

Remove Double.NaN from Driver workflows
https://github.com/openpnp/openpnp/issues/255

# Instructions for Use
Enable bottom vision on a part and pre-rotate in "Machine Seup / ReferenceMachine / Vision / Bottom Vision". Configure the multi-pass vision criteria (more info in the tooltips):
grafik

# Implementation Details
1. Several testings were done in the meantime. Thank you to all the testers!
3. No changes were made in the `org.openpnp.spi` or `org.openpnp.model` packages.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -47,6 +47,15 @@ public class ReferenceBottomVision implements PartAlignment {
     @Attribute(required = false)
     protected boolean preRotate = false;
 
+    @Attribute(required = false)
+    protected int maxVisionPasses = 3;
+
+    @Element(required = false)
+    protected Length maxLinearOffset = new Length(1, LengthUnit.Millimeters);
+
+    @Attribute(required = false)
+    protected double maxAngularOffset = 10;
+
     @ElementMap(required = false)
     protected Map<String, PartSettings> partSettingsByPartId = new HashMap<>();
 
@@ -77,120 +86,120 @@ public class ReferenceBottomVision implements PartAlignment {
                     partSettings);
         }
     }
+    
+    public Location getCameraLocationAtPartHeight(Part part, Camera camera, double angle) {
+        return camera.getLocation()
+                .add(new Location(part.getHeight()
+                        .getUnits(),
+                        0.0, 0.0, part.getHeight()
+                        .getValue(),
+                        0.0))
+                .derive(null, null, null, angle);
+    }
 
-    private static PartAlignmentOffset findOffsetsPreRotate(Part part, BoardLocation boardLocation,
+    private PartAlignmentOffset findOffsetsPreRotate(Part part, BoardLocation boardLocation,
             Location placementLocation, Nozzle nozzle, Camera camera, PartSettings partSettings)
             throws Exception {
-        double angle = placementLocation.getRotation();
+        double wantedAngle = placementLocation.getRotation();
         if (boardLocation != null) {
-            angle = Utils2D.calculateBoardPlacementLocation(boardLocation, placementLocation)
+            wantedAngle = Utils2D.calculateBoardPlacementLocation(boardLocation, placementLocation)
                            .getRotation();
         }
-        angle = angleNorm(angle, 180.);
-        Location location = camera.getLocation()
-                                  .add(new Location(part.getHeight()
-                                                        .getUnits(),
-                                          0.0, 0.0, part.getHeight()
-                                                        .getValue(),
-                                          0.0))
-                                  .derive(null, null, null, angle);
-        MovableUtils.moveToLocationAtSafeZ(nozzle, location);
+        wantedAngle = angleNorm(wantedAngle, 180.);
+        // Wanted location.
+        Location wantedLocation = getCameraLocationAtPartHeight(part, camera, wantedAngle);
+                
+        Location nozzleLocation = wantedLocation;
+        MovableUtils.moveToLocationAtSafeZ(nozzle, nozzleLocation);
+        final Location offsetZero = new Location(maxLinearOffset.getUnits());
 
         try (CvPipeline pipeline = partSettings.getPipeline()) {
 
-        	Location offsets = new Location(location.getUnits());
-        	// Try getting a good fix on the part in multiple passes.
-        	for(int pass = 0;;) {
-	            RotatedRect rect = processPipelineAndGetResult(pipeline, camera, part, nozzle);
-	            camera=(Camera)pipeline.getProperty("camera");
-	            
-	            Logger.debug("Result rect {}", rect);
-	            
-	            // Create the offsets object. This is the physical distance from
-	            // the center of the camera to the located part.
-	            offsets = VisionUtils.getPixelCenterOffsets(camera, rect.center.x, rect.center.y);
-	            
-	            // OpenCV can only tell us the angle of the recognized rectangle in a   
-	            // range of 0° .. 90° as it has no notion of which rectangle side 
-	            // is the true "bottom" side for instance. The angle will abruptly wrap 
-	            // around from 90° to 0° (and vice versa), as another of the four sides 
-	            // becomes the "bottom" side, as perceived by OpenCV.    
-	            // We can assume that the part is never picked more than +/-45º rotated.
-	            // So we change the wrapping-around into one that produces a range from 
-	            // -45° .. +45°. See angleNorm():
-	            angle = angleNorm(location.getRotation() + rect.angle);
-	            
-	            // When we rotate the nozzle later to compensate the angle, the X, Y offsets 
-	            // will change too, as the off-center part rotates around the nozzle axis.
-	            // So we need to compensate for that.
-	            // TODO: verify angle sign here. 
-	            offsets = offsets.rotateXy(-angle)
-	            			.derive(null, null,	null, -angle);
-	                
-	            if (++pass > 2) {
-	            	// Maximum number of passes. 
-	            	break;
-	            }
-	            
-	            if (Math.sqrt(offsets.getX()*offsets.getX() + offsets.getY()*offsets.getY()) < 2.0
-	            		&& Math.abs(angle) < 10.0) {
-	            	// We have a good enough fix - go on with that.
-	            	break;
-	            }
-	            
-	            Logger.debug("Large offsets {}", offsets);
-	            
-	            // Not a good enough fix - try again with corrected position.
-	            location = location.subtractWithRotation(offsets);
-                nozzle.moveTo(location);
-            }
+            // The running, iterative offset.
+            Location offsets = new Location(nozzleLocation.getUnits());
+            // Try getting a good fix on the part in multiple passes.
+            for(int pass = 0;;) {
+                RotatedRect rect = processPipelineAndGetResult(pipeline, camera, part, nozzle);
+                camera=(Camera)pipeline.getProperty("camera");
 
+                Logger.debug("Bottom vision part {} result rect {}", part.getId(), rect);
+
+                // Create the offsets object. This is the physical distance from
+                // the center of the camera to the located part.
+                offsets = VisionUtils.getPixelCenterOffsets(camera, rect.center.x, rect.center.y);
+
+                // OpenCV can only tell us the angle of the recognized rectangle in a   
+                // wrapping-around range of 0° .. 90° as it has no notion of which rectangle side 
+                // is which. We can assume that the part is never picked more than +/-45º rotated.
+                // So we change the range wrapping-around to -45° .. +45°. See angleNorm():
+                double angleOffset = angleNorm(VisionUtils.getPixelAngle(camera, rect.angle) - wantedAngle);
+
+                // When we rotate the nozzle later to compensate for the angle offset, the X, Y offsets 
+                // will change too, as the off-center part rotates around the nozzle axis.
+                // So we need to compensate for that.
+                offsets = offsets.rotateXy(-angleOffset)
+                        .derive(null, null,	null, angleOffset);
+                nozzleLocation = nozzleLocation.subtractWithRotation(offsets);
+
+                if (++pass >= maxVisionPasses) {
+                    // Maximum number of passes reached. 
+                    break;
+                }
+                
+                // We not only check the center offset but also the corner offset brought about by the angular offset. 
+                Location corner = VisionUtils.getPixelCenterOffsets(camera, rect.size.width/2, rect.size.height/2);
+                if (offsetZero.getLinearDistanceTo(offsets) <= getMaxLinearOffset().getValue()
+                        && offsetZero.getLinearDistanceTo(corner)*Math.sin(Math.toRadians(angleOffset)) <=  getMaxLinearOffset().getValue()
+                        && Math.abs(angleOffset) <= getMaxAngularOffset()) {
+                    // We have a good enough fix - go on with that.
+                    break;
+                }
+
+                Logger.debug("Offsets too large {}", offsets);
+
+                // Not a good enough fix - try again with corrected position.
+                nozzle.moveTo(nozzleLocation);
+            }
+            Logger.debug("Offsets accepted {}", offsets);
+            // Calculate cumulative offsets over all the passes.  
+            offsets = wantedLocation.subtractWithRotation(nozzleLocation);
             Logger.debug("Final offsets {}", offsets);
             displayResult(pipeline, part, offsets, camera);
             return new PartAlignment.PartAlignmentOffset(offsets, true);
         }
     }
 
-    private static PartAlignmentOffset findOffsetsPostRotate(Part part, BoardLocation boardLocation,
+    private PartAlignmentOffset findOffsetsPostRotate(Part part, BoardLocation boardLocation,
             Location placementLocation, Nozzle nozzle, Camera camera, PartSettings partSettings)
-            throws Exception {
+                    throws Exception {
         // Create a location that is the Camera's X, Y, it's Z + part height
         // and a rotation of 0, unless preRotate is enabled
-        Location startLocation = camera.getLocation();
-        Length partHeight = part.getHeight();
-        Location partHeightLocation =
-                new Location(partHeight.getUnits(), 0, 0, partHeight.getValue(), 0);
-        startLocation = startLocation.add(partHeightLocation)
-                                     .derive(null, null, null, 0.);
-
-        MovableUtils.moveToLocationAtSafeZ(nozzle, startLocation);
+        Location wantedLocation = getCameraLocationAtPartHeight(part, camera, 0.);
+        
+        MovableUtils.moveToLocationAtSafeZ(nozzle, wantedLocation);
 
         try (CvPipeline pipeline = partSettings.getPipeline()) {
             RotatedRect rect = processPipelineAndGetResult(pipeline, camera, part, nozzle);
             camera=(Camera)pipeline.getProperty("camera");
-    
-            Logger.debug("Result rect {}", rect);
-    
+
+            Logger.debug("Bottom vision part {} result rect {}", part.getId(), rect);
+
             // Create the offsets object. This is the physical distance from
             // the center of the camera to the located part.
             Location offsets = VisionUtils.getPixelCenterOffsets(camera, rect.center.x, rect.center.y);
-    
-            // OpenCV can only tell us the angle of the recognized rectangle in a  
-            // range of 0° .. 90° as it has no notion of which rectangle side 
-            // is the true "bottom" side for instance. The angle will abruptly wrap 
-            // around from 90° to 0° (and vice versa), as another of the four sides 
-            // becomes the "bottom" side, as perceived by OpenCV.    
-            // We can assume that the part is never picked more than +/-45º rotated.
-            // So we change the wrapping-around into one that produces a range from 
-            // -45° .. +45°. See angleNorm():
-            double angle = angleNorm(rect.angle);
-    
+
+            // OpenCV can only tell us the angle of the recognized rectangle in a   
+            // wrapping-around range of 0° .. 90° as it has no notion of which rectangle side 
+            // is which. We can assume that the part is never picked more than +/-45º rotated.
+            // So we change the range wrapping-around to -45° .. +45°. See angleNorm():
+            double angleOffset = angleNorm(VisionUtils.getPixelAngle(camera, rect.angle));
+
             // Set the angle on the offsets.
-            offsets = offsets.derive(null, null, null, -angle);
+            offsets = offsets.derive(null, null, null, angleOffset);
             Logger.debug("Final offsets {}", offsets);
-    
+
             displayResult(pipeline, part, offsets, camera);
-    
+
             return new PartAlignmentOffset(offsets, false);
         }
     }
@@ -312,6 +321,30 @@ public class ReferenceBottomVision implements PartAlignment {
 
     public void setPreRotate(boolean preRotate) {
         this.preRotate = preRotate;
+    }
+
+    public int getMaxVisionPasses() {
+        return maxVisionPasses;
+    }
+
+    public void setMaxVisionPasses(int maxVisionPasses) {
+        this.maxVisionPasses = maxVisionPasses;
+    }
+
+    public Length getMaxLinearOffset() {
+        return maxLinearOffset;
+    }
+
+    public void setMaxLinearOffset(Length maxLinearOffset) {
+        this.maxLinearOffset = maxLinearOffset;
+    }
+
+    public double getMaxAngularOffset() {
+        return maxAngularOffset;
+    }
+
+    public void setMaxAngularOffset(double maxAngularOffset) {
+        this.maxAngularOffset = maxAngularOffset;
     }
 
     @Override
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -7,6 +7,7 @@ import javax.swing.Action;
 import javax.swing.Icon;
 
 import org.apache.commons.io.IOUtils;
+import org.opencv.core.Point;
 import org.opencv.core.RotatedRect;
 import org.openpnp.gui.MainFrame;
 import org.openpnp.gui.support.PropertySheetWizardAdapter;
@@ -112,7 +113,7 @@ public class ReferenceBottomVision implements PartAlignment {
                 
         Location nozzleLocation = wantedLocation;
         MovableUtils.moveToLocationAtSafeZ(nozzle, nozzleLocation);
-        final Location offsetZero = new Location(maxLinearOffset.getUnits());
+        final Location center = new Location(maxLinearOffset.getUnits());
 
         try (CvPipeline pipeline = partSettings.getPipeline()) {
 
@@ -134,7 +135,7 @@ public class ReferenceBottomVision implements PartAlignment {
                 // is which. We can assume that the part is never picked more than +/-45º rotated.
                 // So we change the range wrapping-around to -45° .. +45°. See angleNorm():
                 double angleOffset = angleNorm(VisionUtils.getPixelAngle(camera, rect.angle) - wantedAngle);
-
+                
                 // When we rotate the nozzle later to compensate for the angle offset, the X, Y offsets 
                 // will change too, as the off-center part rotates around the nozzle axis.
                 // So we need to compensate for that.
@@ -147,16 +148,29 @@ public class ReferenceBottomVision implements PartAlignment {
                     break;
                 }
                 
-                // We not only check the center offset but also the corner offset brought about by the angular offset. 
-                Location corner = VisionUtils.getPixelCenterOffsets(camera, rect.size.width/2, rect.size.height/2);
-                if (offsetZero.getLinearDistanceTo(offsets) <= getMaxLinearOffset().getValue()
-                        && offsetZero.getLinearDistanceTo(corner)*Math.sin(Math.toRadians(Math.abs(angleOffset))) <=  getMaxLinearOffset().getValue()
-                        && Math.abs(angleOffset) <= getMaxAngularOffset()) {
+                // We not only check the center offset but also the corner offset brought about by the angular offset
+                // so a large part will react more sensitively to angular offsets.
+                Point corners[] = new Point[4];
+                rect.points(corners);
+                Location corner = VisionUtils.getPixelCenterOffsets(camera, corners[0].x, corners[0].y)
+                        .convertToUnits(maxLinearOffset.getUnits());
+                Location cornerWithAngularOffset = corner.rotateXy(angleOffset);
+                if (center.getLinearDistanceTo(offsets) > getMaxLinearOffset().getValue()) {
+                    Logger.debug("Offsets too large {} : center offset {} > {}", 
+                            offsets, center.getLinearDistanceTo(offsets), getMaxLinearOffset().getValue()); 
+                } 
+                else if (corner.getLinearDistanceTo(cornerWithAngularOffset) >  getMaxLinearOffset().getValue()) {
+                    Logger.debug("Offsets too large {} : corner offset {} > {}", 
+                            offsets, corner.getLinearDistanceTo(cornerWithAngularOffset), getMaxLinearOffset().getValue()); 
+                }
+                else if (Math.abs(angleOffset) > getMaxAngularOffset()) {
+                    Logger.debug("Offsets too large {} : angle offset {} > {}", 
+                            offsets, Math.abs(angleOffset), getMaxAngularOffset());
+                }
+                else {
                     // We have a good enough fix - go on with that.
                     break;
                 }
-
-                Logger.debug("Offsets too large {}", offsets);
 
                 // Not a good enough fix - try again with corrected position.
                 nozzle.moveTo(nozzleLocation);
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -149,7 +149,7 @@ public class ReferenceBottomVision implements PartAlignment {
                 // We not only check the center offset but also the corner offset brought about by the angular offset. 
                 Location corner = VisionUtils.getPixelCenterOffsets(camera, rect.size.width/2, rect.size.height/2);
                 if (offsetZero.getLinearDistanceTo(offsets) <= getMaxLinearOffset().getValue()
-                        && offsetZero.getLinearDistanceTo(corner)*Math.sin(Math.toRadians(angleOffset)) <=  getMaxLinearOffset().getValue()
+                        && offsetZero.getLinearDistanceTo(corner)*Math.sin(Math.toRadians(Math.abs(angleOffset))) <=  getMaxLinearOffset().getValue()
                         && Math.abs(angleOffset) <= getMaxAngularOffset()) {
                     // We have a good enough fix - go on with that.
                     break;
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -87,7 +87,6 @@ public class ReferenceBottomVision implements PartAlignment {
                            .getRotation();
         }
         angle = angleNorm(angle, 180.);
-        double placementAngle = angle;
         Location location = camera.getLocation()
                                   .add(new Location(part.getHeight()
                                                         .getUnits(),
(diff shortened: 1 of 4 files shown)


src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -87,7 +87,6 @@ private static PartAlignmentOffset findOffsetsPreRotate(Part part, BoardLocation
                            .getRotation();
         }
         angle = angleNorm(angle, 180.);
-        double placementAngle = angle;
```
markmaker: We take the angle from the `location` instead. Works smoother with the multi-pass vision loop, where the `location` is gradually adapted.

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -99,36 +98,52 @@ private static PartAlignmentOffset findOffsetsPreRotate(Part part, BoardLocation
...
+	            
+	            // OpenCV can only tell us the angle of the recognized rectangle in a   
+	            // range of 0° .. 90° as it has no notion of which rectangle side 
+	            // is the true "bottom" side for instance. The angle will abruptly wrap 
+	            // around from 90° to 0° (and vice versa), as another of the four sides 
+	            // becomes the "bottom" side, as perceived by OpenCV.    
+	            // We can assume that the part is never picked more than +/-45º rotated.
+	            // So we change the wrapping-around into one that produces a range from 
+	            // -45° .. +45°. See angleNorm():
+	            angle = angleNorm(location.getRotation() + rect.angle);
```
markmaker: Simplification of the angle offset calculation. `angleNorm()` is in effect a "phase shifted" modulo operation and if I'm not completely mistaken [Distributive Equivalence](https://en.wikipedia.org/wiki/Modulo_operation#Equivalences) applies:

> (a + b) mod n = [(a mod n) + (b mod n)] mod n.

Also the 
> `(rect.size.width < rect.size.height) ? 90 + rect.angle : rect.angle`

expression is pointless as it will be removed by `angleNorm()` being a mod 90° function.

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -160,26 +175,20 @@ private static PartAlignmentOffset findOffsetsPostRotate(Part part, BoardLocatio
...
+            // We can assume that the part is never picked more than +/-45º rotated.
+            // So we change the wrapping-around into one that produces a range from 
+            // -45° .. +45°. See angleNorm():
+            double angle = angleNorm(rect.angle);
     
             // Set the angle on the offsets.
             offsets = offsets.derive(null, null, null, -angle);
             Logger.debug("Final offsets {}", offsets);
     
-            offsets = offsets.derive(null, null, null, offsets.getRotation());
```
markmaker: If I'm not mistaken, this does nothing.

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -160,26 +175,20 @@ private static PartAlignmentOffset findOffsetsPostRotate(Part part, BoardLocatio
...
-            }
+            // OpenCV can only tell us the angle of the recognized rectangle in a  
+            // range of 0° .. 90° as it has no notion of which rectangle side 
+            // is the true "bottom" side for instance. The angle will abruptly wrap 
+            // around from 90° to 0° (and vice versa), as another of the four sides 
+            // becomes the "bottom" side, as perceived by OpenCV.    
+            // We can assume that the part is never picked more than +/-45º rotated.
+            // So we change the wrapping-around into one that produces a range from 
+            // -45° .. +45°. See angleNorm():
+            double angle = angleNorm(rect.angle);
```
markmaker: Code reuse of equivalent `angleNorm()`.

src/main/java/org/openpnp/machine/reference/vision/wizards/ReferenceBottomVisionPartConfigurationWizard.java
```diff
@@ -147,7 +147,7 @@ private void testAlignment() throws Exception {
                 throw new Exception("Offset too big");
             }
             nozzle.moveTo(nozzle.getLocation()
-                                .subtract(alignmentOffset.getLocation()));
+                                .subtractWithRotation(alignmentOffset.getLocation()));
```
markmaker: Here I'm unsure. Must dicuss this.

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -99,36 +98,52 @@ private static PartAlignmentOffset findOffsetsPreRotate(Part part, BoardLocation
...
+	            // We can assume that the part is never picked more than +/-45º rotated.
+	            // So we change the wrapping-around into one that produces a range from 
+	            // -45° .. +45°. See angleNorm():
+	            angle = angleNorm(location.getRotation() + rect.angle);
+	            
+	            // When we rotate the nozzle later to compensate the angle, the X, Y offsets 
+	            // will change too, as the off-center part rotates around the nozzle axis.
+	            // So we need to compensate for that.
+	            // TODO: verify angle sign here. 
+	            offsets = offsets.rotateXy(-angle)
```
markmaker: I'm completely unsure of the `angle` sign here in `rotateXy()`. 

Really mind-boggling with bottom vision reversing everything. Best to test this empirically. :-)

cncmachineguy:
Does this fix the shared C issue or just provide the framework to fix it?

markmaker:
> 
> 
> Does this fix the shared C issue or just provide the framework to fix it?

It should fix it. 

I can't test this, only have a single nozzle machine. So If you could help with that, very welcome! :-)

_Mark

cncmachineguy:
I can try it out for sure. Is it pulled and merged or do I get it from somewhere else?

vonnieda:
I will merge it to the test branch tonight so that you can download it. Will write here with the URL when it's ready.

vonnieda:
This is now merged into the test branch, which can be downloaded from http://openpnp.org/test-downloads/

markmaker:
@cncmachineguy, thanks for your offer to help!

Just be warned that this is really fresh and untested. See the description at the top of what changed.
 
To test pre-rotation it would be helpful to set the pickup location and angle of a large part feeder off quite a bit on purpose. So the needed compensation both in angle and X, Y becomes visible. Look out for faulty 90° rotations (pin 1 orientation), as I radically simplified the modulo 45° calculation. 

The new code now tries to align with just one vision shot and therefore needs to anticipate the rotation in the observed X, Y offsets. I've tried hard to think it through but I'm still not 100% sure about the sign of the angle in this rotation. 

I'll try to help quickly... :-)
_Mark

cncmachineguy:
Well I only tested it with 2 nozzles and parts so far, but it seems to work Just great!! part 1 is at 90dg and part2 is 170deg. It picks up part 2 first then part one, looks at part 2 then part 1, places part 2 then part 1. So since it goes back and forth between the 2, I would say it is successful so far!

markmaker:
Great! Could you please send me your log? I'm interested if it works with just one vision shot (formerly two needed). 
Thanks!
_Mark

cncmachineguy:
attached

markmaker:
> attached

Got no attachment, Is probably swallowed by GitHub. Pls send it to mailto:mark@makr.zone Thanks!

_Mark

markmaker:
Hi @cncmachineguy 

Got your log and it looks good on all four parts. Nice to see how it restores the angles when switching back to the nozzles. Made an excel sheet to calculate the correct values and those match the log perfectly. See attached. 

Simulated   Bottom Vision C Axis Correction |   |   |   |  
-- | -- | -- | -- | --
Angle in pre-vision G0 C axis | -100.186839 | 179.813161 | -100.186839 | 179.813161
Angle observed | -78.94358826 | -0.58762836 | -86.7295074 | -0.47547272
Offset calc (sum) | -179.1304273 | 179.225533 | -186.916346 | 179.337688
angleNorm() | 0.869572743 | -0.77446736 | -6.91634645 | -0.66231172
Offset (negated) | -0.869572743 | 0.77446736 | 6.91634645 | 0.66231172
Compensated | -99.31726626 | 179.038694 | -107.103185 | 179.150849
Angle in pre-place G0 C axis (norm 360°) | 260.6827337 | 179.038694 | 252.896815 | 179.150849

[Bottom Vision C Axis Compensation.xlsx](https://github.com/openpnp/openpnp/files/2991420/Bottom.Vision.C.Axis.Compensation.xlsx)

> The first time both parts were twisted 90 degrees I think

Well, if that is the case, then I can't see the reason from the log. 

[OpenPnP.log](https://github.com/openpnp/openpnp/files/2991413/OpenPnP.log)

Also: in your test all parts worked with just one bottom vision shot. As the offset angles were all relatively small, it's probably hard to visually see if X, Y coordinates got corrected the right way around. 

Thanks again for your help in testing.

_Mark

netzmark:
Hi Jason,
I have tested Mark's mod today and can also confirm, like Bert did, it looks good working. I have sent some logs to Mark for his rating. My tests were short and I will continue to be more sure then come back with oppinion.

vonnieda:
Hi @markmaker,

Apologies again for taking so long to get to this. I'm finally freeing up and want to work on merging this. Has your testing uncovered any issues, or the need for any changes, or are you still happy with the code as it is now?

Thanks,
Jason

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -99,36 +98,52 @@ private static PartAlignmentOffset findOffsetsPreRotate(Part part, BoardLocation
...
+	            			.derive(null, null,	null, -angle);
+	                
+	            if (++pass > 2) {
+	            	// Maximum number of passes. 
+	            	break;
+	            }
+	            
+	            if (Math.sqrt(offsets.getX()*offsets.getX() + offsets.getY()*offsets.getY()) < 2.0
```
vonnieda: Is the `2` here in millimeters? Or is it unitless and I don't quite understand? If millimeters, this should be corrected to check against the correct units. Additionally, 2mm or 10 degrees of offset seems like quite a lot. Should we make this much smaller, or perhaps configurable?

markmaker: This is still TODO. See the reply to [your question](https://github.com/openpnp/openpnp/pull/815/files/6ec8856de8f50e810c26fa95ed1547e34238aea4#issuecomment-483909834).

markmaker:
> Hi @markmaker,
> Apologies again for taking so long to get to this. I'm finally freeing up and want to work on merging this. Has your testing uncovered any issues, or the need for any changes, or are you still happy with the code as it is now?

Apologies likewise. While I'm happy with the actual functionality there are still some TODOs open, as stated in the PR description. These are:
 
> - The maximum number of passes are hard-coded (3 passes). A setting should be added to the GUI.
> - The "good fix" criteria are currently hard coded. X, Y offset distance < 2mm and angle offset < 10° are considered good enough. The defaults are subject to field testing and these settings should be added to the GUI.
> - findOffsetsPreRotate() and findOffsetsPostRotate() should be made non-static to access these settings. Need to check if something speaks against it.

> Is the 2 here in millimeters? Or is it unitless and I don't quite understand? If millimeters, this should be corrected to check against the correct units.

You're right, the Millimeters Unit was hardccoded as well as the value. This will be fixed when I add a Length property. 

> Additionally, 2mm or 10 degrees of offset seems like quite a lot. 

The question here is whether a new vision shot is warranted or if we can just blindly apply these remaining offsets. I think the errors introduced by 2mm camera offset and/or 10° of post-vision nozzle rotation will be very small and everything can be anticipated from the last vision shot. 

Remember that post-rotate Bottom Vision will always just take one shot and have an angle offset of up to 180°! 

Additionally Marek's test showed that the Vision multi-pass "good fix" criteria is wrongly formulated, as it takes the final rotation offset (to be applied to the part) instead of just the difference between the last anticipated and vision detected angles. It now does either just one pass or all of them. Need to revisit that.

Sorry I'm no further along the path. 

Can you perhaps shed some light as to why findOffsetsPreRotate() and findOffsetsPostRotate() are static? Am I missing something? 

_Mark

vonnieda:
> > Additionally, 2mm or 10 degrees of offset seems like quite a lot.
> 
> The question here is whether a new vision shot is warranted or if we can just blindly apply these remaining offsets. I think the errors introduced by 2mm camera offset and/or 10° of post-vision nozzle rotation will be very small and everything can be anticipated from the last vision shot.

Thanks @markmaker, that makes sense. I wasn't thinking it through enough.

> Can you perhaps shed some light as to why findOffsetsPreRotate() and findOffsetsPostRotate() are static? Am I missing something?

Just because they do not currently have any instance dependencies. I prefer to make methods static if they can be because it makes it less likely that a variable will be shadowed or an instance variable will be used incorrectly. If we need to add some instance variables for GUI settings it's completely fine to make these no longer static.

Thanks again for your work on this. My understanding is that you are still in test mode and are not ready to merge this due to the remaining TODOs, correct? 

Thanks,
Jason

markmaker:
Hi @vonnieda 

Finally got around to testing this on my machine. It is a single nozzle machine, so shared C operation is not tested, However I think this use case has already been covered by @netzmark and @cncmachineguy. 

In accordance with feedback the new merges contain the follwoing:
* Fixed multi-pass cumulative offset handling. 
* Added GUI settings for maximum vision passes, linear and angular offsets.
* Correct unit handling.
* Added encapsulation for angles' tranformation from pixel/OpenCV to the OpenPNP coordinate system.
* Cleaned up code and made it more readable.
* Detailed large part testing on my machine.
* Fixed a bug in pre-existing code (post-rotate) where the part height was not used in testAlignment (potentially shoving tall parts into the bottom camera!).

The GUI now allows configuration of the multi-pass vision:
grafik

Additional config is properly disabled if pre-rotate is off:
grafik

In testing I deliberately used a large quadratic part and put the grid feeder at a large linear and angular offset:
grafik

grafik

Used my old banged-up nozzle tip holder with huge runout:
grafik

Tested alignment with and without runout calibration with `ReferenceBottomVisionPartConfigurationWizard.testAlignment()`. After fixing the multi-pass offset cumulation issue, it worked like a charm. 

The maximum linear offset now limits both the detected center offset and the corner rotation offset. So larger parts react more sensitively to angular offsets as even small angle offsets can introduce large linear offsets through the leverage. The maximum angular offset can also be limited directly in degrees for the smaller parts. All lengths now use proper unit conversion. 

It should be noted that these are only criteria to determine if an additional bottom vision pass is needed. Even if no further vision is applied, the remaining offsets are always compensated. Setting very small maximum offsets, the effectiveness of the final compensation can be demonstrated through the log. Also the multi-pass vision works as expected. 

It seems quite robust, even when the offset is so large that the part overlaps the camera edge it still nails it in the next pass:
grafik
grafik

Switching off nozzle runout calibration one can nicely see how the multi-pass vision process has to work harder to iron out the errors. If runout calibration is switched on, prediction of the final offsets (applying the large 13° angular offset in anticipation) is quite good. All open transformation questions were resolved. 

grafik

Post-rotated `ReferenceBottomVisionPartConfigurationWizard.testAlignment()`also still works as is expected.

I also tested placing on a PCB. I still get some offsets there but I think this is an issue of my machine. It happens the same way with the old code version. It is very hard to find the culprit because I have no fiducials on my test BCP and I need to hand align the board. I also suspect that the huge runout (radius=0.2mm i.e diameter=0.4mm!) will still introduce some errors when the Z axis is offset. Part height will introduce a parallax in bottom vision that is not compensated and as I use Z probing, my part height has some leeway. 

As far as I can wrap my brain around it, the numerics from the log are fine. I hope someone else with a proven machine/PCB fiducials can test this. Subject to this last point I think it is finished. 

I found a number of other issues while testing. Will document/PR those separately. 

_Mark

netzmark:
Hi Mark,
I have given up today with some issue (excellent head nozzles offsets, perfect nozzles calibration but each placement always moved x or y if any rotation is applied. Good effect only when placement 0deg). But will try clean up this mess tomorrow and make some test of your code.

markmaker:
Hi @vonnieda 

To answer your question directly: 

> Thanks again for your work on this. My understanding is that you are still in test mode and are not ready to merge this due to the remaining TODOs, correct?

Yes and no. I'm quite confident with the code now and it is feature-complete in my eyes. However unfortunately I was not able to 100% positively confirm real PCB placement. See my report above. 

I also noticed in the meantime, that my branch comes from an older develop commit. There were quite some changes since then. I now switched back to develop and some issue that I wanted to report just vanished ;-) 
(Capturing the Park position and Parking did no longer work correctly). 

I saw that there were heave changes involving PCB affine transform, so maybe the PCB placement issue is related... ?

I'm a bit "green" with github so I avoided merging the newest develop into fix/pre-rotate_store_angle in order not to break the PR. Is it ok to do that? Will it correctly sort things out?

Thanks,
Mark

netzmark:
If it was addressed to me "I saw that there were heave changes involving PCB affine transform, so maybe the PCB placement issue is related... ?" - No. The effect gets also with no fiducials checking at all. In my oppinion there is something wrong with experimental NozzleTipCalibration but I must make more experiments yet.

netzmark:
Mark, I will deliver new logs after the Easter holiday.
I have tested it little today and found nothing wrong there however on Tue I will put it into some regular production so the log will be more valuable to rate it.
The thought is, could it be not better if option (pre-rotate) is selectable for part or feeder not the whole jobProcessor? In case of multi-C machines it is not important at all. But in case of shared-C pre-rotation is slowering the machine (there is more C rotations and it slows down XY moves).

markmaker:
Thanks @netzmark for your ongoing help. 

> The thought is, could it be not better if option (pre-rotate) is selectable for part or feeder not the whole jobProcessor? 

Yes, good point. 

It probably needs a selection on the part and perhaps on the nozzle tip as well:
- Always off
- Default
- Always on

Will look into it.

_Mark

markmaker:
Hi 

Found the culprit why I couldn't place on PCB. After searching my head off in machine setup and software it finally dawned on me. Acceleration was simply too high for the off-senter part to remain stable on the nozzle tip. It did not help to reduce (part) speed in OpenPNP. Needed to tweak the controller max acceleration config and upped the setpoint vacuum level. 

Also merged develop into the bracnh. I get good placement:
grafik

Learned a lot about OpenPNP, Java and whatnot. 

Methinks this is it. Ready for merge.  

_Mark

markmaker:
Hi

Followed @netzmark 's suggestion and added a part-individual bottom vision pre-rotate override.

grafik

The global default "Enabled?" setting no longer disables the controls, as the config might still be useful for the individual parts that have pre-rotate "always on". 

grafik

_Mark

vonnieda:
Hi @markmaker, can you explain how these new options work please? 

In particular, are you saying that the AlwaysOn option overrides the global bottom vision enabled option? If that is the case, I think we need a different way. 

The purpose of that is to be able to quickly turn off bottom vision for all parts in case your bottom vision camera is broken, or you aren't using bottom vision due to another type of problem. Having to go in and change the config for every part would be tedious.

Thanks,
Jason

markmaker:
Hi @vonnieda 

Yes that was the idea. The thinking was this;

1. I assumed that after these fixes and the adaptive passes (same speed as post-rotate in most cases), pre-rotate will be the default option. I see no valid reason, why post-rotate should be used anymore. In my tests I get much better results. But maybe I'm missing something.  

2.  Now with the default open "on", you may choose "Always off" on some parts, maybe large parts that don't fit the camera view while rotated etc. That's a "blacklisting" scheme. 

3. The opion "AlwaysOn" is there for completeness more than anything else. I'd expect it would only be used for very exotic parts where switching off pre-rotate would simply negate bottom vision completely, perhaps because the accuracy of pre-rotate is simply needed. 

4. I don't expect this option to be used in a "whitelisting" scheme, with the pre-rotate option being off as a standard and individual part being switched on.

Of course the global setting could also be made into a drop-down with "Default On", "Default Off", "Force On", "Force Off" instead of the checkbox.

However then you need to teach me how to migrate old `machine.xml` settings. :-)

_Mark

vonnieda:
Hi @markmaker, I think perhaps we're not talking about the same thing. I was referring to your comment "The global default "Enabled?" setting no longer disables the controls, as the config might still be useful for the individual parts that have pre-rotate "always on"."

I've highlighted the setting of concern:

Screen Shot 2019-04-21 at 5 25 14 PM

This enabled setting controls whether bottom vision is enabled at all - nothing to do with whether post or pre-rotate is used. As long as that functionality wasn't changed, then I think we're okay.

I'll download the code and take a look over it and let you know if I see any issues. For the record, I agree that there will be no reason to use post rotate but I do think the options are somewhat complicated and confusing, so I may want to simplify a bit.

Thanks,
Jason

markmaker:
Hi @vonnieda 

Ah.. stupid mistake in my description. I meant the "Rotate parts prior to vision?" checkbox, of course. It only concerns the choice of pre-rotate or not. It does not affect bottom vision on/off. The highlighted "Enabled?" checkboxes still shut down bottom vision as before (globally and per part). 

>  As long as that functionality wasn't changed, then I think we're okay.

Then we're okay. :-)

> I do think the options are somewhat complicated and confusing, so I may want to simplify a bit.

Yeah of course, feel free to change anything or just leave out the last merge. 

Please also look at the English in tooltips etc.. ... not a native speaker and these descriptions should be short but consise. Not easy for me. 

_Mark

netzmark:
Hi Mark,
Could you explain pls how works relation max passes to angular and linear offsets? It looks like there is something not in agree with highlighted descriptions.
At default setting linear=1, angular=10 system takes repeated picture at parts angled for values like 2-4deg shown in the pipeline. At setting 10/10 the system has stopped to make new snapshots at the degrees like that. Description suggests that new pictures are not made if angle detected is below declared. Is not linear and angular changed each other?

netzmark:
Hmmm, I'm little confused. Sometimes it like normal but sometimes strange :-).
At 1/10 new pic at lin <0.5 and angle 2-4;
10/10 new pic in practice is never made (had no bigger offsets found) - normal probably;
at 10/0.5 it makes new pic at 2-4deg - normal;
at 0.1/10 it makes new pic every time - normal;
So only option 1/10 looks "strange". Like some bigger angles "telling" the linear is exceeded. I'll take a log to send you afternoon.

markmaker:
Hi @netzmark 

pre-rotate bottom vision uses three criteria to decide whether to have another pass:

1. center offset is larger that the max. linear offset
2. angular offset is larger than the max. angular offset
3. the corner offset due to the angular offset is larger than max. linear offset

The corner offset is calculated by taking the sinus of the angular offset and multiplying it with half the part diagonal (radius of the bounding circle). This way a large part will react more sensitively to angular offsets as the large radius will quickly create considerable corner offsets. 

The illustration here shows the part as it is seen on the first pass superimposed (a bit fainter) the part as it should be: 

bottom vision multi-pass criteria v2

It is the effect of the corner offset that you are observing, I think. Do you have the log? Pls. send it to my PM.

_Mark

netzmark:
Thanks for great explanation with picture.
Mailed you the log to PM.

markmaker:
Hi Marek

it does indeed look strange. 
> 2019-04-24 11:08:39.273 ReferenceBottomVision DEBUG: Bottom vision part 402-100n result rect { {802.0487670898438, 600.5609130859375} 27x14 * -83.6598129272461 }
> 2019-04-24 11:08:39.275 ReferenceBottomVision DEBUG: Offsets too large (0.067131, -0.010667, 0.000000, -6.282969 mm)

This small 0402 part should easily be inside the tolerance.

Are you absolutely sure the settings were default 1mm/10°?

Sometimes I forget to press "Accept" and the new setting is not stored. OpenPNP does not automatically accept changed values when you leave a wizard (as I'm used to).  

_Mark

netzmark:
I'm sure for 99% but can repeat it tomorrow yet and send you a new one.

markmaker:
And do you have the latest merges?

netzmark:
echhhh, I wanted but forogotten to tell you, sorry. It was the version with no individual pre-rotation assignment. I wanted to wait with this for Jason's revision who mentioned that "may want to simplify a bit.". So my test were with version 4 days old not absolutely latets. Does it change a lot about this case?

markmaker:
No, the last merge is not important but this one is:
https://github.com/openpnp/openpnp/pull/815/commits/ac19874c93ce44a2046c7479fddcc6aeb56ae90e

_Mark

netzmark:
My version is after this change so ok.

netzmark:
I have just compiled individual menu to my version.
Understand it works like:
AlwaysOff - forcess no preroation for this part even if global flag (in BottomVision) is ON.
AlwaysOn - forcess preroation for this part even if global is OFF.
Default - follows global selection.
For the first look it seemed be little "over-combinated". But after some short considering - it makes sense over one checkbox turning on/off the prerotation for the part. I will show this tomorrow to my employees what the think about.

netzmark:
Mark I have sent you new log to PM.
Regarding the AlwaysOff/On/Default, my engineer said. Could be good to add combo to the BottomVision where is "rotate prior to vision" checkbox allowing to choose: "use for default" or "use for all". But it's crazyness never ending. As it is now is also quite good.

markmaker:
> Mark I have sent you new log to PM.

Thanks, will have a look.

@vonnieda said:
> I do think the options are somewhat complicated and confusing, so I may want to simplify a bit.

@netzmark said:
> Regarding the AlwaysOff/On/Default, my engineer said. Could be good to add combo to the BottomVision where is "rotate prior to vision" checkbox allowing to choose: "use for default" or "use for all". But it's crazyness never ending. As it is now is also quite good. 

So your engineer is actually saying he wants it _more_ complicated!?  
;-)

_Mark

markmaker:
@netzmark 

found the bug and fixed it. But need to test first.

_Mark

netzmark:
This is great!
(It's no sarcasm. I hate situations when everything is ok but something does not work. 
If there is visible something wrong - it is good ;) ).

markmaker:
Hi @netzmark 

fixed and tested it, 

Furthermore you will now see _which_ criteria prompted an additional vision pass in the DEBUG log.

```
2019-04-25 20:44:06.842 ReferenceBottomVision DEBUG: Bottom vision part QFP50P1200X1200X160-64N-MK20DX256VLH7 result rect { {606.2051391601562, 375.8307800292969} 476x483 * -78.12374114990234 }
2019-04-25 20:44:06.842 ReferenceBottomVision DEBUG: Offsets too large (-0.812531, -0.623029, 0.000000, -11.876259 mm) : corner offset 1.933100912513336 > 1.06
```

The corner offset calculation was wrong. Misunderstood `VisionUtils.getPixelCenterOffsets()` that returns a Location _relative_ to the camera center but takes _absolute_ pixel coordinates (relative to the upper left corner of the camera). 

_Mark

netzmark:
Applied!
I will do all my best to perform some test tomorrow and provide a new log.

markmaker:
Added a little video that shows some of the features:
https://youtu.be/zfJVG5YJ-y8
Please see the [PR Description and Justification](https://github.com/openpnp/openpnp/pull/815#issue-260885864).
_Mark

netzmark:
Hi @markmaker,
After hours of experiments bringing me to crazyness I haven't taken a new log. I will do after this weekend.
But at least it looks the linear limit works properly now.

netzmark:
@markmaker,
Pls update #815 or #825 to be compatible.
- ReferenceBottomVision,
- ReferenceBottomVisionPartConfigurationWizzard,
- VisionUtils.

Frankly speaking you could close #815 probably and just include its changes into #825.
I can't reasonable test two PRs touching similar problem but having not compatible codes, this way my feedback is worth nothing and information you get are misleading.

markmaker:
Hi @vonnieda 

Do you think it would be reasonable/acceptable for you to review if I merge #815 into #825 (again)? 

Thanks,
_Mark

markmaker:
... or ...

@vonnieda could you merge  #815 into #825 into `test` and release that again?

Many thanks,
_Mark

netzmark:
@markmaker,
Can you not include the changes from 815 into 825 and abandon 815? This way 825 could be tested reasonable and later can decide whether your calibration improvement makes sense of not and eventually just extract needed and tested code from 825 to 815.
Merging into the test makes no sense until you don't know it works on multi. Don't believe anybody will test it. We need continue developing of both.

markmaker:
@netzmark 

The two PRs are separate features. It is one of the OpenPNP developer guidelines to try and include only one feature per PR.  
https://github.com/openpnp/openpnp/wiki/Developers-Guide#granularity

I's understandable, Jason needs to be able to review those changes. Also - in this  case - the two PRs are different in terms of how critical the changes are.

- #815 changes functionality that is widely used by the community and needs extra scrutiny to be acceptable. Easier to do that if these changes are isolated and relatively small. 
- #825 touches an experimental feature (#812) that is not widely adapted (I believe) and - as we both learned - has various issues to begin with. So I expect Jason to be a bit more relaxed in accepting changes there. 

My hope is, that @vonnieda will at least accept #815 into openpnp:develop very soon. Once this is done, we can integrate openpnp:develop into #825 (and resolve the predicted merge issues of the `camera.getLocation(tool)` usage). 

I can merge and make a .jar if that helps you. But what about other machine specific hacks?

_Mark

netzmark:
I understand what you mean.
But don't understand why there is a problem to transfer changes being basically tested/finished in 815 to 825. And then continue 825.
Making jar at your side is not good way. Better send me PM mentioned 3 files from 815 properly updated to be compatible with actual 825.
Waiting for Jason seems be losing of time, he seems be "out of the system" for longer time.

markmaker:
@netzmark 

I see your point. Expecially after wrangling with git, trying to merge the two PRs. Naively I thought that merge --> revert merge -> merge again resulted in simply the original merge applied. However I needed to _revert_ the _revert merge_ otherwise it stubbornly refused whatever came from the other branch. 

I guess git does not really merge branches but individual commits. So while you think you tell it to merge a full branch (i.e. the diff to where it "branches off" the common denominator), you actually tell it to merge the commits in the branch... but apparently it only includes a commit if it was never merged before .i.e. it is not smart enought to track the reverted commits and to redo them. 

Professionally I work with svn and always thought git was much smarter. Apparently not. 

I guess it is reasonable to assume that @vonnieda will merge PR #815 first. So it is acceptable after all to merge #815 into #825. If he merges #815 first, develop will move ahead and when he then compares PR #825 against develop, the #815 changes won't show up anymore. No need to review changes twice. 

At least that's what I hope. :-)

_Mark

vonnieda:
I'm not out, but I was under the impression that #815 was not complete
since there were still commits being added to it. If it's complete and
ready for merge I'll review it and get it merged ASAP. Just let me know.

Thanks,
Jason

markmaker:
Hi @vonnieda 

Thanks for your answer. 

I consider this one ready for merge into openpnp:develop. No further bugs surfaced in tests of mine and @netzmark 's. 

And I just reviewed the [code changes](https://github.com/openpnp/openpnp/pull/815/files) again, after having gained more insight into OpenPNP in the last week :-) 

_Mark

vonnieda:
Sounds good, thanks Mark. I'll work on getting this merged in the next day
or so.

Jason

netzmark:
There is small chance it is not really finished. Don't forget I was testing two PR merged with official and mine. So finally it is possible that clean 815 applying is not final fix for this issue.
But, ok, we will improve it together with 825 :-).

Jason, this is great you are in not out :-). Could you confirm?: Now within 815 git is showing changes to official release, and 825 the same. If you will merge 815 then the changes visible in 825 PR will be automatically shown relating to updated official, or still to the version being actual on the moment of PR creation? Just to know how it works... I hope you understand the question.

markmaker:
Hi @netzmark 

> There is small chance it is not really finished. Don't forget I was testing two PR merged with official and mine. So finally it is possible that clean 815 applying is not final fix for this issue.

Mistakes are always possible. I just like to clarify, that I myself have tested this PR in its final form and completely **isolated from #825**. I switched git branch and `machine.xml` for completely separate settings. 

The multi-nozzle aspect was also [tested by @cncmachineguy albeit not in the last version](https://github.com/openpnp/openpnp/pull/815#issuecomment-475161495). 

Much "back and forth" has come out of #825 and this should not be confused with #815.

Again: Errors are still possible.  :-) 

_Mark

netzmark:
This is strange if sometimes are no errors, just strange if that happens :-)

markmaker:
Hi @vonnieda 

just to make it clear, in my eyes this is ready for merge. If it is a time issue, just disregard. If something is not right, I'd like to know and learn. :-]

_Mark

vonnieda:
Hi @markmaker,

It's time, but maybe we can shortcut it. The problem is that this is a big, complicated, far reaching change that affects everyone, so I really want to test the heck out of it, but I am having a hard time finding enough time to really do that.

But! I trust your judgement. If you feel that this has had enough testing, under enough different situations, and that it won't significantly break existing configs, I'll just go ahead and merge it and we can deal with any problems as they come. Does that sound good?

Jason

markmaker:
Hi @vonnieda 

Thanks for the vote of confidence :-)

I'll take the opportunity to ask a few things in return. There are different styles of revising code. I'd be interested how you see it. Just two examples:

**Fix bugs found along the way:** Even if not directly the topic of the PR. Example: TestAlign did not add part height to nozzle Z during bottom vision, so the focal plane was not right and for very tall parts it could even endanger the camera lens. Should I fix it or not? I took the opportunity to refactor that into a separate method getCameraLocationAtPartHeight() which at the same time uncluttered the two bottom vision methods (post- and pre-rotate). A speparate PR would have complicated that and doubled the testing effort needed. Right or wrong?

**New utility methods:** The rules of abstraction led me to add the new VisionUtils.getPixelAngle() analog to getPixelCenterOffsets() & Co. instead of just coding that into the user code. Such additions will blow up the PR and reach into central parts of OpenPNP. However it just feels right, will clear up things for the next guy and allows for future changes. Right or wrong?

That's just two examples. This tendence to clean up things progressively along the way comes out of my experience with successful long term software product maintenance. But I respect different styles :-) The following depends a bit on how you judge such things.  

1. I don't want to rush things. I trust your judgement very much so if it just needs more time for you to review the changes that's fine and much appreciated. 
2. I wasn't under the impression that this PR is _that_ big (especially comparing it to #825). I just looked at the code changes again. The changed math was the most critical and once this was empirically tested with deliberately large offsets, I grew quite confident due also to the mostly _linear_ character of it all. You can't have linear math match up the right way in one case and then not in the other. The handling stuff (pre-rotation setting per part) I tested quite thoroughly. The rest of the chatter comes from the GUI changes. The way Window Builder generates code just creates endless diffs blowing up the PR. 
3. The assumption is that `openpnp:develop` is what the name implies - the forefront, pioneer spirit, some trial and error may occur. People not willing to risk glitches use the stable release, right?
4. I'm here if things break. I can usually react in 24h or so. 

I'm torn: A code review would be very welcome for the reasons outlined above and future work - even if it means waiting. On the other hand and under the above  assumptions, yes I think its ready and I'm also willing to take a measure of responsability. 

_Mark

vonnieda:
> 
> **Fix bugs found along the way:** Even if not directly the topic of the PR. Example: TestAlign did not add part height to nozzle Z during bottom vision, so the focal plane was not right and for very tall parts it could even endanger the camera lens. Should I fix it or not? I took the opportunity to refactor that into a separate method getCameraLocationAtPartHeight() which at the same time uncluttered the two bottom vision methods (post- and pre-rotate). A speparate PR would have complicated that and doubled the testing effort needed. Right or wrong?

Here is how I personally handle this: If the bug is related to the feature, I will go ahead and include it. I'll even go refactor other features if that bug fix affects them. The idea is that all of the code is related and linked together.

Inversely, let's say I am working on feature A and I happen to notice a bug in feature B with an easy fix. I will almost always create a new branch from root, fix the bug in B, commit it, and then go back to working on A.

I do this because I want to make it easy to revert a change that breaks things. I was burned early on in this project by people submitting huge multi-feature PRs that touch dozens of files. I would merge these PRs and then someone would notice that one of the many features was broken so my choice was either to revert multiple features, or try to figure out which of the many changes caused the problem.

So, to answer your specific question, if I were working on improvements to bottom vision and I found that test align didn't work right, sure, I'd include that.

A minor correlary to that is: I find it much easier to merge small PRs than large ones. Look at the history from the past few weeks. Several people have submitted little PRs that it's very easy to take a quick look at and approve. I've merged those while still mulling this one over because this one has much further reaching effects.

If I can take a quick look at a PR and think to myself "This looks okay, and even if it's not, it's not going to break the world, so let's merge it and see." then I will generally just merge it. If I have to fire up my machine, set up a job, test multiple configurations (AlwaysOn, AlwaysOff, Default) then I need to find several hours to do that testing and it's been hard recently to find that kind of block of time.

So, to summarize: If you can submit a PR that is easy for me to merge without worrying - I am going to merge it fast. If it's going to take time to really review and test, then I'm going to need to find time to do that.

> 
> **New utility methods:** The rules of abstraction led me to add the new VisionUtils.getPixelAngle() analog to getPixelCenterOffsets() & Co. instead of just coding that into the user code. Such additions will blow up the PR and reach into central parts of OpenPNP. However it just feels right, will clear up things for the next guy and allows for future changes. Right or wrong?

Right. Adding code almost never worries me. Changing code that everyone relies on worries me. Changing ReferenceBottomVision affects every user. Adding a new utility method doesn't.

> 
> That's just two examples. This tendence to clean up things progressively along the way comes out of my experience with successful long term software product maintenance. But I respect different styles :-) The following depends a bit on how you judge such things.
> 

I don't want to disuade this at all! This is definitely a good thing. My "mantra", if you will, is "Every time I touch the code, leave it better than it was before." This is why I refuse to merge quick hacks, or bug fixes that "work for me but haven't really been tested."

This is, of course, complicated by the maintainer / contributor situation. It's relatively easy for me to spot a problem that might affect a handful of users using an odd configuration because chances are I helped them with that problem 5 years ago. So part of my worry is that when someone is changing a core part of the system I don't know if they've considered all the different possibilities.

>     1. I don't want to rush things. I trust your judgement very much so if it just needs more time for you to review the changes that's fine and much appreciated.

Thank you. I'm going to go ahead and merge it now and if there are problems we'll solve the problems. 

> 
>     2. I wasn't under the impression that this PR is _that_ big (especially comparing it to #825). I just looked at the code changes again. The changed math was the most critical and once this was empirically tested with deliberately large offsets, I grew quite confident due also to the mostly _linear_ character of it all. You can't have linear math match up the right way in one case and then not in the other. The handling stuff (pre-rotation setting per part) I tested quite thoroughly. The rest of the chatter comes from the GUI changes. The way Window Builder generates code just creates endless diffs blowing up the PR.

When I said "big" I suppose I meant in "effect" rather than in sheer code size. This touches a very core part of OpenPnP so to me it's a "big" change because it affects so many people.

> 
>     3. The assumption is that `openpnp:develop` is what the name implies - the forefront, pioneer spirit, some trial and error may occur. People not willing to risk glitches use the stable release, right?

Agreed. My hesitancy here is my fault. For many years I've treated develop as production because I wasn't good about regularly merging to master. I've gotten better about that, so now I agree that I should be more willing to merge potentially problematic changes to develop so we can get more people using and testing it.

> 
>     4. I'm here if things break. I can usually react in 24h or so.
> 

Thanks!

> 
> I'm torn: A code review would be very welcome for the reasons outlined above and future work - even if it means waiting. On the other hand and under the above assumptions, yes I think its ready and I'm also willing to take a measure of responsability.

Thanks for that. Let's get it in and see how it goes. I still don't have time to do a proper review and testing, but other are clearly willing and you have put a lot of time and effort into this, so let's get it in. 

**I do have one small request:** Would you please update CHANGES.md to explain the change and how users can use it? It doesn't have to be extensive, just a quick note. You can submit a second PR for that and I'll merge it right away.

Thank you again for all the work you've put into this. I very much appreciate the time and effort and your attention to detail.

Thanks,
Jason

markmaker:
This animation quickly explains Bottom Vision thresholds. I post this here as a repo (as the Wiki can't store images).

See https://github.com/openpnp/openpnp/wiki/Bottom-Vision#tips 

BottomVisionThreshold

markmaker:
More Wiki images, see 
https://github.com/openpnp/openpnp/wiki/Bottom-Vision
___

Bottom Vision General Settings

Offsets Explained

Part Bottom Vision Settings

Pre-Rotate Option

Rotation Option

Part Size Check

PR #829, opened 2019-05-07
added test buttons for autofeeder actuators

aneox:
added 2 buttons for auto feeders actuators, to easy test, setup and debug feeders.


vonnieda:
Thanks @aneox!

PR #838, opened 2019-05-14
Initialize CVPipelineEditor with last stage selected and stage grid focused

JulienDevillers:
# Description
Initialize CVPipelineEditor with last stage selected and stage grid focused

# Justification
User see immediately the result of the pipeline.
He can immediately navigate through stages using up/down arrow keys.

# Instructions for Use
Open any pipeline editor, press up/down keys.

# Implementation Details
Manual tests.
Mini refactored the actual code to remove redundant parametering of the dialog.

Yes ! love them.

No

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/vision/wizards/ReferenceFiducialLocatorPartConfigurationWizard.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/wizards/ReferenceFiducialLocatorPartConfigurationWizard.java
@@ -94,7 +94,6 @@ public class ReferenceFiducialLocatorPartConfigurationWizard extends AbstractCon
         CvPipelineEditor editor = new CvPipelineEditor(pipeline);
         JDialog dialog = new CvPipelineEditorDialog(MainFrame.get(), "Fiducial Locator Pipeline", editor);
         dialog.setVisible(true);
-        editor.initializeFocus();
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


markmaker:
Hi Julien

Nice to have this initialization in a central place. For instance now the initial window size could be managed (store/restore or make a percentage of the main frame). So the window is not so unnecesary small on a large monitor. 

Coming from C++ I'm a bit unsure whether it is good to do all this initialization in the constructor. Maybe a static method would be more appropriate? I guess @vonnieda needs to take a look.

More questions in the code.

_m

JulienDevillers:
Hi Mark,

My primary goal was UX. Refactoring is a side effect... :)

What do you mean by 'more question in the codes ?

src/main/java/org/openpnp/machine/reference/vision/wizards/ReferenceFiducialLocatorPartConfigurationWizard.java
```diff
@@ -92,12 +92,11 @@ private void editPipeline() throws Exception {
         pipeline.setProperty("package", part.getPackage());
         pipeline.setProperty("footprint", part.getPackage().getFootprint());
         CvPipelineEditor editor = new CvPipelineEditor(pipeline);
-        JDialog dialog = new JDialog(MainFrame.get(), "Fiducial Locator Pipeline");
-        dialog.getContentPane().setLayout(new BorderLayout());
-        dialog.getContentPane().add(editor);
-        dialog.setSize(1024, 768);
+        JDialog dialog = new CvPipelineEditorDialog(MainFrame.get(), "Fiducial Locator Pipeline", editor);
         dialog.setVisible(true);
+        editor.initializeFocus();
```
markmaker: Shouldn't this line be removed?

JulienDevillers: Hi Mark,
You're right, it's done.

vonnieda:
Looks good to me, and a nice cleanup. Thank you!

 @markmaker, in Java doing this kind of initialization in the constructor is more a matter of style than anything. There aren't the potential pitfalls as there are in C++. Doing it in a static method would also be fine, but either is good.

PR #840, opened 2019-05-15
fix error message 'Unable to parse CSV file names' to 'Unable to find relevant headers' names.'

JulienDevillers:
fix error message 'Unable to parse CSV file names' to 'Unable to find relevant headers' names.'

The previous message was unclear to an unaware person.
Added a link to wiki too in the message.


vonnieda:
Nice work, thanks @JulienDevillers!

PR #841, opened 2019-05-15
open csv file as ISO-8859-1 to allow the use of 8 bits ascii char.

JulienDevillers:
# Description and justification
Users may use extended ascii char. In particular the µ char for µF.
As a result the µ char is not correctly rendered in the gui:

image

Current implementation opens csv file as 7bits ascii file (US-ASCII).
Switching to ISO-8859-1 does not change anything for people using only 7bits chars but allows the µ char to be displayed as it should:

image

# Implementation Details
Imported files with µ char in.

### DIFF
--- a/src/main/java/org/openpnp/gui/importer/NamedCSVImporter.java
+++ b/src/main/java/org/openpnp/gui/importer/NamedCSVImporter.java
@@ -267,7 +267,7 @@ public class NamedCSVImporter implements BoardImporter {
     private static List<Placement> parseFile(File file, boolean createMissingParts,
             boolean updateHeights) throws Exception {
         BufferedReader reader =
-                new BufferedReader(new InputStreamReader(new FileInputStream(file)));
+                new BufferedReader(new InputStreamReader(new FileInputStream(file), "ISO-8859-1"));
         ArrayList<Placement> placements = new ArrayList<>();
         String line;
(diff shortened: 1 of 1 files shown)


vonnieda:
Hi @JulienDevillers - thanks for the PR. I think this is a good idea, but I'd prefer UTF-8 if it's all the same. The rest of the system uses UTF-8 when character conversion is required, and UTF-8 is as close to we have as a universal standard. Would that work for you?

JulienDevillers:
Hi Jason,

I like the idea but I did try UTF-8  with an altium csv file and using
UTF-8 as charset did not solve the issue because altium csv file is not
utf-8.
That's why I choose iso8859. Certainly not perfect for the whole planet but
a bit better for some people.

I'm not 100% fluent when dealing with charsets, maybe I'm wrong but that's
my experience.

Le mer. 15 mai 2019 à 17:50, Jason von Nieda <notifications@github.com> a
écrit :

vonnieda:
Works for me. Thanks for checking in to it!

JulienDevillers:
Maybe a future solution would be to include a charset selection in the
"choose import file" dialog.

Le mer. 15 mai 2019 à 21:19, Jason von Nieda <notifications@github.com> a
écrit :

PR #842, opened 2019-05-15
Correct logic in maskHsv when hueMin is less than hueMax

tonyluken:
Description:
Since hue is a circular quantity it is perfectly reasonable to mask hues that cross the 255 - 0 boundary.  For instance, red hues range roughly from 235 through 21 (passing through the 255 - 0 boundary).  To mask such a range, the user would enter a hueMin (235) that is greater than hueMax (21).  Unfortunately the logic in mashHsv is incorrect in that case.  While it does correctly mask the desired hue range under that condition, the logic for masking saturation and value get inverted so that the returned image only contains pixels that have saturations and values that are in the range that should have been masked.

The fix to this problem is to create two masks, one for the hueMin to 255 range and one for the 0 to hueMax range, OR them together and then proceeding as in the case where hueMin < hueMax.

While I was in there, I also added descriptions to each of the parameters to maskHsv.

Justification:
Fixes a bug that makes the result of maskHsv incorrect and difficult to understand by the user (under some conditions).

Instructions for Use:
No special instructions - just use maskHsv in the pipeline as normal and it will work exactly the same as before when hueMin < hueMax and will return correct results when hueMin > hueMax

Implementation Details:
This change was tested by placing objects of various colors (read, green, and blue) in the camera's field of view and trying to mask them individually.  Prior to this change, attempting to mask red hues would result in many other parts of the image being masked incorrectly.
  
The coding style was followed.

No changes were made in the org.openpnp.spi or org.openpnp.model packages.

Maven test was run and passed.

### DIFF
--- a/src/main/java/org/openpnp/vision/pipeline/stages/MaskHsv.java
+++ b/src/main/java/org/openpnp/vision/pipeline/stages/MaskHsv.java
@@ -104,8 +104,8 @@ public class MaskHsv extends CvStage {
         }
         else {
             //Hue range wraps past 255 back through 0 so the mask needs to include the range from hueMin
-            //to 255 in addition to the range from 0 to hueMax so a mask for each separate range is created
-            //and ORed together to form the actual mask.
+            //to 255 in addition to the range from 0 to hueMax.  To accomplish this, a mask for each separate
+            //range is created and then ORed together to form the actual mask.
             min = new Scalar(hueMin, saturationMin, valueMin);
             max = new Scalar(255, saturationMax, valueMax);
             Core.inRange(mat, min, max, mask);
--- a/src/main/java/org/openpnp/vision/pipeline/stages/MaskHsv.java
+++ b/src/main/java/org/openpnp/vision/pipeline/stages/MaskHsv.java
@@ -10,7 +10,9 @@ import org.openpnp.vision.pipeline.CvPipeline;
 import org.openpnp.vision.pipeline.CvStage;
 import org.openpnp.vision.pipeline.Property;
 import org.openpnp.vision.pipeline.Stage;
+import org.pmw.tinylog.Logger;
 import org.simpleframework.xml.Attribute;
+import org.simpleframework.xml.core.Commit;
 
 @Stage(description="Remove color from an image based on the HSV color space. Pixels that fall between (hueMin, saturationMin, valueMin) and (hueMax, saturationMax, valueMax) are set to black in the output image. This stage expects the input to be in HSV_FULL format, so you should do a ConvertColor with Bgr2HsvFull before this stage and ConvertColor Hsv2BgrFull after. These are not applied internally as to not complicate the use of multiple instances of this stage in series. Note that this stage can be used with any 3 channel, 8 bit per channel color space. The order of the filtered channels is hue, saturation, value, but you can use these ranges for other channels.")
 public class MaskHsv extends CvStage {
@@ -38,6 +40,10 @@ public class MaskHsv extends CvStage {
     @Property(description="Maximum value to be masked.  Note values range from 0 to 255 (inclusive). Setting valueMax less than valueMin will result in no pixels being masked.")
     private int valueMax = 255;
 
+    @Attribute(required=false)
+    @Property(description="Inverts the selection of pixels to mask.")
+    private Boolean invert;
+    
     public int getHueMin() {
         return hueMin;
     }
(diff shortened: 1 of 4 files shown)


vonnieda:
Hi @tonyluken, thank you for the PR! I *think* there may be a conflict with a previous change to the behavior of this stage. Can you take a look at the change in https://github.com/openpnp/openpnp/commit/21883c47473fd87987413ff7a8b9ad64c0d4fd77#diff-ceaf7c1944cf848c03547f6b8d027839 and see if you think your changes are compatible. 

I believe the issue is that the way things work now hueMin > hueMax changes it from a "inclusive" match to an "exclusive" match, while your changes would remove "exclusive" match and replace it with an inclusive wrapping match. I may be wrong about this - I haven't tried it, but this is how it looks at first glance.

What do you think?

Thanks,
Jason

tonyluken:
Hi Jason,

Well, without any real documentation as to the intent of the previous change I can’t say for sure – all it says is “Improves MaskHsv to allow interior or exterior ranges.”  See [hsvColorSpace1.pdf](https://github.com/openpnp/openpnp/files/3188729/hsvColorSpace1.pdf)
for how the current version works compared to how I think it should work.  If the intent of hueMin > hueMax is to simply invert the entire set of pixels being masked (including saturation and value ranges) then it does as intended but in a confusing way and doesn’t provide anyway to mask hues that are in a range that cross the 255 – 0 boundary _while still being within specified saturation and value ranges_.  I would contend that if the intent is to provide a means to invert the entire selection, the addition of a Boolean “invert” checkbox would much clearer to the user and together with my changes would allow for hue ranges that cross the 255 – 0 boundary.  I could easily add the invert option if that is the consensus of the group and generate a new pull request.  Below are  some simple examples anyone can try that might illustrate what I’m talking about:

Example 1:
hueMax = 1
hueMin = 0
saturationMax = 1
saturationMin = 0
valueMax = 1
valueMin = 0

In this example, very little of an image should get masked since all the ranges are very narrow.  I’ve tested it and in fact it does just that  with both the current version and with my version.


Example 2:
hueMax = 255
hueMin = 254
saturationMax = 1
saturationMin = 0
valueMax = 1
valueMin = 0

Again, in this example very little of an image should get masked and it should mask nearly the same pixels as Example 1 since the hue ranges are directly adjacent.  I’ve also tested this and in fact it does just that  with both the current version and with my version.  So far so good.


Example 3:
hueMax = 1
hueMin = 254
saturationMax = 1
saturationMin = 0
valueMax = 1
valueMin = 0

Now this is the case I’m trying to address.  I contend that this example should just mask the pixels that were masked in Example 1 plus those in Example 2.  However, with the current version almost all pixels get masked.  The confusing part is having hueMin > hueMax changes the meaning of how the saturation and value ranges work:

When hueMin < hueMax the masked pixels are the **intersection** of the sets of pixels that are within the hue range,  those **within** the saturation range, and those **within** the value range.

However when hueMin > hueMax that gets changed to the **union** of the sets of pixels that are within the reversed hue range, those **outside** the saturation range, and those **outside** the value range.  At least to me, that’s not at all obvious from the description.

Regards,
Tony

vonnieda:
Hi @tonyluken - I agree entirely. I think your improvement is the "right" functionality. My only concern is that there may be users that are already using the existing inverted functionality, and I don't want to break existing users' configurations.

I think we can just upgrade settings on the fly to fit the new model. If you'll add the invert functionality and push a new commit I'll merge it and add the functionality to upgrade the configuration in the field. 

If you'd prefer to do the configuration work too, here is how I would go about it:

1. Add the new invert option as a Boolean with `@Attribute(required =false)`. Note Boolean, not boolean. This will allow the option to be loaded as null the first time.
2. In an `@Commit` method, check if invert == null and hueMin > hueMax. If both are true, swap hueMin and hueMax, and set invert = true. Otherwise set invert = false.

When the XML is first loaded for a MaskHsv stage this will run and "upgrade" the config. After that invert will always be != null, so the upgrade won't happen a second time.

The `@Commit` method is run automatically by the XML deserializer. The prototype looks like this:

```
@Commit
public void commit() {
}
```

tonyluken:
Hi @vonnieda, I made the changes and verified that the old machine.xml file gets updated to the new format.  I also verified that the image produced by the old maskHsv with hueMin > hueMax is bit-for-bit identical to the that produced by the new maskHsv with the hue limits swapped and the invert flag set.  That should make this 100% backward compatible with any previously defined pipelines.   I did add some trace level logging to the commit method so you can see when it is making the updates.  I was bit surprised to see some updates still being logged after I restarted openpnp after the machine.xml file had already been updated.  I tracked that down to when default pipeline .xml files that still contain the old format are loading.  While it works fine as is, I assume we should fix that as well. I'm not sure exactly where those files are located so I didn't make any updates to them.  If you can point my to them, I can make those updates as well.

tonyluken:
Ahh, just found the default pipeline .xml files and fixed them as well.

vonnieda:
Great work @tonyluken, thank you for taking care of the config and backwards compatibility, and for the very nice change in the first place!

PR #845, opened 2019-05-16
Added missing hint and removed UCase char in pipeline editor window.

JulienDevillers:
With only the icon, I was thinking that the first button was "rollback to original settings..."  :D


vonnieda:
Looks good, thanks @JulienDevillers!

PR #849, opened 2019-05-29
Adding a german translation to OpenPnP

pfried:
# Description

Adding a german translation to OpenPnP
Adding three new keys so that at least the top navigation is translatable
Removing some out of date comments
Sorting keys alphabetically to translate manually having a side by side comparison because of some shortcomings of the translation tool
Remove unused string key: "JogControlsPanel.Label.HomeIcon.Description"

# Translations in general
I think we should replace all string literals from the software. I would have started, but I wanted to open the PR to discuss.

Code tested


vonnieda:
Looks good, thank you @pfried!

pfried:
@vonnieda Do you think replacing all strings within the software makes sense? Maybe we should wait until 2.0 is done

vonnieda:
Yes, I think so. The user facing ones at least. It's a little bit of a slow
process because you have to be careful not to catch strings that are used
for internal functionality, but there aren't too many of those.

I'll be checking in one very big change tonight but after that I think it
will be okay to get started.

Jason

PR #852, opened 2019-06-02
Placements table appearance

aneox:
Hi! I like openpnp2, but i think placements table appearance should be changed. 
Please take a look [screenshot](https://yadi.sk/i/GjbR0aZcC2bp6A)

I have moved "Enabled" column far away from "Placed" column.
And i have changed placement color to white.
In my opinion, reading this table now become much better.

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
@@ -78,7 +78,7 @@ public class JobPlacementsPanel extends JPanel {
     private JobPanel jobPanel;
 
     private static Color typeColorFiducial = new Color(157, 188, 255);
-    private static Color typeColorPlacement = new Color(157, 255, 168);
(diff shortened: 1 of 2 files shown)


vonnieda:
I like it! Thanks for the nice improvement @aneox!

vonnieda:
@aneox Do you think we should adjust others tables to be similar? Primarily, moving the enabled column to the left?

aneox:
@vonnieda 
I think not necessarily.

main reason is that "Placed" column looks similar to "Enabled", same column with checkboxes, but column "Placed" is changing while job is running and it was hard to monitor it.

I thought about status column, it will be good to enhance it, add status "Done" with dark green color, also add status "Skipped" with orange color. But it need a lot of code changes(

vonnieda:
Okay, thanks! 

No need to spend more time on status column. Both status column and placed column are moving out of this table soon anyway. My plan here is that the current Job tab becomes Job Edit and is only used for setting up the job. Job status information will be moved to a new panel called "Production".

Issue #183, opened 2016-01-03
Refactor NozzleTips to be global


vonnieda:
Currently you have to duplicate nozzle tips for nozzles as this implies that in a multi tip system you have a bank of tips for each nozzle. I think this is unlikely and makes the system inflexible. This should be changed so that nozzle tips are a machine level object (instead of nozzle) and we just mark which ones are compatible with which nozzles. This makes it possible to load any compatible nozzle tip into any nozzle without having to have duplicate tips.

This will also make it easier to add better support for specifying packages compatible with nozzle tips without repeating data, and the same for vision information.
- Move NozzleTips to Machine (Global).
- Create a way to configure which NozzleTips are compatible with which Nozzles.
- Consider that some NozzleTips may be fixed to a specific Nozzle.
- Add NozzleTip / Part mapping and configuration to Part, instead of buried in NozzleTip.
- Consider the case where a machine has more than one of the same NozzleTip so that it can pick up two of the same parts at once. When mapping a Part to a NozzleTip we may need to do it by name or by allowing a multi-select.

cri-s:
There is only some remarks to be considered:
It can be that nozzles have fixed nozzle-tip.
As example two nozzle machine where one nozzle is fixed and the other can be changed.
Reason for doing that, non changeable nozzleTips have higher precision that automatic
changeables nozzles.
Some nozzleTips are compatible only with certain nozzle. 
Example are higher vacuum needed nozzles that can be used only on one specific nozzle
because machine construction gives that nozzle additional vacuum.

Annother consideration is NozzleChanger.
It's handled on machine specific implementation (machine/reference)
This is bad, because JobPlanner don't know how many Nozzles can be changed during Job processing as the api don't allow it.

vonnieda:
Thanks @cri-s, these are all good points. Please see my comments on #188 for how to get access to all the information about NozzleTips.

As part of the refactoring for this ticket I will make a new NozzleTipChanger interface that can be extended for different types of changers: manual, 3 position, carousel, etc.

cri-s:
Jason, i have extended the driver api a bit,
void dwell()  , wait until finished
void dwell(double)  wait time in seconds.
boolean GCode(string)
boolean GCode(string,time)

Actually the only not right is on firepick driver, where i probably
need community help if this
get accepted by you. Then i suggest to use GCode for driving it. I
suggest then to use GCode
for driving changer and i think, manual should be highest priority.
Currently i have tested only dwell(double) in order to go away with
[...]

PR #854, opened 2019-06-08
Feature/global nozzle tips

vonnieda:
Hi @markmaker, this is the finished up global nozzle tips code with your calibration changes merged in. Overall there were no issues merging, but there are three small things I'd like you to take a look at as it affects calibration. I will add comments on the individual lines below.

This PR fixes https://github.com/openpnp/openpnp/issues/183. NozzleTips now live on Machine and Nozzles each maintain a list of which NozzleTips are compatible.

Thanks,
Jason


src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
```diff
@@ -171,7 +161,7 @@ public void place() throws Exception {
     }
     
     private ReferenceNozzleTip getUnloadedNozzleTipStandin() {
-        for (NozzleTip nozzleTip : this.getNozzleTips()) {
+        for (NozzleTip nozzleTip : this.getCompatibleNozzleTips()) {
```
vonnieda: Since Nozzle no longer owns NozzleTips, I changed this to getCompatibleNozzleTips. It's not an exact analog of what you were doing before, since now the standin might be shared.

src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
```diff
@@ -264,7 +254,7 @@ public void moveToSafeZ(double speed) throws Exception {
     @Override
     public void home() throws Exception {
         Logger.debug("{}.home()", getName());
-        for (NozzleTip attachedNozzleTip : this.getNozzleTips()) {
+        for (NozzleTip attachedNozzleTip : this.getCompatibleNozzleTips()) {
```
vonnieda: Same here, as above. This will potentially cause the code below to run multiple times per nozzle tip, since multiple nozzles may have the nozzle tip configured as compatible. Perhaps it would make sense to do this on Machine.home() since that is where NozzleTips live now?

src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java
```diff
@@ -603,7 +603,7 @@ public static void resetAllNozzleTips() {
         // reset all nozzle tip calibrations, as they have become invalid
         for (Head head : Configuration.get().getMachine().getHeads()) {
             for (Nozzle nozzle  : head.getNozzles()) {
-                for (NozzleTip nt : nozzle.getNozzleTips()) {
+                for (NozzleTip nt : nozzle.getCompatibleNozzleTips()) {
```
vonnieda: And again, as above, changed from getNozzleTips to getCompatibleNozzleTips.

markmaker:
Hi @vonnieda 

Thanks! 

Regarding the review: we need to take one step back to what we discussed in PM. 

I'm on it, I'll push a proposal.  

_Mark

vonnieda:
Sounds good - looking forward to it!

-Jason

markmaker:
Hi @vonnieda 

I'm running out of time and don't want to block this. This is my proposal. I wasn't able to test it on the machine yet and as I have only a single nozzle machine this will be limited anyway. 

Like we agreed in the PMs before, I moved the calibration interface from NozzleTip to Nozzle. It combines the nozzle and the currently loaded NozzleTip to form the calibration target. Remember: we discussed how both the nozzle and the nozzle tip can have runout. As the same nozzle tip can change from one nozzle to the other, we need to be specific. In order to specifically store calibration for each nozzle on the nozzle tip, I changed it into a Map. 

While the calibration interface is now clearly and justifyably on the nozzle programmatically, it would be cumbersome to move the GUI too. All settings for the calibration are still stored on the nozzle tip and it would be a hassle for the user to move back and forth to do the actual calibration. 

Therefore the GUI must work backwards from the nozzle tip to identifiy the nozzle. For normal tips this is easy: it's just the nozzle that has the current nozzle tip loaded. Otherwise the user is reminded to load the right one. But for my "unloaded" stand-in it's a bit more complicated as multiple nozzles can be "unloaded" at a time.  In this case (and just for the GUI!) I favor the currently selected nozzle from the machine controls. 

Consider the review you opened covered by these changes. I guess it will become clear when you look at the code. 

**There is one thing known to be open:** if the user actually does as told and loads the right nozzle tip on the nozzle, the calibration "Status" info label does not change. Like I said my time ran out and frankly I don't know off-hand how to update the GUI by external events like that.  

Hope this is still useful in its untested / slightly unfinished form. 

_Mark

src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
markmaker: This is correct, with the additions I made.

src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
markmaker: Again correct with the additions I made.

src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java
markmaker: It can go directly to the nozzle tips. See my push.

src/main/java/org/openpnp/machine/reference/ReferenceNozzleTip.java
```diff
@@ -95,6 +94,11 @@ public String toString() {
         return getName() + " " + getId();
     }
 
+    @Override
```
markmaker: NOTE: I reverted this in the next commit.

markmaker:
Best to view the changes in one fell swoop:
https://github.com/openpnp/openpnp/pull/854/files/14992da6ab44f34a81b4cab67ed34628e22a13e2..7ca2d7044ff890020163d6c422aa8fac603d94fd

There was some some back-and-forth so the individual commits are not useful.

_Mark

markmaker:
Hi @vonnieda 

As you see in #855 (independent vacuum valve actuation) I performed first tests. It all worked just like that! :-)

As #855 is based on OpenPNP 2.0 + #854 (global nozzle tips) + #825 (extended nozzle tip runout and bottom camera offset compensation) this test also covers all these enormous changes. I'm quite happy in deed, 8-D

Congratulations and a big thank-you to you, @vonnieda!  

For my machine the `machine.xml` migration was not that difficult. Some hints for other early adopters: 

- copy the big `<nozzle-tips/>` XML block to the right place (I took a default OpenPNP 2.0 `machine.xml` as a guide). 
- start OpenPNP again and again, note the elements/attributes no longer supported and delete them (would be so nice to have a command-line option for the XML parser to ignore unknown elements and attributes, but it seems the parser in OpenPNP has no such thing*)
- define nozzle tip to nozzle compatibility on the Nozzle
- learn the new way to change nozzle tips (by clicking the checkbox in the list)
- define vacuum levels not forgetting the fact that isPartOff is now measured with opened valve for a moment (but see #855).
- might have forgotten some stuff 

 _Mark

*) BEGIN OT-RANT: unfortunately the IT mainstream soon forgot what the "X" in "XML" stands for, or falsely thought is was a one-way proposition. The industry engaged in a schema definition orgy completely ignoring the fact that XML is cool because it can best loosely couple systems on different and independent evolutionary paths, such as a web server and client (i.e. HTML). If you want to regulate everything with watertight schemas, so everybody has to use validators, generators and stubs and WTH else, you can just as well use the old binary RPC facilities or file formats. At least those were much faster. END OT-RANT.

markmaker:
Hi @vonnieda 

I'd like to fix the remaining issue here.

> There is one thing known to be open: if the user actually does as told and loads the right nozzle tip on the nozzle, the calibration "Status" info label does not change. Like I said my time ran out and frankly I don't know off-hand how to update the GUI by external events like that.

The "Status" label should be updated whenever:

1. A new tip is loaded or unloaded
2. The [selected tool changes](https://github.com/openpnp/openpnp/blob/fee9728215b6174da838a636539370c401a2287c/src/main/java/org/openpnp/gui/MachineControlsPanel.java#L129) in the MachineControlsPanel

i.e. it should do same as here:
https://github.com/openpnp/openpnp/blob/1c5073ca350c8d2cbc0eb3050cc4c55d175f1681/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java#L845

I looked into the code and I found so many possibilities, that I don't know what is the correct way to do it. I suppose it's bad style to spread the knowledge about one class into others, so direct calling from these places is bad. But vice versa is also bad, the nozzle tip calibration should not need to care about how the MachineControlsPanel GUI is made up. Machine head activity seems too broad and doesn't cover the selected tool (I think). 

Any hints?

_Mark

vonnieda:
Hey Mark, I'll look at it in detail this evening but the right way is to use a `PropertyChangeListener`. In general, every data holder object in OpenPnP extends `AbstractModelObject`, which includes `PropertyChangeSupport`: https://docs.oracle.com/javase/7/docs/api/java/beans/PropertyChangeSupport.html

Every setter should then `firePropertyChange("propertyName", oldValue, newValue)`.

Then, in the UI, when you want to monitor something for changes you do something like:

```
modelObject.addPropertyChangeListener("propertyName", e -> {
    view.updateWithNewValue(e.newValue);
});
```

So the general idea is you find a property to monitor, and set up a listener for it. If there is no property to monitor, or if the property isn't using property change events, that's a good opportunity to fix that property. In general there is no downside to adding property change support to an object.

That's the general gist. This is how most, but not all, GUI updates work in OpenPnP. Many of the Swing components automatically monitor for property changes, so you don't have to deal with the glue, and if you are using any of the bindings common in the wizards, the bindings do it for you.

In the case of setSelectedTool, that does not currently fire a property change event, but it should. Feel free to add it.

And a few other random notes, so that you know the direction things are going:

1. My next big set of changes, probably mid next week, is to remove Apply / Reset and begin implementing Undo. Part of that work will be making sure that every single GUI update happens from a property change event. It's necessary, because if a user performs an Undo the only way we'll know about it is if the property change fires an event.

2. The concept of the selected tool will probably go away, or be immensely changed, sometime this year. I want to get to the point where OpenPnP "does the right thing" rather than the user having to click all over the place for every operation. You will still be able to perform tool specific operations from the Machine Setup panel, but in general my hope is that once the machine is set up you don't have to worry about a selected tool. As a simple example, if you tell OpenPnP to pick a certain part, or a certain feeder, it should use a compatible nozzle and compatible nozzle tip, not whatever you happen to have selected.

So, the reason I mention it is that you may want to spend a minute thinking about whether monitoring that property is the best way to do it. Perhaps that status should instead be on a tab of the Nozzle setup, so that the "current" nozzle is obvious.

markmaker:
Hi Jason

Thanks, `PropertyChangeListener` helps.

Regarding the planned changes: 

I don't pretend to have the same holistic grasp of OpenPNP but I just suggest to postpone the removal of "selected tool" to the end off all changes and re-evaluate then. 

Yes, I agree things such as picking a part should employ all the pre-configured logic. I also saw the post where you announced that we will be able to execute individual placements from the job. Very nice!

But improving that, does not automatically mean, we should get rid of "selected tool", IMHO.  

The user can only open one wizard at a time, so some operations will become very cumbersome, if you have to work through the tool's wizard (e.g. the nozzle) to do something with it all the time, especially during machine setup. One example is the now missing nozzle tip "load" and "unload" buttons formerly available from the nozzle tip wizard. Now, I always have to go to the nozzle and back, including changing to the right tab in both wizard stacks (should stick to the front). I was actually planning a proposal to put the nozzle tip "load" and "unload" buttons back - not surprisingly using the "selected tool" to specify the nozzle. Try imagining setting up the nozzle tip changer without these button. Not funny.

To put it drastically: it's a bit like chopping off your left arm and then having to do everything one-handed, including tying your shoes. I know: not funny. :-/ 

_Mark

vonnieda:
Yep, I agree. Right now a lot of the setup tasks are painful with all the switching. Note, though, that I did add a feature in the last commit or so that keeps the same tab selected if you select the same type of object. So if you select Nozzle Tip 1, then tab 2, then select Nozzle Tip 2, tab 2 should remain selected. It helps, some.

It has been a long time since I've had to setup a machine from scratch, but right now I'm setting up the CHMT36VA over and over to force myself to see the problems, and the switching back and forth between dozens of tabs and panels is what is driving a lot of my work right now. It's terrible - especially Apply.

I'd be fine with putting the load and unload buttons back temporarily, and we can revisit how that works when selected tool goes away. It's all going to involve a lot of rethinking how the UI works with the goal of making it better, not worse :)

And just for some food for thought: When setting up the nozzle tip changer, why should you need a specific nozzle? Any "compatible" nozzle should do - so just use the first one that is compatible.

Jason

markmaker:
> And just for some food for thought: When setting up the nozzle tip changer, why should you need a specific nozzle? Any "compatible" nozzle should do - so just use the first one that is compatible.

Well, it could sometimes work with defaults but it would have to be smarter than "just use the first one that is compatible". For a "load" it should favour the first compatible nozzle that has no tip attached yet, right?  

Personally I don't like these asymmetrical and somewhat unpredictable automatisms when operating manually. I value automatic _suggestions_, e.g in spell checking, but I hate it when it just "auto-corrects" me. 

What about all the capture buttons? Will they capture the first nozzle with a tip loaded? Makes sense? What if you set up the nozzle changer (which is proably the hairiest capture situation)? There you need to capture the coordinates with the nozzle tip **off**. Will OpenPNP be smart enough to alter the capture button behaviour? Will OpenPNP **users** be smart enough to **predict** the changed behaviour? 

People won't know these hidden caveats especially when the are new and setting up their machines. Unfortunately, the first impression of OpenPNP always comes from setting up the machine. :-) 

Of course it will be a completely different experience for CHMT36VA users where you can simply download the ready-made `machine.xml` (what's the fun in that? :-)

But I just sincerely hope OpenPNP will not abandon the makers among us. :-/ 

_Mark

vonnieda:
> > And just for some food for thought: When setting up the nozzle tip changer, why should you need a specific nozzle? Any "compatible" nozzle should do - so just use the first one that is compatible.
> 
> Well, it could sometimes work with defaults but it would have to be smarter than "just use the first one that is compatible". For a "load" it should favour the first compatible nozzle that has no tip attached yet, right?

In my opinion, no. In fact, I'd like to get to the point where nozzle tips are not loaded at all unless they are being used. Eventually I'd like it if when a job ends the machine puts all the nozzle tips away. Easier to get to them for cleaning, service, etc.

So in this case, if we need a nozzle to set up a nozzle tip, we grab the first compatible nozzle, load the nozzle tip (unloading one if there's already one loaded) and get to it. The user doesn't have to think about which nozzle to use, whether it's got something loaded already, etc. We avoid "yak shaving". The user is not trying to select a nozzle, unload the nozzle tip, load the correct nozzle tip, and then check a loader position. They are trying to check a loader position. Everything else is wasted work.

> 
> Personally I don't like these asymmetrical and somewhat unpredictable automatisms when operating manually. I value automatic _suggestions_, e.g in spell checking, but I hate it when it just "auto-corrects" me.

I don't think there's anything unpredictable about this. It uses the first compatible nozzle - bottom line.

> 
> What about all the capture buttons? Will they capture the first nozzle with a tip loaded? Makes sense? What if you set up the nozzle changer (which is proably the hairiest capture situation)? There you need to capture the coordinates with the nozzle tip **off**. Will OpenPNP be smart enough to alter the capture button behaviour? Will OpenPNP **users** be smart enough to **predict** the changed behaviour?

We'll have to think through exactly how this works, but I don't see any reason it has to be complicated. I do want to get rid of excessive "capturing". I'm not looking to design this entire change right now - as I said, it's later this year. When the times I'll certainly post and people can weigh in.

> 
> People won't know these hidden caveats especially when the are new and setting up their machines. Unfortunately, the first impression of OpenPNP always comes from setting up the machine. :-)

I don't see these as hidden caveats. I think right now there is far too much to be aware of when performing any action. There is too much in the way of doing the thing you are trying to do. You have to think far too much about what the ramifications of any action are when the software should be doing that for you.


> 
> Of course it will be a completely different experience for CHMT36VA users where you can simply download the ready-made `machine.xml` (what's the fun in that? :-)

People will always need to tweak and calibrate a given machine. The `machine.xml` will get you the setup, but you will still need to tweak nearly every position and touch nearly every wizard.

> 
> But I just sincerely hope OpenPNP will not abandon the makers among us. :-/

OpenPnP is and always will be for makers. Right now OpenPnP is basically a giant pick and place toolkit. It's very hard to use to do the main thing that it should do: build boards. My goal over the course of this year is to make it good at building boards. Machine setup will always be a necessary evil, but it should not be the focus of the program.


It might help if I explain a bit where I'm coming from on this:

1. I'm currently working on a production SMT line for the first time in my life. I'm seeing where the machine and the software work to make the operator's job easier. I'm finding OpenPnP very lacking in this regard. Things that need to happen all the time take many clicks and tab switches and it's so frustrating.

2. I have been thinking about OpenPnP's UI for *years*. The UI has not changed much since I started the project nearly a decade ago. At the time I was building each piece as I went and no thought was really given to how it all worked together.

    Around 2015 I started to think a lot more about how the program operated as a whole. I started moving setup specific tasks into machine setup and removing them from the main, in your face UI. Before then things like nozzles, cameras, nozzle tips, etc. were all very front and center. This is also when I started prototyping the nav view.

    When I decided to release OpenPnP 2.0 I saw the opportunity to make massive changes that, again, I have been thinking about for years. I've held back from doing this for a long time because I knew it would break backwards compatibility for a lot of people. I realize there might be some fallout from this, and not everyone will be on board, but that is the case for how things are now, so I think it's worth trying some new ideas. This is open source - we can always change it if everyone hates it :)

Jason


> 
> _Mark

markmaker:
Thanks.

> OpenPnP is basically a giant pick and place toolkit.

And that's very cool! I love LEGO!

> very hard to use to do the main thing that it should do: build boards. 

Why not separate it in two GUIs?

1. A production GUI
2. A setup GUI

Each optimized for the job at hand, not compromising each other? 

_Mark

vonnieda:
> Thanks.
> 
> > OpenPnP is basically a giant pick and place toolkit.
> 
> And that's very cool! I love LEGO!

Legos are toys. This is a tool :)

(Yes I know people build hacks with Legos. Then if they want to use that thing more than a few times they rebuild it with not Legos.)

> 
> > very hard to use to do the main thing that it should do: build boards.
> 
> Why not separate it in two GUIs?
> 
>     1. A production GUI
> 
>     2. A setup GUI

Well, if you mean two separate programs my answer is because there is a lot of functionality that has to be shared and using two different programs will make that no easier. 

If you mean in the same program, with better separation, I agree completely. That's my goal for the "Production Panel" that is coming soon. It will bring together many of the things you need to do while running a job like resetting a feeder, seeing errors, adjusting a pick location, adjusting a place location, etc. 

Jason

markmaker:
> If you mean in the same program, with better separation, I agree completely.

Yes. An full screen overlay panel even. It could have a significantly different GUI approach for an entiely  different audience. But users could still open the shiny hood and get their hands dirty with the engine beneath.

To be honest the PR #825 moving target experience, the latest exchange and some of your announcements in the group left me quite insecure. Will OpenPNP 2.0 be a construction site with almost no chance to land new contributions (such as the feeder I'm developing)? While clearly being the "bleeding edge" with risk of bugs etc. will it still remain usable for assembling boards during the next half year? 

A clear separation of the new production GUI would significantly allay these worries for me. IMHO only such a clear abstraction would enable a truly fresh approach. Some of the "opinionated" stuff could be implemented on this higher level on top of the "**un**opinionated" machine that should more or less just offer the orthogonal capabilities of the machine. At the same time - by adding this new level on top - there would be less need to plow up the existing **fine** machine model underneath, so it remains a reliable target against which new stuff can be developed. 

Not to misunderstand me: global nozzle tips are a good thing on the machine level, as they better reflect the physical world. But there are other examples we discussed :-]

Had trouble formulating that in English, hope it's clear what I mean ... with all the best of intentions!

_Mark

vonnieda:
> > If you mean in the same program, with better separation, I agree completely.
> 
> Yes. An full screen overlay panel even. It could have a significantly different GUI approach for an entiely different audience. But users could still open the shiny hood and get their hands dirty with the engine beneath.

I think there is a misunderstanding here. The Production Panel will just be another tab. It's not a UI overhaul, and I'm not removing or changing the Machine Setup tab or anything like that. It will mostly be additional, new UI, not much in the way of modification of the existing UI.

> 
> To be honest the PR #825 moving target experience, the latest exchange and some of your announcements in the group left me quite insecure. Will OpenPNP 2.0 be a construction site with almost no chance to land new contributions (such as the feeder I'm developing)? While clearly being the "bleeding edge" with risk of bugs etc. will it still remain usable for assembling boards during the next half year?

I'm sorry about the moving target. We were both developing large, invasive features that overlapped. That's pretty rare in general and I don't expect it to be an ongoing problem.

I intend for 2.0 to be stable enough for daily use in the next 2-3 weeks. I put up a roadmap here: https://github.com/openpnp/openpnp/wiki/OpenPnP-2.0-Roadmap. Everything down to Production tab will be done in those next 2 - 3 weeks. After that it's mostly UI changes with very few changes to machine functionality.

In general, a feeder is two files and one modification to ReferenceMachine to add the class name. Integrating it should be trivial. If it requires SPI changes, that's a different story of course.

> 
> A clear separation of the new production GUI would significantly allay these worries for me. IMHO only such a clear abstraction would enable a truly fresh approach. Some of the "opinionated" stuff could be implemented on this higher level on top of the "**un**opinionated" machine that should more or less just offer the orthogonal capabilities of the machine. At the same time - by adding this new level on top - there would be less need to plow up the existing **fine** machine model underneath, so it remains a reliable target against which new stuff can be developed.

I think I've addressed this. There is no plowing up. In general I think that if you compare the .spi and .model packages a few weeks from now you will find there are very few changes in total compared to 1 month ago and even 3 years ago.

I want to be clear: The changes I'm making now are almost all entirely based on issues that have been filed for years and now is just a good chance to knock out a bunch of them that are interdependent. I'm not turning OpenPnP into a shiny black box that you can't hack. The UI will still remain separate from the machine model and you will still be able to (and required to) adjust every single property of every single object when configuring a machine.

My goal here is just to make OpenPnP good for production and not *just* for machine hacking. Think "shortcuts" rather than completely new maps.

Jason

vonnieda:
One more quick thought on this: None of the changes I'm making or planning are set in stone. These are just my ideas and opinions. If I make an announcement and you have concerns about it, let's discuss it.

markmaker:
> I intend for 2.0 to be stable enough for daily use in the next 2-3 weeks.

> I'm not turning OpenPnP into a shiny black box that you can't hack. 

> good for production and not _just_ for machine hacking. Think "shortcuts" rather than completely new maps.

Very glad to hear this! Phew. Thanks a lot.

_Mark

Issue #98, opened 2015-06-23
Add Z Touch Off


vonnieda:
This is several new features that should probably be broken out into their own issues.

1. The ability to detect a Z probe strike and create an event.
2. UI to show when a Z probe is triggered.
3. Automatic Z probe touch off.

vonnieda:
Initial thoughts regarding this feature:

	
	/**
	 * TODO: Add a way to z-probe. Ideally this would be automatic but we
	 * probably also want to be able to give UI feedback. So, perhaps we
	 * add:
	 * boolean isTouched()
	 * 
	 * something to the machine listener to fire an event when touch happens
	 * On that note, we should get rid of MachineListener and add events
	 * directly to the primary objects. Or we could add more discrete events
	 * to MachineListener such as nozzleTouched(Nozzle) and such.
	 * 
	 * void touchOff() which will run a zprobe and return when it's finished. It
	 * should adjust the nozzle's location to account for any required offset.
	 * In other words, if the zprobe requires compressing a spring, that
	 * compression offset should be subtracted from the location before it
	 * returns so that the final location is where the nozzle is just touching
	 * the probed location. This may require per-nozzletip configuration, so
	 * need to think about that too.
	 */

vonnieda:
New thinking on this is to make a new type of HeadMountable called Probe or Zprobe that would have it's own API, a la Nozzle, Camera or Actuator. This would allow the probe to be selected, deployed, etc. Likely any move against the probe would deploy it first.

vonnieda:
Current thinking, based on some chatter in IRC is to do this with pressure first. The problem of accidentally sucking up parts can be solved by touching off against the tape instead of the part, since it should be the same height. We can also always use the smallest nozzle tip for z-probe to reduce the risk.

vonnieda:
We'll probably need object specific z-probe systems. For instance, when setting the height of a board you probably don't want to probe the board corner because the board may be rounded - there might not be anything there. So instead it would make sense to probe the nearest placement, or something like that.

For feeders we'll want to probe the tape instead of the part so that we don't suck up a part.

We also need the ability to measure part height automatically, and that might include a pick, probe against a defined location and then discard. Or perhaps we just probe parts during their first usage in a job. Probably better to do the former, since we need height for other things like setting up bottom vision.

vonnieda:
Good ideas from this video https://www.youtube.com/watch?v=3kkf6bktIHA&feature=youtu.be&t=37s

[...]

PR #855, opened 2019-06-10
Feature/Independant Valve Switching for Nozzle.isPartOff()

markmaker:
# Description

ReferencePnpJobProcessor.Place.checkPartOff() currently uses Nozzle.pick() and Nozzle.place() for valve switching.

https://github.com/openpnp/openpnp/blob/fee9728215b6174da838a636539370c401a2287c/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L776-L790

This assumes that PICK_COMMAND / PLACE_COMMAND only contains G-Code to actuate the valve. That may not be true for users that add additional G-Code under the assumption that a “pick” really is a “pick” and not just a valve actuation.

E.g. my machine performs a Z-probe, valve actuation, dwell, slow (relative) retraction sequence there. I guess other users might have similarly elaborate pick and/or place commands.

This proposal actuates the valve inside Nozzle.isPartOff() using the already existing vacuum actuator (currently only used for reading) for switching the valve independently. It uses the ACTUATE_BOOLEAN_COMMAND in order to switch the valve. The same dwell time as in the Pick() operation is used. 

# Justification

PICK_COMMAND / PLACE_COMMAND should only be used for what their names imply. Separate valve switching should be independent of these. 

NOTE: this is upwards compatible. People that do not have the valve switching set up will simply not have valve switching in isPartOff() until they add it in the Actuator, so it will simply work as it did in OpenPNP 1.0 with the valve closed.  

# Instructions for Use

On your vacuum Actuator, define the ACTUATE_BOOLEAN_COMMAND. An example (for Smoothieware) could be:
```
{True:M10 ; open solenoid valve}
{False:M11 ; close solenoid valve}
```
This addition is also useful for the user to experiment with the valve and vacuum levels, using the Actuators tab and buttons in the Machine Controls.

grafik

NOTE: For the Nozzle.pick() and Nozzle.place() the valve still needs to be actuated using the PICK_COMMAND / PLACE_COMMAND G-Code as before. It may be embedded in the middle of the G-Code sequence there, so extracting it is not generally possible. In other words Nozzle.pick() and Nozzle.place() do still not use the Actuator to actuate the valve. This slight inconsistency is justified by the potential difference between a pick/place sequence down on the feeder/PCB with the part on the nozzle vs. simply actuating the valve on/off up at safe Z. 

# Implementation Details

1. This is a proposal for early OpenPNP 2.0. EDIT: it was tested on the machine, using the proven mini test job, runout calibration and multi-pass bottom vision pre-rotate. 
OpenPNP2_1stPnP
    Additionally the machine controles Actuator dialog was tested manually for mixed boolean and reading operation.   
3. No changes in the `org.openpnp.spi` or `org.openpnp.model` packages
4. It passed `mvn test`.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
@@ -555,13 +555,21 @@ public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMoun
     public boolean isPartDetectionEnabled() {
         return vacuumSenseActuatorName != null && !vacuumSenseActuatorName.isEmpty();
     }
-    
-    private double readVacuumLevel() throws Exception {
+
+    private Actuator getActuator() throws Exception {
         Actuator actuator = getHead().getActuatorByName(vacuumSenseActuatorName);
         if (actuator == null) {
             throw new Exception(String.format("Can't find vacuum actuator %s", vacuumSenseActuatorName));
         }
-        return Double.parseDouble(actuator.read());
+        return actuator;
+    }
+    
+    private void actuateVacuumValve(boolean on) throws Exception {
+        getActuator().actuate(on);
+    }
+
+    private double readVacuumLevel() throws Exception {
+        return Double.parseDouble(getActuator().read());
     }
 
     @Override
@@ -574,7 +582,18 @@ public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMoun
     @Override
     public boolean isPartOff() throws Exception {
         ReferenceNozzleTip nt = getNozzleTip();
-        double vacuumLevel = readVacuumLevel();
+        double vacuumLevel = Double.NaN;
+        try {
+            // We need vacuum on to determine the vacuum level.
+            actuateVacuumValve(true);
+            // Use the same dwell time as a pick.
+            Thread.sleep(this.getPickDwellMilliseconds() + nozzleTip.getPickDwellMilliseconds());
+            // Now read the vacuum level.
+            vacuumLevel = readVacuumLevel();
+        }
+        finally {
+            actuateVacuumValve(false);
+        }
         return vacuumLevel >= nt.getVacuumLevelPartOffLow() && vacuumLevel <= nt.getVacuumLevelPartOffHigh();
     }
 }
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
@@ -587,7 +587,7 @@ public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMoun
             // We need vacuum on to determine the vacuum level.
             actuateVacuumValve(true);
             // Use the same dwell time as a pick.
-            Thread.sleep(this.getPickDwellMilliseconds() + nozzleTip.getPickDwellMilliseconds());
+            Thread.sleep(this.getPickDwellMilliseconds() + nt.getPickDwellMilliseconds());
             // Now read the vacuum level.
             vacuumLevel = readVacuumLevel();
         }
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -751,14 +751,7 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
                 return;
             }
             try {
-                // We need vacuum on to determine the vacuum level.
-                nozzle.pick(part);
-                
-                boolean partOff = nozzle.isPartOff();
-                
-                nozzle.place();
-                
-                if (!partOff) {
+                if (!nozzle.isPartOff()) {
                     throw new JobProcessorException(nozzle, "Part detected on nozzle after place.");
                 }
             }


vonnieda:
I didn't intend to close this. I didn't notice that it had been submitted against the global-nozzle-tips branch instead of develop.

vonnieda:
It seems like maybe my comment I sent via email didn't get posted, so repeating it here:

Hi Mark, I'll comment fully on this tonight, but I wanted to make a quick comment now in case you are intending to work more on this today:

In short, pick() and place() are just poorly named right now. They should really be vacuumOn and vacuumOff. In the olden days of OpenPnP the pick() and place() functions actually were responsible for the entire pick sequence, but then that turned out to be too inflexible.

I understand the issue, but I'm not happy with this solution. I don't want to add "another way" to do something. If we need to refactor the driver interface so that there is a clear definition of how we turn the vacuum on and off, let's do that. And if we also need to think through the pick and place processes, we can do that too. But this solution feels like a bandaid and I'm not interested in bandaids when we are in the process of breaking the world :)

Also, as a minor actual concrete issue: This change will not actually turn on the pump in most configurations. Most people using dual nozzles use PUMP_ON and PUMP_OFF to control the pump, and these are fired from pick() and place().

markmaker:
Hi @vonnieda 

Glad you looked at it critically. 

> If we need to refactor the driver interface so that there is a clear definition of how we turn the vacuum on and off, let's do that. And if we also need to think through the pick and place processes, we can do that too. But this solution feels like a bandaid and I'm not interested in bandaids when we are in the process of breaking the world :)

Ok, in the spirit of "breaking the world" with the best of intentions, some questions/suggestions:

1. Do you agree that the machine building blocks are easiest to understand and manage for the user (as well as the developer), if they are actual UI and Java objects instead of just an artefact of some code behaviour? 
2. Do you agree that Actuators should be used for valve and pump? (and camera lights for that matter)
3. They could both act as BOOLEAN for the actuation and READ for the vacuum level. The pump could even act as WRITE for the setpoint (for machines that run a reservoir bottle on an hysteresis, like mine).
4. It would simplify/unify the G-code list of commands and simplify/unify scripting etc. 
5. Once the user understands config for one actuator, the next one is easy. 
6. Independant test actuation/read is readily available through machine controls. 
7. The code is free to use the building blocks for other puposes such as isPartOff() or say Z-Probe "sniffling". 
8. Caveat: Using Actuators would mean we'd need to split the PICK/PLACE commands in two, one before the valve actuation, one after. 
9. Suggestion: keep the PICK_COMMAND / PLACE_COMMAND, add two new PICKED_COMMAND, PLACED_COMMAND. Actuate the valve in between (this way it is upwards compatible).
10. Do not Thread.sleep() with the dwell times but define PickDwell% PlaceDwell% variables for G-code instead. This way the dwell can be placed where the user likes, and also in the middle of a sequence of G-code. This way the timing is done by the controller, synchronized to the machine actions and much more precise, which matters for speed-optimized short dwell times and/or fast vacuum decay rates (weak pump + large nozzle tip + now open valve in isPartOff()).  

_Mark

vonnieda:
>     1. Do you agree that the machine building blocks are easiest to understand and manage for the user (as well as the developer), if they are actual UI and Java machine model objects instead of just an artefact of some code behaviour?

Yes, agreed.

> 
>     2. Do you agree that Actuators should be used for valve and pump? (and camera lights for that matter)

Yes, on all three. Making valve and pump actuators would let us easily move pump control out of GcodeDriver and into Head or Machine, which is where it should be. Camera lights also yes, I will probably take an initial stab at that in the next day or two as I am working on a large set of camera changes.

> 
>     3. They could both act as BOOLEAN for the actuation and READ for the vacuum level. The pump could even act as WRITE for the setpoint (for machines that run a reservoir bottle on an hysteresis, like mine).

Good idea. I hadn't thought of it before but I suppose I sort of modeled the actuator after modbus a bit. I like the idea of using a single one as a register, a switch, a sensor, etc.

> 
>     4. It would simplify/unify the G-code list of commands and simplify/unify scripting etc.

Yes.

> 
>     5. Once the user understands config for one actuator, the next one is easy.

Yes.

> 
>     6. Independant test actuation/read is readily available through machine controls.

Yes, this is a nice benefit that people have asked for. 

> 
>     7. The code is free to use the building blocks for other puposes such as isPartOff() or say Z-Probe "sniffling".

Yes.

> 
>     8. Caveat: Using Actuators would mean we'd need to split the PICK/PLACE commands in two, one before the valve actuation, one after.

Hmm... 

> 
>     9. Suggestion: keep the PICK_COMMAND / PLACE_COMMAND, add two new PICKED_COMMAND, PLACED_COMMAND. Actuate the valve in between (this way it is upwards compatible).

I was leaning towards just getting rid of ReferenceDriver.pick() and ReferenceDriver.place() altogether, along with PICK_COMMAND and PLACE_COMMAND. If these are done by the actuator now, why do we need the additional API?

I think the argument will be "Because my nozzle does a z-probe, a short dwell, etc." and I think my answer to that is that you should use a ReferenceNozzle subclass, like the ones I showed Bert and Marek. I am open to including such subclasses in the main distribution if people don't want to maintain forks.

In general, a pick and place machine nozzle operates by lowering to a height, turning on vacuum, and raising to safe-z. That's the most common case by far. If you need more specialized functionality I think it should be done in a subclass.

> 
>     10. Do not Thread.sleep() with the dwell times but define PickDwell% PlaceDwell% variables for G-code instead. This way the dwell can be placed where the user likes, and also in the middle of a sequence of G-code. This way the timing is done by the controller, synchronized to the machine actions and much more precise, which matters for speed-optimized short dwell times and/or fast vacuum decay rates (weak pump + large nozzle tip + now open valve in isPartOff()).

I think this should also be done in a nozzle subclass. In fact, in general, I want to be moving as much code *out* of the driver as possible. Visual homing, squareness compensation, backlash compensation (maybe), axis mapping, etc. should all happen at a higher level, making it easier to override and extend. 

Remember, the drivers are also "Reference" classes. The concept of a driver doesn't even exist outside of the ReferenceMachine, there is no SPI interface for the driver. My intention always was that if a user needed extreme performance or complicated functionality then they should subclass Reference* to get there.

By moving more functionality out of the driver and into the high level code it becomes easier for someone to have a high performance ReferenceNozzle or even a ReferenceMachine that marshalls commands in a different way.

So, my counter-proposal:

* Remove ReferenceDriver.pick() and .place().
* Change ReferenceNozzle.pick() and .place() to use the vacuum actuator as you've described.
* Move pump control into ReferenceHead (probably) via actuator. The logic is basically "actuator.on = any nozzle has a part on"
* Move the full pick cycle and place cycle into ReferenceNozzle.pick() and .place(). By this I mean instead of the JobProcessor being responsible for lowering the nozzle, let ReferenceNozzle.pick() do that, and likewise with raising the nozzle. This solves a  bunch of problems for people who have complicated heads. This allows them to override these cycles in a subclass rather than having to modify the job processor. It also codifies the concept of a "pick" or "place" instead of that code being spread all over the place.

Thanks,
Jason

markmaker:
Very nice to ping and pong this out with you :-)

Agree on all points.

I'll think about a Z-probing ReferenceNozzle sub-class. For this I'd like to break out the actual Z-probing from pick() and place() too. Likely one more Actuator. Again, available from Machine controls, so people can capture feeder heights etc. everywhere. 

Cheers,
_Mark

vonnieda:
Thanks @markmaker!

BTW, I quietly added the first stab at z-probing the other day: 

https://github.com/openpnp/openpnp/blob/develop/CHANGES.md#openpnp-20-ongoing-changes

https://github.com/openpnp/openpnp/commit/326365b95d645aa3ba55cfb5d339bd82f3bad3f3

You set an actuator on the machine and then there is some basic API support for reading a value from it. I haven't spent much time on it yet but I wanted to get a start on it. Main thing that is left is to let the user set the units for the value.

Of note, I added it to the standard location buttons panel, so now if you have z-probing enabled and you "capture camera location" it will capture the camera coordinates and then go perform a z-probe at that location to fill in Z.

I'm not completely clear on why/how you do a z-probe in your pick:

> E.g. my machine performs a Z-probe, valve actuation, dwell, slow (relative) retraction sequence there. I guess other users might have similarly elaborate pick and/or place commands.

So if there is something else we need to consider WRT z-probing for that, let me know.

Thanks,
Jason

markmaker:
> I'm not completely clear on why/how you do a z-probe in your pick

It's largely driven by how the Liteplacer is designed. I never used it, but AFAIK you never ever need to set up part or feeder heights in the Liteplacer software. It is just built into the pick and place Z move to stop the nozzle when a certain counter-force is "felt". It's a soft long-way spring-loaded design with microswitch. A nod to the true prototyping use case where you dont' want to spend a lot of time setting up parts and feeders. **I do like this notion**.  

grafik
https://www.liteplacer.com/the-machine/assembly-instructions/pnp-head-step-11-attach-z-high-limit-switch/

It may also be due to the fact that it uses CP40 rather than Juki nozzle tips. CP40 tips are solid metal. Only the tip **holder** has a little spring, but that may be shorter-way and "harder" than Juki's. So the Liteplacer added its own spring (this is just a guess). 

grafik

> BTW, I quietly added the first stab at z-probing the other day:

Hmmm... that complicates things. It's a different concept of Z-probing. 

It seems the Liteplacer use case confounds the Z-probing with other nozzle tip action. For one thing, you associated the Z probe with the head and not the nozzle. This would be odd in a multi-nozzle machine that supports nozzle tip probing similar to Liteplacer. 

What if you wanted to Z-probe a feeders with one specific nozzle? The feeder might be at the far left/right and not reachable by the other nozzle (or using a different, non-compatible nozzle tip). Now that you have added the Nozzle to NozzleTip compatibility map, such a distinction might matter. 

> Of note, I added it to the standard location buttons panel, so now if you have z-probing enabled and you "capture camera location" it will capture the camera coordinates and then go perform a z-probe at that location to fill in Z.

Maybe the Z probing Actuator should be put on the HeadMountable instead of the Head. So in your use case (using the laser ranger you discussed in the group), you would associate it with the camera. It would be the natural thing, as you write the "capture camera location" button actually does the Z-probing.

In my use case I would put the actuator on the Nozzle. Equally the "capture nozzle tip location" could do the actual Z-probing. Though the spring-loaded Z-probe is rather "rough" and you don't want that always to happen (such as when capturing tip changer way-points). 

Hmmm... 

_Mark

vonnieda:
Thanks Mark, I understand now. Does the LitePlacer do this for every single pick and place cycle, or does it capture the Z coordinate for each feeder / part / board combination until it knows that combination and then run without it for the rest of the job?

markmaker:
> Does the LitePlacer do this for every single pick and place cycle

Yes. I think the original Liteplacer software fully depends on that.

However I should perhaps clarify.

Bottom vision is not used in the Liteplacer software other than for nozzle tip calibration AFAIK. The missing bottom vision is also the principal reason I went for OpenPNP from the start. 

But for bottom vision you need at least a rough part height. So in my setup I set a rough part height and use the Z probing just for the "final approach". It should be noted that this last millimeter or so happens at a reduced Z feedrate. Otherwise the momentum of the rather large and heavy nozzle column is too high (poor parts). 

Ideally of course the Z probe would only be used for the first pick to measure and record the feeder height (could be done after each machine homing). Then the machine could put the first part down on a flat ESD proof surface to measure and record the part height (I saw that in a video somewhere). From then on the operation could be standard. 

I'm also still thinking about stereoscopic depth measurements for both down-looking and bottom camera. The beauty would be that only software is needed and practically all the users out there could benefit. 
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/openpnp/S7pDpZGIjmk/SxLrj-xmAQAJ

Can you please tell me what lens you have on your cameras? 3.6mm? It seems only those are avaible lately under the link you proivide here:
https://github.com/openpnp/openpnp/wiki/OpenPnpCaptureCamera#troubleshooting 

My goal is a machine that recalibrates everything at least once per homing. So you can use maker quality cheap mechanics, keep hacking the machine, use non-precision dockable feeders, etc.  

_Mark

vonnieda:
Thanks for the clarification Mark. My hope for Z probing was to do something similar to ( https://www.youtube.com/watch?v=3kkf6bktIHA&feature=youtu.be&t=37s ) in the long run, and in the short term just make it easier to capture Z coordinates.

I sort of forgot that people would want to use nozzle vacuum sensing for Z probe, so I think I will need to go back to my original proposal in https://github.com/openpnp/openpnp/issues/98 which is that a Z probe is a type of HeadMountable, similar to a Nozzle or Actuator.

So, I think there will be an interface called Zprobe and I will write one or two concrete implementations. The first will be ActuatorZprobe which will work like what I have now. The next would be VacuumZprobe, and for this you would configure a Nozzle, NozzleTip, and maybe vacuum levels. Users will choose an implementation like they do now for Nozzles, Cameras, etc. This would allow you to write a StereoscopicZprobe, too, of course.

We'll have the same kind of situation as now where Head has getDefaultZprobe() and this will be used in most cases. But custom code or scripts would also be able to loop through the list of Zprobes and get the one most applicable to whatever task you want to perform. 

This will also help solve a problem, which is that depending on the type of Z probe you use, you may need to specify offsets for different objects. This is mentioned in #98 as well. With my laser sensor, I can't probe on the solder mask of a green board because the laser is red, so I need to probe on a fiducial. Having a an extensible object model for the probe will allow it to make intelligent decisions about how to should probe a given object.

So, I think the API will look something like: public Location Zprobe.probe(Location target, Object hint), where hint might be a BoardLocation, Feeder, Part, etc.

None of this has too much to do with the LitePlacer case, at least in the form that the original software takes. In that case you really do need a custom Nozzle type that just stops when it hits. Now, given what I've described above, I think it will be more efficient to just have a setting like "Measure Z heights on first cycle" or something and it would instead just use the Z probe system I've described above to record the Z heights and then operate at full speed.

Sounds good?

As for lenses: My DIY machine is in pieces - I'm using the CHMT36VA for development now. I'm not sure what lenses it uses. I *think* my DIY machine was using 8mm for top and 3.2mm for bottom.

Thanks,
Jason

markmaker:
Sounds good! And yes, that is exactly the video I meant.

_Mark

PR #857, opened 2019-06-20
Nozzle Tip Changing and Calibration UI Touches

markmaker:
# Description

- Bringing back the buttons to load/unload nozzle tips from the NozzleTip Wizards. The nozzle currently selected in the machine controls panel is used to perform the load/unload (or the default, if no nozzle is selected).
- Updating the Calibration Status info-text when a nozzle tip is loaded/unloaded or the if the currently selected nozzle changes in the machine controls panel. The latter is only relevant for the "unloaded" stand-in nozzle tip used to calibrate the naked nozzle. 
- Fix: when a nozzle tip is requested to be loaded on a nozzle tip, it needs to be unloaded first, if it is currently loaded on another nozzle.  
- EDIT: Removing the choice of the calibration system from the UI. On update, all configurations are forced to the "Model & Camera Offset" system. The calibration system can later still be changed through the `machine.xml`.

# Justification
The nozzle tip load/unload buttons are very useful when setting up the nozzle tip changer or runout calibration (pipelines etc.). 

The removal of the calibration system choice simplyfies the usage significantly. All other calibrations systems have pitfalls that can now be evaded for future users. For exotic cases, the other systems (e.g. the "Table" system) can still be enabled in the `machine.xml`. 

# Instructions for Use
- Use the buttons to load/unload nozzle tips (see screen shot). 
- On the Nozzle Tip Calibration Wizard try loading/unloading the current tip. See if the Status info changes correctly.
- Using a multi-nozzle machine, test if the Machine Control Panel selected nozzle is properly used in all load/unload operations.
- Using a multi-nozzle machine, test if the Nozzle Tip Calibration Wizard Status info of the "unloaded" stand-in nozzle tip is updated properly for the Machine Control Panel selected nozzle (even if other nozzles are unloaded too). 
- Using a multi-nozzle machine, try loading an already loaded tip on another nozzle. 
- Using a multi-nozzle machine, on the Nozzle Tip Calibration Wizard try calibrating the current nozzle tip if is loaded on a nozzle different that the Machine Control Panel selected nozzle (it should still work).

grafik

# Implementation Details
1. Dry tested as far as possible, but not on machine. Changes affect multi-nozzle machines, so my testing would be mostly inconclusive anyway. 
2. I did follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)
4. Ran `mvn test` before submitting the Pull Request.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java
@@ -416,8 +416,10 @@ public class ReferenceNozzleTipCalibration extends AbstractModelObject {
     }
 
     @Attribute(required = false)
-    private ReferenceNozzleTipCalibration.RunoutCompensationAlgorithm runoutCompensationAlgorithm = RunoutCompensationAlgorithm.Model;      // modelBased or tableBased? Two implementations are available
-
+    private ReferenceNozzleTipCalibration.RunoutCompensationAlgorithm runoutCompensationAlgorithm = RunoutCompensationAlgorithm.ModelCameraOffset;
+    
+    @Attribute(required = false)
+    private double version = 1.0;
 
     @Attribute(required = false)
     private RecalibrationTrigger recalibrationTrigger = RecalibrationTrigger.NozzleTipChangeInJob;
@@ -432,6 +434,13 @@ public class ReferenceNozzleTipCalibration extends AbstractModelObject {
     @Commit
     public void commit() {
         angleIncrement = null;
+        
+        // OpenPNP Version update
+        if (version < 2) {
+            version = 2.0;
+            // Force ModelCameraOffset calibration system.
+            runoutCompensationAlgorithm = RunoutCompensationAlgorithm.ModelCameraOffset;
+        }
 
         // We need to invoke this even later than the commit(), as the Configuration is not yet ready.
         SwingUtilities.invokeLater(() -> {
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java
@@ -8,6 +8,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
+import javax.swing.SwingUtilities;
+
 import org.apache.commons.io.IOUtils;
 import org.opencv.core.KeyPoint;
 import org.opencv.core.RotatedRect;
@@ -36,9 +38,6 @@ import org.simpleframework.xml.core.Commit;
 
 @Root
 public class ReferenceNozzleTipCalibration extends AbstractModelObject {
-    // Reference back to the nozzle tip.
-    private ReferenceNozzleTip nozzleTip;
-
     public static interface RunoutCompensation {
 
         Location getOffset(double angle);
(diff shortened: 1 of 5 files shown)


src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java
```diff
@@ -433,6 +432,25 @@ public Location getCameraOffset() {
     @Commit
     public void commit() {
         angleIncrement = null;
+
+        // We need to invoke this even later than the commit(), as the Configuration is not yet ready.
+        SwingUtilities.invokeLater(() -> {
```
markmaker: Hi @vonnieda 

Please review this use of `invokeLater()`. I got a null pointer exception when adding the listeners immediately. I'm still fresh with java :)

Thanks,
_Mark

markmaker:
Includes finishing touches from #854 (over from #825) as discussed here:
https://github.com/openpnp/openpnp/pull/854#issuecomment-501726138
and here:
https://github.com/openpnp/openpnp/pull/854#issuecomment-501796938

markmaker:
See the Wiki:
https://github.com/openpnp/openpnp/wiki/Nozzle-Tip-Calibration-Setup

src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java
vonnieda: Hi @markmaker, in general, the UI should listen to the model, not the other way around. The reason for this is that in some cases the UI will not exist - unit tests being the primary one. When this code is running as a unit test it will fail because SwingUtilities will not have an event thread, and MainFrame.get() will return null.

I'm not sure I'm following what the listener is intended to do exactly, but it looks like the goal here is to have the wizard update the calibration status when the selected tool changes. If that's the case then simply have the wizard add the listener from it's constructor. At that time the UI is guaranteed to exist, and the configuration is guaranteed to have been loaded.

Other than that, I'm good with the rest of the PR.

markmaker: Hi @vonnieda 

Thanks for reviewing this. I made another proposition, see the commit. 

My conclusion was to move most of the `"runoutCompensationInformation"` stuff from `ReferenceNozzleTipCalibration` to `ReferenceNozzleTipCalibrationWizard`. I renamed it "calibrationStatus". Now the listeners work model-to-UI and UI-to-UI as is appropriate, I guess. 

Note, "model" is meant as in "model view control" here, **not** as in "org.openpnp.model". 

## Why the move?

The `"runoutCompensationInformation"` that needs updating was a property of the `ReferenceNozzleTipCalibration` i.e. a model property, not UI. When I moved the `addPropertyChangeListener()` to `ReferenceNozzleTipCalibrationWizard` (as you suggested) it felt wrong. I was wiring a model-to-model change notification through the GUI. If the same Wizard were to be created multiple times, the same listeners would also added multiple times (don't know if this can ever happen in OpenPNP but it certainly feels wrong). By moving the property to the Wizard, this is fixed. This also nicely sorted things out with `getUiCalibrationNozzle()` that is no longer needed on the model side but only in the UI where it belongs. 


_Mark

vonnieda:
Nice work @markmaker. This looks much cleaner, and has a much better separation of concerns. 

Some thoughts below that might help with future work:

In general, there should be no UI related code in any of the org.openpnp.spi, .model, or .reference classes. This rule is somewhat broken by design because of WizardConfigurable and PropertySheet interfaces. When I wrote these, I should have used dependency injection or a global registry of UI classes. I intend to correct that "some day".

So, given that, what I try to make sure of is that none of the packages listed above ever call INTO the UI during normal operation. In other words a wizard calling a function from a reference class is fine, but a reference class calling a function in a wizard is not. As I mentioned before, this is important when it comes to unit testing, since no UI exists when the tests run. It's also important for maintaining MVC so that it's (relatively) easy to replace the UI if needed.

One final note: Wizard are indeed created multiple times - in fact, every time you open one it is created. This does result in listeners being added multiple times and currently this is the cause of a memory leak. Once a wizard has added itself as a property listener, it means that wizard will never be garbage collected. I have a global fix for this that I am working on as part of my "Remove Apply/Reset" branch. The fix is that listeners will be proxied through a weak reference, which will allow them to be garbage collected once no other references remain.

For the time being this is fine - it's been this way since the start - and since during normal operation most people aren't opening hundreds of wizards over and over it doesn't really cause a problem. But, as part of removing apply/reset, the listeners will now be added directly to much longer lived objects, so it will become more of an issue going forward. That's why I'll include the fix at that time.

Jason

markmaker:
Thanks!

>One final note: Wizard are indeed created multiple times - in fact, every time you open one it is created.

Are you sure? Oftentimes I forget to press Apply. When I go back to the Wizard later, it is still there, complete with the changed content, so it seems the Wizard is not deleted/recreated wheny I navigate away from it and back.

_Mark

vonnieda:
It depends on where in the interface you are, and how you get to and from it. If you are looking at a feeder wizard, from the Feeders tab for instance, moving to another tab does not reload the feeder wizard because the load process happens when you select a feeder. But selecting a different feeder will create a new wizard and the previous one will be lost. Likewise, in the machine setup tab, selecting a different object always creates a new wizard.

Easiest way to think of it is that the container that holds the wizard creates the wizard, and if the wizard in that container changes, the previous one is lost and a new one is created.

PR #859, opened 2019-06-23
Feature/Actuator based Pick+Place, Contact Probing Nozzle and independant Valve Switching

markmaker:
# Description

EDIT: this PR has changed guite a bit. See the discussion.

* removed pick() and place() methods from the drivers.
* removed GCodeDriver PICK_COMMAND/PLACE_COMMAND fragments.
* vacuum valves actuated by Actuators, controlled by Nozzle pick() and place()
* vacuum pump added to Head, actuated by Actuator
* valve actuation in turn actuates the pump if needed
* removed GCodeDriver PUMP_ON_COMMAND, PUMP_OFF_COMMAND.
* adapted testing to new actuator based valve (as much as I understand)
* changed all Wizard Actuator references into ComboBoxes for easy selection
* added new ContactProbeNozzle subclass to OpenPNP allowing pick and place with contact sensing Actuator. 

# Justification

As discussed previously in PR #855, a solution is needed for machines that use Z contact probing nozzles for pick an place. When the nozzle is lowered to the part on a pick / to the PCB on a placement, the nozzle will stop automatically when a certain contact force is sensed (Liteplacer style).

The OpenPNP 1.0 solution involved adding probing G-code to the PICK_COMMAND and the PLACE_COMMAND. But this was no longer possible for OpenPNP 2.0 because the pick() and place() methods are also used for valve switching during the isPartOff() vacuum test. Obviously, no probing should happen then. 

To cleanly separate the valve switching from the pick() and place() semantics, valve switching is moved to the vaccum actuator. Both reading the vacuum level and switching the vacuum valve is now implemented there. Users can then use the actuator to test vacuum valve switching and to determine vacuum levels for the isPartOn() and isPartOff() ranges they must enter in the configuration. 

At the same time Nozzle pick() and place() are really what the name implies and can be overridden to do more than just valve switching in a subclass. We're back where this PR originated: the ContactProbingNozzle sub-class. 

Some machines switch the vacuum pump on and off as needed (as long as at least one part is on a nozzle). In this PR the pump is therefore also modelled as a separate actuator with boolean control. The goal is to move towards a design, where more machine elements are modelled as separate objects rather than hide them inside GCode and entangle them with other functionality. A more general Actuator based redesign is achieved, where the user can test and use each machine element separately (for setup in the Machine Controls and for custom scripting). The user must only once learn how to setup an actuator and can reuse that knowledge many times. See the discussion for more detail on how this solution evolved.

The PR contains both the redesign and the new ContactProbingNozzle as its justification. The second is clearly separated in its own new file (plus one line added for registration). 

# Instructions for Use

## Vacuum actuator
On your GCode driver choose the existing vacuum Actuator, define the ACTUATE_BOOLEAN_COMMAND. An example (for Smoothieware) could be:
```
{True:M10 ; open solenoid valve}
{False:M11 ; close solenoid valve}
```
This addition is also useful for the user to experiment with the valve and vacuum levels, using the Actuators tab and buttons in the Machine Controls.

grafik

## Pump actuator
Create a new actuator on your head. 
grafik
grafik
grafik

Set it as the pump actuator:
grafik

On your GCode driver choose the new pump Actuator and define the ACTUATE_BOOLEAN_COMMAND. 

grafik
An example (for Smoothieware) could be:
```
{True:M106 ; turn on pump}
{False:M107 ; turn off pump}
```

## ContactProbeNozzle

Instead of using the `ReferenceNozzle`, create a `ContactProbeNozzle`. 
grafik

If you already have one or more `ReferenceNozzle`s set up, close OpenPNP, edit your machine.xml and replace 

`<nozzle class="org.openpnp.machine.reference.ReferenceNozzle" 
`
with 
 
`<nozzle class="org.openpnp.machine.reference.ContactProbeNozzle" 
`

Start OpenPNP again and add a new probing Actuator for each Nozzle. On the Nozzle, assign the probe:
grafik

On your GCode driver choose the new contact probing Actuator and define the ACTUATE_BOOLEAN_COMMAND for probing. 

grafik

An example (for Smoothieware) could be (Note, we are using M400 followed by M114.2 instead of plain M114 because Smoothieware does not report the rotation axis with M114):
``` 
{True:G38.2 Z-4 F1200   ; Probe down max. 4mm for contact with picked/placed part }
{True:M400          ; Wait until machine has stopped }
{True:M114.2        ; Report current realtime position, as M114 does not report rotation }
{False:G91 G0 Z1 G90; Retract 1mm }
{False:M400         ; Wait until machine has stopped }
{False:M114.2       ; Report current realtime position, as M114 does not report rotation }
```
You need to define the POSITION_REPORT_REGEX to read back the probed Z i.e. to get OpenPNP back in sync with the machine position after probing. 

grafik

For Smoothieware this could be:

`^ok MCS: X:(?<x>-?\d+\.\d+) Y:(?<y>-?\d+\.\d+) Z:(?<z>-?\d+\.\d+) A:(?<rotation>-?\d+\.\d+).*`

To test the contact probing, move your nozzle tip on top of an object not quite touching it. Then go to the machine controls, switch to the Actuators tab and click on the Probe Actuator:
grafik

You can now use the On and Off buttons to probe and retract. If you are too far above the contact surface, the machine will halt because it did not reach the contact within the 4mm given in the Gcode example above (adapt this value to your machine i.e. the contact probe travel/spring). This is another safety check halting the machine if something is not right (such as the feeder Z position not set right).

You should set your PCB Z and the feeder pick location Z so that the nozzle tip is not quite touching the part. The probing will then allow for variations within the probing distance, so you don't need super precision in your feeders etc. and your setup. 

Note the probed part Z height is not yet used within OpenPNP. A future PR could use the ContactProbingNozzle to calibrate part heights on the fly (first use) to improve precision in bottom vision (perfect focal plane) and in placement. 

# Implementation Details

1. This has evolved to be merged soon, IMHO. Final testing is pending.  
3. The spi Head was expanded to get the pump actuator. The spi Nozzle has changed valve methods. 
4. It passed `mvn test`.


markmaker:
Hi @vonnieda 

As you see I tried avoiding mixing in Z-probing per se. If Z-probing using the nozzle is later needed, it can be added on top, as an implementation of the ZProbe interface you discussed.

I left the driver pick() and place() intact, including the PICK_COMMAND/PLACE_COMMAND G-Code fragments, so a transition is possible. Also I don't know how other drivers will handle this (e.g. Neoden4). I guess it could be difficult for them to implement this with generic actuators. 

_Mark

vonnieda:
@markmaker I'm not comfortable with the dilution of the Nozzle interface here. Please see https://github.com/openpnp/openpnp/pull/855#issuecomment-500890284 for how this can be implemented. As I said before, I don't want to add "another way" to do this. We can choose a way that works and go with that, and users can migrate their settings.

In short, the changes should look quite a lot like:

1. Remove ReferenceDriver.pick() and ReferenceDriver.place(), and all that goes along with them.
2. Add configuration for the vacuum actuator into ReferenceNozzle.
3. Move the pump on/off logic from GcodeDriver to an actuator on ReferenceHead. 
4. Change Nozzle.pick() and Nozzle.place() to both take the Location that the action should happen at.

I believe this covers all known machine configurations for types of pick and place actions, and it keeps the interface semantically clear.

markmaker:
Hi @vonnieda

Thanks. I see I must have misunderstood the turn things took with the Z probing discussion. I had the impression, the probing motion needed to be extracted from pick(), place(). My mistake. I still don't really see how ZProbe will integrate with the subclass, but I trust you have a solution in mind.  :-)

Everything was crystal clear back at the https://github.com/openpnp/openpnp/pull/855#issuecomment-500890284 however, so I will proceed as you suggest. Though not any time soon ... the weekend is over. 

I've since seen how drivers other than the GCodeDriver (e.g. the NeoDen4Driver) handle generic Actuators. They just seem to interpret **reserved names** for Actuators in order to assign correct roles and trigger the right propriatory protocol (or whatever they need to do). 

_Mark

vonnieda:
> I still don't really see how ZProbe will integrate with the subclass, but I trust you have a solution in mind. :-)

Hi @markmaker, are you referring to the ZProbe feature in general, or with regards to your contact probing nozzle? If the latter - I don't think they are related per se. In other words, I don't think the contact probe nozzle would have to use a ZProbe.

The ZProbe feature will just be a new HeadMountable class, unrelated to nozzle or anything else. It will be up to concrete implementations of the ZProbe interface to determine how it works.

> I've since seen how drivers other than the GCodeDriver (e.g. the NeoDen4Driver) handle generic Actuators. They just seem to interpret **reserved names** for Actuators in order to assign correct roles and trigger the right propriatory protocol (or whatever they need to do).

Yes, this is one way to use the ReferenceActuator without dealing with implementing a new subclass specific to the driver. A more "correct" way would be to have an e.g. Neoden4Actuator and it would carry whatever information it needed to do it's job, but hardcoding actuator names is a quick a dirty way to get it done.

Jason

markmaker:
> are you referring to the ZProbe feature in general, or with regards to your contact probing nozzle? 

The latter. We [spoke about it](https://github.com/openpnp/openpnp/pull/855#issuecomment-501282321), you said:

> So, I think there will be an interface called Zprobe and I will write one or two concrete implementations. The first will be ActuatorZprobe which will work like what I have now. The next would be VacuumZprobe, and for this you would configure a Nozzle, NozzleTip, and maybe vacuum levels. Users will choose an implementation like they do now for Nozzles, Cameras, etc. This would allow you to write a StereoscopicZprobe, too, of course.

I must have misunderstood that. I thought it means a Nozzle or Camera subclasse could implement the ZProbe interface on top to provide Z-Probing as a secondary service. 

But I guess the right understanding is that the Probe would still be a separate machine model object, just _referencing_ the other object that actually provides the service, e.g. _reference_ the nozzle for vacuum sniffing or contact sensing or _reference_ the camera for stereoscopic capture. 

Hope I get it this time.

_Mark

vonnieda:
> > are you referring to the ZProbe feature in general, or with regards to your contact probing nozzle?
> 
> The latter. We [spoke about it](https://github.com/openpnp/openpnp/pull/855#issuecomment-501282321), you said:
> 
> > So, I think there will be an interface called Zprobe and I will write one or two concrete implementations. The first will be ActuatorZprobe which will work like what I have now. The next would be VacuumZprobe, and for this you would configure a Nozzle, NozzleTip, and maybe vacuum levels. Users will choose an implementation like they do now for Nozzles, Cameras, etc. This would allow you to write a StereoscopicZprobe, too, of course.
> 
> I must have misunderstood that. I thought it means a Nozzle or Camera subclasse could implement the ZProbe interface on top to provide Z-Probing as a secondary service.
> 
> But I guess the right understanding is that the Probe would still be a separate machine model object, just _referencing_ the other object that actually provides the service, e.g. _reference_ the nozzle for vacuum sniffing or contact sensing or _reference_ the camera for stereoscopic capture.

Yes, this is correct. ZProbe will be a "first class" interface, like Nozzle, Camera, etc. Not an extension of those. This keeps the interface simple and extensible. 

> 
> Hope I get it this time.
> 
> _Mark

markmaker:
> Yes, this is one way to use the ReferenceActuator without dealing with implementing a new subclass specific to the driver. A more "correct" way would be to have an e.g. Neoden4Actuator and it would carry whatever information it needed to do it's job...

Well, the more I think about it, the more I'm convinced that the semantics of an actuator is quite well-defined by **it's assignment**. If you take the vacuum sensing actuator assigned to Nozzle 1, it is actually quite clear what it represents and what's it supposed to do. 

We just need to be careful to assign actuators to machine model objects in a semantically unambiguous way. 

_Mark

markmaker:
Hi @vonnieda 

Note, Travis failure seems to be JDK11 related (it tries to install the JDK and fails, seem not related to the code).
https://travis-ci.org/openpnp/openpnp/jobs/549960857

So this is another try. Tried to follow 1:1 what you suggested. Quite a change!

One known deviation: In pick() I used the feeder as a parameter instead of the location. Why? Because there are feeders that need special motion to approach them. One example was discussed here:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/openpnp/3GhQhlIK-vw/xoUKbnQ3CQAJ
My new feeder that I'm developing will need that too. The idea is to later make it a method of the feeder to move the nozzle over the pick location. I'm very tired, so maybe that's not well described, sorry.

I also changed all the Actuator Name TextEdits into ComboBoxes: 
grafik

There is a question with the camera light Actuator. Vision can capture an image with or without settling. The light needs to be ON before settling. Of course I could just duplicate code in settleAndCapture() and in capture(), but it feels wrong and the light would be switched on/off twice with settling. Suggestions? 

Had a very hard time debugging the unit tests. 
- How can I force Eclipse to log all exceptions (including stack traces)? 
- How can I get LogUtils.isDebugEnabled() == true in tests? 
    (the mvn "Debug Output" option does not help)

**This is work in progress. No machine tests so far.** 

_Mark

src/main/java/org/openpnp/gui/MachineControlsPanel.java
```diff
@@ -419,8 +419,8 @@ public void mouseClicked(MouseEvent e) {
                     }
                     else if (e.getOldValue() != null && e.getNewValue() == null) {
                         for (int i = 0; i < comboBoxHeadMountable.getItemCount(); i++) {
-                            NozzleItem item = (NozzleItem) comboBoxHeadMountable.getItemAt(i);
-                            if (item.getNozzle() == e.getOldValue()) {
+                            HeadMountableItem item = (HeadMountableItem) comboBoxHeadMountable.getItemAt(i);
```
markmaker: This is just a bugfix. I think you added Actuators to the dropdown. Now those will create a cast exception.

vonnieda:
> Hi @vonnieda
> 
> Note, Travis failure seems to be JDK11 related (it tries to install the JDK and fails, seem not related to the code).
> https://travis-ci.org/openpnp/openpnp/jobs/549960857

Okay, I will look into this.

> 
> So this is another try. Tried to follow 1:1 what you suggested. Quite a change!
> 
> One known deviation: In pick() I used the feeder as a parameter instead of the location. Why? Because there are feeders that need special motion to approach them. One example was discussed here:
> https://groups.google.com/d/msg/openpnp/3GhQhlIK-vw/cD0CnAFxAwAJ
> My new feeder that I'm developing will need that too. The idea is to later make it a method of the feeder to move the nozzle over the pick location. I'm very tired, so maybe that's not well described, sorry.

I have some concerns about this, and it interferes with plans I have for part measurement. Two questions:

1. The thread you linked seems to be related to the user having put something in the way of their home location. The solution is to simply move the home switches, or not put something in the way. Changing the way pick works for one user's misplaced home location is not a good solution.

    The entire point of homing is that we don't know where the machine is yet and we need a reference, so having locations involved in homing is not logical.

2. Can you describe why your feeder will need a special pick motion? A special feed motion, I understand. That's what feed() is for. And a special post pick motion I also understand, and we have postPick() for that. Both are specific to the feeder. If we truly need another step here, it should be on the feeder, not the nozzle.

This seems to conflate picking with feeding, and I don't want to do that.

> 
> I also changed all the Actuator Name TextEdits into ComboBoxes:
> grafik
> 
> There is a question with the camera light Actuator. Vision can capture an image with or without settling. The light needs to be ON before settling. Of course I could just duplicate code in settleAndCapture() and in capture(), but it feels wrong and the light would be switched on/off twice with settling. Suggestions?

Are you adding light control via actuator here? That seems to be a completely unrelated feature and should be a different PR, I think?

> 
> Had a very hard time debugging the unit tests.
> 
>     * How can I force Eclipse to log all exceptions (including stack traces)?

You can add an exception breakpoint: https://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.jdt.doc.user%2Freference%2Fviews%2Fbreakpoints%2Fref-addexception_viewaction.htm


> 
>     * How can I get LogUtils.isDebugEnabled() == true in tests?
>       (the mvn "Debug Output" option does not help)

Look at the code in the log window that sets the global log level. You can use that in a `@Before` annotation to set it before the tests run: https://stackoverflow.com/questions/531371/what-are-junit-before-and-test

> 
> 
> **This is work in progress. No machine tests so far.**
> 
> _Mark


Before you put further work into this PR let's discuss it and make sure we're both on the same page. I don't want you to waste time just for me to ask you to change everything again.

Thanks,
Jason

markmaker:
Hi @vonnieda 

Thanks for the review.

> I have some concerns about this, and it interferes with plans I have for part measurement. Two questions:
> 1. The thread you linked ...

Like I said I was tired. :-/ I was just using this example to show that some feeders need to be approached carefully from one side. It has nothing to do with homing, that collision was just the foremost symptom discussed in the group. I was just assuming that - as a generalized case - collisions might also happen when picking parts.  

I remember seeing a DIY automatic feeder that fed the tape by the nozzle pushing down on a lever right in the pick (can't seem to find that now). It worked "naturally" with a single nozzle machine. But if the guy had used a multi nozzle machine, moving from one pick location to the next in a straight line would have collided with the feeder mechanics. Yes it can probably be done with Scripting and perhaps some reverse-feeder-for-part finding hack. 

> Can you describe why your feeder will need a special pick motion? ...

My new feeder is pushing away a cover from the cut strip. The nozzle is tenths of a millimeter above the tape doing that. True, it can be done in the feed operation, at least as long as the processor groups the feed operation with the pick operation on multi-nozzle machines (which is not self-evident, IMHO and should not be taken for granted in a feeder design). 

But I really want to keep the nozzle lowered right on the feeder and I do **not** want to move to safe Z between cover opening and pick. It would actually be safer to do the cover pushing as part of the pick motion for smoothest operation (to avoid vibration and parts to jump out, it's really, really delicate for 0402s). Note this is cheap 3D printed "mechanics", the cover flexes slightly and keeps some tension from being pushed. 

Admittedly these are all cheapo DIY examples abusing the nozzle for purposes, you might wrinkle your nose about. Having said that I'm quite sure there are very professional machines out there that need special approach paths to some of their specialty feeders as well. 

Ultimately that's why I really think the move to the feeder pick location should be made a method of the feeder for override. Sometimes, nozzle and feeder really need to work together closely and some "conflation" is inevitable. Unfortunately Java has no [multimethods](https://en.wikipedia.org/wiki/Multiple_dispatch) :-). 

I also think the feeder (of even the PlannedPlacement) should be much more involved and available in the whole placement process (all steps) and generally be avaible in the scripting parameters. Like [you said in a related matter](https://groups.google.com/d/msg/openpnp/7SckWf-sWkE/WvCqWcQBBwAJ);

> One thing missing is you can't tell exactly which feeder was used, but if you use one feeder per part you can use the part to find it. If you develop this in a script and find it works well, then we could integrate it further so we know the exact feeder.

Different approaches to providing that are conceivable.

> Are you adding light control via actuator here? That seems to be a completely unrelated feature and should be a different PR, I think?

Well, I thought you [agreed on this, there](https://github.com/openpnp/openpnp/pull/855#issuecomment-500890284):
> > 2. Do you agree that Actuators should be used for valve and pump? (and camera lights for that matter)
> > 
> 
> Yes, on all three. Making valve and pump actuators would let us easily move pump control out of GcodeDriver and into Head or Machine, which is where it should be. Camera lights also yes, I will probably take an initial stab at that in the next day or two as I am working on a large set of camera changes.

As it is the same pattern of refactoring things away from GCode and Scripts towards a set of Actuators linked from the right machine model object plus as I was introducing the ComboBox model for Actuators, it seemed the right moment to just take this off your shoulders. 

But yes it can be made a separate PR, if you wish. 

_Mark

vonnieda:
Thanks @markmaker, I understand now. Let me think more about this and get back to you soon. I do have a few quick thoughts:

1. It seems you really actually need something on Feeder, not on Nozzle, right? If you want the feeder to decide how it's approached, it seems that perhaps Feeder.pick(Nozzle) is needed? I am wondering if we can perhaps merge feed, pick, postPick, etc. all into Feeder.pick(Nozzle). This what I'll spend some time thinking about, but feel free to comment on the idea. It requires some thought because this is actually how it used to work a long time ago, and I changed it with presumably good reason, so I need to go revisit those decisions.

2. Regarding camera lighting: Yes, I agree this is the way to go, but I prefer to have PRs contain as little as possible, especially when new features are introduced. This makes it easier to discuss the feature, reason about it, merge it, and revert it if need be without also reverting unrelated features.

    I expect camera lighting to be a bit of a complex feature, as I'd like to offset control over color and mode, as well as just on and off.

Thanks,
Jason

markmaker:
Hi @vonnieda 

>  it seems that perhaps Feeder.pick(Nozzle) is needed

I fear that just reverses the problem, because the Nozzle wants to have its say too. Remember I actually wanted to create a ContactProbeNozzle subclass to create a Liteplacer style nozzle that stops on contact. **Both** the nozzle **and** the feeder are instrumental in the pick() step. It's one of these cases where the classical OOP single method dispatch fails to grasp the real world. 

https://en.wikipedia.org/wiki/Multiple_dispatch#Examples

In the absence of true multimethods, I think the practical solution is splitting up the pick() and place() into its sub-steps and letting the machine objects most closely associated with each sub-step override the sub-step methods. 

A pick() is composed of:

1. Move to the feeder pick location, first at safe Z, but may in the end involve a special approach path. Does not yet go down in Z (at least not all the way).
2. Lower the nozzle to the part, may involve contact probing.
3. Switch on the vacuum valve.
4. Retract nozzle minimally from feeder, may involve releasing the probe switch. 
5. Move up to safe Z, overrides may at the beginning involve a special retract path. 

Consequently I suggest wrapping all this up into the Nozzle.pick() but calling methods `feeder.moveToPickLocation(Nozzle)` and `feeder.retractFromPickLocation(Nozzle)` for sub-steps 1 and 5. The default implementation for a feeder is just a `moveToAtSafeZ(nozzle, pickLocationAtSafeZ)` for the former and a `moveToSafeZ()` for the latter. 

Sub-step 2 would be a protected nozzle method for override in subclasses. It would simply do the final Z down move in the plain nozzle. Override with contact probing in ContactProbeNozzle.  

Sub-step 4 would likewise be a protected nozzle method. It might do nothing in the plain nozzle. Or perhaps raise Z by part height at a lower feedrate to increase the chances that the part remains on the nozzle tip while untangling it from the feeder (tape pocket etc.). Which in turn would allow us to crank up the nominal Z speed safely. The override in ContactProbeNozzle would first retract until the probe switch is released. This could also be used to auto-learn the true feeder height, so maybe the probing could be limited to the first pick after homing for speed. The override would then call the super method.

The place() is simpler and follows from sub-steps 2 through 3.

If I'm not mistaken, you are strongly motivated to keep the top level interfaces very, very simple. I agree it makes sense to put up a hard fight before a more complex interface is allowed in.  For what it's worth, I would just argue that oversimplification can be as bad or worse than overcomplexity. 

What do you think?

> Regarding camera lighting...

I'll remove that. 

_Mark

markmaker:
Hi Jason

assuming that this bit is uncontroversial, I looked around a bit and found more Actuator links. Added the ActuatorsComboBoxModel to ReferenceAutoFeeder.

grafik

But now I saw the SwitcherCamera and SwitcherCameraConfigurationWizard. It links Actuators **by Id**! 

It would undoubtedly be better to link Actuators by Id rather than by Name! This could be the opportunity to change that. 

Side note: SwitcherCameraConfigurationWizard uses a static combobox i.e. Actuators are not updated, if created/deleted after the Wizard was first opened. Could be fixed by using ActuatorsComboBoxModel, once its Id based. 

_Mark

markmaker:
I edited [the proposal post](https://github.com/openpnp/openpnp/pull/859#issuecomment-505754412). Please read it online.

_Mark

markmaker:
Hi @vonnieda 

travis fails due to som certificate missing, it seems:
`ln: failed to create symbolic link ‘/home/travis/openjdk9/lib/security/cacerts’: No such file or directory
`

_Mark

markmaker:
Hi @vonnieda 

I thought about how to break up this PR. But while it would be cosmetically more pleasing to do some chopping up, I found little **real** reduction in complexity:

1. Separating the removal of LinuxCNC and OpenBuildsDriver would be easy, but it helps little. Just four files removed and two lines in the class registration. It actually simplifies this PR by eliminiating needed changes in these obsolete drivers. 
2. Separating my ContactProbeNozzle would be easy but it's just two completely separate files and one line registering the class. There can be no side effects from that. Conversely the new class demonstrates, why I _had to_ break out [two protected methods in ReferenceNozzle](https://github.com/openpnp/openpnp/pull/859/files#diff-b380475a49f8df054452d2236b8bb376R133). 
3. Separating the ComboBox Actuator UI would simplify the PR but adding new String edits just to replace them with ComboBoxes in a follow-up PR seems a bit silly. Otherwise if I base this PR on top of the ComboBox PR the diff remains the same until the former is merged. 
4. There [is a bugfix](https://github.com/openpnp/openpnp/pull/859/files#diff-a39fdabedaf60299b982407a991bf4fcR426) here, but we agreed earlier that small bugfixes can be integrated. 
5. In order to [integrate sub-classes' wizard(s)](https://github.com/openpnp/openpnp/pull/859/files#diff-c57561d7f72f7008ff3d9cc02de5afabR57) with those of the super class, I propose [merging the arrays](https://github.com/openpnp/openpnp/pull/859/files#diff-652192740849a85859e2d8587abaa20cR49). But is it really worth separating this small utility addition? 
6. Everything else is a quite straight-forward implementation of [the stuff we agreed upon plus your counter-proposal](https://github.com/openpnp/openpnp/pull/855#issuecomment-500890284). I reviewed it again just now and I'm actually even more assured its the way to go. 
7. The `feeder` param in `pick()` issue we can discuss. I could leave it out for now and propose it again, when I PR my new feeder. But I would like for you to first asses [my last response to that matter, if possible](https://github.com/openpnp/openpnp/pull/859#issuecomment-505754412).

Thanks for all your guidance and patience in this matter.

_Mark

markmaker:
Hi @vonnieda 

The latest commits should resolve remaining concerns about changing the `pick()` and `place()` semantics (see [point 7 here](https://github.com/openpnp/openpnp/pull/859#issuecomment-533788356)). 
* Methods `pick()` and `place()` will revert to practically the old meaning.
* They will **not** do the move to and away from the pick/place location, instead the job processor will do it (as it has in the past).
* No new `feeder` parameter to `pick()` needed. No new `location` parameter to `place()`.
* No change in the `Nozzle.pick() Nozzle.place()` interface.
* Changes in calling code reverted.
* ContactProbeNozzle does probing in straight overrides of `pick()` and `place()`
* No special Z motion/positioning is involved "pre-probing", we simply assume the user to capture the part/PCB Z heights slightly above the parts where the probe is not yet in contact/triggered.
* Special feeder approach motion will be solved later.

If you see "greener light" here, I will go on and test this on the machine, as soon as possible. 

Thanks for reviewing. 

_Mark

markmaker:
I've updated the description to reflect the latest state of this PR and to document how to use.

I will now go to the machine and test the latest changes.

markmaker:
Referenced here:
https://groups.google.com/d/msg/openpnp/763b0vw_PaM/-GRXeiXfBwAJ

A workaround was found, see the updated Description at the top.

markmaker:
Hi @vonnieda 

I have tested everything except the pump on/off governing on my machine. 

1. Using the Actuators freely on the machine controls.
2. Picking manually on the feeder.
3. Manual discard. 
4. Runing a mini job, including vacuum sensing, pick and place. 
5. ContactProbingNozzle working as expected.

I cannot reasonably test the pump on/off because my machine runs the pump underpressure on a hysteresis and I have a single nozzle machine, so I cannot test having multiple parts on the nozzles i.e. keeping the pump running as long as needed. 

I will report back when the M114 discussion is resolved.

_Mark

vonnieda:
Thanks @markmaker. I am working on merging PRs today and this is next on the list. I responded to your M114 question, so let's sort that out and then we'll get to merging.

markmaker:
Pushed the proposed change, so we know what it is about. 

https://github.com/openpnp/openpnp/pull/859/commits/1d4b8bda2c7ca5ad1575e439d7cf94af615a920e

markmaker:
I think both M114 issues were resolved, thanks:
https://groups.google.com/d/msg/openpnp/763b0vw_PaM/5jshx_XmBwAJ

It would be nice to be able to merge this. 

I'll be offline for the next time. Dinnertime. :-)

_Mark

vonnieda:
@markmaker Great work on this Mark! Thank you for patience and your persistence. This is a really nice improvement to OpenPnP.

I am ready to merge this as it is, but I have two asks:

1. Please add BREAKING CHANGE information to CHANGES.md and explain what users will need to do to migrate their configuration. This will be shown to users when they update and should help ease the pain of migration. 
2. Consider whether it might be worth not removing the GcodeDriver commands so that existing configurations will load without error. I think, as it stands now, if someone upgrades to this they will not be able to start OpenPnP. If you leave the commands in (but not any of the code that uses them) then users could start OpenPnP and copy and paste the commands to the new areas from the UI without having to edit machine.xml by hand.

Thanks,
Jason

vonnieda:
@markmaker Nevermind the above. I'm going to do those two things and then merge this.

markmaker:
> 
> 
> @markmaker Nevermind the above. I'm going to do those two things and then merge this.

Oh, hope we don't do it both! :-)

_Mark

vonnieda:
Thanks @markmaker!

markmaker:
@vonnieda 

Please feel free to use your own version and/or improve my text. Explaining such complex stuff in English is still veeery hard and I won't feel put off at all if you optimize it :-) 

Also I think it would help, if you were to quickly announce the breaking changes _in person_ on the list. Thanks a lot for this great project and the great way you are leading it!

Bed time for me...

_Mark

CircuitSetup:
@markmaker I have my vacuum solenoids on one controller and the vacuum sensors on another.  I can't combine the actuators because the G code would be sent to the wrong controller. 

I previously wasn't using the sensors because they were inconsistent. However, I see that with the latest build you have to have something in the Vacuum Actuator. How can I set this up properly?

CircuitSetup:
Actually, I figured out how to configure 1 actuator with 2 controllers. However, there is no way to not use vacuum sensing that I cam see, because an error is thrown that there is no vac actuator or the part errors if it doesnt get a value for the vac sensor. 

In my case I cant figure out why the vac sensor values are not working, and parts error every time. They vac sensor values are negative. Would that have anything to do with it?

markmaker:
Added a Wiki Page:
referenced from
https://github.com/openpnp/openpnp/wiki/Setup-and-Calibration%3A-Nozzle-Setup

PR #878, opened 2019-07-24
Allow escaped character sequences in Gcode strings

tonyluken:
# Description
This feature allows usage of Java style escape sequences to insert arbitrary Unicode characters in Gcode strings.  

# Justification
Some machine controllers may require special character sequences to perform certain actions.  For instance, recovering a TinyG from a hard alarm state (such as after a limit switch hit) can be accomplished by sending a ctrl-x character (\u0018) to its serial port (typically as part of the DISABLE_COMMAND).

# Instructions for Use
This feature is disabled by default and needs to be enabled  prior to use by checking the "Backslash Escaped Characters" checkbox on the GcodeDriver General Setting tab.  Then any occurrences of  \uxxxx (that is a backslash followed by a lower or upper case u followed by exactly four hexadecimal characters) in any Gcode string will be replaced by the Unicode character represented by that hexadecimal value prior to the string being sent to the controller.  In addition, the following escape sequences are also recognized: \t for tab, \b for backspace, \n for line feed, \r for carriage return, and \f for form feed.  Any other occurrences of backslashes will not be treated as an escape sequence and will be passed as-is to the controller. 

# Implementation Details
1. This change was tested both with the featured enabled and disabled as well to ensure no impact to those not desiring to use it.  
2. The coding style was followed.
3. No changes where made in the `org.openpnp.spi` or `org.openpnp.model` packages.
4. Maven tests were run and passed.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/AbstractReferenceDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/AbstractReferenceDriver.java
@@ -265,8 +265,15 @@ public abstract class AbstractReferenceDriver extends AbstractModelObject implem
                   if (i<len) {
                        c = s.charAt(i++);
                        if ((c == 'u') || (c == 'U')) {
-                            c = (char) Integer.parseInt(s.substring(i,i+4),16);
-                            i += 4;
+                          try {
+                               c = (char) Integer.parseInt(s.substring(i,i+4),16);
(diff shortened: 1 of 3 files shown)


vonnieda:
Looks great @tonyluken, thanks for the PR! This is something people have wanted for a long time. Great work!

PR #888, opened 2019-09-04
Feature/add scalars

codehero:
# Description
The Add pipeline only does a simple add and does not support subtraction.
This PR adds scalar parameters so that each stage output may be scaled.

# Justification
Subtraction lets me remove phantom edges from an image.
Default behaviour is not altered as the default scalar value is 1.

# Instructions for Use

The first scalar scales the first stage and must be non-negative.
The second scalar scales the second stage can can be negative.

# Implementation Details
1. How did you test the change? I added my change to a pipeline and verified the expected result.


codehero:
Something weird happened. I only wanted to PR 2b317ff. I will have to figure this out later, though I will accept any git advice.

codehero:
Closing and making new pull request

PR #887, opened 2019-09-04
Feature/set color transform

codehero:
# Description
Extend SetColor pipeline stage to apply camera transform.

# Justification
The edges of a camera transform create a nuisance when developing vision algorithms.
This PR is one step in eliminating that nuisance by creating a transformed SetColor block.
Using this stage, one can do edge detection and create a perimeter of the image.
This perimeter can be used to mask out the edge data from an image, eliminating phantom edges.

# Instructions for Use
Just tick the transform box in SetColor

# Implementation Details
1. How did you test the change? I applied a rotation in Image Transform to a camera. I tested by ticking/unticking the transform box. Default behaviour was preserved, and expected behaviour was correct.
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? Yes
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.

I modified the .spi for this specific purpose, but the function transformImage() is generic in scope.
There was no interference in existing functionality, and modifying ReferenceCamera was sufficient to implement the interface change.

PR #890, opened 2019-09-04
Feature/set color transform2

codehero:
Supersedes #887 

# Description

Extend SetColor pipeline stage to apply camera transform.

# Justification

The edges of a camera transform create a nuisance when developing vision algorithms.
This PR is one step in eliminating that nuisance by creating a transformed SetColor block.
Using this stage, one can do edge detection and create a perimeter of the image.
This perimeter can be used to mask out the edge data from an image, eliminating phantom edges.

# Instructions for Use

Just tick the transform box in SetColor

# Implementation Details
1. How did you test the change? Tested with USB camera. Applied a rotation, then checked that transformed image produced by SetColor matched the dimensions and visual appearance of the capture stream.

I modified the .spi for this specific purpose, but the function transformImage() is generic in scope.
There was no interference in existing functionality, and modifying ReferenceCamera and AbstractCamera was sufficient to implement the interface change.

Extension of the interface is justified by code working at the calibration level. While masking and templating works at the "user" level of working fiducials and footprints, code that tackles calibration needs to be aware of more parameters.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceCamera.java
@@ -218,6 +218,26 @@ public abstract class ReferenceCamera extends AbstractCamera implements Referenc
         }
         return height;
     }
+ 
+    @Override
+    public synchronized int getCaptureWidth() {
+        if (captureWidth == null) {
+            BufferedImage image = safeInternalCapture();
+            width = image.getWidth();
+            height = image.getHeight();
+        }
+        return captureWidth;
+    }
+
+    @Override
+    public synchronized int getCaptureHeight() {
+        if (captureHeight == null) {
+            BufferedImage image = safeInternalCapture();
+            width = image.getWidth();
+            height = image.getHeight();
+        }
+        return captureHeight;
+    }
 
     @Override
     public Location getHeadOffsets() {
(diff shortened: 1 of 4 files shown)


markmaker:
Hi @codehero 

can you elaborate, why you want to do it this way rather than crop the camera image or mask it or filter the vision result set, like I pointed out [here](https://github.com/openpnp/openpnp/pull/887#issuecomment-527758780)?

Your changes seem quite far reaching (and btw. they may result in an endless recursion/crash if used within a filter). 

The typical camera is optically inferior/distorted at the edges, and the angles start to go wide, so even small object distance errors create big pixel offsets. You can't get good results from anywhere near the edges, so might as well crop them.  

Most computer vision happens in the center of the camera and most objects have to be recognized at arbitrary angles, so the work area tends to be circular in the center of the camery anyway, i.e. MaskCircle is the natural thing to do. 

Why does this not work with your calibration? Otherwise, why can't you filter out the edges from the result set? 

Just to be clear we're not talking lens calibration, which happens **before** transformation. This is IMHO the only case where the full raw camera view area is really needed. 

_Mark

codehero:
I will complete the calibration work I am doing, then you can show me how you would do it.

vonnieda:
Closing this for now pending the above discussion. This can be reopened once there is agreement.

PR #891, opened 2019-09-13
Add optional capability to MaskHsv to automatically set mask limits

tonyluken:
**# Description** 
This change adds an optional capability (turned off by default) to have the min and max limits for each of the three channels set automatically such that the masked pixels are those about the channel's histogram peak.  The fraction of the image pixels masked is controlled by a single additional parameter.

**# Justification**
It can be difficult to find and tune the six current parameters to achieve a desired result.  This option allows the parameters to be quickly set to values that give reasonable results.

**# Instructions for Use**
To use this feature, set the auto checkbox and adjust the fractionToMask parameter to a value between 0.0 and 1.0 (inclusive).  Note that the auto checkbox can later be unchecked to freeze the min and max limits at the computed values (may need to select a different pipeline stage and reselect the MaskHsv stage to see the updated limits - I couldn't figure out an easy way to automatically cause them to update).  Note that this can be used to achieve a "green screen" effect: set the auto checkbox with fractionToMask=1.0 and pass an image of just the background color (it doesn't need to be green - just different that the objects you're trying to detect) through the pipeline.  Then uncheck the auto checkbox.  Now any images with objects of a different color than the background image will have the background masked.

**# Implementation Details**
**1. How did you test the change?** Ran various images through the pipeline on my machine and observed the affect on the resulting images.  Also verified that with the auto feature disabled, it works as previously.
**3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.**  No changes were made in those packages.
**4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.**  The Maven tests were run and all passed.

### DIFF
--- a/src/main/java/org/openpnp/vision/pipeline/stages/DetectLinesHough.java
+++ b/src/main/java/org/openpnp/vision/pipeline/stages/DetectLinesHough.java
@@ -1,0 +1,117 @@
+package org.openpnp.vision.pipeline.stages;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.opencv.core.MatOfPoint;
+import org.opencv.core.Point;
+import org.opencv.core.Mat;
+import org.opencv.imgproc.Imgproc;
+import org.openpnp.vision.pipeline.CvPipeline;
+import org.openpnp.vision.pipeline.CvStage;
+import org.openpnp.vision.pipeline.Property;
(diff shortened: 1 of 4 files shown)


vonnieda:
Really nice work @tonyluken, looking forward to using this! Thanks for the PR!

PR #892, opened 2019-09-14
Charmhigh  CHM-T48VB Integration

c-riegel:
# Description
These changes primarily add support for the features of the Charmhigh CHM-T48VB based on the firmware developed by Matt Baker.  Other tweaks were made along the way of implementing these features that should help with speed and safety. 

I'm new to git, so I apologize that these are all bundled together.  I thought git would let me make multiple pull requests based on my different commits and there were several issues that would come up as I was doing the integration testing.  My commits to git are more atomic if you can see them individually as part of the pull request.  However, I'm not sure how to break them up on my side.

# Justification
The changes were made so that the machine does not have to  be a T48VB.  Other machines with similar sensors can benefit.  Those without the specific sensors (photo-interrupters for the drag pin and Z-axis) will be unaffected.

# Instructions for Use
The part check was changed in ReferencePnpJobProcessor.
- This is worthy of review.  My tests have shown that the vacuum does not need to be turned back on in order to check for a part still being on the nozzle after placement.  In my experience if the part is still attached to the nozzle it is because the air line is pressurized and the reading indicates this without needing to turn the vacuum back on.

Buttons for turning the vacuum on and off were added to the Feeders tab.
- This is worthy of review.  This feature was only implemented for the Gcode driver.  As I was adding many feeder definitions it was helpful to have a quick button to control the vacuum.  The added icons attempt to reuse the openpnp color scheme.

Changes made to ReferenceDragFeeder:
 - Support was added to handle 0402 parts when not used with vision. Previously it would not get the second part that was feed from an 0402 feed when vision was not enabled.
 - The actuators section now has fields for identifying the drag pin actuator, the value the actuator will read when the pin is safely retracted, and the actuator for peeling off the film.  The feed process will try to confirm that the pin is safely retracted after backing off and if not it will try to back off again to release tension on the pin.  
 - Support was added so that the user can identify which well to pick the next part in the tape from.  This allows users to tell the machine which slots already have exposed parts (primarily useful for 0402 parts, but allows the user to not have to retract the drag feeder for parts).  The Locations section now contains a button labeled 'Set next part offset to:' that will identify which well to pick from.  Setting it to 0 will tell the machine to pick from the pick location without performing a drag operation.  Setting it to 1 will tell the machine to pick from the location in front of the pick location.  Setting it to negative values will tell it to perform a standard feed operation that would include a drag step.

Changes made to ReferenceHeadConfigurationWizard:
[...]

PR #893, opened 2019-09-17
Charmhigh Safety Checks

c-riegel:
# Description
Implements safety checks supporting PnPs from Charmhigh

# Justification
The machine should not crash into itself during operation.

# Instructions for Use
The Machine Setup > ReferenceMachine > Heads > ReferenceHead > Nozzles > ReferenceNozzle now has two new fields under "Safe Z": "Safe Z Actuator Name" is the actuator to read for a safety check to confirm that the nozzle is at safe-z, "Safe Z Value" is the value that will be compared with the reading from "Safe Z Actuator Name" to confirm the nozzle is at safe-z.

ReferenceDragFeeder now has an "Pin Retracted Value" under the "Actuators" section.  After a drag operation the drag pin actuator will be read and compared with the "Pin Retracted Value" to confirm that the pin has been safely retracted.

Edit by @vonnieda, this looks like it was reverted: ~~ReferenceHead was changed to go from saying "Z Probe" to "Z Contact" to better identify what it is for.~~

# Implementation Details
By initially breaking a 502 tip and bending my drag pin and then working to correct this on my machine.
No, minimal changes were used instead of following the coding style.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
@@ -157,6 +157,8 @@ public class ReferenceDragFeeder extends ReferenceFeeder {
 	        }
 		}
 
+        head.moveToSafeZ();
+
         if (vision.isEnabled()) {
             if (visionOffset == null) {
                 // This is the first feed with vision, or the offset has
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
@@ -81,6 +81,8 @@ public class ReferenceDragFeeder extends ReferenceFeeder {
     protected String actuatorName;
     @Attribute(required = false)
     protected String peelOffActuatorName;
+    @Attribute(required = false)
+    protected String pinReturnActuatorValue;
     @Element(required = false)
     protected Vision vision = new Vision();
     @Element(required = false)
@@ -137,9 +139,9 @@ public class ReferenceDragFeeder extends ReferenceFeeder {
          * it would have already hit.
          */
 
-        Actuator actuator = head.getActuatorByName(actuatorName);
+        Actuator dragPinActuator = head.getActuatorByName(actuatorName);
 
-        if (actuator == null) {
+        if (dragPinActuator == null) {
             throw new Exception(String.format("No Actuator found with name %s on feed Head %s",
                     actuatorName, head.getName()));
         }
@@ -154,8 +156,6 @@ public class ReferenceDragFeeder extends ReferenceFeeder {
 						peelOffActuatorName, head.getName()));
 	        }
 		}
-
-        head.moveToSafeZ();
 
         if (vision.isEnabled()) {
             if (visionOffset == null) {
@@ -189,16 +189,15 @@ public class ReferenceDragFeeder extends ReferenceFeeder {
 	        }
 
 	        // Move the actuator to the feed start location.
-	        actuator.moveTo(feedStartLocation.derive(null, null, Double.NaN, Double.NaN));
+            dragPinActuator.moveTo(feedStartLocation.derive(null, null, Double.NaN, Double.NaN));
 
 	        // extend the pin
-	        actuator.actuate(true);
-
-	        // insert the pin
-	        actuator.moveTo(feedStartLocation);
+            dragPinActuator.actuate(true);
 
 	        // drag the tape
-	        actuator.moveTo(feedEndLocation, feedSpeed * actuator.getHead().getMachine().getSpeed());
+            dragPinActuator.moveTo(feedEndLocation, feedSpeed * dragPinActuator.getHead()
+                                                                               .getMachine()
+                                                                               .getSpeed());
 
 			if (peelOffActuator != null) {
 				// peel off before backoff
@@ -206,14 +205,46 @@ public class ReferenceDragFeeder extends ReferenceFeeder {
 				peelOffActuator.actuate(false);
 			}
 
+            // retract the pin
+            dragPinActuator.actuate(false);
+            
 	        // backoff to release tension from the pin
 	        if (backoffDistance.getValue() != 0) {
-	            Location backoffLocation = Utils2D.getPointAlongLine(feedEndLocation, feedStartLocation, backoffDistance);
-	            actuator.moveTo(backoffLocation, feedSpeed * actuator.getHead().getMachine().getSpeed());
+                Location backoffLocation = Utils2D.getPointAlongLine(feedEndLocation,
+                        feedStartLocation, backoffDistance);
+                dragPinActuator.moveTo(backoffLocation, feedSpeed * dragPinActuator.getHead()
+                                                                                   .getMachine()
+                                                                                   .getSpeed());
+            }
+
+            // if possible, confirm drag pin retraction
+            if (pinReturnActuatorValue.length() > 0) {
+
+                if (!dragPinActuator.read()
+                                    .equals(pinReturnActuatorValue)) {
+                    // if we're here then the pin is stuck, so what follows tries to back off a
+                    // second
+                    // time by the same amount to try to get the pin free from binding
+                    Logger.debug("\n\nfeed: move to pin backoff 2");
+                    if (backoffDistance.getValue() != 0) {
+                        Location backoffLocation2 = Utils2D.getPointAlongLine(feedEndLocation,
+                                feedStartLocation, backoffDistance.add(backoffDistance));
+                        dragPinActuator.moveTo(backoffLocation2,
+                                feedSpeed * dragPinActuator.getHead()
+                                                           .getMachine()
+                                                           .getSpeed());
 	        }
 
-	        // retract the pin
-	        actuator.actuate(false);
+                    // try retracting the pin again (maybe the command has a sleep that would
+                    // help)
+                    dragPinActuator.actuate(false);
+                }
+
+                if (!dragPinActuator.read()
+                                    .equals(pinReturnActuatorValue)) {
+                    throw new Exception("Unable to retract drag pin");
+                }
+            }
 
 	        if (this.isPart0402() == true) {
 				// can change it to "feededCount = parts_count_userSettings;"
@@ -221,11 +252,8 @@ public class ReferenceDragFeeder extends ReferenceFeeder {
 	        }
         } 
         else {
-			Logger.debug("Multi parts drag feeder: skipping drag " + feededCount);
+            Logger.debug("Multi parts drag feeder: skipping drag feededCount = " + feededCount);
         }
-
-
-        head.moveToSafeZ();
 
         if (vision.isEnabled()) {
             visionOffset = getVisionOffsets(head, location);
@@ -396,6 +424,16 @@ public class ReferenceDragFeeder extends ReferenceFeeder {
     public void setPeelOffActuatorName(String actuatorName) {
         String oldValue = this.peelOffActuatorName;
         this.peelOffActuatorName = actuatorName;
+        propertyChangeSupport.firePropertyChange("actuatorName", oldValue, actuatorName);
+    }
+
+    public String getPinReturnActuatorValue() {
+        return pinReturnActuatorValue;
+    }
+
+    public void setPinReturnActuatorValue(String actuatorName) {
+        String oldValue = this.pinReturnActuatorValue;
+        this.pinReturnActuatorValue = actuatorName;
         propertyChangeSupport.firePropertyChange("actuatorName", oldValue, actuatorName);
     }
 
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
@@ -56,6 +56,12 @@ public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMoun
     @Element(required = false)
     protected String vacuumSenseActuatorName;
     
+    @Element(required = false)
+    protected String zSafeActuatorName = "";
+
+    @Element(required = false)
+    protected String zSafeActuatorValue = "";
+    
     /**
      * If limitRotation is enabled the nozzle will reverse directions when commanded to rotate past
      * 180 degrees. So, 190 degrees becomes -170 and -190 becomes 170.
@@ -304,6 +310,20 @@ public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMoun
                 safeZ.getValue(), Double.NaN);
         getDriver().moveTo(this, l, getHead().getMaxPartSpeed() * speed);
         getMachine().fireMachineHeadActivity(head);
+        
+        if (this.zSafeActuatorName.length() > 0 && this.zSafeActuatorValue.length() > 0) {
+            Actuator zSafeActuator = getHead().getActuatorByName(zSafeActuatorName);
+            if (zSafeActuator == null) {
+                throw new Exception(String.format("No Actuator found with name %s on Head %s",
+                        zSafeActuatorName, this.getName()));
+            }
+            String zProbeValue = zSafeActuator.read();
+            if (!zProbeValue.equals(zSafeActuatorValue)) {
+                throw new Exception(String.format(
+                        "Safe Z Actuator %s says Nozzle %s is not at safe Z (actuator is reading %s and not %s)",
+                        zSafeActuatorName, this.getName(), zProbeValue, zSafeActuatorValue));
+            }
+        }
     }
     
     @Override
@@ -618,5 +638,21 @@ public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMoun
         ReferenceNozzleTip nt = getNozzleTip();
         double vacuumLevel = readVacuumLevel();
         return vacuumLevel >= nt.getVacuumLevelPartOffLow() && vacuumLevel <= nt.getVacuumLevelPartOffHigh();
+    }
+    
+    public String getzSafeActuatorName() {
+        return zSafeActuatorName;
+    }
+
+    public void setzSafeActuatorName(String zSafeActuatorName) {
+        this.zSafeActuatorName = zSafeActuatorName;
+    }
+
+    public String getzSafeActuatorValue() {
+        return zSafeActuatorValue;
+    }
+
+    public void setzSafeActuatorValue(String zSafeActuatorValue) {
+        this.zSafeActuatorValue = zSafeActuatorValue;
     }
 }
--- a/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceDragFeederConfigurationWizard.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceDragFeederConfigurationWizard.java
@@ -44,8 +44,6 @@ import javax.swing.border.EtchedBorder;
 import javax.swing.border.TitledBorder;
 
 import org.jdesktop.beansbinding.AutoBinding.UpdateStrategy;
-import org.jdesktop.beansbinding.BeanProperty;
-import org.jdesktop.beansbinding.Bindings;
 import org.openpnp.gui.MainFrame;
 import org.openpnp.gui.components.CameraView;
 import org.openpnp.gui.components.ComponentDecorators;
@@ -79,13 +77,13 @@ public class ReferenceDragFeederConfigurationWizard
     private JTextField textFieldFeedEndY;
     private JTextField textFieldFeedEndZ;
     private JTextField textFieldFeedRate;
-    private JLabel lblActuatorId;
-    private JLabel lblPeelOffActuatorId;
     private JLabel lbl0402PartDetected;
     private JTextField textFieldActuatorId;
     private JTextField textFieldPeelOffActuatorId;
+    private JTextField textFieldRetractActuatorValue;
     private JPanel panelGeneral;
     private JPanel panelVision;
+    private JPanel panelActuators;
     private JPanel panelLocations;
     private JCheckBox chckbxVisionEnabled;
     private JPanel panelVisionEnabled;
@@ -127,8 +125,12 @@ public class ReferenceDragFeederConfigurationWizard
                 new ColumnSpec[] {FormSpecs.RELATED_GAP_COLSPEC, FormSpecs.DEFAULT_COLSPEC,
                         FormSpecs.RELATED_GAP_COLSPEC, FormSpecs.DEFAULT_COLSPEC,
                         FormSpecs.RELATED_GAP_COLSPEC, FormSpecs.DEFAULT_COLSPEC,
-                        FormSpecs.RELATED_GAP_COLSPEC, FormSpecs.DEFAULT_COLSPEC,},
+                        FormSpecs.RELATED_GAP_COLSPEC, FormSpecs.DEFAULT_COLSPEC,
+                        FormSpecs.RELATED_GAP_COLSPEC, ColumnSpec.decode("left:default:grow"),},
                 new RowSpec[] {FormSpecs.RELATED_GAP_ROWSPEC, FormSpecs.DEFAULT_ROWSPEC,
+                        FormSpecs.RELATED_GAP_ROWSPEC, FormSpecs.DEFAULT_ROWSPEC,
+                        FormSpecs.RELATED_GAP_ROWSPEC, FormSpecs.DEFAULT_ROWSPEC,
+                        FormSpecs.RELATED_GAP_ROWSPEC, FormSpecs.DEFAULT_ROWSPEC,
                         FormSpecs.RELATED_GAP_ROWSPEC, FormSpecs.DEFAULT_ROWSPEC,}));
 
         JLabel lblFeedRate = new JLabel("Feed Speed %");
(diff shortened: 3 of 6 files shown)


markmaker:
Hi @c-riegel 

Some questions/comments:

1. I understand why a machine would want to perform a safety check for Z being retracted before moving in X-Y, especially in manual jogging mode. But this would then have to be a part of the general `moveTo()` method and in OpenPNP it would have to be coupled with some logic to allow some legitimate (small) X-Y moves with Z down such as over the bottom camera or inside the nozzle tip changer (or the new feeder I'm developping). If you add it to `moveToSafeZ()` then the safety check is redudant or it actually means you don't trust the Z motor to perform as told (not trusting your open-loop motors?). Conversely adding the safety check in `moveToSafeZ()` wouldn't help preventing a collision if you jog in X-Y while the nozzle is down. Or maybe this is not a real "safety check" after all?

2. Why a new actuator to read back Z? Can't we trust the recorded Z in OpenPNP and why? And if not, why can't we use the existing GCode [POSITION_REPORT_REGEX](https://github.com/openpnp/openpnp/wiki/GcodeDriver#position_report_regex)?

3. If I understand correctly, "Z Probe" is not the same as "Z Contact". OpenPNP Z Probe wants to measure a distance using either a laser ranger, or moving a probe down etc. We should talk about your intentions first, especially as I'm also proposing "Contact Probing Nozzle" that we determined is not to be confused with Z probing in general (see the #859 discussion). 

_Mark

src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
```diff
@@ -155,8 +157,6 @@ public void feed(Nozzle nozzle) throws Exception {
 	        }
 		}
 
-        head.moveToSafeZ();
```
markmaker: Why remove this? I believe you can't blindly assume safe Z when `feed()` is called.

c-riegel:
@markmaker Thanks for your feedback.

1. I think you are correct: the intent is to not trust the open-loop stepper motors.  There is a photo-interrupter that gets blocked when the cam motor extends one of the nozzles in z.  My intent was to confirm the success of the moveToSafeZ by checking the status of the photo-interrupter.  If the motor skips steps for whatever reason then asking the motor control for the location will not be correct.  My intent was not to prevent all z-extended movement in x & y, but to confirm the success of the moveToSafeZ with the hardware.  My intent was not to prevent jogging moves where a human is trying to do something, but to do a sanity check as part of the running of a job where a human is not in the loop.

2. The actuator does not read back Z. It only confirms that Z is at the home height.  Any movement away from that position would trigger the photo-interrupter.  You can trust the z location in OpenPnP, but why not verify it?  You can't trust the open-loop state if the stepper's steps are skipped for whatever reason.

3. I don't think I understand what you're trying to say here.  I'm very ignorant with these other issues.  I think the term "z probe" is highly ambiguous, which is why I thought it was meant to be something for checking for safe-z when I forked.  I would suggest that things be named as unambiguously as possible.  I don't understand your intent with "z probe".  If an actuator is for measuring distance in z, then I think it should be called something like "z range" or "z rangefinder".  If it's to report contact on a part or a board, "z contact" would be a better name.  They are all separate concerns in my mind.  My intent was to put things back the way I found them so as to not get in the way of PR#859.  Maybe my comments here don't address at all what you were trying to relay.  If I seem to have taken a tangent it's because I didn't understand what you were trying to say and not that I don't value your comments.

For what it's worth, I didn't design the machine or the head's electronics.  I did however break a 502 tip because of the wrong cam settings on the z-axis, which is what drove me to add the sensor check.

src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
c-riegel: I will revisit after I get done with another PR.

markmaker:
Thanks @c-riegel for the clarifications.

1. I see I misunderstood the function. Now knowing of the photo-interrupter it makes a lot of sense. We all need to remind ourselves to describe features more profoundly in the PRs (and in comments in the code) so things become very clear and will stay that way in years to come :-).  I say "we" because I noticed my own #859 is also much less clear that I thought it was, when I wrote it up. 

2. Again, makes a lot of sense now. I wonder if it would make sense to set the machine in the unhomed state on failure. 

3. When Jason added the Z probe facility it was clearly intended as a measuring device. The intent is to automatically capture the height of parts/feeders/pcb etc. Also note that the Z probing actuator is on the head, **not the nozzle**, so it somewhat makes it clear that this cannot be a "Z contact" probe. I guess you then added your own zSafeActuator on the nozzle because of that. 

Also you should never change the name of `@Element` properties as these are saved to the `machine.xml` and changing these element's names will break all the users' configuration. 
 
> For what it's worth, I didn't design the machine or the head's electronics. I did however break a 502 tip because of the wrong cam settings on the z-axis, which is what drove me to add the sensor check.

Just for clarification: is the photo-interrupter a Charmhigh feature or did you add it? And do you know why the Z axis might get stuck? It does not seem to be a very _loaded_ axis. 

Personally I would have found a "pre X-Y move" safe Z check much more universally useful :-) 

_Mark

c-riegel:
Good feedback @markmaker.  I see that it was not obvious that it was a photosensor, but just like the z-probe it's hard to know what the person your communicating with doesn't know. 

It's a Charmhigh feature.  That's why I made the comment that I didn't design it.  I can't speak to their design thoughts.  If I had to guess I would assume it's the analog to the homing switch on x and y.  It could potentially be used to confirm that the z axis cam is homed.  For example, if you started the machine after rotating the cam motor one of the z axis nozzles would be down initially.

src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
c-riegel: Reverted

vonnieda:
@c-riegel and @markmaker: How would both of you feel about these changes, and maybe some of the other PRs being pushed into extension classes? Since a lot of this functionality is relatively specific to the CharmHigh machines, I think it makes sense. So, instead of modifying ReferenceNozzle, AbstractHead, and ReferenceDragFeeder, we just make ChmtNozzle, ChmtHead, and ChmtDragFeeder. We get the functionality from the PR, and in fact can make it even more specific to the CHMT machines, but we don't affect other non-CHMT users.

@c-riegel Explanation, in case this concept is new to you: machine.xml loads the classes that are specified in it. So a nozzle can be a ReferenceNozzle, a MyNozzle, a ChmtNozzle, etc. just by adding a class and putting the name in the machine.xml. This is used most extensively in the feeders and cameras, where each type of feeder or camera is a class. Adding machine specific functionality is actually the reason I wrote it that way. The Reference* classes were intended to be a reference as to how to write a class for a specific machine.

The reason I'm suggesting this is mostly related to ReferenceDragFeeder. I'm frustrated by how complex that class has become and I am loathe to add even more to it. It has become a big mess of half baked features and it's difficult to use for everyone. Making a clean break and adding this functionality to a new class that can grow to meet the needs of CHMT users would be beneficial, I think.

No need to jump right into code changes - let's discuss. Thoughts?

c-riegel:
@vonnieda I don't have enough information to answer you.  I don't even have enough knowledge to formulate an opinion at this point, so right now I'm totally neutral and indifferent.  I would have to ask:

What is the criteria you would use to decide a change goes in to the Reference class vs being in a subclass?<br><br>I think this is a very important question that will direct the future of the project, so some points to think about that pop to mind: <br>
- You could say that the Reference class is just that: a reference and **all** real machines "should" use inherited classes and no real, physical machine "should" use the Reference class.  Its intent would be a starting point that should be subclassed away from immediately; a teaching aid to assist with the fundamental architecture of the program; a naive implementation.
- But if you agree to the above then what about functionality that is used by used by multiple  machines? So as to no violate [don't repeat yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) would you put _anything_ that is used multiple places in the Reference class?
- However, the changes in this PR were intended to be safety related, and for those types of changes one might argue that even though Charmhigh machines do have the hardware natively to support these features you want ideas, concepts, and features that support **safety** to be a core part of OpenPnP.  If a machine does not have these actuators then at least they can not get away from seeing them and thinking, "hmm, a photo-interrupter here would be a good idea", or "you know what, checking the status of the drag pin **is** a good thing to check before trying to move on".  The price the user pays for this is an empty text field on their UI for a machine that doesn't have the actuators.

Other thoughts:
- pro: going this way means I would have to worry less about having to make stubs for methods in, for example, the neoden4 and openbuilds classes.  With this statement I'm thinking more about #895.  But then, I don't think machine.xml would let me load a different FeedersPanel.
- con: but if you like the functionality of #895 then would you be the one taking on the effort to implement it upstream in Reference?  I cannot test on machines that I don't have.

Sorry to mention #895 here, I hope it's not confusing my meaning, but I'm in need of an example where neoden4 and openbuilds get touched just because I'm playing around with features for CHMT, and I think the answer to your question really has to consider a bigger idea than just the features in this specific PR.  #895 is not really a good example, so hopefully you can imagine a better hypothetical PR.

Maybe based on your feedback I could formulate an opinion, but at this point I'm, "meh, whatever is fine".

Can you touch a bit more on your issues with ReferenceDragFeeder?  Charmhigh machines are mostly drag feeders, and since I have #897 and potentially other work in that area I'd like any input you could offer before doing work on the drag feeder.  Let's assume that everything vision related was removed from ReferenceDragFeeder: if this were the case, then would you still have an issue with it?  Could you mention two specific things you don't like about ReferenceDragFeeder that _don't_ relate to vision?

markmaker:
Hi @vonnieda 

for my #859 this discussion already took place and was reflected when I closed #855 and resubmitted #859 ... _I hope!_ 

All the specific "Liteplacer style" features are now in a subclass "ContactProbeNozzle". 

The other changes are all general and some parts of it implement plans, you yourself proposed in the group over the time. Such as removing LinuxCNC and OpenBuildsDriver and many other things.

We discussed a general problem with subclassing that is Java's lack of multiple ineritance. If two contributers add two features in two subclasses and I want to build a machine that benefits from both features, I'm stuck. That is one reason for making the machine model more filigrane, moving logic out from drivers into the model, making the problems more numerous but also smaller and simpler to solve. 

I first tried to do it using more subtle steps and you said:

> But this solution feels like a bandaid and I'm not interested in bandaids when we are in the process of breaking the world :)

The culmination of that discussion with you agreeing to most of my proposals and stating further conditions is here:

> By moving more functionality out of the driver and into the high level code it becomes easier for someone to have a high performance ReferenceNozzle or even a ReferenceMachine that marshalls commands in a different way.
> 
> So, my counter-proposal:
> 
> * Remove ReferenceDriver.pick() and .place().
> 
> * Change ReferenceNozzle.pick() and .place() to use the vacuum actuator as you've described.
> 
> * Move pump control into ReferenceHead (probably) via actuator. The logic is basically "actuator.on = any nozzle has a part on"
> 
> * Move the full pick cycle and place cycle into ReferenceNozzle.pick() and .place(). By this I mean instead of the JobProcessor being responsible for lowering the nozzle, let ReferenceNozzle.pick() do that, and likewise with raising the nozzle. This solves a  bunch of problems for people who have complicated heads. This allows them to override these cycles in a subclass rather than having to modify the job processor. It also codifies the concept of a "pick" or "place" instead of that code being spread all over the place.
> 
> Thanks,
> Jason
https://github.com/openpnp/openpnp/pull/855#issuecomment-500890284

That's how - following your advice and feedback - I made #859, and yes I was "breaking the world :)". These changes are substantial but they also already cover @c-riegel 's #895, so you can intependently actuate the valve and/or check for part on/off.  

We got stuck on two details, that is passing the feeder to the pick() method and actuating camera LEDs using an actuator. 
https://github.com/openpnp/openpnp/pull/859#issuecomment-505575958

I complied with removing the LEDs actuator to be done in a seperate PR.

But the feeder param to pick() issue remained. It is clashing with the demand to subclass my nozzle. This was my last explanation that never got an answer:
https://github.com/openpnp/openpnp/pull/859#issuecomment-505754412

Then I went to a vacation and ever since my work load was brutal.... :-(

I hope we can resolve this comparably small remaining difference, because it would really, really bring OpenPNP a big step further and it would prevent newer PRs from building on outdated models. 

_Mark

c-riegel:
I agree with @markmaker: I am duplicating part of #859 with #895 (but #859 does not add a button control for priming or purging to FeederPanel, which I make great use of).

#859 has lots of goodies that I want as well, such as the migration of Thread.sleep dwelling to looking at the vacuum level.  I have another branch for that as well as you've seen from cancelled #892.

I did both #895 and #892 without even realizing that #859 existed, so no ill intent from my side.  Just ignorance on my part and if anything they are good features and great minds think alike, no ;)

I see the last comment on #859 is coming up on being 3 months old, and it is a major overhaul.  I did spend some time yesterday going through it a bit looking for inspiration, but it encompasses more than I could work on, for example I have no way of providing value to ContactProbeNozzle.  

Does #859 still have work to do on it?  I don't know what is left.  Would it make sense to break it up in to smaller PRs?  That seems like a tremendous effort as well, but perhaps your skills with git are at a ninja level that I only dream of.  

However, with all of that and after reading the comments of #859 I don't think my question is directly addressed.  If you believe that the question has firmly been answered then could you restate the guidance directly or quote the comment(s)?  My question was: what are the guidelines for when to subclass vs. integrating into a Reference class?  

Maybe I seem dense when you read this, but the problem I have is that @vonnieda has suggested creating ChmtNozzle, ChmtDragFeeder, etc. (so _subclassing is the guidance_) which would seem to suggest that the Reference classes should be locked down, but then @markmaker has included this comment above from @vonnieda:

> So, my counter-proposal:
> - Remove ReferenceDriver.pick() and .place().
>
> - Change ReferenceNozzle.pick() and .place() to use the vacuum actuator as you've described.

where we are editing Reference classes (so _not subclassing is the guidance_).  Why were these changes not pushed to sub-classes?  What is the thought process that should be used with future changes?  What would you have to consider before giving the guidance, "oh, that should go in a ChmtGCodeDriver subclass".  Does this sound silly?  If yes, then why?

I do worry that subclassing things such (ChmtNozzle or what I'll call Chmt*) will get out of hand.  If I were working on ChmtNozzle because I have a Charmhigh machine I would not go and study NeodenNozzle or OpenBuildsNozzle, so I think the code base could get fragmented.  I think you will end up having separate teams clustered around the machines they have and not contributing to OpenPnP's core.  But maybe that's what you want?

markmaker:
Hi @c-riegel 

personally I think there is both in your PR: Chmt specific and general stuff. It probably needs to be listed by feature and then discussed.

Just a two examples:

- The photointerrupter seems Chmt specific to me, so a subclass of the Nozzle seems in order
- The 2 x pick per drag feed seems not Chmt specific to me but feature specific. It could be subclassed as "MultiPickDragFeeder". Though frankly I don't like the way it is implemented now or before. It should not be reacting to "0402" but have a more generic "picks per drag" plus pitch vector GUI setting. Then use a feed counter and calculate the pick location and feed operation in "modulo" style rather than "hacking" offsets back and forth. To be able to feed multiple parts per drag could be useful for other machines and a could improve the average speed of a properly designed DragFeeder for parts other than 0402 and pick counts larger than 2. 

_Mark

c-riegel:
First off, I worry that I might come off as "snarky" or "sarcastic" or that you might read some "tone" in to my comments, so let me say that I am a [logician](https://www.16personalities.com/intp-personality), and I don't have any feelings associated with the decisions made here.  Also, **I will operate as if you don't have feelings associated with the decisions made here.**  My goal is the best technical solution, and I am bad about trying to say things in a nice way to protect the feelings of people.  If you read any "tone" in my comments please know that there was none intended.  I honestly do not feel emotionally invested and only want the best technical solution.  If you can read my comments and pretend that a child is making the comment with an inflection of curiosity instead imagining the inflection of a know-it-all with an inflection of superiority you will be much closer to my intended "tone".

@markmaker Yes, I agree the 0402 capabilities of ReferenceDragFeeder are a hack.  It seems to me that ReferenceDragFeeder needs to be aware of the part spacing on the tape somehow.  However, this would be a parameter of the tape and doesn't seem to fit under Parts or Packages, unless you want to hack it in to one or the other.  Do you have thoughts on this?  It's in no way related to this PR, but I am working in ReferenceDragFeeder for other things and would appreciate any feedback.

I still don't understand your design strategy at all.  Help me to understand your reasoning.  Please let me know if I misrepresent your position.

- Why do you suggest making MultiPickDragFeeder a subclass of ReferenceDragFeeder rather than making an architecture where a drag feeder that only does 1 drag op (as ReferenceDragFeeder does) is a special case of a general feeder class that can support _n_ drag ops?  Why would you want to keep ReferenceDragFeeder if it is only a special case that only handles single drags?  I think this goes back to my original question: when to subclass and when to modify Reference classes?  In this scenario I don't see why you would want to keep ReferenceDragFeeder if you can do MultiPickDragFeeder.  It seems you should only implement MultiPickDragFeeder and then name it ReferenceDragFeeder, so I'm curious if you disagree.

- Why do you see the photointerrupter as being chmt specific?  I agree that chmt does implement it, but is that the only criteria you're using to make this subclass suggestion?  Or do you see a larger idea that could be generalized as a guideline.  For example, maybe your generalized position is that you don't like having text fields in the wizards that some users, lacking the necessary actuators, will have to leave blank and your general position is that all fields that will not always be used by OpenPnP should be moved to subclasses.  If you agree with this, then should we eliminate the field that you enter the length of the PCB in, since OpenPnP does not use this for anything?  If you wanted a field for pcb length for some reason would you want that in a subclass?

I'm very interested in the answer to the _why?_  I'm trying to find the edges of the criteria to best decide how to develop new features.  The question about if and where to put the field for pcb length is rhetorical, but at some point you cross a line as to "well that's obviously best in a subclass" or "that's obviously best in the Reference class".  I want to know exactly where that line is, or if you can fully specify where the line is then as much information about what's on this side of the line and what's on that side of the line so I can try to make the same technical decisions that you would.

Should we try to get to guidelines first?  I mean real generalized guidelines without bias or consideration to any specific PR, and then try to use those guidelines to make the decision on specific PR(s) or see how to modify the guidelines based on the attempted application of the guidelines to specific PR(s).  And feel free to dictate what those guidelines should be, I won't be hurt if you just mandate what the guidelines are.  And I'm ok if the answer is _no guidelines_, but if the answer is "we can't do guidelines because each case is too specific," then can we do something to try to get around the:

1. Developer implements change
2. Group disagrees about subclass vs. edit to Reference decision that developer made to implement change
3. Developer gets to redo changes

cycle?  Do you have thoughts on this?  Tribunal review prior to feature development?  

And I don't mean a large set of developer guidelines or anything like that; I mean an answer, as concrete as possible, to the question of when to subclass and when to modify a Reference class.  I still don't feel any closer to a sense of when something should be in a subclass vs. being implemented as a modification to a Reference class.

vonnieda:
I'll weigh in fully this weekend, but I thought I'd offer a couple quick
thoughts that might help provide some clarification:

1. In general, I try not to merge things that break existing functionality.
That means that I have to think very hard about changes to Reference* and
org.openpnp.spi* classes. These classes not only affect existing machines
by the code they contain, the interfaces affect existing extensions of
OpenPnP. There are a lot of people that extend OpenPnP with machine
specific classes and scripting to do what they want. If I change an
interface, that likely breaks all of that code. So, interface changes need
to be considered carefully and my default is to reject them unless it is a
change that is overwhelmingly positive.

2. The reason I mentioned moving things into subclasses is one of
expediency. Changes to Reference classes and SPI interfaces take a lot of
time to review, test, document, etc. for the reasons described above, but
adding a class that won't affect anyone except you is trivial. I'll just
click merge. I don't have to worry about it breaking other people's
machines or code and if you want to keep changing it, adding to it, etc.
it's still easy to merge.

So, one simple guideline could be: If you don't have to change core code -
don't. We can *always* decide down the road that a non-core feature is so
useful that everyone will want it and pull it into the core code.

Again, I'll comment more this weekend, or will work on merging things.
Since you both seem to be in agreement that Mark's PR covers some of this
better, then perhaps focusing on getting that merged is the better use of
time. The only issue there is that Mark's PR is so large that every time I
pull it up to work on it I'm overwhelmed with the number of changes.

Thanks,
Jason

markmaker:
> The only issue there is that Mark's PR is so large that every time I pull it up to work on it I'm overwhelmed with the number of changes.

Yes, I know. If I had the time I'd break it up. But it doesn't look good on the "spare time front"... sorry. 

_Mark

markmaker:
>  If you read any "tone" in my comments please know that there was none intended. 

Yes, absolutely the same here!

> Why do you suggest making MultiPickDragFeeder a subclass of ReferenceDragFeeder rather than making an architecture where a drag feeder that only does 1 drag op (as ReferenceDragFeeder does) is a special case of a general feeder class that can support n drag ops? Why would you want to keep ReferenceDragFeeder if it is only a special case that only handles single drags? 

Good questions. But the answer is also good :-) 

If you have only one pick per drag, you don't need to bother with part pitch and other settings, you can leave that to the machanics of the feeder. The setup is quite a bit easier. And the feeder has no "saved state", It is always in the same state. No worries if the uncovered but unused second part was lost or a new reel loaded in the meantime etc.

All that added complexity is neatly tucked away with a sub-class. 

_Mark

c-riegel:
@markmaker I think we are probably in complete agreement.  I agree with your drag feeder comments though I believe 0402 parts, where you have multiple parts feed per drag, throw a monkey-wrench in to the design, but I believe you would agree with this and that we are on very much the same page.

@vonnieda Your feedback is extremely helpful, and I understand your position.  Based on your comments I think anyone could make well informed decisions about how to contribute new code.  However, based on this philosophy let's assume that someone wanted to follow my foot prints and use OpenPnP with their CHMT 48VB (or someone with a similar machine to @markmaker wanted to follow his foot prints): how should we make this easy for such a person.  It sounds like you need a combobox in the installation procedure where they can select a machine specific default configuration.  Maybe you just need a wiki page for suggested setups, but as a recent newbie I want to bring this topic up as I don't want people that follow us to have a learning curve that has to be any steeper than it has to be.

@vonnieda said:

> Since you both seem to be in agreement that Mark's PR covers some of this
better, then perhaps focusing on getting that merged is the better use of
time. The only issue there is that Mark's PR is so large that every time I
pull it up to work on it I'm overwhelmed with the number of changes.

However, I never said that Mark's PR covers things better.  I'm not saying don't spend time doing the merge; feel free to do it if you want to.  I'm not saying Mark's PR is _not_ better, perhaps it is.  Just don't do it because you have a sense of agreement that I never agreed to.  Based on your feedback here you said you don't like changes to Reference* or SPI interfaces, both of which are in @markmaker PR.  If you believe that #859 

> is so useful that everyone will want it and pull it into the core code

then yes, absolutely try to merge the PR.  Just do it because _you_ believe this to be true.  I can not comment on this.  I can not speak to the value, bad or good, of ContactNozzle or other features in #859.  If you want my thoughts on any specific feature then I'm happy to review/comment, but I'm rather long winded because I want my view to be concretely understood, so I'm also fine with staying out of it.

Just to state my position concretely: 
- I believe that the actuator/event driven approach is correct and blind timers are incorrect.
- I believe the process flow of: turn on vacuum, move to pick location, go down in z is correct and the process flow of: move to pick location, go down in z, turn on vacuum is incorrect.<br><br>I implemented a hack (though I think it's quite clever) where I could use a Job.Placement.Starting script to call nozzle.actuate(true) which triggers a GCodeDriver call to the nozzle's ACTUATE_BOOLEAN_COMMAND to actuate solenoids and turn on the vacuum to the nozzle before the feed operation.  I have not accessed Mark's approach to this or how #859 would allow this to happen.  Based on previous posts I believe that @vonnieda would disagree with my belief here (or at least that my approach is not the way it is typically done), which is fine.  However, this got me to where I wanted to be, and so if the guidance is that this approach would be best done in a Chmt* subclass arrangement then I'm cool with that.  
- Also, I want to throw in a plug for a FeederPanel set of buttons to do nozzle.actuate and nozzle.purge.  Wow, how have you lived without these?

markmaker:
Hi @c-riegel 

I feel unhappy about this situation. It is not about whose PR is better, its more about which comes first. 

In an ideal world my PR would long ago have been integrated (with whatever adjustments left to be done). But we're all doing this in our spare time, with Jason having put an incredible work sum in here. 

You could build your improvements on top of mine. I _do see improvements_ in your PR ... damn, you just built it on top of yesterdays' blueprints. 

It is just that some of these new blueprints were proposed by Jason a long time ago, we had some in-depth discussions and I invested quite a lot changing things to follow Jason's guidance. This is still the OpenPNP 2.0 initiative where we are "breaking the world :)". And yes, I still feel strongly that the proposed changes are _good_ and they in no way prevent you from improving on top of it.

As an aside: I agree that turning on and off the pump with each pick might not be the best way. Personally I use a vacuum reservoir bottle and controller driven hysteresis to keep the underpressure up all the time. I don't use the pump actuator, just the valve. However in my PR I modelled the pump on-off scheme according to Jason's guidance which is just what people in the field use and are accustomed to. 

But the fact remains that the pump actuator should first be properly modelled and the logic (and GCode) removed from the driver. That is what my PR does (among many other similar things). And after all the discussions we had, I must assume it does it in accordance with Jason's long term plans (except for the necessary final touches, I hope). 

If you think about it: you could easily build your changes on top of mine and change the way the pump actuator is governed. And this is much cleaner than on top of the old blueprint with the pump logic half hidden in and dominated by the driver. With this preperly modelled you could even make the pump governing scheme a dropdown choice, so users can have it _their_ way.

This might sound a bit self-serving but there is nothing to prevent you from basing your PR on top of my branch. As long as my PR is not merged, Githup will show all changes cumulated, which is a hassle, but you (or Jason) could do diffs across branches for assessment in a sandbox. Once my PR would be merged, the diffs magically disapear on Github. 

_Mark

c-riegel:
Hi @markmaker 

I want to make sure that you don't feel I am in any way mocking your PR.  I'm not.  @vonnieda assumed I was saying it was better.  I clarified that I had no vote for or against your PR due to ignorance on my part of the full scope and details of your PR.  And please let me further clarify that **I have no vote for or against my PR either**.  I only wanted guidance on how to continue to contribute moving forward.  I just want to know how to help...  I hope you see that.

I think we are moving in the same direction.  You see from my comment what my goals are.  I'm perfectly happy building on your work.  

> It is not about whose PR is better, its more about which comes first.

In general I would never agree with this statement.  I don't know why you would logically take an "earlier" option over a "better" option.  And please don't assume that I'm specifically talking about my PR and trying to insinuate that it's somehow "better".  Please remember that I said:

> I'm not saying Mark's PR is not better, perhaps it is.

But please, let's agree that that a "better" option should **always** win over an "early" option.

> You could build your improvements on top of mine.

I'm perfectly happy to do so.  Would you wish for me to try to break up your PR and submit the parts individually?  I'm happy to go that route.  However, as I've said before, I don't feel confidant doing work related to ContactNozzle because I don't have the hardware for testing it.  There might be other portions of your PR that I could not contribute to.  I don't know the full scope of your PR.  

> I do see improvements in your PR ... damn, you just built it on top of yesterdays' blueprints.

I'm not sure what you mean here.  I'm under the understanding that I have to modify all of my PRs and create a bunch of new Chmt* subclasses.  All of my PRs have been invalidated.  Am I wrong?  And I'm not going to be mad either way (meaning if I have to redo things with a bunch of subclasses that's ok with me).  It sounds like the subclasses will make Jason's job much easier and that's great.  I just don't want you thinking that I'm somehow _winning_ or even _trying to win_.  I just want to know how to help...

vonnieda:
@c-riegel @markmaker Please hold off on any further changes for now. I am working today to fully review all the existing PRs. I'm going to look at these from an individual feature perspective, decide what features can be merged already and then will merge as much as I can. Once that's complete we can work further to get other parts merged as needed.

vonnieda:
Hi @c-riegel,

I would like to apologize for letting these PRs rot as long as I have. I don't know if it helps or not, but I would like to explain: When there are changes to so many core classes it is very difficult to test the results and to consider how it will affect all the existing users. I generally think it's better to fail fast and let bugs expose themselves, but in OpenPnP it can mean a machine causing physical damage to itself. So, this comes down to me not having the bandwidth to really fully test changes of this magnitude. I don't know a good solution.

So, to make matters worse, I've let these PRs go so long that they are full of conflicts and some of the functionality has been superseded. If you would like to try to get these features in, I would ask that you add them to new Charmhigh specific classes (extends ReferenceNozzle, extends ReferenceFeeder, etc.) which will make them very easy to merge as there is no chance of breaking existing machines.

I am going to be using this guideline going forward for all PRs, for what it's worth. I will be asking everyone to default to custom classes instead of changes to core classes.

So, I'm sorry for letting your hard work go wasted. If you are still interested in these features, I hope you'll consider moving them to subclasses.

Thanks,
Jason

PR #894, opened 2019-09-17
Simple text changes to address things I found ambiguous when starting

c-riegel:
It was ambiguous if you were wanting the distance from one point on a pcb to the same point on an adjacent pcb on a panel or the gap between the pcbs.


vonnieda:
@c-riegel Thank you for taking the time to split this up. Looks good!

c-riegel:
@vonnieda I'm about to do another PR, but now the change from this PR shows up in the diff between openpnp:develop and I'm afraid that my next PR will undo the changes from this PR.  Do I need to do something in the branch for my new PR to grab this change from openpnp:develop?  Do you know how to do that within eclipse?  Or do I not need to worry?

PR #898, opened 2019-09-18
Script event Job.Placement.Starting

c-riegel:
# Description
Adds a new script event called Job.Placement.Starting that occurs before the feed operation during a job.

# Justification
I want to add this so I can call nozzle.actuate(true) in a script so I can have the nozzle purged of air by the time I get to doing the actual pick operation.  I don't think it should be a requirement that the vacuum can only come on after the tip has made contact with the part.

This is just why I want it.  Other users may not need this event in which case it will cause them no harm.

# Instructions for Use
It works the same as the other scripting events.  

# Implementation Details
By running it on my machine.
No, minimal changes were made to ease merge efforts.
Tests pass.


vonnieda:
Looks good, thanks @c-riegel!

PR #902, opened 2019-10-03
Added pipeline stage to write actuators

jarekkt:
# Description
I have added Cv pipeline stage which allows writing to actuators. 

# Justification
This allows controlling actuators directly from Cv pipeline, without using external scripts ( there is a Cv pipeline stage which can execute such script). It helps e.g. to control light intensity directly (if you have such actuator - it is quite common case) 
 
# Instructions for Use
Setup actuator name, value and type ( double of boolean) - in similar way as all other parameters for any stage.

# Implementation Details
This code was copied mainly from part controlling feeder actuators, it was just formatted to follow Cv stage classes implementation
Tested for positive (proper actuator) and negative cases ( wrong actuator name).
It supports all actuators ( machine and head based)

### DIFF
--- a/src/main/java/org/openpnp/vision/pipeline/stages/ActuatorWrite.java
+++ b/src/main/java/org/openpnp/vision/pipeline/stages/ActuatorWrite.java
@@ -1,0 +1,91 @@
+package org.openpnp.vision.pipeline.stages;
+
+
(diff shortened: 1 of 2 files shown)


vonnieda:
Looks good, thank you @jarekkt!

PR #905, opened 2019-10-18
Bugfix/bottom fiducial

### LINKED ISSUE #819: Bottom Side Rotation Error w/ 3 Point Fid Check
The new fiducial check code seems to be calculating a bad angle for bottom side boards.

A fix has been posted by @aneox in
https://github.com/openpnp/openpnp/commit/77f539aaf194d743be2a1a34f58c2dd57b6a7978#commitcomment-32973251 but I think this just hides the problem. 

It appears the fiducial check is adding 90 degrees to the result for some reason. @aneox's fix removes the 90 degrees but it would be better to find out where the 90 degrees is coming from in the first place.

This bug can be reproduced easily by loading the pnp-test job and running a fiducial check on the bottom board with angle 36.609. After the check completes the crosshair shows the correct XY but the rotation is +90.

bilsef:
# Description
Fixes #819.

# Justification
Needs fixing
# Instructions for Use
None required
# Implementation Details
Invert X for bottom side fiducials and placements.

Tested by doing 2 point and 3 point fiducial checks on all of the boards in the sample job.  Verified that the camera moved to the board origin and is in the correct orientation.  Moved to each component and verified that the camera is centered over the component and is in the correct orientation.  Did the same testing on actual machine using 6 board panel.
Yes
N/A
All tests passed

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
@@ -41,6 +41,7 @@ import org.openpnp.util.QuickHull;
 import org.openpnp.util.Utils2D;
 import org.openpnp.util.VisionUtils;
 import org.openpnp.vision.pipeline.CvPipeline;
+import org.pmw.tinylog.Logger;
 import org.simpleframework.xml.Attribute;
 import org.simpleframework.xml.Element;
 import org.simpleframework.xml.ElementMap;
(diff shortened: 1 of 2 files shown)


vonnieda:
I don't have time at the moment to test this myself, but it passes all our automated tests and it sounds like it solves the issues, so I'm going to go ahead and merge it. If anyone finds any issues, please let us know.

Thanks for doing this @bilsef!

bilsef:
There is a issue with bottom side panel fiducial check which I am looking into.  I think it is related to this bug.

PR #907, opened 2019-10-19
Fix bottom side panel fiducial check

bilsef:
# Description
Fixes panel bottom fiducial detection.

# Justification
Bugfix #905 fixed the board bottom fiducial detection, but the panel fiducials were still incorrect.  This is because the panel must always be treated as top side for the purpose of fiducial detection.  This will update the origin of each board in the panel to the lower left corner of the boards.

# Instructions for Use
Set up the panel the same as for the top side.  If the fiducial locations are different, set them also.  Fiducials are referenced to the lower left corner of the lower left board (board 0 origin).

# Implementation Details
1. How did you test the change? I created a 2 x 3 panelized version of the pnp-test board with top and bottom panels.  Panel and board fiducials were properly located for top and bottom.  Components appeared in the correct positions.  I also tested on a live system with a panelized board.
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.  All tests passed

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceFiducialLocator.java
@@ -74,10 +74,14 @@ public class ReferenceFiducialLocator implements FiducialLocator {
     public Location locateBoard(BoardLocation boardLocation, boolean checkPanel) throws Exception {
         List<Placement> fiducials;
 
+        Side boardSide = boardLocation.getSide();  // save for later
+       
         if (checkPanel) {
             Panel panel = MainFrame.get().getJobTab().getJob().getPanels()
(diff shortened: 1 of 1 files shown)


vonnieda:
Hey @bilsef, this looks good and I will go ahead and merge it, but I'd like to ask that you consider a condition where the fiducial check throws an error. This will leave the board side set incorrectly. Ideally I think it would be better if we didn't have to modify the board location side, but if that's the best way to do it, we probably need a try/catch/finally to ensure we don't leave the board location side set wrong.

bilsef:
Hi Jason,
I agree, it would be best if we didn't need to modify the side.  I think it could be done by passing the side to the sub-functions, but it would be a bigger change.  I will look into handling the fiducial check error.
--Bill

Issue #909, opened 2019-10-21
Row index out of range error


gxurma:
# Row index out of range error
## Describe the Issue
Stepping through the placements list of a board, when stepping from the last placement, it ends in "Error: Row index out of range" 

## Steps to Reproduce
clean install
load the pnp-test job.
click on any board
click on any placement
step through the placements with the icon "Position the camera at the next placements location"
step over the end of the list. You'll get the error

## Expected Result
It should just wrap to the first placement. Like it was in the v1 openpnp.

## Actual Result
throws an error
"Error: Row index out of range"

K2AAE:
Just got Row index out of range error on 2019-12-5 (then long number I can't copy) same conditions gxurma

PR #911, opened 2019-10-23
Bugfix/select next placement

bilsef:
# Fix for bug #909
The method selectNextTableRow in gui/support/Helpers.java calls getRowCount, but this returns the total number of placements on the board.  We need the count of placements on the currently selected side of the board.  I created a overload method for selectNextTableRow that counts the placements on the selected side.

# Justification
Bug fix.

# Instructions for Use
load a job
click on any board
click on any placement
step through the placements with the icon "Position the camera at the next placements location"
step over the end of the list
It will wrap back to the first placement.

# Implementation Details
1. How did you test the change? Tested with the pnp-test job.
Tests run: 20, Failures: 0, Errors: 0, Skipped: 0

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
@@ -498,7 +498,7 @@ public class JobPlacementsPanel extends JPanel {
                 // used after the initial click. Otherwise, button focus is lost
                 // when table is updated
                	Component comp = MainFrame.get().getFocusOwner();
-               	Helpers.selectNextTableRow(table);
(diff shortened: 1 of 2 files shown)


vonnieda:
Hi @bilsef - we should be able to fix this just by querying the TableModel, which will be generic. I can look into it in a few days, or you can dig in if you like. The reason to do it that way is that it also needs to handle when other filters are applied, such as search.

bilsef:
Hi @vonnieda, Thank you for the feedback.

vonnieda:
Fixed in https://github.com/openpnp/openpnp/commit/0279d657105b620c689b326c6cd9761e441e76e5

Issue #913, opened 2019-11-11
Nozzle tip Toolchanger blank third position line and not shown in machine.xml


gxurma:
# Problem Report
## Describe the Issue
When adding new nozzle tip, on the Toolchanger page it show a blank line for Third location.
Also saving entered data does not work. button never greys out when hitting Apply

## Steps to Reproduce
How can a developer reproduce the issue you are experiencing? Use a step by step format like this:
1. Start fresh OpenPnP
2. add new Nozzle Tip
3. go to page Toolchanger
4. try to enter and apply plausible data. It will not work.
5. change to the other Nozzle tip, and back to the new tip.
6. there will be no change in the list, blank line remains.

## Expected Result
entered data shall get accepted and written to the machine.xml

## Actual Result
entered data does not  get accepted and written to the machine.xml

## fault mitigation:
manually edit machine.xml with plausible data

gxurma:
![kép](https://user-images.githubusercontent.com/2029708/68574504-52c66680-046a-11ea-8e94-66f312467b37.png)

gxurma:
Tried to solve this by modifying the initialization constructor in the 
src\main\java\org\openpnp\machine\reference\ReferenceNozzleTip.java
from
private Location changerMidLocation2;
to
private Location changerMidLocation2;  new Location(LengthUnit.Millimeters);

But that breaks the tests. 
I dont yet understand the testing system.

vonnieda:
Fixed in #915. Thanks @gxurma!

PR #915, opened 2019-11-11
getting rid of toolchanger config backward compatibility due to issue #913

gxurma:
Solves Issue #913

# Description
Removing backward compatibilty stuff from the 
src\main\java\org\openpnp\machine\reference\ReferenceNozzleTip.java
and added missing middle 2 positions to 
src\test\resources\config\BasicJobTest\machine.xml
to fulfill the automatic tests 

# Justification
it solves issue #913 

# Implementation Details
used it and added new nozzle tip and the problem of missing line and not able to write the machine.xml is gone. now it starts with 0.000.

AFAIK just copy and pasted the new constructor...
done. it says Build Success.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTip.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTip.java
@@ -24,7 +24,6 @@ import org.openpnp.spi.base.AbstractNozzleTip;
 import org.openpnp.util.UiUtils;
 import org.simpleframework.xml.Attribute;
 import org.simpleframework.xml.Element;
-import org.simpleframework.xml.core.Commit;
 
 public class ReferenceNozzleTip extends AbstractNozzleTip {
     @Attribute(required = false)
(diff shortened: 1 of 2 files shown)


vonnieda:
I'll have to look closer at this, but at first glance, I think this is an error. The change seems to do the thing that the comment says not to do - which is insert a move to 0,0,0,0.

gyurmainnio:
Hi Jason! Did you have a chance to check it?
Questions is, what is the error? Constructor or backwards compatibility stuff, that I removed? Do we really need the backwards compatibility only for the tests?

gxurma:
Hi Jason! Did you have a chance to check it?
Questions is, what is the error? Constructor or backwards compatibility stuff, that I removed? Do we really need the backwards compatibility only for the tests?

vonnieda:
Hi @gxurma: The issue is that removing the backwards compatibility means that someone who opens an old configuration will suddenly have a move to 0,0,0,0 in the middle of their nozzle change and it will cause a nozzle crash. The backwards compatibility is important, and causes no harm, and it's not related to the issue you are seeing.

The issue you are seeing is because of where / how the backwards compatibility happens. When a config is loaded the @Commit method is called, which checks if the position is null and sets it. In the case of creating a new nozzle, this is not called until the config is re-loaded, so it remains null.

This is a little tricky because with the XML framework it's hard to tell the difference between a freshly created object and one that is in the process of being loaded. Let me think about this and see if I can come up with a fix this evening.

vonnieda:
I reconsidered this and will go ahead and merge it. The backwards compatibility was added long ago and by now I expect everyone that is going to update will have.

PR #919, opened 2019-11-18
Remove duplicated vision offset correction

tjanuszewski:
# Description
getPickLocation() is using visionLocation value to calculate pick location. visionLocation already contains value that was previously corrected in updateVisionOffsets().

l value (pick location) is calculated from lineLocations, which may be calculated using visionLocation in getIdealLineLocations().

# Justification
Vision offset was added twice and pick location error was increasing with feed count so much, that it was almost impossible to use strips longer than ~20-30 parts.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java
@@ -189,11 +189,6 @@ public class ReferenceStripFeeder extends ReferenceFeeder {
         // so that the part is picked at the right angle
         l = l.derive(null, null, null, angle + getLocation().getRotation());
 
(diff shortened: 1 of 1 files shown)


vonnieda:
Nice find @tjanuszewski! This may explain the problems many users have had with the strip feeders. I'll do some testing on this soon and get it merged.

vonnieda:
I worked through the code on this, and did quite a bit of testing, and it looks to solve this issue. Nice work @tjanuszewski. Additionally, with the removal of that block of code the visionOffsets is no longer used at all, so I will remove it completely after merging.

Again, nice find @tjanuszewski - this is going to help a lot of people!

CircuitSetup:
Yes, thanks @tjanuszewski . This was driving me craving thinking my pipelines were off.

PR #920, opened 2019-11-20
Removing legacy drivers

markmaker:
# Description
The legacy drivers LinuxCNC and OpenBuildsDriver are removed to ease further development. All their functionality is covered by GCodeDriver. 

The removal is actually basis for/of #859 and breaking this out (and merging it) will hopefully simplify code review there.

# Justification
As discussed repeatedly, the legacy drivers LinuxCNC and OpenBuildsDriver should be removed. There are sometimes users mistakenly trying to still use them. The removal has been discussed and no objection voiced AFAIK. 

Just two example references:
https://groups.google.com/d/msg/openpnp/GiIKbgy4Soc/Z5a-epVSCwAJ
https://groups.google.com/d/msg/openpnp/1PT4ODY5ClU/LW1WI4mFAAAJ

# Instructions for Use
The drivers are no longer available for Machine Setup. 
grafik

Users with a `machine.xml` still based on these drivers must manually modify the `<driver/>` section and upgrade to the GCodeDriver. 

# Implementation Details
1. The driver source code files were removed 
2. Registration of the classes was removed from the machine configuration. 
3. `mvn test` was run succesfully.


vonnieda:
Nice work, thanks @markmaker!

PR #922, opened 2019-11-26
Create binary MaskHsv for true "green screen" effect

markmaker:
# Description
This PR adds a "binaryMask" option to MaskHsv to return a black and white stencil of the "green screened" image.

# Justification
Juki nozzle tips are one application example of using a "green screen" effect to recognize objects in from of a green backdrop, in this case the broad green collar of the nozzle tip. As long as the object itself is not green, its presence can be detected simply by the pixel not being green. 

NOTE: The _real_ reason for this PR is my upcoming new feeder PR that uses the green screen effect extensively. When this description says "green" it effectively means any vivid color, the hue of which can be set in the MaskHsv stage. 

MaskHsv currently blackens the color areas in a given full color image. Consequently, black that was present in the source image cannot be distingiushed from the black created by the mask. Black objects (including shadows) cannot be recognized. 

This PR changes this by simply using the HSV mask directly, switching everything not green to white, including formerly black areas. The full outline of the object is recognized rather than just the shiny parts. This might help with some difficult parts where asymmetric or dull pads/pins/balls are present. 

# Instructions for Use
Switch on the "binaryMask" in the MaskHSV stage: 

grafik

Move your MaskCircle after the MaskHSV stage, it needs to block the non-green backdrop outside the Juki nozzle tip collar. Better yet: mount a green backdrop screen behind your nozzle tip to support recognizing large parts, 

You can remove/disable your ColorConvert stages because the result is already a grayscale image. 

You can remove/disable the Threshold stage, because the result is already a binary mask.

## Example 
The problem shown here is _staged_ for illustration. Given proper adjustment, the speck could be suppressed, of course. The intention is to show how the binary mask - for some parts - might be more robust because it simply takes the whole part body into consideration. Thanks to Mike M. for the nice example image. 

grafik

## Results - Standard Pipeline

grafik

grafik

grafik

## Results - Binary Pipeline

grafik

grafik

# Implementation Details
1. The code was tested in the pipeline (and while developping the upcoming new feeder).
4. Successful `mvn test` before submitting.

### DIFF
--- a/src/main/java/org/openpnp/vision/pipeline/stages/MaskHsv.java
+++ b/src/main/java/org/openpnp/vision/pipeline/stages/MaskHsv.java
@@ -19,7 +19,7 @@ import org.pmw.tinylog.Logger;
 import org.simpleframework.xml.Attribute;
 import org.simpleframework.xml.core.Commit;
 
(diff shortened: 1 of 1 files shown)


CircuitSetup:
Awesome, thanks for adding this @markmaker

vonnieda:
Nice work @markmaker!

PR #923, opened 2019-11-27
Method to determine if a location is inside soft-limits.

markmaker:
# Description
A method on the Head can be called to determine whether a HeadMountable can go to the specified location within the soft-limits. 

# Justification
The soft-limits check is currently only implemented inside ReferenceHead.moveTo(). In order to proactively check whether a location is within soft-limits, the check is refactored into a separate method.

The method is used in my upcoming feeder to check the eligibility of fiducials on the feeder. 

# Instructions for Use
Use `head.isInsideSoftLimits(hm, location)` to check if the HeadMountable `hm` can go to `location`.

# Implementation Details
1. The refactored soft-limit checking on moveTo() was successfully tested on purpose. It was also very useful in developing, when things went wrong. The new method was also successfully used for the feeder I'm developing, to select eligible fiducials for feeders that are on the edge of the machine.  
2. I did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style).
3. The method is in `org.openpnp.spi`.
4. Successful `mvn test` before submitting.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceHead.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceHead.java
@@ -31,6 +31,7 @@ import org.openpnp.machine.reference.psh.NozzlesPropertySheetHolder;
 import org.openpnp.machine.reference.wizards.ReferenceHeadConfigurationWizard;
 import org.openpnp.model.Configuration;
 import org.openpnp.model.Location;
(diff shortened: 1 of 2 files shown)


vonnieda:
Nice work @markmaker! This has the nice side effect that it makes the soft limits more overridable now, too, for people who want more complex soft limit checking.

PR #925, opened 2019-11-27
New vision stages

markmaker:
# Description
The following three vision stages were a byproduct of the development of my upcoming feeder.
* CreateShapeTemplateImage.java
* FitEllipseContours.java
* DrawEllipses.java

It turned out I don't use them (replaced by better/simpler solutions). Rather than just throw them away I thought I post them here. 

The PR or parts of it could simply be closed but kept to be reactivated later. 

# Justification
CreateShapeTemplateImage is useful to create a specific image based on (complex) vector shapes in absolute units scale. This can then be used to search for it in the captured frame. One idea I'm keeping in mind, is using it to recognize the shape and calibrate the location of the nozzle tip changer shoe. Similar to CreateFootprintTemplateImage.

FitEllipseContours takes contours and searches for ellipses. Standard OpenCV.

DrawEllipses draws them. Standard OpenCV.

# Instructions for Use
Similar to the other stages.

# Implementation Details
1. FitEllipseContours and DrawEllipses were used for a time in the vision for the new feeder. They worked fine. CreateShapeTemplateImage never made it to be really used or tested. 
4. Done `mvn test` before submitting the Pull Request.


vonnieda:
Nice contribution Mark! The template stage should be fun to play with.

PR #926, opened 2019-11-30
Add scripting event when using the "Position the Camera" button.

bilsef:
Add scripting event after camera move to position.

# Description
Add a scripting event after moving the camera to position by pressing the Position the Camera Icon.

# Justification
Most of the time when pressing the Position the Camera Icon, I want to look at something, so I have to turn on the down light.  I thought it would be nice if the light could be turned on automatically so I added this scripting event.

# Instructions for Use
Put a Camera.AfterPosition script in the scripts/Events directory.  For example:
Camera.AfterPosition.js
> /**
>     Controls lighting for the down camera 
>  */
>  var downCamLights = machine.getActuatorByName("DownCamLights");
>  downCamLights.actuate(true);

# Implementation Details
1. How did you test the change?  Added above script and verified light was turned on when using the Position the Camera Icon.
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.  N/A
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.  Tests run: 20, Failures: 0, Errors: 0, Skipped: 0

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -35,7 +35,9 @@ import java.beans.PropertyChangeListener;
 import java.io.File;
 import java.io.FilenameFilter;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.prefs.Preferences;
 
 import javax.swing.AbstractAction;
@@ -95,6 +97,7 @@ import org.openpnp.spi.Machine;
 import org.openpnp.spi.MachineListener;
 import org.openpnp.util.MovableUtils;
 import org.openpnp.util.UiUtils;
+import org.pmw.tinylog.Logger;
 
 import com.google.common.eventbus.Subscribe;
 
@@ -1254,6 +1257,14 @@ public class JobPanel extends JPanel {
                         MainFrame.get().getCameraViews().ensureCameraVisible(camera);
                         Location location = getSelection().getLocation();
                         MovableUtils.moveToLocationAtSafeZ(camera, location);
+                        try {
+                            Map<String, Object> globals = new HashMap<>();
+                            globals.put("camera", this);
+                            Configuration.get().getScripting().on("Camera.AfterPosition", globals);
+                        }
+                        catch (Exception e) {
+                            Logger.warn(e);
+                        }
                     });
                 }
             };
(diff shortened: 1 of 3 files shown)


src/main/java/org/openpnp/gui/JobPanel.java
```diff
@@ -1254,6 +1257,14 @@ public void actionPerformed(ActionEvent arg0) {
                         MainFrame.get().getCameraViews().ensureCameraVisible(camera);
                         Location location = getSelection().getLocation();
                         MovableUtils.moveToLocationAtSafeZ(camera, location);
+                        try {
+                            Map<String, Object> globals = new HashMap<>();
+                            globals.put("camera", this);
```
vonnieda: This should probably be `globals.put("camera", camera);` and not `globals.put("camera", this);`, otherwise the scripts are going to get the instance of the action listener instead of the camera. Same with the other instances of this call in the PR.

src/main/java/org/openpnp/gui/JobPanel.java
```diff
@@ -1282,6 +1293,14 @@ public void actionPerformed(ActionEvent arg0) {
                         
                         MovableUtils.moveToLocationAtSafeZ(camera, location);
                        
+                        try {
+                            Map<String, Object> globals = new HashMap<>();
+                            globals.put("camera", this);
```
vonnieda: And here.

src/main/java/org/openpnp/gui/JobPlacementsPanel.java
```diff
@@ -480,6 +483,14 @@ public void actionPerformed(ActionEvent arg0) {
                 Camera camera = MainFrame.get().getMachineControls().getSelectedTool().getHead()
                         .getDefaultCamera();
                 MovableUtils.moveToLocationAtSafeZ(camera, location);
+                try {
+                    Map<String, Object> globals = new HashMap<>();
+                    globals.put("camera", this);
```
vonnieda: And here.

src/main/java/org/openpnp/gui/JobPlacementsPanel.java
```diff
@@ -506,6 +517,14 @@ public void actionPerformed(ActionEvent arg0) {
                         .getDefaultCamera();
                 MovableUtils.moveToLocationAtSafeZ(camera, location);
                 
+                try {
+                    Map<String, Object> globals = new HashMap<>();
+                    globals.put("camera", this);
```
vonnieda: And here.

bilsef:
All fixed.

vonnieda:
Thanks @bilsef! Would you mind updating the docs with the new event at https://github.com/openpnp/openpnp/wiki/Scripting#scripting-events

bilsef:
Sure, I will do that.

PR #929, opened 2019-12-03
New ReferenceLeverFeeder

bilsef:
# Description
ReferenceLeverFeeder supports feeders where a lever is pushed one or more times to feed the parts (such as Alex Feeder).  Made from ReferenceDragFeeder.

# Justification
While the ReferenceDragFeeder can be used for this type of feeder, it has some issues.  It doesn't support part pitch greater than 4 and the feed operation is not optimal.

# Instructions for Use
You will need a Head Mountable Actuator defined.  It doesn't have to do anything for this feeder.  You can also define a cover take-up actuator if needed.  Add a new ReferenceLeverFeeder and use the wizard to set it up.  Setting the part pitch to 2mm will pick 2 parts per feed.  If the pitch is greater than 4mm, the appropriate number of feeds will be performed for each pick.

# Implementation Details
1. How did you test the change? I tested it on my machine set up with 15 feeders of various sizes.
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.  All tests passed.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceMachine.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceMachine.java
@@ -41,6 +41,7 @@ import org.openpnp.machine.reference.driver.NullDriver;
 import org.openpnp.machine.reference.feeder.AdvancedLoosePartFeeder;
 import org.openpnp.machine.reference.feeder.ReferenceAutoFeeder;
 import org.openpnp.machine.reference.feeder.ReferenceDragFeeder;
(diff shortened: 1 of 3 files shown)


vonnieda:
Great work @bilsef! This will help a lot of people!

If you'd like to update the CHANGES.md in the same format as the other entries there with some information about this feeder, it will be shown to users when they update. Just submit another PR with the CHANGES.md and it will go out in the release.

PR #931, opened 2019-12-11
Feature/stage improvements

tonyluken:
# Description
1. Modified pipeline stage ImageCapture to have the capability to average multiple camera captures.
2. Modified pipeline stage HistogramEqualize to be able to operate on color images as well as gray scale images.
3. Added new pipeline stage HistogramEqualizeAdaptive to provide Contrast Limited Adaptive Histogram Equalization (CLAHE) capability to the pipeline.

# Justification
All of these changes add new capabilities while being backward compatible with existing pipeline definitions.  The change to ImageCapture can be used to reduce noise in the captured image (at the expense of extra time) - most useful in training pipelines.  Both HistogramEqualize and HistorgramEqualizeAdaptive can be applied to selected channels of color images to generate false color images aiding in feature extraction.  HistogramEqualizeAdaptive is also useful for uneven illumination compensation.

# Instructions for Use
1.  ImageCapture - now has a "count" parameter that controls the number of camera captures that are performed and averaged together.  This parameter defaults to one (1) so that existing pipeline definitions utilizing ImageCapture behave exactly the same as they did prior to this change (unless the user actively modifies the value).
2.  HistogramEqualize - now has a "channelsToEqualize" parameter that controls on which of the image channel(s) equalization is performed.  Equalization can be selectively applied to any one, two, or all three channels of a color image.  This parameter has no effect on gray scale images so that existing pipeline definitions utilizing HistogramEqualize behave exactly the same at they did prior to this change.
3.  HistogramEqualizeAdaptive - This new stage has four user parameters: "channelsToEqualize" functions as described above in HistogramEqualize; "clipLimit" controls the amount of enhancement to be applied - lower values will limit the amount of contrast enhancement while higher values will allow for more enhancement but at the risk of increasing noise in homogeneous regions of the image; "numberOfTileRows" and "numberOfTileCols" control the number of tiles into which the image is subdivided.

# Implementation Details
**1. How did you test the change?** Modified various pipeline definitions to utilize these changes on my machine and observed the expected behaviors.
**2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?** Yes, the coding style was followed.
**3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.**  No changes were made to these packages.
**4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.**  The Maven tests were run and all passed.


vonnieda:
Looks good, thank you @tonyluken! I'm going to go ahead and merge this now, but I'd like to ask that you double check one thing: Is there a potential for a memory leak on `workingMat` in the two histogram stages? I looked briefly and I'm not sure if there is any copying being performed that might require a `.release()`.

PR #934, opened 2019-12-20
Improvements to the Feeders panel

tonyluken:
# Description
This change fixes a couple of annoyances associated with the Feeders panel.  The displayed contents of the feeders table now automatically update whenever a feeder wizard's Apply button is pressed.  Prior to this change, the user was forced to select a different row in the table to see the update.  Also, when changing feeder selections in the feeders table, the user is now warned via a dialog box if there are any un-applied configuration changes.  Prior to this, selecting a different feeder would result in the loss of any un-applied changes without any warning to the user.

# Justification
Less user confusion and less risk of losing intended configuration changes.

# Instructions for Use
No special instructions - the usage of this should be self-evident.

# Implementation Details
**1. How did you test the change?** Tested various combinations of selecting/adding/deleting feeders with and without un-applied changes to a feeder's configuration.  All combinations tested worked as expected.
**2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?**  Yes, the coding style was followed.
**3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.**  No changes were made in either package.

### DIFF
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -103,8 +103,9 @@ public class FeedersPanel extends JPanel implements WizardContainer {
 
     private Preferences prefs = Preferences.userNodeForPackage(FeedersPanel.class);
     
+    private JTabbedPane configurationPanel;
     private int priorRowIndex = -1;
-    private Feeder priorFeeder;
+    private String priorFeederId;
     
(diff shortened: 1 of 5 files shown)


vonnieda:
Nice work @tonyluken! Works really well.

PR #937, opened 2020-01-19
Improving the translation capabilities of the the JobPanel

pfried:
#496 Description
Adding a translation key for the Undo / Redo Commands
Fixing an issue in the PartsPanel where the tooltip was not shown
Removed unused strings

# Implementation Details

Well the issue is that strings and keys get quite long, I added a new line where appropriate, but it comes down to if you will accept that coding style


vonnieda:
Looks good, thank you @pfried!

PR #940, opened 2020-01-23
Feature/A More Job Like Manual Pick (Repeat)

markmaker:
Note, this is a repeat of what the #939 was intended to be. The changes were reverted in the merge, due to a git misunderstanding on my side.

---

# Description

A pick in the Feeder Panel does not behave like one would in a Job. So this PR aims to implement a more "job like" behavior for the manual pick operation:

- The selected feeder is prepared (like all of them would in a Job).
- This results in the desired behaviour with the BlindsFeeder, otherwise it would sometimes end up in a ridiculous NozzleTip changing frenzy.
- You can now switch on automatic NozzleTip change on manual feed in the "Tool Changer" wizard of the Nozzle.
- If enabled, the NozzleTip is automatically changed for one that is compatible with the package in the feeder.
- If disabled, the pick is aborted if the NozzleTip is incompatible (of course, this is up for discussion, like everything).
- The vacuum check is also done.

# Justification
The user can better experiment with picks, if the behavior is more like in a Job. Most importantly the vacuum setup (vacuum ranges, dwell times) can now be tested outside a job. 

# Instructions for Use
Switch on the "Change On Manual Feed?" Checkbox in the "Changer Tool" WIzard of the Nozzle. 
grafik

Then use the pick button in the Feeder Panel:
grafik

# Implementation Details
1. Tested it with the BlindsFeeder, most specifically by forcing it to change the nozzle tip for cover opening. While testing, I had some pick issues and only noticed then (after some frustration), that the vacuum check is not currently performed in the manual pick. 
3. Added the  `org.openpnp.spi.Nozzle.isNozzleTipChangedOnManualFeed()` method.

### DIFF
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -77,6 +77,8 @@ import org.openpnp.model.Part;
 import org.openpnp.spi.Camera;
 import org.openpnp.spi.Feeder;
 import org.openpnp.spi.Nozzle;
+import org.openpnp.spi.NozzleTip;
+import org.openpnp.spi.JobProcessor.JobProcessorException;
 import org.openpnp.spi.PropertySheetHolder.PropertySheet;
 import org.openpnp.util.MovableUtils;
 import org.openpnp.util.UiUtils;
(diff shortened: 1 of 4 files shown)


markmaker:
Hi @vonnieda 

Seeing that you merged the original and having double-checked the commits, I'll merge that myself.

_Mark

PR #941, opened 2020-01-23
Revert "Feature/A More Job Like Manual Pick (Repeat)"

vonnieda:
Reverts openpnp/openpnp#940

@markmaker I think the history of this commit got tied up in the BlindsFeeder PR so that when I merged the BlindsFeeder PR Github thought this had been merged as well, so it closed the PR and marked it merged, but it was not actually merged.

I'm not quite ready to merge this yet, so reverting and will discuss in the PR.

PR #942, opened 2020-01-23
Feature/A More Job Like Manual Pick (Repeat 2)

markmaker:
Note, this is a revert of the revert of the repeat of what the #939 was intended to be. 

---

# Description

A pick in the Feeder Panel does not behave like one would in a Job. So this PR aims to implement a more "job like" behavior for the manual pick operation:

- The selected feeder is prepared (like all of them would in a Job).
- This results in the desired behaviour with the BlindsFeeder, otherwise it would sometimes end up in a ridiculous NozzleTip changing frenzy.
- You can now switch on automatic NozzleTip change on manual feed in the "Tool Changer" wizard of the Nozzle.
- If enabled, the NozzleTip is automatically changed for one that is compatible with the package in the feeder.
- If disabled, the pick is aborted if the NozzleTip is incompatible (of course, this is up for discussion, like everything).
- The vacuum check is also done.

# Justification
The user can better experiment with picks, if the behavior is more like in a Job. Most importantly the vacuum setup (vacuum ranges, dwell times) can now be tested outside a job. 

# Instructions for Use
Switch on the "Change On Manual Feed?" Checkbox in the "Changer Tool" WIzard of the Nozzle. 
grafik

Then use the pick button in the Feeder Panel:
grafik

# Implementation Details
1. Tested it with the BlindsFeeder, most specifically by forcing it to change the nozzle tip for cover opening. While testing, I had some pick issues and only noticed then (after some frustration), that the vacuum check is not currently performed in the manual pick. 
3. Added the  `org.openpnp.spi.Nozzle.isNozzleTipChangedOnManualFeed()` method.

### DIFF
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -77,6 +77,8 @@ import org.openpnp.model.Part;
 import org.openpnp.spi.Camera;
 import org.openpnp.spi.Feeder;
 import org.openpnp.spi.Nozzle;
+import org.openpnp.spi.NozzleTip;
+import org.openpnp.spi.JobProcessor.JobProcessorException;
 import org.openpnp.spi.PropertySheetHolder.PropertySheet;
 import org.openpnp.util.MovableUtils;
 import org.openpnp.util.UiUtils;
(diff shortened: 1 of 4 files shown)


markmaker:
Resubmitted for review and discussion, see #941, #940, #939.

vonnieda:
Thanks @markmaker - so there are two things I want to discuss:

1. Do we really need to add the "nozzle change on manual feed" option? If the user wants to feed a part, and they have a nozzle changer, why not just go ahead and change it and do what the user intended to do without bugging them?
2. The general "job like" feed is good - I'm doing the same thing in my visual training code so mainly I think we should make this a utility method that is used in the job, the manual feed button, and my visual training code instead of duplicating it everywhere. I can refactor that once I merge my stuff, so this isn't as big of a deal, but it relies on #1. My preference would be to not add the "nozzle change on manual feed" option so that the utility method can be simple.

Thoughts?

Thanks,
Jason

markmaker:
Hi @vonnieda 

Thanks for looking into this.

> 1. Do we really need to add the "nozzle change on manual feed" option? If the user wants to feed a part, and they have a nozzle changer, why not just go ahead and change it and do what the user intended to do without bugging them?

I just wanted it to be no surprise for users updating to the new revision. It is potentially dangerous, if the machine suddenly goes to change the nozzle tip, completely unexpected by an experienced user. However, because this is 2.0 and if you say so, we could also just remove it. ;-)

> 2. The general "job like" feed is good - I'm doing the same thing in my visual training code so mainly I think we should make this a utility method that is used in the job, the manual feed button, and my visual training code instead of duplicating it everywhere. I can refactor that once I merge my stuff, so this isn't as big of a deal, but it relies on #1. My preference would be to not add the "nozzle change on manual feed" option so that the utility method can be simple.

I completely agree, that the best solution would be to not imitate the job behavior but to put both on the same basis. And I see how that idea should definitely be fulfilled, albeit in a different form (see below).

For the feeder panel pick action, however, I see several problems.

1. Step granularity I: a feed and a pick are wrapped in one step. We only want the pick, so we can't wrap up the whole Pick step in the utility method. Looking at the code, there is also an exception handling issue, which is best solved by splitting up the utility method once more. Not much "separation of concerns" gained when in fact two method calls remain per utility method.   
2. Step granularity II: the feeder preparation and loading the right nozzle tip is done in completely different steps in a real job. Plus the nozzle tip loading is entangled in planner logic, that is completely unwanted here and cannot be reused. Consequently, it is not possible to use one utility method to wrap up all the bits needed, and to still also use that method from the job. You would have to break this up into even more utility methods, again not gaining much "separation of concerns". 
3. Dependencies: There should be no requirement to have a job loaded, board location defined, placements defined etc, before you can try your first pick (especially when you set up a new machine).  However, the primary parameter of `stepImpl()` is the `PlannedPlacement` and the utility method would have to artificially forfeit that knowledge to be able to work outside a job (there are script parameters already using that info). 
4. In the end you end up with almost as many utility method calls in the feeder panel pick action, as I propose in this PR. You still need custom code to select the right nozzle tip outside a planner regimen. And you burden the utility methods, and therefore the JobProcessor with the constraint to not use any knowledge about the planned placement, the board, retry exception handling, etc.

Like I said, the thought of (re)using the true Job behavior for testing is definitely the right one! However, it might be better channeled into a different, very useful function:

- Performing one isolated Job placement ad-hoc i.e. regardless of and without any effect to the placement's and the job's status.

This would be available for one placement inside a job. It would run the job as if just this one part was to be placed. The true behavior could be tested:

grafik

_Mark

markmaker:
Hi @vonnieda 

Found a mistake in my point 1: It is not true that we don't want the `feed()`, so that argument is invalid. 

Also, now re-reading your post, I'm not so sure anymore, if I completely understood, what you `really` intend to do, so maybe my response is off the mark too. 

_Mark

PR #947, opened 2020-02-01
Nozzle Offset Wizard optional Z offset

markmaker:
# Description
For the most commonly used OpenPNP coordinate system with Z = 0 = Safe Z = Camera Z, the Offset Wizard should only capture X, Y and not Z. Therefore, this PR adds a checkbox to include Z that is switched off by default. 

If switched on, the guide reminds the user to move the nozzle to precisely the desired Z offset height. There is also a **Warning** that only machines with Safe Z > Camera Z should use this option.

grafik

The PR also swaps around the last two steps of the former guide text. Before pressing "Apply", the new offset is not applied to the Nozzle and testing it on the Machine Controls Panel would be futile.

# Justification
Using the former nozzle Offset Wizard step-by-step will capture the Z height of the mark down on table as the nozzle Z offset. With the most commonly used OpenPNP coordinate systems i.e. with Z = 0 = Safe Z = Camera Z (more generally with Safe Z ≤ Camera Z), this results in crashing the tip on the table with any subsequent move at (supposedly) Safe Z.  

## Why not just leave out Z altogether?
Capturing the Z offset _does_ make sense for coordinate systems anchored on the table surface:

- the down-looking Camera Z (usually Z = 0) is assumed to be the focal plane, which is at the PCB / machine table surface height (i.e. where the mark is!) 
- Safe Z of all the other HeadMountables is above that (e.g. a positive Z value)
- capturing the Z offset will make sure each nozzle Z = 0 is precisely at the PCB / machine table surface height

Therefore, adding the checkbox and allowing for both systems _does_ make sense. 

# Instructions for Use
Most users will just leave the "Include Z?" checkbox off. The mark position Z value is then ignored i.e. the new nozzle offset Z will be reset to the old Z, already stored on the Nozzle. 

Users that want to store a Z offset, switch the "Include Z?" checkbox on. They then need to make sure the Nozzle Z is at the desired offset height. 

# Implementation Details
1. Unfortunately I haven't been able to test the changes, due to no access to the machine. The intention is to allow others to test and to discuss this.
4. Did run `mvn test` before submitting the Pull Request. However, this is hardly conclusive (see 1.).


vonnieda:
Has anyone had a chance to test this and see that it works? I'm ready to merge it if others have tested. I am unable to test it currently.

vonnieda:
Hi @markmaker, I think there is some confusion about Z head offset - I'm not sure if it's on my part or yours :)

The way that I use OpenPnP, and the way I *think* most people use it, is the Z offset is used to account for differences in the Z positioning or length of nozzles in relation to each other. This is why we use something as a reference surface. The height of the surface doesn't matter, as we're measuring the difference between nozzles.

When OpenPnP positions, it subtracts the offsets, so if one nozzle is longer than the other, from OpenPnP's perspective when moving, they are the same.

Given this, the addition of the checkbox seems of limited use, and, if added, should default to on, I think.

markmaker:
Hi @vonnieda 

It is well possible that I'm mistaken. 

Given my recent RTFM moment, I now tried to reread the Wiki and find where I might have veered off. But I haven't found anything about where zero is supposed to be in Z and about safe Z. Might have tomatoes on my eyes :)

https://github.com/openpnp/openpnp/wiki/Setup-and-Calibration%3A-Nozzle-Setup#head-offsets
https://github.com/openpnp/openpnp/wiki/User-Manual#coordinate-system

I also looked at the [default machine.xml](https://github.com/openpnp/openpnp/blob/develop/src/main/resources/config/machine.xml) and the [OpenBuilds machine.xml](https://github.com/openpnp/openpnp-openbuilds/blob/develop/OpenPnP/machine.xml), both are useless as guides, because they don't include a proper Safe Z config.

So these were my thoughts and assumptions:

1. OpenPNP, while being extremely universal, still was created with a dual nozzle shared Z axis designs in mind. 
2. This influences how one thinks about Z and more specifically about safe Z namely that there is a definitive height at which the nozzle is "retracted", namely when both are at the same height. 
3. The assumption is also reflected in the design of OpenPNP in the way that Safe Z is a single value rather than a "greater than" condition or a safe Z range. 
4. There is no such "natural" choice in a single nozzle machine or a multi nozzle machine with dedicated (non-shared) Z axes i.e. in such a machine you would choose your safe Z freely, namely at a height that is sufficient to not collide with anything (having the tallest part on the nozzle tip). You would choose Safe Z as low as possible for speed. 
5. In a shared Z axis machine, the Safe Z value must be valid and equal (after transform) for both nozzles because otherwise a `moveToSafeZ()` on the head would lead to unnecessary wiggle moves when one nozzle's safe Z overrides the other's. If Safe Z was a "greater than" condition or range, this constraint could be removed. Even a shared Z axis machine could optimize safe Z and allow limping nozzles for faster retract speed, I think I've seen this in commercial machine videos.
6. **Now comes the crux of the matter:** I had the impression that most people choose a safe Z of zero and have negative Z values for feeder heights etc. See here for [one independent example](https://github.com/ErichStyger/McuOpenPnP_Machine/blob/master/OpenPnP/machine.xml).
7. Zero has the added benefit that it will remain stable through AxisTransform. Other Safe Z values might create small floating-point errors when transforming back and forth resulting in unnecessary Z moves, because OpenPNP uses exact compares on axis coordinates (i.e. without any machine tolerance or resolution).
https://github.com/openpnp/openpnp/blob/6954a2a22f5f0cea8f8bfdbe6cb9eaac5b115a03/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java#L591-L603
8. Documentation says [that the camera position is the origin/reference of the coordinate system](https://github.com/openpnp/openpnp/wiki/Setup-and-Calibration%3A-Nozzle-Setup#head-offsets), i.e. specifically the camera Z offset is also zero. 
9. If you have a zero Safe Z (6), (7) and a zero camera Z (8), this means that the nozzle to camera offset **must not** include the Z down on the table (assuming we're still talking about the Offset Wizard, where we just made a mark in putty on the table).
10. If you include the Z offset, a `moveToLocationAtSafeZ()` results in the nozzle tip scratching on the table. Done that, been there.
11. Hence, the checkbox (and why I believed **default off** is right).

Tell me we're I went wrong :-) I guess we should then amend the Wiki to make all things Z much clearer. 

_Mark

vonnieda:
Hey Mark,

My usual setup, and I *think* the most usual setup by most is:

Camera Z Head Offset: 0
Nozzle Z Head Offset: Negative value, on my CHMT36VA this is -8.5mm
Safe-Z: 0 in all cases

Now, with that said, on my CHMT36VA which is a two nozzle cam system, there
is the cam wheel radius of +8.5. So, the two essentially cancel each other
out to get 0, but if the -8.5 is not specified then it will not work
correctly.

I usually recommend to people that Safe-Z be 0 in all cases, simply because
it's a poorly implemented feature and adds too much complexity.

I disagree with the statement "If you include the Z offset, a
moveToLocationAtSafeZ() results in the nozzle tip scratching on the table."
since this is exactly how I, and I believe many others, use it. So, perhaps
there is a difference in our setups? If I remember correctly, you are using
a LitePlacer, which has a single Z and no cam.

So, I wonder if it will help if you list your settings as I have above?

Thanks,
Jason

markmaker:
Hi Jason

Thanks for looking into this.

My setup currently is

- Camera Z Head Offset: 0
- Nozzle Z Head Offset: 0 (would be roughly +35.2mm by Wizard, guessing from feeder heights)
- Nozzle Safe Z is: -22mm, so 13.2mm above deck. 

My thinking goes like this: 

1. the Nozzle Head Offsets are Camera location minus Nozzle mark location (not the other way around), so the Z offset should be _positive_ for machines where absolute Z (i.e. the camera relative logical Z coordinate) is in the negative on deck level:
https://github.com/openpnp/openpnp/blob/fdca1a72221f6b5fa3bf2f1bd3aa6ab284e4aa4d/src/main/java/org/openpnp/machine/reference/wizards/ReferenceNozzleCameraOffsetWizard.java#L286
2. Safe Z is stored in Nozzle coordinates, it is **not** in absolute coordinates. If you look at moveToSafeZ() in the Nozzle ... 
https://github.com/openpnp/openpnp/blob/6954a2a22f5f0cea8f8bfdbe6cb9eaac5b115a03/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java#L324-L331
... it goes directly to the driver which in turn subtracts the `hm` (i.e. Nozzle) Head Offsets:
https://github.com/openpnp/openpnp/blob/6954a2a22f5f0cea8f8bfdbe6cb9eaac5b115a03/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java#L490
3. Subtracting the positive Z offset from (1) will move the nozzle further down. 
4. This means that changing the Head Offsets but not changing Safe Z will change the physical Safe Z. Assuming the Z Offset was zero, for all machines that have absolute Z < 0 on deck level the shift of Safe Z is downward for those with absolute Z > 0 upward. For your machine with the cam wheel radius cancelling out the offset, the effect might be invisible (?). 
5. So unless I am missing something vital here, it is completely natural that you can't just change the Z Offset without danger.
6. An alternative would be to let the Wizard adjust Safe Z along with the Head Offsets.

I should go back to the machine and retry this. It definitely did the scratching when I didn't manually delete the Z value from the measured Z offset. 

AlexH in the group [reported the same thing in detail](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/openpnp/dFFIqhHxlpc/wUbTPGQRFgAJ) (which prompted this PR). 

_Mark

vonnieda:
Thanks Mark - I will also look and see if the wizard does something
different than my manual instructions used to. I've never personally used
the wizard.

My guess is this has something to do with Safe-Z. I never use, and I never
recommend its use as I think the feature is broken by design. I'd just as
soon rip it out and redo it.

Jason

markmaker:
Thanks, Jason

> My guess is this has something to do with Safe-Z. I never use, and I never recommend its use as I think the feature is broken by design. I'd just as soon rip it out and redo it.

I've used my custom/non-zero Safe Z for some time now, and I don't think it is broken, per se. Like I suggested, I believe it could be improved by being a "greater than" condition or range, but mathematically you can't really do it fundamentally different. 

About the Wizard. 

The mark location is recorded as absolute coordinates by taking the nozzle location and subtracting the _old_ offsets:
https://github.com/openpnp/openpnp/blob/fdca1a72221f6b5fa3bf2f1bd3aa6ab284e4aa4d/src/main/java/org/openpnp/machine/reference/wizards/ReferenceNozzleCameraOffsetWizard.java#L274

That is equivalent to the current camera location in X and Y (if camera offsets are zero as recommended). 

Looking at this line again...
https://github.com/openpnp/openpnp/blob/fdca1a72221f6b5fa3bf2f1bd3aa6ab284e4aa4d/src/main/java/org/openpnp/machine/reference/wizards/ReferenceNozzleCameraOffsetWizard.java#L286
... the idea behind the subtraction is "how far did I need to move the camera to be at the same location the nozzle was before?" It is obvious that correctly gives us the offsets in X and Y.

However, the camera can't be moved in Z. If you call `camera.getLocation()` it will have no Z axis mapped and constantly return the camera's Head Offset Z which should be 0. 

The camera simply can't be used as a reference for Z. 

Instead, the subtraction gives us the answer to "what is the needed Z offset for the nozzle to be at Z=0 (Camera Z) when the nozzle tip touches the table surface". 

It only makes sense to store this Z offset if you want table surface Z = 0 (Camera Z). 

Needless to say: you need to set a Safe Z higher than 0 in that case. Otherwise, "crash-screeech". 

_Mark

markmaker:
Hi @vonnieda 

I tested the old code. Confirmed that it crashes the nozzle if you follow the steps with a machine that has Camera Z = 0 and Safe Z = 0 or even below. Removed the slate to test, it dives right down when you press the Nozzle Tip to Camera Location Button on the Machine Controls as suggested by the Wizard:

NozzleOffsetWizardCrash

Also tested the new PR code. I fixed some additional issues in the old version.

* Do not set the new head offsets on the nozzle until "Apply" is pressed
* Adjust Safe Z if Z is included
* Hide Z fields if Z is not included
* Added `firePropertyChange()`s to prevent lost updates if both the "Configuration" and "Offset Wizard" pages of the Nozzle are opened and modified (both have the `headOffsets` and now the  `safeZ` properties buffered)
* initialize the mark location in the proxy so it will not throw on Apply with manual data entry
* added Decorators
* enlarged fields for larger/negative Millimeter numbers from 5 to the standard 10 dialog units

NozzleOffsetWizard

_Mark

vonnieda:
Thanks @markmaker!

PR #960, opened 2020-03-03
Neoden4: First usable version of driver

tlan:
This version of the driver will allow you to move around, use nozzles, rails, lights etc,
It also honors the status flag from Neoden motion controller when moving around,
preventing damage to nozzles, and not skipping commands etc.

Still work in progress, tho.


vonnieda:
Great work @tlan! This is going to be really awesome for NeoDen 4 owners!

PR #965, opened 2020-03-13
Add support for vacuum & blow threshold

tlan:
If set on the part's package vacuum threshold and / or blow threshold
will be used in pick and place operations instead of just using on / off.

Changes Package and Nozzle.


vonnieda:
This has been merged with some changes. See https://github.com/openpnp/openpnp/commit/0aada0eb69bae5bea3939d092ec28c30e85c4ace for the commit.

The changes are:

- Change "blow" terminology to "blow off".
- Move vacuum and blow off settings from package table to new package tab.
- Packages panel now remembers the selected tab when switching packages. Helpful if you need to change a bunch of vacuum or blow off levels at once.

The functionality is otherwise the same. You'll need to edit or reset your machine.xml and packages.xml to fix the field names.

Thanks,
Jason

PR #973, opened 2020-03-30
job auto save option

aneox:
# Description
The checkbox in PNP processor settings, allow saving job each part place.

Sometimes, rarely, my computer freeze or reboot. So this option must-have for me.
May be some users want it too.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -76,6 +76,9 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
     @Attribute(required = false)
     protected int maxVisionRetries = 3;
 
+    @Attribute(required = false)
+    protected boolean autoSaveJob = false;
+
     @Element(required = false)
     public PnpJobPlanner planner = new SimplePnpJobPlanner();
 
@@ -94,7 +97,15 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
     
     public ReferencePnpJobProcessor() {
     }
-    
+
+    public boolean isAutoSaveJob() {
+        return autoSaveJob;
+    }
+
+    public void setAutoSaveJob(boolean autoSaveJob) {
+        this.autoSaveJob = autoSaveJob;
+    }
+
     public synchronized void initialize(Job job) throws Exception {
         if (job == null) {
             throw new Exception("Can't initialize with a null Job.");
@@ -754,6 +765,16 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
             // Mark the placement as "placed"
             boardLocation.setPlaced(jobPlacement.getPlacement().getId(), true);
             
+            if (autoSaveJob) {
+                try {
+                    Configuration.get().saveJob(job, job.getFile());
+                }
+                catch (Exception e) {
+                    // TODO Auto-generated catch block
+                    e.printStackTrace();
+                }
+            }
+
             totalPartsPlaced++;
             
             scriptComplete(plannedPlacement, placementLocation);
--- a/src/main/java/org/openpnp/machine/reference/wizards/ReferencePnpJobProcessorConfigurationWizard.java
+++ b/src/main/java/org/openpnp/machine/reference/wizards/ReferencePnpJobProcessorConfigurationWizard.java
@@ -25,6 +25,7 @@ import javax.swing.JLabel;
 import javax.swing.JPanel;
 import javax.swing.JTextField;
 import javax.swing.border.TitledBorder;
+import javax.swing.JCheckBox;
 
 import org.openpnp.Translations;
 import org.openpnp.gui.components.ComponentDecorators;
@@ -43,6 +44,7 @@ public class ReferencePnpJobProcessorConfigurationWizard extends AbstractConfigu
     private final ReferencePnpJobProcessor jobProcessor;
     private JComboBox comboBoxJobOrder;
     private JTextField maxVisionRetriesTextField;
+    private JCheckBox autoSaveJobCheckBox;
 
     public ReferencePnpJobProcessorConfigurationWizard(ReferencePnpJobProcessor jobProcessor) {
         this.jobProcessor = jobProcessor;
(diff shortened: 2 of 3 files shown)


vonnieda:
Hi @aneox - this was in OpenPnP for a while but removed because it causes performance problems. See https://github.com/openpnp/openpnp/blob/develop/CHANGES.md#openpnp-20

> BREAKING CHANGE: Removed job auto save and config auto safe from the job processor. This feature caused a serious performance hit and cluttered up the job processor. This feature will be added back at a later date with a new implementation that does not harm performance.

I am open to adding this feature back if it's done in such a way that it doesn't cause a delay each time the job and/or config is saved. One idea would be to override the writing of the config and job so that it writes to a RAM buffer and then launch a thread to write it out to disk in the background.

aneox:
@vonnieda 
Hi, sure, I remember you said about performance issue, so at this time, I have added the checkbox, which can disable this feature. In my case, no performance problems, with Samsung nvme SSD.

vonnieda:
Hi @aneox - the old version had the checkbox too. The problem is that people *want* to use this, but then they enable it and run into performance problems. With small jobs it's not bad, but if you have 100 placements an an 8 panel board suddenly it can hundreds or thousands of milliseconds to save the files. 

I'm sorry, I won't merge it in this form. If you want to modify it to solve the performance issue, I will look at it again.

markmaker:
Hi

One thing to look at, would be to `<part-settings-by-part-id>`. This replicates both a custom bottom vision pipeline and a fiducial locator pipeline for every part (not package!) encountered by OpenPNP (i.e. whenever you pressed "Reset All Parts" on one of the machine vision wizards after tuning the default pipeline). This creates a huge avalanche of 99.9% repetitive data in the `machine.xml`, half of it useless (fiducial and bottom vision) and this just grows and grows, obsolete parts never removed. Frankly I don't see the point in defining part specific pipelines at all, package specific would be enough (you can always create a new package, if one part is peculiar). At least these pipelines could be set to null / removed from the hash map if identical to the default pipeline (i.e. also when using the "Reset All Parts" button). 

There's a good chance that getting rid of these would eliminate the performance problem for most if not all real world users.

@vonnieda, is there a reason why these are not in the `parts.xml` in the first place (or better yet in the `packages.xml`)?

_Mark

netzmark:
On the moment I also have Samsung Evo but I didn't test this feature on it yet. But for sure it issued on the job performance on my previous disk, also SSD.

vonnieda:
@markmaker Agreed that the bottom vision pipelines should be on packages. I put it on part originally because I thought that people would want to have as few packages as possible and I worried about color variations on different parts with the same package. But I've since shifted to think, as you do, that you can just define a new package in that case.

With that said, I've also been thinking a lot about named pipelines, and I believe @tlan has some code in the works for that. In that case, instead of having pipelines stored on part or package or anywhere else we just have a (hopefully small) global lists of named pipelines and you just select the one you want for a given operation.

The reason they are in machine.xml and not parts.xml or packages.xml is that early in the project I wanted to keep all machine configuration in machine.xml and keep parts and packages relatively generic. The intention was for users to be able to share information from parts and packages, and in time, have a cloud database of all the common ones. Since pipelines are quite machine specific, I thought they belonged in machine.xml. My inclination these days is that nearly everything belongs in the job (feeders, parts, packages, vision settings, etc.) so there's a lot to think about before big changes are made :)

All that aside - the feature under discussion here has a pretty easy path to being performant. We just need to buffer in RAM and then write to disk from a worker thread. I think that is the right approach for this feature - not trying to change the entire config layout to suit the feature.

netzmark:
@vonnieda, and when do you want to save RAM containment to the disk?
"when" I mean which point of job?

aneox:
@vonnieda Today openpnp freezed with a lot of errors 
`2020-04-02 23:06:21.132 SystemLogger ERROR: LOGGER ERROR: Failed to write log entry (java.lang.NullPointerException)`
and then crashed

markmaker:
> All that aside - the feature under discussion here has a pretty easy path to being performant. We just need to buffer in RAM and then write to disk from a worker thread. I think that is the right approach for this feature - not trying to change the entire config layout to suit the feature.

Are you sure it's the disk IO? The serializer uses reflection to save the data and I would imagine, this is a costly operation, given the sheer number of objects it recurses through (pipeline stages!). 

It is just my experience as a 30-year professional developer, that proper buffered disk IO usually cannot be significantly sped up by rolling our _own version_ of buffered disk IO. 
   :-) 

I was shortly wondering if the `close()` should be deferred. I haven't found any doc in the API whether Java waits for a _physical_ flush on `close()` or not. It seems not:
https://stackoverflow.com/questions/730521/really-force-file-sync-flush-in-java

But maybe in Java everything is different. 

_Mark

vonnieda:
@netzmark: I think save any time the job is stopped or paused, and then maybe every minute. This should really handle almost anything except a computer or software crash. If it's a crash, there's no guarantee any of this works anyway, so I think better to handle the common cases and also try to fix bugs that cause crashes.

@markmaker I believe I tested this when we went through this the last time and it was disk. SSDs are still between tens and hundreds of times slower than writing to RAM, and as Marek originally said, 200ms of delay in a job with hundreds of parts can add minutes to a job.

The thread where this was all discussed is here: https://groups.google.com/forum/#!msg/openpnp/SKcP9xNd2_E/9gsBRTwnAgAJ

I want to note that in that thread I suggested that this is something that happens quite rarely and people agreed with that.

My reasoning on this is:
1. The feature was added via PR and [I suggested during the review](https://github.com/openpnp/openpnp/pull/740#pullrequestreview-135211079) that it might cause peformance issues.
2. Eventually [someone found that it did cause performance issues](https://groups.google.com/forum/#!msg/openpnp/SKcP9xNd2_E/9gsBRTwnAgAJ) so changes were made to improve that.
3. It was still problematic, and people seemed to agree that it was not actually needed, so it was removed.

As I said above, if someone is willing to do this in a way that does not cause performance issues, whether that is saving to RAM, fiddling with close(), saving a small "state" file, or anything else, I will be happy to merge it. But as this PR stands now, it is just the re-addition of a feature that caused support headaches, so it will not be merged in this form.

vonnieda:
Sorry, didn’t mean to close it.

netzmark:
Obligatory saving at the job stopped-finished or paused and optional (checkbox) for additional saving every minute" - for me personally sounds good. 
But fast machine may do many parts in 4 minutes, so I can understand @aneox's point of view.

PR #975, opened 2020-04-03
Feature/improved affine transform

tonyluken:
# Description
 Changed to use a least-squared-error method for computing affine transforms.  Updated the referenceFiducialLocator code to use all fiducials that are enabled in the placements panel.  Also updated (and renamed) the MultiPlacementBoardLocationProcess code to semi-automate it and allow more than two placements to be used.  Both now perform nominal checks on the results.  If the results are out of nominal tolerance limits, referenceFiducialLocator will throw an exception and MultiPlacementBoardLocationProcess will pop-up an error dialog to inform the user.

# Justification
This change allows all available fiducials (or multiple placements) to be used when computing the board placement affine transform.  This should help reduce placement errors due to measurement errors of the fiducial (or placement) locations.

# Instructions for Use
On the Placements panel of the Job tab, enable the fiducials (two or more) that are to be used and click on the Perform Fiducial check button. The camera will move to each fiducial's expected location and the vision system will attempt to measure its actual location.  Alternatively, if fiducials are not available, click on the Multi Point Placement (which used to be the Two Point Placement) button to perform a manual alignment.  This procedure has changed slightly in that all the desired placements are selected in the first step (using ctrl-click or shift-click in the Placements panel) and then, on the following steps, the camera will move to the expected location of each selected placement and wait for the user to manually jog the camera to its exact position.  In either case, if all goes well, the board's placement affine transform is set and the camera moves to the board's origin.

# Implementation Details
**1. How did you test the change?** Tested on an actual board and on print-outs of boards that had deliberate distortion applied to them.  Tried various quantities of fiducials (and placements) and in different configurations to see the effect on the results.  Tests were run on both sides of a board to ensure that both the top and bottom of a board work correctly.  A test was also performed on a print-out of a panelized board to verify panelization still works correctly.
**3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing**.  No changes were made to these packages.


vonnieda:
This is really impressive work @tonyluken! It passes all the Maven tests, which is great, and I did some manual testing and it looks good too. I really like the new MultiPlacementBoardLocationProcess. Much easier to use than the old one. Great work!

PR #976, opened 2020-04-05
Feature/fix nozzle pirouettes and add backlash compensation to Z and Rotation axis

tonyluken:
# Description
By default, nozzle rotations are now performed the shortest way around unless the Nozzle Limit Rotation check box is checked in which case nozzle rotations are still limited to +/- 180 degrees.  Also adds support for backlash compensation on the z-axis and the rotation axis.

# Justification
Eliminates a lot of wild nozzle rotations that can occur during a job.  On some machines (Liteplacer), the nozzle rotation is driven via a belt.  On the Liteplacer, the belt must be loose enough to allow the z-probe function to operate correctly but that looseness can result in significant rotational backlash.  Since x and y already support backlash compensation, it only seems reasonable to support it on the other axis as well.

# Instructions for Use
No new instructions for the limiting nozzle rotations.  The new backslash compensations work just like x and y.  If they are to be used, the MOVE_TO_COMMAND should be modified the use the BacklashOffsetZ and BacklashOffsetRotation variables in a similar manner as for x and y.

# Implementation Details
**1. How did you test the change?** Ran some test jobs on my machine and observed far fewer rotations of the nozzle and angular placement improved with rotation backslash compensation enabled.  Note I did not test z-axis backslash compensation because I did not need it.
**3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.**  No changes were made to either of these packages.


vonnieda:
Nice work @tonyluken. I think this functionality existed *somewhere* way back in the day - possibly in the GrblDriver, but it probably got lost during the big push for GcodeDriver. This is a nice change. Thank you!

PR #977, opened 2020-04-10
Feature/ReferencePushPullFeeder

markmaker:
# Description
Generic implementation for a feeder that can feed parts by performing push and pull (and other) motions using a head-mounted Actuator.  

- Uses a semantic vision concept based on sprocket holes and knowledge of the EIA 481 standards. 
- Simple OCR vision stage that can read the part from a label on the feeder. 
- OCR based automatic feeder slot swapping, part changing or even new feeder creation. 
- Automatic and learning "only when needed" calibration for feeders that are mounted with large tolerances (+/-2mm). 
- Any part pitch, any feed pitch, any tape width.
- Multi-feed: 2mm pitch/0402 support. For speed, you can also use a multiplier for any pitch. 
- Easy to clone & sync settings between feeders, by package or by (new) Tape&Reel Specification.
- One click setup from the second feeder.
- Handles geometric transformations for all the relevant settings (allows cloning settings from a "west" to a "south" feeder, for instance). 

While this feeder was developed for an upcoming Open Hardware all-3D-printed tape reel feeder, it was an important goal to make it universal. If you like the added functionality, you should also be able to cover most use cases for ReferenceLeverFeeder and ReferenceDragFeeder. 

vlcsnap-2020-04-10-22h34m30s611

This PR encloses these major building blocks that are carefully modelled for independent reuse by other software and/or pipelines: 

1. the ReferencePushPullFeeder class
2. the SimpleOcr vision stage to read text from the camera
3. the RegionOfInterestProcess guided process to stake out a region of interest (e.g. for OCR) in the camera view using the mouse and without having to go into the pipeline. 
4. the AffineWarp vision stage that is used to extract and rotate the OCR region of interest from the image (it can define the ROI in the stage or get it from the pipeline caller by property)
5. the AffineUnwarp vision stage that transform geometrical vision results back to full camera image coordinates 

The AffineUnwarp is not actually used by the feeder, but it makes the AffineWarp ready for use in BottomVision and FiducialLocator pipelines, where it can speed up subsequent stages by reducing image size. As a benefit, you can let the pipeline end with the warped image and get a nice magnification effect. 

# Justification
## Why yet another feeder class? 
This PR started out from trying to extend ReferenceLeverFeeder. But trying to provide both backwards compatibility and new features, quickly turned out impractical. ReferenceLeverFeeder uses a reference image for vision, that cannot be used for the semantics I needed. Such as "understanding" the tape location and feed direction, cloning and transforming geometry from another feeder, changing from white to black tape, from 2mm to 4mm part pitch without invalidating the vision setup. On the other hand, ReferenceLeverFeeder should also not be burdened by complexity that some (existing) users don't want to use.  

## Why not break up the PR?
As can easily be seen, the building blocks (1 through 4, listed in the Description) all rely on each other. In order to test this code, including the stages using pipeline properties set by the feeder, I couldn't break up the PR. Building block 5 just makes AffineWarp complete for other use.

## Risk of introducing this code
Most of the changes happen in new classes. They are merely registered in existing classes. Unless they are actually instantiated by the user, they are without effect. 

## Changes in existing code
Some words on the changes in existing code (most important ones, roughly ordered by how GitHub lists the changed files):

`MainFrame`: The instructions' pane underneath the camera views, that can show/hide to guide through a process, is a bit hard to notice. I added a blinking "prompt" icon to it. Also, the font was always an out-of-place Serif font (setting the font on the pane does not work with `html` content, it seems). It is now told to use the standard dialog font.  

`CameraView`: I refactored the view pixel to length units calculation out, so I can use it in my RegionOfInterestProcess to calculate the ROI from clicks. See `getCameraViewCenterOffsetsFromXy()`

`LocationButtonsPanel`:  added the `positionActuatorNoSafeZAction` (incl. new icon) to enable No-Safe-Z motion with actuators (useful when setting up the feeder's push-pull motion).

`Package`, `PackagesTableModel`: added the tapeSpecification to the Package in order to group packages further, e.g. a "C" and "R" packages should be treated the same when it comes to feeding them from a tape, as well as various package sizes (0603, 0805, etc.) that still have the same part pitch. The user can now set a common tag such as "8x2" or "8x4" to code the tape width/pitch and other relevant specs. The ReferencePushPullFeeder will then be able to clone/sync settings across this group and automatically create new feeders with one click.  

`Feeder`, `ReferencePnpJobProcessor`, `BlindsFeeder`: reworked the job preparation for feeders that I introduced with the BlindsFeeder. Instead of letting each feeder class do its own Travelling Salesman journey for prep, the job processor now does it globally, so all feeders needing visits by the head will be prepped alongside each other, on a single optimized path. In case of the ReferencePushPullFeeder, the feeder's location is calibrated and the OCR tag verified.

`ReferenceFiducialLocator`: the locator now displays the final pipeline image (like bottom vision). I tested the AffineWarp/AffineUnwarp with visual homing, now showing a magnified fiducial lock. The code tests the presence of the MainFrame to work inside tests. 

`CvStage`, `ResultsPanel`: the results panel displays mouse coordinates converted into camera location offsets in proper length units, if the selected stage defines a unit by the new `getLengthUnits()` method. 

# Instructions for Use
For now there is a movie demonstrating the main usage: 
https://youtu.be/5QcJ2ziIJ14
Note, the movie does not explain the 3D printed feeder featured there. The feeder hardware will be documented later. 

vlcsnap-2020-04-10-22h35m03s333

I will create a Wiki entry when this PR has been merged (with whatever tuning yet required). 

# Implementation Details
1. The feeder class, region of interest process,  AffineWarp, SimpleOcr stages have been tested as shown in the movie. If you look closely you see the AffineWarp is also used in BottomVision, to test the AffineUnwarp stage i.e. transforming the coordinates back to full camera image pixel coordinates. It was also used/tested in the Visual Homing fiducial locator. **EDIT: This now also includes a Job test, mixed BlindsFeeder/ReferencePushPullFeder testing the latest code, mixed Travelling Salesman feeder prep and also btw. including the latest merge from `develop` with new OpenCV.**
3. I made changes in the `org.openpnp.spi` or `org.openpnp.model` as listed and justified above.

grafik


markmaker:
I wanted to note, that this PR is prepared for @tonyluken 's #943 as I made the following provisions:

- separated the part rotation in tape from the Rotation of the feeder, same signature as tony's base class implementation (once merged/inherited, it can then be removed from ReferencePushPullFeeder).  
- use the same feeder local transformation as #943
- access all Locations using setters/getter, so a first port should be trivial
- can be simplified, once #943 is merged, to directly work with the feeder local coordinate system

markmaker:
Hi @vonnieda 

Concluded the Job test with the latest code: mixed BlindsFeeder/ReferencePushPullFeder Travelling Salesman feeder prep and also btw. including the latest merge from `develop` [with new OpenCV 4.3.0](https://groups.google.com/d/msg/openpnp/pxMpm9CvVBU/HO4oeV0UAQAJ). 

This is IMHO ready for review/merge. 

Thank you!

_Mark

vonnieda:
Hi @markmaker - this is a massive PR. Is there any way you could minimize your changes to just your new feeder without making significant changes to the rest of the system? Perhaps work within the limitations of how things work now and then make incremental changes to other subsystems as needed? As this stands now, I can barely imagine how I would go about testing all of these changes.

markmaker:
Hi @vonnieda 

Thanks for looking into this.

I agree that just looking at the number of files changed looks daunting. :-(  But if you ignore all the new classes, icons, pipeline and their simple registration, I would strongly argue that this PR is _not_ massive at all.  

I've tried to explain why I can't break it up, in the section "Why not break up the PR?" (see Description of PR), as the building blocks are all interconnected and can only be tested together. I would have to commit untested code, to break that up. 

I've also detailed why the "Risk of introducing this code" is small, unless new classes are instantiated by the user, nothing bad can happen. The vision stages are all new, so they can't harm existing users. While I _do_ think these new stages will let users improve their fiducial locator and bottom vision pipelines, their use is completely voluntary, i.e. I have _not_ changed the default pipelines. 

Furthermore, I've tried to document the "Changes in existing code" and like to argue that changes are minimal, cosmetic and/or additive, i.e. just adding features rather than changing existing behavior. Some of the more cosmetic changes like the blinking icon in the description pane under the camera views, or supporting mouse cursor coordinates in length units in the pipeline results pane, I _could_ break out, if this _really_ helps to clear the way. 

The largest actual change concerns the `BlindsFeeder` but that is hardly in broad use by now. While the feeder preparation has changed, it is very easy to see that it can only ever affect the new `ReferencePushPullFeeder `and the `BlindsFeeder`, both fresh and still evolving, and actively being supported by me in the group, with users waiting to test and use them. 

I'm aware that I am asking much of you and cannot thank you enough for continuing to lead this great project forward.  I just hope you could give this another look, and maybe point to _more specific_ areas, i.e. where _exactly_ you are uneasy with the changes. 

_Mark

markmaker:
> Some of the more cosmetic changes like the blinking icon in the description pane under the camera views, or supporting mouse cursor coordinates in length units in the pipeline results pane, I _could_ break out, if this _really_ helps to clear the way.

Done that. And two more, see the PRs and commits here. I hope this will all be properly re-untangled by git later. :-/

Down by 10 files, hope its worth all this work and helps the PR along. :-D

Thanks, 
_Mark

vonnieda:
Hi @markmaker,

Thanks for breaking the PR up. The other chunks were much easier to review and merge.

I accidentally merged this before I was ready, so I have reverted it temporarily. I'm going to take a closer look at a few of the more global changes and will get back to you soon.

Thanks,
Jason

PR #978, opened 2020-04-13
Bugfix/Pump Actuation

markmaker:
# Description
Fixes a small logic error in Revision: bd3a78a5f2890aa430858fed540963a5797cc93e which leads to the pump constantly being switched on and off. Observed this in the log while testing something else (my machine does not actually actuate the pump, it runs on a hysteresis instead).

The pump must only be enabled before the valve switches on and only be disabled after the valve switches off. Same for the value setting variant with blow off. 

See the past change:
https://github.com/openpnp/openpnp/commit/bd3a78a5f2890aa430858fed540963a5797cc93e#diff-b380475a49f8df054452d2236b8bb376

# Justification
Obvious.

# Instructions for Use
Look into the log with tracing level. Do a small job. 

# Implementation Details
1. Tested on machine. Observed log. My machine does not actually actuate the pump. 


vonnieda:
Thanks @markmaker!

PR #986, opened 2020-04-17
Revert "Revert "Feature/ReferencePushPullFeeder""

vonnieda:
This will be the new #977 after the revert of the revert. #977 was accidentally merged, then reverted, and this is the revert of that which will eventually (actually) merge #977. I'm creating this so that there is an open PR for #977, so it doesn't get forgotten.

### DIFF
--- a/src/main/java/org/openpnp/gui/components/CameraView.java
+++ b/src/main/java/org/openpnp/gui/components/CameraView.java
@@ -1280,10 +1280,7 @@ public class CameraView extends JComponent implements CameraListener {
         }
     }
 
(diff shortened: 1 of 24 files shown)


vonnieda:
Note to self: Explore what is happening with the tape type in packages.

markmaker:
@vonnieda I'm closing this, as I opened #987 instead.

This way I'm able to work with the fork/push workflow I'm used to (git is at times still a bit unpredictable for me). I also merged and adapted the latest in develop (thank you for merging those!).

markmaker:
Closing it.

PR #987, opened 2020-04-17
Feature/ReferencePushPullFeeder (Reinstated)

markmaker:
_This is the reinstated #977 (after mistaken merge & revert) for continued discussion and improvement._

# Description
Generic implementation for a feeder that can feed parts by performing push and pull (and other) motions using a head-mounted Actuator.  

- Uses a semantic vision concept based on sprocket holes and knowledge of the EIA 481 standards. 
- Simple OCR vision stage that can read the part from a label on the feeder. 
- OCR based automatic feeder slot swapping, part changing or even new feeder creation. 
- Automatic and learning "only when needed" calibration for feeders that are mounted with large tolerances (+/-2mm). 
- Any part pitch, any feed pitch, any tape width.
- Multi-feed: 2mm pitch/0402 support. For speed, you can also use a multiplier for any pitch. 
- Easy to clone & sync settings between feeders, by package or by (new) Tape&Reel Specification.
- One click setup from the second feeder.
- Handles geometric transformations for all the relevant settings (allows cloning settings from a "west" to a "south" feeder, for instance). 

While this feeder was developed for an upcoming Open Hardware all-3D-printed tape reel feeder, it was an important goal to make it universal. If you like the added functionality, you should also be able to cover most use cases for ReferenceLeverFeeder and ReferenceDragFeeder. 

vlcsnap-2020-04-10-22h34m30s611

This PR encloses these major building blocks that are carefully modelled for independent reuse by other software and/or pipelines: 

1. the ReferencePushPullFeeder class
2. the SimpleOcr vision stage to read text from the camera
3. the RegionOfInterestProcess guided process to stake out a region of interest (e.g. for OCR) in the camera view using the mouse and without having to go into the pipeline. 
4. the AffineWarp vision stage that is used to extract and rotate the OCR region of interest from the image (it can define the ROI in the stage or get it from the pipeline caller by property)
5. the AffineUnwarp vision stage that transform geometrical vision results back to full camera image coordinates 

The AffineUnwarp is not actually used by the feeder, but it makes the AffineWarp ready for use in BottomVision and FiducialLocator pipelines, where it can speed up subsequent stages by reducing image size. As a benefit, you can let the pipeline end with the warped image and get a nice magnification effect. 

# Justification
## Why yet another feeder class? 
This PR started out from trying to extend ReferenceLeverFeeder. But trying to provide both backwards compatibility and new features, quickly turned out impractical. ReferenceLeverFeeder uses a reference image for vision, that cannot be used for the semantics I needed. Such as "understanding" the tape location and feed direction, cloning and transforming geometry from another feeder, changing from white to black tape, from 2mm to 4mm part pitch without invalidating the vision setup. On the other hand, ReferenceLeverFeeder should also not be burdened by complexity that some (existing) users don't want to use.  

## Why not break up the PR?
As can easily be seen, the building blocks (1 through 4, listed in the Description) all rely on each other. In order to test this code, including the stages using pipeline properties set by the feeder, I couldn't break up the PR. Building block 5 just makes AffineWarp complete for other use.

## Risk of introducing this code
Most of the changes happen in new classes. They are merely registered in existing classes. Unless they are actually instantiated by the user, they are without effect. 

## Changes in existing code
Some words on the changes in existing code (most important ones, roughly ordered by how GitHub lists the changed files):

`CameraView`: I refactored the view pixel to length units calculation out, so I can use it in my RegionOfInterestProcess to calculate the ROI from clicks. See `getCameraViewCenterOffsetsFromXy()`. Tip: use "ignore white-space" setting for diff.

`Package`, `PackagesTableModel`: added the tapeSpecification to the Package in order to group packages further, e.g. a "C" and "R" packages should be treated the same when it comes to feeding them from a tape, as well as various package sizes (0603, 0805, etc.) that still have the same part pitch. The user can now set a common tag such as "8x2" or "8x4" to code the tape width/pitch and other relevant specs. The ReferencePushPullFeeder will then be able to clone/sync settings across this group and automatically create new feeders with one click.  

`Feeder`, `ReferencePnpJobProcessor`, `BlindsFeeder`: reworked the job preparation for feeders that I introduced with the BlindsFeeder. Instead of letting each feeder class do its own Travelling Salesman journey for prep, the job processor now does it globally, so all feeders needing visits by the head will be prepped alongside each other, on a single optimized path. In case of the ReferencePushPullFeeder, the feeder's location is calibrated and the OCR tag verified.

This PR is well prepared for @tonyluken 's #943 as I made the following provisions:

- separated the part rotation in tape from the Rotation of the feeder, same signature as tony's base class implementation (once merged/inherited, it can then be removed from ReferencePushPullFeeder).  
- use the same feeder local transformation as #943
- access all Locations using setters/getter, so a first port should be trivial
- can be simplified, once #943 is merged, to directly work with feeder local coordinates, where needed

# Instructions for Use
For now there is a movie demonstrating the main usage: 
https://youtu.be/5QcJ2ziIJ14
For  _one possible_ use case of this class, employing an all-3D-printed reeled tape feeder, see here:
https://makr.zone/new-all-3d-printed-tapereel-feeder/399/

vlcsnap-2020-04-10-22h35m03s333

I will create a Wiki entry when this PR has been merged (with whatever tuning yet required). 

# Implementation Details
1. The feeder class, region of interest process,  AffineWarp, SimpleOcr stages have been tested as shown in the movie. If you look closely you see the AffineWarp is also used in BottomVision, to test the AffineUnwarp stage i.e. transforming the coordinates back to full camera image pixel coordinates. It was also used/tested in the Visual Homing fiducial locator. I also tested a simle Job, mixing BlindsFeeder/ReferencePushPullFeder, testing the latest Travelling Salesman feeder prep and also btw. including the latest merge from `develop` with new OpenCV.
3. I made changes in the `org.openpnp.spi` or `org.openpnp.model` as listed and justified above.

grafik

### DIFF
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -510,9 +510,10 @@ public class FeedersPanel extends JPanel implements WizardContainer {
                 Feeder feeder = getSelection();
 
                 // Simulate a "one feeder" job, prepare the feeder.
-                List<Feeder> feedersToPrepare = new ArrayList<>();
-                feedersToPrepare.add(feeder);
-                feeder.prepareForJob(feedersToPrepare);
+                if (feeder.getJobPreparationLocation() != null) {
+                    feeder.prepareForJob(true);
+                }
+                feeder.prepareForJob(false);
 
                 // Check the nozzle tip package compatibility.
                 Nozzle nozzle = MainFrame.get().getMachineControls().getSelectedNozzle();
--- a/src/main/java/org/openpnp/gui/components/CameraView.java
+++ b/src/main/java/org/openpnp/gui/components/CameraView.java
@@ -1280,10 +1280,7 @@ public class CameraView extends JComponent implements CameraListener {
         }
     }
 
-    private void moveToClick(MouseEvent e) {
-        int x = e.getX();
-        int y = e.getY();
-
+    public Location getCameraViewCenterOffsetsFromXy(int x, int y) {
         // Find the difference in X and Y from the center of the image
         // to the mouse click.
         double offsetX = (scaledWidth / 2.0D) - (x - imageX);
@@ -1297,12 +1294,17 @@ public class CameraView extends JComponent implements CameraListener {
         offsetX *= scaledUnitsPerPixelX;
         offsetY *= scaledUnitsPerPixelY;
 
-        // The offsets now represent the distance to move the camera
-        // in the Camera's units per pixel's units.
+        // The offsets now represent the distance in the Camera's units per pixel's units.
 
         // Create a location in the Camera's units per pixel's units
         // and with the values of the offsets.
         Location offsets = camera.getUnitsPerPixel().derive(offsetX, offsetY, 0.0, 0.0);
+        return offsets;
+    }
+
+    private void moveToClick(MouseEvent e) {
+        // Get the offset from the Camera view center in Camera's units.
+        Location offsets = getCameraViewCenterOffsetsFromXy(e.getX(), e.getY());
         // And move there.
         UiUtils.submitUiMachineTask(() -> {
             if (camera.getHead() == null) {
--- a/src/main/java/org/openpnp/gui/processes/RegionOfInterestProcess.java
+++ b/src/main/java/org/openpnp/gui/processes/RegionOfInterestProcess.java
@@ -1,0 +1,261 @@
+/*
+ * Copyright (C) 2020 <mark@makr.zone>
+ * inspired by TwoPlacementBoardLocationProcess by
+ * Copyright (C) 2011 Jason von Nieda <jason@vonnieda.org>
+ * 
+ * This file is part of OpenPnP.
+ * 
+ * OpenPnP is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ * 
+ * OpenPnP is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public License along with OpenPnP. If not, see
+ * <http://www.gnu.org/licenses/>.
+ * 
+ * For more information about OpenPnP visit http://openpnp.org
+ */
+
+package org.openpnp.gui.processes;
+
+import java.awt.BasicStroke;
+import java.awt.Color;
+import java.awt.Graphics2D;
+import java.awt.Point;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.awt.event.MouseAdapter;
+import java.awt.event.MouseEvent;
+import java.util.HashMap;
+import java.util.Map;
+
+import javax.swing.SwingUtilities;
+
+import org.openpnp.gui.MainFrame;
+import org.openpnp.gui.components.CameraView;
+import org.openpnp.gui.components.reticle.Reticle;
+import org.openpnp.model.LengthUnit;
+import org.openpnp.model.RegionOfInterest;
+import org.openpnp.spi.Camera;
+import org.openpnp.util.UiUtils;
+
+/**
+ * Guides the user through the three point region of interest operation using step by step instructions.
+ */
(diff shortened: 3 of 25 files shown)


markmaker:
@vonnieda [wrote: ](https://github.com/openpnp/openpnp/pull/986#issuecomment-615035723)
> Note to self: Explore what is happening with the tape type in packages.

The use of the Tape Specification is for the feeder to know which parts and packages are the same in terms of the tape feeding perspective. The feeder then knows it can clone settings across this group of feeders and even create new feeders solely from the part detected by OCR. 

Watch it in action here:
https://youtu.be/5QcJ2ziIJ14?t=294

And here:
https://youtu.be/5QcJ2ziIJ14?t=449

I'm aware that adding a column there, used only by _one_ feeder, is a bit cheeky. The only reason, I dared do it, is because the package table seems underused and it is nice to see the tape spec in an overview and to be able to sort (group) and search by it. 

grafik

I figured that the column could still be moved inside a Wizard tab, once more important columns compete for the space. 

_Mark

vonnieda:
Hi @markmaker,

It looks like I caused conflicts by merging your BlindsFeeder change earlier today. I've attempted to resolve the conflict but it's not looking right. So, two things:

1. It looks like the tape spec is just a free form string. Is the meaning you are using here 4mm feed by 2mm part pitch? And would it make sense to formalize this somehow? I could see it being useful for things like auto feeders and strip feeders, too.

2. Can you see if you can get the conflicts resolved? I'm about ready to merge this.

Thanks,
Jason

markmaker:
Hi @vonnieda 

> Can you see if you can get the conflicts resolved?

I should have done the merge immediately, sorry. It's done, now. 

> It looks like the tape spec is just a free form string. Is the meaning you are using here 4mm feed by 2mm part pitch? And would it make sense to formalize this somehow? I could see it being useful for things like auto feeders and strip feeders, too.

Yes its just a dumb String for now. All packages with identical tape specification make a group, that's all. 

In the screenshot I made it up to contain the tape width, the letter 'x' and the part pitch. I believe that's [a common way to spec what tapes a feeder takes](https://smtnet.com/index.cfm?fuseaction=search_submit&searchstring=8x4%20&collection=site_parts). The tape width gives you the offset from the line of sprocket holes to the line of parts (via EIA 481). The part pitch is also obviously important. 

**But** the feeder does **not** need to parse these properties from the tape specification, instead they are taken from another already set up feeder that is marked as a template and has a part with a package of the same tape specification (if the tape spec is empty, it will just take a template feeder with the same package instead).

The String could be **anything that makes the feeder setup for that tape spec unique.** For instance, [in the video I set up a feeder with transparent tape](https://youtu.be/5QcJ2ziIJ14?t=474), where I switch off calibration, because it can't see the sprocket holes (it still does the OCR). For such a package, I could now set the tape specification to "8x4T" as in "**T**ransparent", for instance. If another part with a package with the same tape specification comes along, it will automatically be cloned from that feeder.  

There are many properties that could make a certain type of tape unique for the feeder. Frankly I don't have enough experience yet ;-). 

One reason this is just a string, is that I plan to import it from the ECAD (Eagle for the moment), using a custom attribute. 

The tape specification could one day become an ID with an object behind it, where you could set these properties formally. But for the moment, I believe that would be overkill. 

> I'm about ready to merge this. 

Cool!  :-D

_Mark

markmaker:
Wow, thanks Jason, that was quick!
_Mark

markmaker:
**NOTE**: this comment is just repository for the images used to document the ReferencePushPullFeeder Wiki page. To understand, please consult the [ReferencePushPullFeeder Wiki page](https://github.com/openpnp/openpnp/wiki/ReferencePushPullFeeder). 
___
grafik
grafik
grafik
grafik
grafik
(more to come)

uncle-buzz:
I found an issue with the number of faces of the friction wheel and the spool left who uses the number of wings instead of the number of faces, so creating a friction wheel with 7 wings and 6 faces will build a spool left with 7 faces, I replace the friction_wheel variable with the hex_face at line 3018 and the build is now conform, sorry I don"t know how to propose a PR :
`                   union() {
                        translate([0, 0, -e]) {
                            beveled_extrude(height=spool_wall_left+2*e, bevel=bevel_z, angle=135) {
                                circle_p(d=friction_hex_diameter,
                                    $fn=hex_face);
                            }
                        }
                    }`
instrad of 
`                   union() {
                        translate([0, 0, -e]) {
                            beveled_extrude(height=spool_wall_left+2*e, bevel=bevel_z, angle=135) {
                                circle_p(d=friction_hex_diameter,
                                    $fn=friction_wings);
                            }
                        }
                    }`

PR #989, opened 2020-04-19
Fix/Z-Probe reading conversion error handling

markmaker:
# Description
User might create and set a Z Probe actuator just to see what happens (I did). Much later your Location Button Panel Capture Camera Coordinates Button will "inexplicably" stop working i.e. a cryptic NullPointerException stack trace is displayed.
grafik

Reason: If you leave an Actuator's GCodeDriver ACTUATOR_READ_REGEX empty, it will return a null String on read. 

This Bugfix adds an Exception message pointing to the Z probe and conversion error.

In case of null:
grafik

In case the actuator return a non-number:
grafik

# Justification
See also:
https://groups.google.com/d/msg/openpnp/Ew7drDA3J60/kwVUS_FnAwAJ

# Instructions for Use
Steps to reproduce: Using GCodeDriver add a fresh Actuator as the Z Probe. Press one of the capture buttons:
grafik

# Implementation Details
1. I tested this with the NullDriver (I'm currently not at the machine). The NullDriver will always return a random number, therefore I used the debugger to insert a null/non-number reading. 

### DIFF
--- a/src/main/java/org/openpnp/util/Cycles.java
+++ b/src/main/java/org/openpnp/util/Cycles.java
@@ -31,7 +31,15 @@ public class Cycles {
         Location lOld = zProbe.getLocation();
         Location lz = zProbe.getLocation();
         MovableUtils.moveToLocationAtSafeZ(zProbe, l);
-        double z = Double.parseDouble(zProbe.read());
+        double z;
+        String reading = zProbe.read();
+        try {
+            z = Double.parseDouble(reading);
+        }
+        catch (Exception e) {
+            throw new Exception("Head "+Configuration.get().getMachine().getDefaultHead().getName()+" Z Probe "+zProbe.getName()
+            +" conversion failed ("+reading+")", e);
+        }
         lz = lz.add(new Location(LengthUnit.Millimeters, 0, 0, z, 0));
         MovableUtils.moveToLocationAtSafeZ(zProbe, lOld);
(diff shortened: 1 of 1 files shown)


vonnieda:
Hi @markmaker - this looks good, but I think it may also be solved by what I was working on in https://github.com/openpnp/openpnp/commit/45bf82958afb57c4da4dbd430fe0b9f64cd8cd83. Would you be willing to take a look at that and see if you think it will address the problem, and maybe merge it? At the time I was working on it I had not yet fully tested it so I didn't merge it.

I'll merge this for now, and we can revisit if the larger scale fix takes care of the problem.

markmaker:
Hi @vonnieda 

I agree that your larger scale fix will cover this too and more. I've since seen that this null case may happen in other places such as vacuum level reading. 

The fix in this PR could then be reverted. Or maybe not, because capturing camera coordinates is something every beginner needs a lot and extra feedback is valuable as it may still apply, e.g. when the regex is faulty and extracts something that is not a number. Some readouts might also have non-number cases, if I'm not mistaken, Smoothie sometimes prints "inf" when it thinks a sensor is unconnected. 

_Mark

vonnieda:
I went ahead and merged this, along with a new framework for running
GcodeDriver tests against a TCP based GcodeServer, and some tests for the
new functionality.

Jason

PR #991, opened 2020-04-21
Feature/Advanced Vacuum Checks

markmaker:
# Description

This PR adds advanced methods of part-on/part-off vacuum testing to OpenPNP, while keeping the previous simple functionality available. 

## Feature List

1. Easier setup with Measurement Method combo box. Detection can be switched off without having to overwrite your values.
2. The method combo box controls the visible fields in the Wizard, so it can be easy or it can be advanced, as users like it. 
3. New "Difference" method allows you to specify the permissible _change_ in the vacuum level, in addition to the absolute level.
4. Therefore, much finer nuances in changes of the vacuum level can be discriminated, which is essential for the very small nozzle tips with high air flow resistance and/or strong pumps and/or vacuum reservoir systems and/or multi-nozzle systems with crosstalk.
6. The vacuum level can be monitored/polled while picking or placing a part, so OpenPNP can go on, as soon as the desired buildup/decay level is reached. The former dwell time becomes a timeout. 
7. Much faster picks and places result in the majority of cases, while gaining robustness for the few exceptions. The timeout/dwell time can now be increased without _general_ speed penalty. The wait will only be incurred in the rare worst case.
8. Graphical diagnostics allow a much more informed approach. How fast do the vacuum levels build up or decay? How fast is your sensor? How fast is OpenPNP in querying the controller? No more guesswork! No more browsing in the endless log. 
9. I wanted to hook up the oscilloscope, but then thought, why not use OpenPNP instead?

grafik

# Justification

Following [some discussions in the group](https://groups.google.com/d/msg/openpnp/gL7uEjKmLzU/oOaAlRBdAgAJ) and negative 0402 tests on my machine, and some long thinking about why the detection actually works (and why not), it appears that the current absolute pressure level part-on/part-off vacuum testing method will not work with fast operation and/or small nozzle tips and/or strong pumps and/or vacuum reservoirs and/or multiple nozzles with crosstalk. The baseline of the pressure is simply too varied, the difference between the part obstruction or not obstructing a fine nozzle tip is too small. 

I did my best to try and ensure that the old settings will be ported over and if people don't like the new way, it should work as before. 

A lot of thought and work has gone into reducing the changes in existing core and reference classes to a minimum and to keeping the outer layers (JobProcessor, spi and model interfaces) almost untouched. Most remaining changes are additive, either truly new classes, methods and fields or high-level ifs in methods to separate new from old ways. Adding a new Nozzle subclass was evaluated but found impractical, frankly not least by being convinced of the general usefulness of the new implementation for many users. 

However, this surely needs some thorough review, and I am prepared to make changes as needed. 

The before pick part-off test needs to be rethought, IMHO. Comments in the code.  

The way I implemented the check options (org.openpnp.spi.Nozzle.PartOnStep and org.openpnp.spi.Nozzle.PartOffStep) is just a proposal. I'm awaiting guidance. 

# Instructions for Use

The user interface complexity scales with the Measurement Method, dynamically:
From "None"...
grafik
to "Absolute" 
grafik
to "Difference"
grafik

As you see in the last image, the settings are two folded for Part On (after pick) and Part Off (after place) sensing. 

The "Establish Level?" checkboxes enable life monitoring of the pressure while the nozzle tip is pressed down on the part, either while picking or placing. As soon as the level reaches the target Vacuum Range, OpenPNP can go on and lift the nozzle. This is also needed if you want the graph. 

With the "Perform Checks?" checkboxes, you can choose when to check for Part On/Part Off. [The need for these was also concluded in the group](https://groups.google.com/d/msg/openpnp/gL7uEjKmLzU/M87TLphqAgAJ). 

The "Vacuum Range" settings are the absolute vacuum readings you want to achieve. In Absolute mode the are the only criteria for the checks.

The "Last Reading" will be set after a pick or place has taken place. This will be the value that either satisfied or violated the last absolute check. 

In Difference mode you also get a Difference Range to specify the allowed _change_ in pressure, as compared to the baseline measurement. 

In Part On checks, the baseline is the pressure that was reached before lifting the nozzle. A fluctuating pump pressure, due to hysteresis regulation or crosstalk with a second nozzle, or a recent dip in the pump pressure due to part loss, will be compensated. 

In Part Off checks, the baseline is the pressure that is measured right before opening the valve for the check. 

Again a "Last Reading" is provided to see the result of the last check.  

The "Probing time" is a new separate dwell time for the Part Off check. The pick dwell time can now be different for obvious good reasons. On my machine the probe time must be quite small to get (somewhat) conclusive results. 

In the following image you see the both "Part Off" checks enabled, the first might be usable, because the rise in pressure in the first 40ms is clearly a reaction to the valve opening. But the second check won't work, as the checks are too close together and the pressure rise is still very much influenced by the first check. 

grafik

As you can see, the graphical diagnostic (some repetitions of it) can also tell you real quick when something _won't_ work and you can make an informed decision to give it up. ;-) 

# Implementation Details
1. This was Job-Tested on my machine. **But this obviously needs to be tested by more people and machines.** 
3. Changes in the `org.openpnp.spi` or `org.openpnp.model` packages (Nozzle) were needed to enable/disable the different JobProcessor checks. 

### DIFF
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -557,7 +557,7 @@ public class FeedersPanel extends JPanel implements WizardContainer {
                 feeder.postPick(nozzle);
 
                 // Perform the vacuum check, if enabled.
-                if (nozzle.isPartOnEnabled()) {
+                if (nozzle.isPartOnEnabled(Nozzle.PartOnStep.AfterPick)) {
                     if(!nozzle.isPartOn()) {
                         throw new JobProcessorException(nozzle, "No part detected.");
                     }
--- a/src/main/java/org/openpnp/gui/components/SimpleGraphView.java
+++ b/src/main/java/org/openpnp/gui/components/SimpleGraphView.java
@@ -1,70 +1,190 @@
+/*
+ * Copyright (C) 2020 <mark@makr.zone>
+ * inspired and based on work by
+ * Copyright (C) 2011 Jason von Nieda <jason@vonnieda.org>
+ * 
+ * This file is part of OpenPnP.
+ * 
+ * OpenPnP is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ * 
+ * OpenPnP is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public License along with OpenPnP. If not, see
+ * <http://www.gnu.org/licenses/>.
+ * 
+ * For more information about OpenPnP visit http://openpnp.org
+ */
+
 package org.openpnp.gui.components;
 
 import java.awt.Color;
 import java.awt.Dimension;
+import java.awt.Font;
+import java.awt.FontMetrics;
 import java.awt.Graphics;
 import java.awt.Graphics2D;
 import java.awt.RenderingHints;
-import java.util.Map;
+import java.awt.geom.AffineTransform;
+import java.awt.geom.Point2D;
+import java.awt.geom.Rectangle2D;
+import java.math.RoundingMode;
+import java.text.DecimalFormat;
+import java.util.List;
 
 import javax.swing.JComponent;
 
-import org.openpnp.machine.reference.ReferenceNozzleTip;
+import org.openpnp.util.SimpleGraph;
+import org.openpnp.util.SimpleGraph.DataRow;
+import org.openpnp.util.SimpleGraph.DataScale;
 
+@SuppressWarnings("serial")
 public class SimpleGraphView extends JComponent {  
-    
-    private Map<Double, Double> graph;
-    private final Dimension PREF_SIZE = new Dimension(100, 50);
-    private ReferenceNozzleTip nozzleTip;
-    
-    public SimpleGraphView(ReferenceNozzleTip nozzleTip) {
-        this.nozzleTip = nozzleTip; 
-        //this.setBorder(new TitledBorder(UIManager.getBorder("TitledBorder.border"), "Graph", TitledBorder.LEADING, TitledBorder.TOP, null, new Color(0, 0, 0)));
-    }
 
-    public Map<Double, Double> getGraph() {
-        return graph;
+    private SimpleGraph graph;
+    private final Dimension PREF_SIZE = new Dimension(100, 100);
+
+    public SimpleGraphView() {
     }
-    public void setGraph(Map<Double, Double> graph) {
+    public SimpleGraphView(SimpleGraph graph) {
         this.graph = graph;
     }
-    public void paintComponent(Graphics g)
-    {
+
+    public SimpleGraph getGraph() {
+        return graph;
+    }
+    public void setGraph(SimpleGraph graph) {
+        this.graph = graph;
+        repaint();
+    }
+    protected String formatNumber(double y) {
+        // format numbers without necessary digits (incredibly difficult in Java) 
+        DecimalFormat format = new DecimalFormat("0.####"); // Choose the number of decimal places to work with in case they are different than zero and zero value will be removed
+        format.setRoundingMode(RoundingMode.HALF_UP); // choose your Rounding Mode
+        return format.format(y);
+    }
+    public void paintComponent(Graphics g) {
         super.paintComponent(g);
         Graphics2D g2d = (Graphics2D) g;
         g2d.setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON);
-        Color color1 = new Color(225, 0, 0);
+        
+        Font font = getFont();
+        FontMetrics dfm = g2d.getFontMetrics(font);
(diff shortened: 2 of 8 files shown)


vonnieda:
Merging this to the test branch so others can download and try it.

markmaker:
Hi Jason

Thanks for making this available for general testing, so quickly. 

I'm still unsure with git/GitHub... if I have new commits for this, how do I push them? Do I just create a new PR, based on my local branch? Or do I need to rename the PR branch, so this does not get confused? Do I need to specifically target this for `openpnp:test` or will you do it on the receiving end?

Thanks,
Mark

vonnieda:
I *think* that if you push more commits to `markmaker:feature/advanced-vacuum-checks` then this PR will see them, and we might be able to merge them via Github, but I'm honestly not sure. Try pushing a test commit and we'll see what happens. If that doesn't work, you can always just create a new PR. For now, go ahead and target test since that saves a step, but it's trivial to change from one target branch to another on my end, too, so either way is fine.

vonnieda:
This has now been merged from test to develop.

markmaker:
Thanks @vonnieda!

doppelgrau:
@markmaker just to be sure: for PartOff testing while the vacuum is still active is not supported?
Reason for me asking: with my setup there seem 20-30ms between vacuum off and the time the sensor is red. so the value is not very stable (and so it is hard to distinguish between a free nozzle and a part that is partially stuck but still allowing more air than a good pick)

vonnieda:
@markmaker Could you please update the wiki at https://github.com/openpnp/openpnp/wiki/Setup-and-Calibration%3A-Vacuum-Sensing for this feature?

markmaker:
@doppelgrau sorry, I'm not sure I understand what you mean with "while the vacuum is still active".

After lifting the nozzle up, the vacuum is quickly switched on and then back off and then the level is measured. I've added a second wait time to allow for sensor reading lag.

@vonnieda yes the Wiki entries (also for Camera Settle) are planned/in draft. I's just a matter of when I'm at the machine, as I need to be there to create new screenshots and vacuum graphs with the latest version. I'd like to do it all together with the new motion testing. Probably Wednesday.

_Mark

doppelgrau:
The values are way lower than reading manually and fluctuating, like it is read after vacuum is switched off or during vacuum switch on.

markmaker:
Yes, that's on purpose, the idea is to measure how far it raised, because with the very fine nozzle tips you have almost no difference once the vacuum level has established. You can influence it with the times.

_Mark

doppelgrau:
I hope it gets clearer when the wiki is done, but currently I have a problem detecting stuck parts that doesn't block the nozzle completely. There the the vacuum level "stabilized" is different enough, but the wildly fluctuation values I get currently do not allow me to detect them.

markmaker:
Hi @doppelgrau 

I'll do the Wiki ASAP. 

I thought this would be more self-explanatory: before my change, it used the pick dwell time as a part off check time. The migratory code cannot fully migrate this as it was the combined Nozzle and Nozzle Tip pick dwell time and while migration the Nozzle Tip I don't know which Nozzle it will be, so I just take the default Nozzle: 

https://github.com/markmaker/openpnp/blob/bd3bee55ed875478fb0a57e8aa0be3e1240929b9/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTip.java#L152-L175

This may not have worked for your migration. 

The point is, this was a very long dwell (slow!!). If you set a long Valve open value again, it will work as before. 

The screenshot shows how this is intended to work with my 0402 nozzle tip. Only a very short pulse of 10ms is given and 100ms wait for signal delay. Then the bump is measured as a **differential**. I had much better results than with the saturated measurement. I said "better" but they are still not reliable enough. I guess I will have to disable 0402 Part-Off detection and rely on Aligment (or pure luck).

grafik

But having said that, it is important to note that nothing should have changed for those who don't want the differential measurement. You can set the absolute method and the former long dwell time. 

Ok, strictly speaking, you can no longer make a difference between two nozzles. i.e. you need to add the former nozzle pick dwell time to the former nozzle tip pick dwell time. But I don't see how differentiating between nozzles would make any sense. 

If this still acts differently than before, please report. 

_Mark

doppelgrau:
image

My problem with 150/0 in absolute mode I'd expect that It reads the value during vacuum on (about 2000 on my system), not the observed 500-1100. See attached picture.
My guess is, that the delay between regex ok of switching off and the reading is too large, I'd need a reading while vacuum is still on, not some (unknown) time already off.

markmaker:
Hi @doppelgrau 

Ah, I see. In deed, I introduced a behavior change with the latest change. It isn't noticable with a faster connection (I get <3ms round trip), and my signal lags behind a bit anyway, so I missed it in my own testing, but your connection seems to be quite slow (guessing ~15ms round trip?) and the signal very fast to react, so this matters.

Sorry about that. Will correct this ASAP.

_Mark

markmaker:
Btw. as you seemed concerned about the Backlash Compensation speed in #998, you might want to have a look at your connection speed.

Workaround: put Backlash move and final move on the same G-code line. Most controllers support multiple commands on the same line but OpenPnP will then only have to do one communications round-trip.
Ref: The example here uses two lines:
https://github.com/openpnp/openpnp/wiki/GcodeDriver#backlash-compensation

_Mark

doppelgrau:
Thanks, that's a good hint. (Both, taking a look on the speed, but that seems fine with 115200, tinyG does not support much more)

markmaker:
This is a repository for images for the Wiki, Please do not delete. Read the WIki to understand what they mean:
https://github.com/openpnp/openpnp/wiki/Setup-and-Calibration%3A-Vacuum-Sensing
______
grafik

grafik

grafik
grafik

grafik

grafik

PR #993, opened 2020-04-24
Feature/dynamic safe z

doppelgrau:
# Description
Enables a dynamic safeZ computation for a nozzle.

# Justification
Saves time on all machines with a slow Z axis. Allows to lower the safeZ to a value where the nozzle will never collide with any part on the table (including the PCB with parts). If a part is attached the nozzle will be raised, until the bottom of the part is at safeZ (if possible, nozzle can never be raised above 0). So a collision even with large parts on the nozzle will be avoided.

# Instructions for Use
Enable the "Dynamic Safe Z" in the Nozzle setup, fill in correct height information in the parts and set "safe Z" to a value where the nozzle will never collide with anything mounted on the table.

# Implementation Details

Tested in the simulator with different part heights.

yes.

no changes

passes.


vonnieda:
Nice work @doppelgrau - I think this will be a fun feature to experiment with!

PR #995, opened 2020-04-26
Feature/Advanced Camera Settle

markmaker:
# Description
## What is Settling the Camera?
Before a Computer Vision Operation can be performed, the camera must be settled, i.e. the machine should have stopped, all vibrations ceased and the camera and communications lag passed. A clean, stable image is needed. 

The simple approach is to just wait a moment. However, in order to cover the worst case, a rather long wait time must be chosen. The machine will be significantly slowed down. 

A better way was developed by @vonnieda, the so-called Autosettle. It compares frame after frame from the camera and when the last two are sufficiently similar, the settling is concluded. The feature was a bit hidden, as it was activated by providing a negative settle time, which then served as the maximum allowed pixel difference between the last two frames. 

## What is new in this PR?
Camera Autosettle is now exposed on the GUI, it was significantly expanded and graphical diagnostics are provided: 

AdvancedCameraSettleSmall
([Larger version](https://makr.zone/AdvancedCameraSettle.gif))

At the same time, the easy way still works, the GUI remains simple until you change the Settle Method to one of the advanced methods:
  
grafik

A commit handler will upgrade existing `machine.xml` and the machine should behave as before. 

## Ideas and Features

- The settle threshold can be combined with a timeout, so the camera will always settle, even if the threshold is never reached (due to noise in the camera image, etc.). The camera will no longer hang.
- Like in electronics, Debouncing makes sure a stable image is truly reached. In tests, it turned out that the machine vibrates strongly, i.e. the images swings back and forth. Surprisingly often, two consecutive frames look the same just by accident (interference frame-rate vs. vibration). An early test revealed that in my 60p camera the nozzle tip was not yet visible/in focus for the first two frames and again the Autosettle was happy with the blurred images not differing that much. Excessive motion blur from the down-looking camera caused the same (I can no longer reproduce the last two, but it might still happen with high latency cameras). 
- The settling can be made Color Sensitive for use cases where color patterns are present, but contrasts are low. 
- The settling can be made Edge Sensitive, so rather than comparing the images directly, it compares gradients/edges. This is useful when you have large (out-of-focus) dark and bright areas in your image, but the contrast in your object is low. 
- Contrast (i.e. dynamic range) can be enhanced: Autosettle has to cope with both harsh contrasted images (such as glaring metal parts on black background) and low-contrast objects like the dark-in-dark black Samsung CP40 nozzle tips, or bright-in-bright paper cut tape strips on typical whitish double-sided adhesive tape. Remember: Auto-exposure must be switched off for stable pipeline operations. Autosettle can compensate, so the same threshold value works across those scenarios. 
- The image can be Denoised, i.e. the level of detail can be reduced to get a more reasonable settling precision. Camera sensor noise, compression artifacts and a benign remainder of vibrations will be smoothed out. 
- The sensitivity can be restricted to a circular center area of the camera. Areas with uncontrolled levels of lighting (reflections, overhead light, etc) and other disruptions in the periphery of the camera's field of view, will be ignored. 
- The Graphical Diagnostics can be switched on to record the last settle frames and plot a graph with the settle data. You can scratch over the graph to watch the recorded frames. A heat map for motion is overlaid:
grafik
- A test buttons panel is provided to generate movements, immediately followed by camera settling. Note, drastic motion pictured above is for show, and will never happen in real-life OpenPnP :-)
grafik
- Extensive Tooltips have been provided.  

EDIT: Note, the diagnostics add minimal processing burden during settling, as the heat map generation etc. is postponed until the settle is complete. If you switch on _both_ Diagnostics _and_ Log level "Debug" or "Trace", all the frames are also written to disk. 

With reasonable settings, I have occasional frames dropped on my 60p ELP cam (see Image), but rarely on the more usual 30p. 

# Justification
See the group discussion:
https://groups.google.com/d/msg/openpnp/jnvon8elGzI/VhDvC8KnDQAJ

I'd like to submit this PR to get testing going. Some further Discussion might be needed alongside. 

# Instructions for Use

Will follow. 

# Implementation Details
1. I tested the PR on my machine, trying hard to test all the combinations. However, I propose a merge into `openpnp:test`.
2. Did follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style),
4. Sucessful `mvn test` before submitting the Pull Request.


vonnieda:
Hey @markmaker - I think this looks really good, and I'm open to merging it directly to develop if you are happy with it. Looks like the merge of the advanced vacuum checks caused a conflict, so if you can resolve that I'll go ahead and merge.

markmaker:
Hi @vonnieda 

Thank you, the confidence is much appreciated. ;-)

I resolved the conflicts, which btw. was an interesting study of the surprising limits of three-way-merging. It go it completely and utterly wrong. ;-) 

I had all conflicts resolved locally and then pushed it, but GitHub still complained about some conflict. Then I pressed a button called "Resolve Conflicts" and nothing happened, it just went away, seemingly without even a trace. Don't know what that was. The branches now show no diffs.

SimpleGraph has evolved a bit in the meantime, most notably I had to switch from Milliseconds to Nanoseconds as a time basis, because I tested with 120 fps camera settings (and yes, its works flawlessly, the settle code now uses almost no extra processing time, even with ambitious settings). As a consequence of Nanotime, I had to adapt the vacuum sensing SimpleGraph time basis as well.

Then I performed a big Job integration test of #991 and #995 with some finishing touches including a small CHANGES.md entry,

**For me it's ready.**

Thanks again,
_Mark

netzmark:
Mark sorry that I still haven't a time to test your new vacuum (nor 2.0 at all), but it could be good to have there a mark on the graph showing the moment when the vacuum valve was commanded to be turned on. I suspect there is a time when value stays at 0 until starts rising. Could be interresting to know how long is "nothing happens" time.
Maybe it is shown there already... ;).

markmaker:
> Mark sorry that I still haven't a time to test your new vacuum (nor 2.0 at all), but it could be good to have there a mark on the graph showing the moment when the vacuum valve was commanded to be turned on. I suspect there is a time when value stays at 0 until starts rising. Could be interresting to know how long is "nothing happens" time.
> Maybe it is shown there already... ;).

This should be in #991 :-)

Yes it is already shown. See the blue line called "Valve" going up and down: 

grafik

vonnieda:
@markmaker Has the information here made it into the Wiki anywhere yet? If not, could you please update https://github.com/openpnp/openpnp/wiki/Setup-and-Calibration%3A-General-Camera-Setup#settle-time?

Thanks,
Jason

markmaker:
This is a container for the images for the Wiki. Please do not delete. Read the Wiki page to understand what they mean.
https://github.com/openpnp/openpnp/wiki/Camera-Settling
_____
grafik

grafik

grafik

grafik

grafik

grafik

grafik

grafik

markmaker:
AdvancedCameraSettleSmall

PR #996, opened 2020-04-26
Change modality of ActuatorControlDialog to MODELESS

electronick-de:
# Description
Changes the behaviour of the small actuator control windows (dialogs). They will not force the user to close them before doing something else.

# Justification
With the modality set to MODELESS the actuator control dialogs may stay opened while doing other tasks in the GUI, i.e. jogging the axes etc. It is primarily usefull when working in manual mode or when testing/calibrating the machine.

# Instructions for Use
Switch to the "Actuators" tab, click on any configured actuator, switch to any other tab, i.e. "Jog". Now you can jog the machine and give commands to actuators simultaneousely and easier than before.

# Implementation Details
Maven test succeeded.
Actuator control dialogs stay visible, don't hold the focus.

### DIFF
--- a/src/main/java/org/openpnp/gui/ActuatorControlDialog.java
+++ b/src/main/java/org/openpnp/gui/ActuatorControlDialog.java
@@ -47,6 +47,8 @@ public class ActuatorControlDialog extends JDialog {
                 FormSpecs.RELATED_GAP_ROWSPEC,
                 FormSpecs.DEFAULT_ROWSPEC,}));
         
+        setModalityType(JDialog.ModalityType.MODELESS);
+        
         JLabel lblBoolean = new JLabel("Set Boolean Value");
         getContentPane().add(lblBoolean, "2, 2");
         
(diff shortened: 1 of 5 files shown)


vonnieda:
I'm okay with giving this a try, but I wonder if users will miss the previous behavior. I think it's worth merging and if there are concerns we can add a checkbox to the window of "Pin Window" or something. Anyone else have thoughts?

markmaker:
I think it should check that the actuator's dialog is not already open, otherwise just focus the opened one instead of opening a second one. 

Perhaps store and manage the dialogs in a HashMap by actuator?

Also what happens if I delete an actuator that has it's dialog open? What if I then press one of the dialog's buttons? 

I know a garbage collected language like Java can cope with many such abuses, but still...

I was envisaging a different solution, where you could individually configure the buttons/fields of the actuators to appear permanently in the "Special" tab of the machine controls. ;-)

_Mark

electronick-de:
> I think it should check that the actuator's dialog is not already open, otherwise just focus the opened one instead of opening a second one.

This is a good point and needs rework.

"Special" tab would lead away from what was initially intended. This mod was made for people who use OpenPnP in manual mode. I mean as a high tech tweezers. No programming, no feeders. Just put a device on the table, jog the camera over it, actuate the pick actuator. Then move over the PCB and actuate the place actuator. With modal dialogs one needs to close them, change tab, jog, change tab, open dialog.
For this scenario two additional buttons "pick" and "place" in the "Jog" tab would be enough. But maybe, there is another scenario or other actuators that someone needs to have quick available?

markmaker:
> "Special" tab would lead away from what was initially intended. 

Agree.

PR #1001, opened 2020-05-05
Flush response buffer before sending a command and fix backlash comp

tonyluken:
# Description
This change forces the GcodeDriver sendCommand to flush the response queue before sending a new command to the controller.  In some cases, prior responses could be queued up and be erroneously interpreted as command completions before the command was actually completed by the controller causing bad things to happen (see the discussion at https://groups.google.com/forum/#!topic/openpnp/Hqo05GerezE).  The Homing regex and Move regex waiting loops were updated to call sendCommandNoFlush which doesn't flush the queue so that responses that the loops are waiting for do not get flushed.  Checking was also added to both loops to throw an exception in the event of a controller error.

This change also fixes backlash compensation that was broken by PR 998 which accidentally set all axis backlash values to that of the x-axis. 

# Justification
See the discussion at https://groups.google.com/forum/#!topic/openpnp/Hqo05GerezE

# Instructions for Use
Flushing of the queue is mainly for controllers that use the MOVE_TO_COMPLETE_REGEX to detect end of moves (like TinyG) .  Note that this fix does not completely eliminate the possibility of a spurious controller response being placed in the queue after it is flushed and before the command is sent but it should greatly reduce the possibility of it occurring.  For TinyG at least, it is also recommended to set the status interval to a very long duration, i.e., set $si=4e9 in the CONNECT_COMMAND to further reduce the possibility of spurious responses being placed in the queue.

# Implementation Details
**1. How did you test the change? Be descriptive.** Found a scenario that was fairly regularly causing an erroneous detection of a move completion before was actually completed (vision processing for a strip feeder was starting before the camera arrived over the strip).  After making this fix, the same scenario was run many times without a problem.
**3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.**  No changes were made in either of these packages.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
@@ -300,6 +300,7 @@ public class GcodeDriver extends AbstractReferenceDriver implements Named, Runna
 
         // Check home complete response against user's regex
         String homeCompleteRegex = getCommand(null, CommandType.HOME_COMPLETE_REGEX);
+        String commandErrorRegex = getCommand(null, CommandType.COMMAND_ERROR_REGEX);
         if (homeCompleteRegex != null) {
             if (timeout == -1) {
                 timeout = Long.MAX_VALUE;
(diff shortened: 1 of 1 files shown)


vonnieda:
This looks good @tonyluken. Thanks for the improvements, and especially spotting that backlash error!

PR #1006, opened 2020-05-14
Feature/fix bug nozzle tip calibration

tonyluken:
# Description
Fixes a bug in nozzle tip calibration that resultes in NaNs propagating into the nozzle location.  This was only detected when running with the NullDriver and simulated up camera so that the measured runout (which should be theoretically zero) was dominated by noise.  This resulted in a square root of an extremely small but negative scale factor.  The fix is to limit the scale factors to be no less than zero before taking the square root.

# Justification
Fixes a bug that is apparent during certain testing.

# Instructions for Use
No special instructions.

# Implementation Details
1. How did you test the change? **Ran the same scenario that was causing the problem and verified that NaNs no longer occur.**
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?  **Yes.**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.  **No changes were made to these packages.**
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.  **Maven tests were run and passed.**


vonnieda:
Thanks @tonyluken!

PR #1007, opened 2020-05-14
Fix/release all mat

doppelgrau:
# Description
Made sure for all temporary Mat* Objects in the Stages MaskHsv and Normalize  the release() function ist called to avoid memory leaks.

# Justification
The Mat-Objects are suspected to create memory leaks. So try to avoid that.

# Instructions for Use
Should change nothing for the user.

# Implementation Details
1. How did you test the change? - Run it on my machine. Hard to tell if really an improvement, but no downsides noticed.
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. - No
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. - Succeed


vonnieda:
Looks good, thanks @doppelgrau!

PR #1008, opened 2020-05-14
Feature/new vision stages

doppelgrau:
# Description

- Added a "Prefix" for the filename to the CV-Pipeline stages ReadPartTemplateImage and WritePartTemplateImage
- Added a CV-Pipeline stage that selects a single rectangle from a list of rectangles.
- Addes a CV-Pipeline stage that checks multiple rotated rectangles if the match a given Template

# Justification
## Filename prefix (ReadPartTemplateImage and WritePartTemplateImage)

Allows for a single part to have multiple templates (e.g. for top and bottom vision). Default empty string, so no change in the behavior.

## SelectSingleRectangle

Bottom vision expects a single rotated rectangle as result. Many stages has as the output format List <rotatedRectangles> (e.g. MatchPart(s)Template). This stage can be used for a simple "type conversion" by selection a entry.

## MatchPartsTemplate

Mostly the same like MatchPartTemplate. Besides some cleanup (e.g. better alignment of the rotated rectangles if the template is not in portrait orientation) main difference is, that this stage processes all rotated rectangles from the input. Output can also now be multiple matches. Useful if not every rectangle that fits the size is a sure match where only the orientation is relevant. (E.g. multiple different parts in the vision, or parts laying upside down)

# Instructions for Use

## Filename prefix (ReadPartTemplateImage and WritePartTemplateImage)

Enter a string in the "prefix" parameter, this will be prepended during load/save of templates.

## SelectSingleRectangle

Select the stage where to take the rectangles from, and which elemt (0=first, 1=second -1=last).

## MatchPartsTemplate

Same as MatchPartTemplate.

# Implementation Details
1. How did you test the change? - Run them on my machine, testing the HeapFeeder branch.
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. - not changed
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. - Succeed

### DIFF
--- a/src/main/java/org/openpnp/gui/wizards/CameraVisionConfigurationWizard.java
+++ b/src/main/java/org/openpnp/gui/wizards/CameraVisionConfigurationWizard.java
@@ -22,6 +22,8 @@ package org.openpnp.gui.wizards;
 import java.awt.Font;
 import java.awt.HeadlessException;
 import java.awt.event.ActionEvent;
+import java.awt.event.ItemEvent;
+import java.awt.event.ItemListener;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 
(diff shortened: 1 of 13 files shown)


doppelgrau:
Added an other change: in the Read/WritePartTemplate I also try now to get the part direct, not only feeder, and in the bottom vision I set the part in the enviroment.
So Vision is possible with the bottom vision. (Use case: detect the orientation of polarized parts, that have good marks on the bottom, but not on the top)

doppelgrau:
After some testing(and a bug fix) it seems to work fine. Only minor change: feed part in the the pipeline during bottom vision, so that can be accessed there.

PR #1010, opened 2020-05-18
Snap the camera jogging angle when holding down the ALT button

pfried:
This lacks information about the board rotation, but as uses generally try to XY align their boards it might still be an improvement. Could also be changed to snapping to 15° angles

# Description
Snaps the visible camera jogging button when holding down the ALT button.

# Justification
I saw the youtube video about the visual jogging and saw Jason struggling with getting the angle right

# Instructions for Use
Use visual jogging and hold down the ALT button. The rotation will snap into multiples of 45°

@vonnieda Feel free to reject. I know that the camera lacks knowledge of the real orientation of the board, but as most machines mount the PCB in some axis aligned fashion this still seems as an improvement to me


vonnieda:
Looks good to me! Nice change!

Issue #843, opened 2019-05-16
FR : manage a list of pipelines definition at a system level and affect them where needed.


JulienDevillers:
# Feature Request
## Describe the Feature
Instead of having a default pipeline and a specific pipeline for each part.
We could have a system that allows has many pipeline settings as the user needs.
Then everywhere where a pipeline is needed (feeder, camera runtout calibration...)
the user can assign one of the predefined pipelines or a specific one.

## How Will It Look?
A new tab in the gui with an incrementable list of named pipelines and the ability to modify any of them.
As I suspect that there won't be a lot of pipelines, maybe the pipeline setup window might be integrated to this tab (using the current camera view(s) would be a nice integration).

For each pipeline, we will be able to see the list of items to which the pipeline is assigned to.
In feeders, it will replace the tape type.
The user will still be able to modify the pipeline in the feeders but this will affect all feeders using this pipeline.

Adding a checkbox to say "this particular feeder is using a specific pipeline or one of the generic pipelines defined in the pipelines tab.

## Why Do We Need It?
1. copy / paste pipelines beetween feeders in not very fun to do and error prone.
3. You never know if 2 feeders are sharing the same pipeline settings.
2. Easier to add new feeder, if an already present feeder uses the same packaging, just need to 
select it in the list. (Generally, we will use (for strip), white, black and clear).

vonnieda:
Agreed - I think this is a good improvement. This concept has been discussed a bit here: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/openpnp/JyZvwIf9A0U/C5lq4VtTAgAJ

One thing I'd change from the above is that I don't think we really need "Adding a checkbox to say "this particular feeder is using a specific pipeline or one of the generic pipelines defined in the pipelines tab.". If a user has a feeder that requires a specific pipeline they can always create one in the global list and assign it. I'm open to arguments on that.

This would be a very large change, so if you are going to go ahead with it I'd suggest breaking it into small chunks, i.e. create the new pipeline manager area (Machine Setup -> Vision -> Pipelines), then convert one class to use it, like Strip Feeder. We'd merge that and let users try it and see what they think before converting everything that uses pipelines.

pfried:
[...]

PR #1013, opened 2020-05-23
WIP: Reusable CVPipelines

pfried:
# Description
Starting the implementation of a reusable CvPipeline configuration system as suggested in openpnp#843

Creating a new class Pipeline
-> includes a CvPipeline
-> includes a Camera
-> is saved to XML with an id and referenced by that

The actual implementations are in ReferenceBottomPipeline and ReferenceStripFeederPipeline as they load a specific default implementation. (ReferenceBottomPipeline is not used but shows the concept of having a bottom down camera)

This solves assigning the same pipeline configuration to multiple feeders. For now this is located in the ReferenceStripFeeder but should be moved up in the hierarchy

This basically also solves the issue with having multiple up cameras since a pipeline is now tied to a camera and thus could be used in job processing

This makes sense to me since the pipeline settings are camera dependent

# Justification
CvPipelines cannot be reused currently, see #843 

# Instructions for Use
Let OpenPnP create  a new machine.xml with the new config
Run OpenPNP
Pipelines are in the machine config under Vision -> Pipelines
Pipelines can be added by using the plus icon on top
Pipelines can then be used within ReferenceStripFeeder

# Implementation Details
Does not brake tests

I hope so

Yeah well not really an option not to touch them, but I was gentle

works

@vonnieda Lets discuss if that makes sense, also the issue of the camera binding (no gui for that yet tough).

### DIFF
--- a/src/main/java/org/openpnp/gui/support/NamedListCellRenderer.java
+++ b/src/main/java/org/openpnp/gui/support/NamedListCellRenderer.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright (C) 2011 Jason von Nieda <jason@vonnieda.org>
+ * 
+ * This file is part of OpenPnP.
+ * 
+ * OpenPnP is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ * 
+ * OpenPnP is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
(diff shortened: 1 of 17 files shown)


JulienDevillers:
Hello @pfried 

I tried your feature, but for now, I'm more interested by bottom vision. As I understand by reading, your instructions, it is only partially implemented. (there's no way to assign a pipeline to a package). Is that right?

pfried:
@JulienDevillers This is not a "working" software but merely a sketched idea. But it should work for ReferenceStripFeeders anyway, but only for those.

pfried:
Unfortunately I wont have time to pursue this further, so I need to close the issue

Issue #1017, opened 2020-06-09
ACTUATOR_READ_WITH_DOUBLE_COMMAND in sub-driver not working


bilsef:
# Problem Report
## Describe the Issue
Using a actuator with "Read With Double" fails if the actuator is serviced by a sub-driver.  The bug was introduced in the Apr 20, 2020 release of Improved Actuator Read Errors.

## Steps to Reproduce
1. You will need a G-Code sub-driver
2. Create a test actuator
3. In the sub-driver assign a ACTUATOR_READ_WITH_DOUBLE_COMMAND to the test actuator
4. In the sub-driver assign a ACTUATOR_READ_REGEX to the test actuator
5. Open the test actuator dialog box
6. Enter a value in Set Double Value
7. Click on Read With Double

## Expected Result
Read Value will display result

## Actual Result
A Error box opens stating Actuator "test" read error: Driver configuration is missing ACTUATOR_READ_COMMAND or ACTUATOR_READ_REGEX.

PR #1018, opened 2020-06-10
Fix for bug #1017

bilsef:
# Description
Fixes bug #1017 where ACTUATOR_READ_WITH_DOUBLE_COMMAND fails for sub-drivers.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. Tested on live machine.
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. n/a
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.  All tests passed


vonnieda:
This looks good @bilsef - although, for bonus points, could you change it so that the error message is specific to the call? i.e. if they passed a double value, specify that the double command is missing.

bilsef:
Good idea Jason.  I will do that.

PR #1021, opened 2020-06-21
allow auto assign known parts

doppelgrau:
# Description
Until now importing a kicad.pos file you had the choice between importing only the positions or import the positions, assign existing parts and if parts does not exist automatically create a new part.
This change allows (default on) assigning only existing parts.

# Justification
I think the use case where only existing parts should be imported is not so uncommon.

# Instructions for Use
Just import the .pos file.
If automatic assignment is not wanted, uncheck the checkbox in the import dialog.
Other check-boxes work unchanged.

# Implementation Details
1. How did you test the change? Unittest and existing .pos
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. No changes
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. ok

### DIFF
--- a/src/main/java/org/openpnp/gui/importer/KicadPosImporter.java
+++ b/src/main/java/org/openpnp/gui/importer/KicadPosImporter.java
@@ -67,7 +67,7 @@ public class KicadPosImporter implements BoardImporter {
         return board;
     }
 
-    static List<Placement> parseFile(File file, Side side, boolean createMissingParts, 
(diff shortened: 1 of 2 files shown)


vonnieda:
Sounds good to me, thanks @doppelgrau!

PR #1022, opened 2020-06-30
Feature/allow full rotation in bottom vision

doppelgrau:
# Description
Introduced a new option for the parts, that the bottom Vision can not only adjust the angle (+-45°) but fully rotate the part (+-180°).
Added a option in the ReferenceBottomVision PartSettings (enum) and made that visible in the Wizzard. Default is "Adjust" => no change in behavior.
Also changed the call of the "angleNorm" function, depending on the setting.

# Justification
In some use cases (LoosePartFeeders, HeapFeeder) the orientation of a part can only be detected using the bottom vision. in that case it must be able to rotate freely if the part is polarized.

# Instructions for Use

in the ReferenceBottomVision Tab of a part select "Rotation = Full" and provide a vision pipeline, that return meaningfull results (e.g. using MatchParts).

# Implementation Details
1. How did you test the change? Yes and an other user.
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. No changes
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. Run fine

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -130,12 +130,18 @@ public class ReferenceBottomVision implements PartAlignment {
                 // the center of the camera to the located part.
                 offsets = VisionUtils.getPixelCenterOffsets(camera, rect.center.x, rect.center.y);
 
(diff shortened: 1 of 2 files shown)


vonnieda:
Looks good to me!

PR #1023, opened 2020-07-16
Add Gstreamer Camera

ian-arkver:
This camera uses a textbox to specify an
arbitrary gstreamer launch pipeline.

Ann appsink is appended to this pipeline and
samples are pulled and fed to openPnP.

The pipeline should end with an explicit capsfilter element.

------

# Description

Adds GstreamerCamera class, based on ReferenceCamera class which allows use of
an arbitrary gstreamer pipeline as a video source. This can be for example a v4l2src,
nvarguscamerasrc, rpicamsrc, rtsp receiver and decoder, media file reader and playbin, etc., etc.

# Justification

Other camera sources are available which can read from v4l2 devices, perhaps even using
gstreamer to do so, however the pipelines used are fixed by the libraries. Although many
offer good controls, none allow arbitrary gstreamer sources to be used. This rules out their
use where a specific, non-standard camera source element is needed.

This camera source might also be useful for testing purposes where a pre-recorded video
sequence could be played back.

# Instructions for Use

GstreamerCamera is added to the list of available camera sources. When added a wizard is
available with a single text box under "Device Settings". This box is used to specify the gstreamer
launch pipeline, in normal gst-launch-1.0 syntax. The pipeline should be capable of producing
video/x-raw,format=BGRx and should finish with a capsfilter element. An appsink is automatically
linked to the end of the pipeline with these caps.

It is recommended to use gst-launch-1.0 to test the pipeline. Once video is correctly working it can
be copy/pasted into openPnP. The pipeline string will be saved in machine.xml.

# Implementation Details

A new dependency is added to pom.xml to bring in org.freedesktop.gstreamer bindings.
A new GstreamerCamera.java and GstreamerCameraConfigurationWizard.java are added.
The new class is added to the list of available camera types in ReferenceMachine.java.

I believe the new files conform to the coding style.
No spi or model changes are introduced.
Maven tests are not affected as they don't use this camera, however they continue to pass.

The camera has been tested on a Jetson Nano using both a MIPI camera with nvarguscamerasrc
and a USB camera with v4l2src.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -194,7 +194,12 @@
         	<groupId>javax.xml.ws</groupId>
         	<artifactId>jaxws-api</artifactId>
         	<version>2.3.0</version>
-    	</dependency>
+		</dependency>
+        <dependency>
+            <groupId>org.freedesktop.gstreamer</groupId>
+            <artifactId>gst1-java-core</artifactId>
+            <version>1.1.0</version>
+        </dependency>
    	</dependencies>
 	<build>
 		<plugins>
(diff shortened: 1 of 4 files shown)


vonnieda:
Looks good @ian-arkver - nice work on your first OpenPnP PR!

ian-arkver:
Thanks @vonneida for the swift review and merge.

PR #1030, opened 2020-08-05
Feature/comms fail catch

ian-arkver:
# Description

With TCP communications and GCodeDriver, if the comms is lost due to the far end closing the socket, openPnP does not recover gracefully and requires a restart. This catches the command write and response read errors and correctly closes the connection. Since this is most likely a fatal error, the machine is disabled.

# Justification

This change is specific to GCodeDriver over TCP comms, so maybe a little specific. The change should be fairly unintrusive though.

# Instructions for Use

If the connection is broken, but then restored, the machine can be restarted and re-homed.

# Implementation Details

Tested using GCodeDriver via local socket to a gcode daemon.
There should be no coding style violations.
No changes to models or spi.
Maven tests do not cover this case, so are not affected.


vonnieda:
@ian-arkver I think it might be better to just disconnect() and then throw an exception so the user can decide what they want to do, but I'll go ahead and merge this and see what others think.

ian-arkver:
OK, thanks Jason. Let me know if you want a follow-up change.

PR #1031, opened 2020-08-06
Fix / Various assorted bugfixes broken out

markmaker:
# Description

In the course of developing and testing the upcoming "Global Axes" PR, I found some bugs and cleaned up some code. These are now broken out in this PR to reduce the size of the upcoming PR.

## Machine Setup Panel

When changing objects in the Machine Setup Panel, the currently opened Wizard Tab is re-selected if the new object has the same class. Due to some race condition, if you click around quickly, this sometimes throws an index out of bounds exception, leaving the Panel in a half crashed way, with redraw issues, permanently. There were various reports in the group, etc.. 

This is fixed and enhanced in this PR: The Panel now remembers which Wizard Tab is selected for which class. Changing between objects will reliably re-select the tab. As it is stored per class, it will now remember the tab even if you visit other classes' objects in between. 

This should solve #961.

## Errored Placements

During (extensive!) testing I was repeatedly irritated by the message "Errored Placements". This message is issued even if there are **none**. This is now remedied. 

## Unnecessary imports

Unnecessary imports in `OpenPnpCaptureCameraConfigurationWizard` were removed. 

## Backlash Compensation correction in BlindsFeeder removed

Realizing that a properly calibrated Backlash Compensation should only take up slack/play of the belt/actuator but should **not** actually physically move the Nozzle (at least not significantly), I removed the corresponding correction in (my own #936) `BlindsFeeder`. 

## Camera Settling lose Ends

Having reworked Camera Settling in #995, I now stumbled over some code that still used the old (positive) settle time in a `Thread.sleep()` to assume a settled camera. If the user selects one of the advanced Settle Methods, the fixed settle time is no longer part of the settings (hidden on the GUI) and may be inadequate. Even before my PR, negative settle times for auto-settling would not have worked. 

Therefore, I removed the settle timeout getter/setter from the `Camera` interface, removed `Thread.sleep()` calls from `ReferenceLeverFeeder` and `ReferenceDragFeeder` and changed the `OpenCvVisionProvider` methods to use `camera.settleAndCapture()` instead of `camera.capture()`. 

Similarly, I changed `SampleJobTest` to properly use the `AbstractCamera` and set the `FixedTime` method and settle time (to 0ms,  the unit test does this to speed up simulation). 

## Firing Property Changes in ReferenceStripFeeder Reference/Last Hole Setters

Missing property change firing led to lost updates if a `ReferenceStripFeeder` is loaded/modified in the Wizard and operated at the same time. This is now fixed. Unfortunately, I don't remember the exact scenario how to reproduce (but it was nasty to diagnose). 

## Formatting DrawTemplateMatches Scores

`DrawTemplateMatches` score numbers were previously printed with full double precision. They are now formatted with reasonable 3 decimal digits. 

# Justification

All these bugs and issues were encountered during rework of the global axis handling and most importantly when testing the GUI and adapting existing and creating new unit test.  

# Instructions for Use

There is no user action needed to enable or control these fixes. Just use the features described above as before and observe.

# Implementation Details
1. Most of these changes were tested using the unit tests or the GUI with the `NullDriver` machine. `BlindsFeeder` changes were tested on the machine (albeit in the context of an early stage of the upcoming PR). The only missing tests are the `ReferenceLeverFeeder` and `ReferenceDragFeeder` camera settling due to not having these on my machine. However, the code changes are simple and I carefully looked at how the pipeline does it. I hope this is acceptable, not least under the premise that `OpenCvVisionProvider` methods are actually deprecated.  
3. Changes in the `org.openpnp.spi.Camera` interface as described.
4. Succesfully ran `mvn test` before submitting the Pull Request.

### DIFF
--- a/src/main/java/org/openpnp/gui/MachineSetupPanel.java
+++ b/src/main/java/org/openpnp/gui/MachineSetupPanel.java
@@ -29,6 +29,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Enumeration;
+import java.util.HashMap;
 import java.util.List;
 import java.util.prefs.Preferences;
 
@@ -90,7 +91,7 @@ public class MachineSetupPanel extends JPanel implements WizardContainer {
      */
     private boolean disableLastSelectedListener = false;
     private Object lastSelectedNode = null;
-    private int lastSelectedTabIndex = 0;
+    private HashMap<Class, Integer> lastSelectedTabIndex = new HashMap<>();
 
     public MachineSetupPanel() {
         setLayout(new BorderLayout(0, 0));
@@ -159,10 +160,12 @@ public class MachineSetupPanel extends JPanel implements WizardContainer {
         
         tabbedPane.addChangeListener(new ChangeListener() {
             public void stateChanged(ChangeEvent e) {
-                if (disableLastSelectedListener) {
-                    return;
+                if (lastSelectedNode != null) {
+                    if (disableLastSelectedListener) {
+                        return;
+                    }
+                    lastSelectedTabIndex.put(lastSelectedNode.getClass(), tabbedPane.getSelectedIndex());
                 }
-                lastSelectedTabIndex = tabbedPane.getSelectedIndex();
             }
         });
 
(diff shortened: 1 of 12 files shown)


vonnieda:
These all look like safe changes to me, and some nice improvements too. Thanks @markmaker!

PR #1040, opened 2020-08-18
Add ability to choose actuators between moves in the tool changer section

cncmachineguy:
# Read This First

If you would like to add a new feature to OpenPnP that changes core or reference functionality,
please create a topic on the [discussion group](http://groups.google.com/group/openpnp) to discuss
the feature first. Also consider if your feature can be implemented as a new subclass of
a Reference class, or an entirely new class, instead of changing or adding to the Reference
class.

will not be accepted.

# Description
Three dropdowns are added to the tool changer GUI. These allow the user to choose (or not) actuators on the machine to fire between moves.

# Justification
There are users with machines that need more than just 4 simple moves to load/unload a nozzle tip. This is an attempt to make that available to machines that need more.

# Instructions for Use
To use it, simply create actuators on the machine that you want to have fire true between the step. Note the actuator will simply actuate (true) for the step. If you are turning a valve on/off, you will need to create an actuator for both. For instance 'valve on' can be one and 'valve off' could be the other. This is different then the normal toggle way we do it now with true/false.

# Implementation Details
I tested this in my install by choosing different actuators and watching the log to verify they are fired.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
@@ -552,8 +552,12 @@ public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMoun
             MovableUtils.moveToLocationAtSafeZ(this, nt.getChangerEndLocation(), speed);
 
             if (changerEnabled) {
-                Logger.debug("{}.unloadNozzleTip(): moveTo Mid Location 2", getName());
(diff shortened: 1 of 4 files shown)


vonnieda:
The only concern I had was if the name is being set directly, or if it's
using the toString() method of the Actuator. If the latter, I'll need to
fix that up. I'm going to take a look at it tonight.

Otherwise it looks great. Nice work!

cncmachineguy:
Please do and let me know.
Thanks Jason :)

PR #1050, opened 2020-08-25
Factorize Communication code

diorcety:
# Description
Factorize serial and tcp serial communication using primitive read/write functions

# Justification
writeLine and readLine are duplicated code which can be factorized using simple abstract read and writeBytes function.

# Implementation Details
Tested over a simple query/reply protocol


vonnieda:
Thanks @diorcety - this looks good to me!

PR #1053, opened 2020-08-27
Feature/OpenPnpCaptureCamera Capture FPS Test

markmaker:
# Description

Adds an FPS test button to the Device Settings Wizard of OpenPnpCaptureCamera.

camera-exposure-and-fps avi

# Justification
It is sometimes useful to know how a specific Camera performs in terms of FPS on OpenPnP. 

I made the Test button in order to evaluate [a new color camera with marvelous 720p@120 fps and global shutter](https://www.aliexpress.com/item/32916630999.html). I just wanted to make sure, the no-name camera actually delivers what the specs promised (it does). 

The button then revealed that FPS are highly exposure dependent. See the [report about the suspected reasons](https://makr.zone/camera-fps-cpu-load-and-lighting-exposure/519/).

This is one more thing, I learned through some added diagnostics (see #995, #991 for other examples). 

# Instructions for Use
Just set the camera Device and Format then press the Test button. The Test button will first apply any changes you made and (re)start the camera in the selected mode. For most accurate measurements, set the Preview FPS to 0. 

The test captures frames as fast as it can. The capturing includes the necessary memory copy and Java `BufferedImage` conversion but no further processing such as applying lens calibration, transforms etc.. This should give you the best possible raw FPS, as obtainable by OpenPnP (i.e. in Java).

The computed FPS is the average achieved over 2 seconds. It will sometimes vary a bit, due to other stuff happening (such as Java garbage collection etc.). On a fast computer it will show you the true device FPS, plus/minus a few frames, at least up to 120 fps (see the animation above). 

The following shows the ELP FullHD camera and its extreme crash to just 2 fps on full exposure: 

camera-exposure-and-fps-elp avi

This clearly shows us that good lighting is needed!

# Implementation Details
1. Tested with the two mentioned cameras. 
4. Successfully ran `mvn test` before submitting the Pull Request.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/camera/OpenPnpCaptureCamera.java
+++ b/src/main/java/org/openpnp/machine/reference/camera/OpenPnpCaptureCamera.java
@@ -318,6 +318,41 @@ public class OpenPnpCaptureCamera extends ReferenceCamera implements Runnable {
         thread.start();
     }
 
+    public double estimateCaptureFps() throws Exception {
+        ensureOpen();
+        if (stream == null || format == null) {
+            throw new Exception("Camera stream not properly initialized."); 
+        }
+        // Start warmup capture timer for 1 second.
+        boolean warmup = true;
+        long t0 = System.currentTimeMillis();
+        long timeout = t0 + 1000;
+        long t1 = 0;
+        int capturedFrames = 0;
+        for (int frames = 0; frames < 120; frames++) {
+            stream.capture();
+            while (!stream.hasNewFrame()) {
+            }
+            t1 = System.currentTimeMillis();
+            capturedFrames++;
+            if (t1 > timeout) {
+                if (warmup) {
+                    // Warmup complete. 
+                    warmup = false;
+                    // Start the real capture timer of 2 seconds.
+                    t0 = System.currentTimeMillis();
+                    timeout = t0 + 2000;
+                    capturedFrames = 0;
+                }
+                else {
+                    break;
+                }
+            }
+        }
+        // Compute the fps.
+        return capturedFrames*1000./(t1-t0);
+    }
+
     private void setPropertiesStream(CaptureStream stream) {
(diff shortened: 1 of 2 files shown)


ian-arkver:
Interesting. I wonder if the low FPS I was seeing on my USB camera was because auto-exposure was cranking the exposure time way up. Now that I've got the camera on the machine with decent lighting and manual exposure set to minimum I must retest it.

Useful gadget Mark.

tritao:
Nice feature, just yesterday I was trying to get some camera FPS measurements, and had to use `openpnp-capture-test` (which does not really measure the Java overhead).

Will be testing this when its merged.

vonnieda:
@markmaker I think this looks good, but I wonder if it could be put on Abstract or ReferenceCamera instead, so that it would work with all camera implementations?

markmaker:
@vonnieda thanks for looking into this.

The Problem is that I want to use `stream.hasNewFrame()` and `stream.capture()` as opposed to the generic `camera.capture()` available in `AbstractCamera`. The lower level `stream` is obviously only available in the `OpenPnpCaptureCamera`.

`camera.capture()` is not guaranteed to capture unique frames (the same frame can be delivered over and over again) and there is no clear indication, when a new frame is available (short of comparing frames).

`camera.capture()` also does too much already, applying lens calibration and transforms, broadcasts frames to CameraPreview etc. so the overhead skews the benchmark. 

In fact, there is already a way to roughly check frame rates based on `camera.capture()` here:
https://github.com/openpnp/openpnp/wiki/Camera-Settling#checking-the-camera-frame-rate

Because of the said overhead, this was not conclusive for the 120 fps camera and even on lower fps cameras you always get irregular measurements, therefore I needed a lower level approach, ergo direct access to `stream`.

Hope this makes sense.

_Mark

vonnieda:
Makes sense, and is reasonable. Thanks for the explanation.

PR #1060, opened 2020-10-04
Feature / Better manual Nozzle Tip Change

markmaker:
# Description

## Intro
For machines that have no automatic nozzle tip changer, one can disable the automatic changer on the nozzle:

Nozzle Tool Changer Settings

OpenPnP will then only use the **Last Location** of the Nozzle Tip Tool Changer setup. A user can just capture a location that is convenient to manually load/unload Nozzle Tips:

Tool Changer Last Location

## Better Unload/Load Messages
This PR improves the usability by naming the nozzle and nozzle tip in the messages. Furthermore, the nozzle is now moved to the **Last Location** and the message shown for _both_ the unload _and_ the load operation (formerly only unload):

Unload

Unload

This now also works, if no nozzle tip was loaded to begin with. 

When these Messages are shown, the current Job or GUI action is aborted. The user needs to resume the Job or re-execute the GUI action (this in itself is not new). However, because the unload / load now shows two separate Messages, a user will have to do this one _more_ time. 

In return, the needed intervention is now very clear, with the nozzle tip and nozzle clearly named in the message. The danger of confusing nozzle tips, nozzles, or a separate unload from a combined unload/load operation is significantly reduced. 

## Usage through the Load and Unload Buttons
As a consequence, this now works with both the GUI buttons (formerly only unload worked):

grafik

## Usage through the Pick Buttons
When used manually from the GUI, the pick buttons
Pick Button
now behave more like the JobProcessor and will now also calibrate a nozzle tip, when manually loaded. 

## Internals
The `PartsPanel` will now use the same refactored `FeederPanels.pickFeeder` function for a more job-like behavior. See also #942 .

The `FeedersPanel.pickFeeder` function now includes explicit nozzle tip calibration as in a job, to make sure calibration is done even after interruption due to manual change. 

Also changed the label of the "Change On Manual Feed?" option to "Change On Manual Pick?", as this is actually more accurate (my own mistake from #942).

Because of indentation, it is recommended to look at the diff with whitespace changes suppressed.

# Justification
See also:
https://groups.google.com/g/openpnp/c/AoED-hR5hTo/m/6gbKnic1BwAJ

# Instructions for Use
See the Description above.

# Implementation Details
1. Tested this in the NullDriver machine. From the GUI and in Jobs. Including different triggers for nozzle tip calibration. Used part incompatible nozzle tips to prompt automatic nozzle tip change both from a Job and from the GUI (using the Change on Manual Pick option).

### DIFF
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -523,62 +523,72 @@ public class FeedersPanel extends JPanel implements WizardContainer {
             UiUtils.submitUiMachineTask(() -> {
                 Feeder feeder = getSelection();
 
-                // Simulate a "one feeder" job, prepare the feeder.
-                if (feeder.getJobPreparationLocation() != null) {
-                    feeder.prepareForJob(true);
-                }
-                feeder.prepareForJob(false);
-
-                // Check the nozzle tip package compatibility.
-                Nozzle nozzle = MainFrame.get().getMachineControls().getSelectedNozzle();
-                org.openpnp.model.Package packag = feeder.getPart().getPackage();
-                if (nozzle.getNozzleTip() == null || 
-                        !packag.getCompatibleNozzleTips().contains(nozzle.getNozzleTip())) {
-                    // Wrong nozzle tip, try find one that works.
-                    boolean resolved = false;
-                    if (nozzle.isNozzleTipChangedOnManualFeed()) {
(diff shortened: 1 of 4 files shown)


markmaker:
Hi @vonnieda 

To clarify, this could go directly into `develop`.

Thanks for having a look if you've got the time (no hurry).

_Mark

vonnieda:
Looks good, thank you @markmaker!

PR #1069, opened 2020-10-21
set list length of combo boxes to 20 items.

JulienDevillers:
# Description
Added more lines on parts combo boxes.

# Justification
It's easier to select with a bigger list.

# Instructions for Use
open a job or a leverFeeder
open the part combobox.

I would have been pleased to use a constant but I did not find any clue on where to put it.


vonnieda:
This looks fine to me - I do wonder if it will behave different on different platforms, but we can see if anyone complains :) Thanks @JulienDevillers!

JulienDevillers:
Ha ha ! 20 Is not that much. It would had been just for me, i.had put 40! :D

PR #1076, opened 2020-11-22
Implemented part count display on the ReferenceStripFeederConfigurationWizard

desertgreg:
# Description
This feature tells the user how many of a the currently selected part are used by the currently loaded job.  It is useful because with strip feeders you need to pull back the tape to uncover enough parts for the job to complete and it can be tedious to go back and forth between panes to add this information up.

# Justification
It saves me a lot of time for people using strip feeders.

# Instructions for Use
See the attached image, next to the part selection combo box, OpenPNP will tell you how many parts are used.

# Implementation Details
1. How did you test the change?  - I added a new label to the wizard and a function to update the label based on the part selected.
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?  Yes
3. No changes to the model were needed.
4. mvn test completes with no errors.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
@@ -64,11 +64,13 @@ import org.openpnp.gui.support.PartsComboBoxModel;
 import org.openpnp.machine.reference.camera.BufferedImageCamera;
 import org.openpnp.machine.reference.feeder.ReferenceStripFeeder;
 import org.openpnp.machine.reference.feeder.ReferenceStripFeeder.TapeType;
+import org.openpnp.model.Board;
 import org.openpnp.model.Configuration;
 import org.openpnp.model.Length;
 import org.openpnp.model.LengthUnit;
 import org.openpnp.model.Location;
 import org.openpnp.model.Part;
+import org.openpnp.model.Placement;
 import org.openpnp.spi.Camera;
 import org.openpnp.util.HslColor;
 import org.openpnp.util.OpenCvUtils;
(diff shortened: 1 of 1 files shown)


vonnieda:
Looks good to me, thanks @desertgreg!

src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
```diff
@@ -399,6 +413,32 @@ public void createBindings() {
         ComponentDecorators.decorateWithAutoSelectAndLengthConversion(textFieldFeedEndZ);
     }
 
+    private void updatePartInfo()
+    {
+        int count = 0;
+        List<Board> boards = Configuration.get().getBoards();
+        for (int i=0; i<boards.size(); i++)    {
+            Board board = boards.get(i);
```
markmaker, line 421: You could use the newer loop form that is much more readable:

```
for (Board board : Configuration.get().getBoards()) {
...
}
```

desertgreg:
Sure, thanks for the feedback!  Would the typical approach be to make
another pull request?

I would like to improve the code as well.  The feature is useful once
youve set up your feeders and you're going back through them to pull the
cover tape off.  However when you edit a feeder to change the current part,
it doesn't update correctly.  I have hooked up a callback to the part combo
box where I re-run the function to count the parts but I noticed its not
working.  This weekend I stepped through the code and at that point in the
code, the feeder object still reports the old part that was selected when
the wizard loaded.  Do you have an idea how I could make this work better?

markmaker:
Hi @desertgreg 

Yes, when the PR is already merged (which I oversaw when I commented, btw.), I believe you need to make a new one (I'm no github expert).

Regarding the combo box: I always use "item changed" listeners:

```
       combobox.addItemListener(new ItemListener() {
            public void itemStateChanged(ItemEvent e) {
                ...
            }
        });
```
This seems to work for all sorts of changes: user changes, initial loading and fired property changes. 

If you are listening to changes back and forth between two components, you need to guard against endless loops using flags or similar.

That's no "Java veteran" advice ;-) My experience stems exclusively from [my OpenPnP work here](https://github.com/openpnp/openpnp/pulls?q=is%3Apr+author%3Amarkmaker). Java documentation is surprisingly bad, IMHO. Many things are documented in highly abstracted interfaces (all auto-generated with no overview or narrative thread) and therefore very vague when it comes to behavior in the implementation classes. If you look at stackoverflow.com advice etc. it seems everybody is just guessing from empirical observation. 

_Mark

Issue #755, opened 2018-07-24
ReferenceStripFeeder Breaks When Part Pitch > Hole to Hole Distance


vonnieda:
Set hole to hole distance to 4mm, set part pitch to 8mm, attempt to feed. Feeder tries to go to -Infinity.

See https://groups.google.com/d/msgid/openpnp/CAAYwRibwxzE8xs9gvVsWvfqyk8hXK_8HzNphxQZLbN8gYuD_ZA%40mail.gmail.com?utm_medium=email&utm_source=footer

doppelgrau:
Can be closed? A PR has been merged in December.

PR #1083, opened 2020-12-04
Fix for issue 755 "ReferenceStripFeeder Breaks When Part Pitch > Hole to Hole Distance"

desertgreg:
# Description
Fixes issue #755 which was caused by a divide by zero when the user incorrectly configures a strip feeder.

# Justification
Fixes a bug

# Instructions for Use
No change in usage of ReferenceStripFeeder

# Implementation Details
1. I configured a strip feeder with holes very close together (less than one partPitch distance) and the new code uses the user-supplied part pitch rather than using a computed "adjustedPartPitch".  This spaces the pick locations along a vector in the direction between the two supplied points.  I also configured a feeder with two holes exactly on top of each other and in this case, all pick locations will be computed as the same as pick location zero.  In all cases "reasonable" behavior for an incorrectly configured feeder.

2. The code follows the standard
3. No model changes
4. mvn test passes with no issues


vonnieda:
Sounds good to me! Thanks @desertgreg!

PR #1088, opened 2020-12-13
ReferenceDragFeeder works with 0402 parts with vision enabled and dis…

jacoburge:
…abled

# Description
Currently when using 0402 parts with ReferenceDragFeeder feeder vision must be enabled to pick two parts per advance of the tape (required for 2mm component pitch tapes common on 0402 components).  This feature is now fully independant of feeder vision features so can be used with or without feeder vision enabled.

# Justification
This enables higher accuracy machines to disable feeder vision and pick 0402 components.  If required feeder vision can still be used in conjunction with this feature.

# Instructions for Use
As before but now 0402 components can be picked without feeder vision enabled

# Implementation Details
1. Tested with physical drag feeder, both with feeder vision enabled and disabled, verifying correct pick location for multiple picks.
2. Yes
3. N/A
4. Tests pass

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
@@ -76,6 +76,8 @@ public class ReferenceDragFeeder extends ReferenceFeeder {
     @Element
     protected Location feedEndLocation = new Location(LengthUnit.Millimeters);
     @Element(required = false)
+    private Length partPitch = new Length(4, LengthUnit.Millimeters);
+    @Element(required = false)
     protected double feedSpeed = 1.0;
     @Attribute(required = false)
     protected String actuatorName;
@@ -99,7 +101,7 @@ public class ReferenceDragFeeder extends ReferenceFeeder {
      * correct feed locations.
      */
     protected Location visionOffset;
-    protected Location partPitch;
+    protected Location partPick;
 
     @Override
     public Location getPickLocation() throws Exception {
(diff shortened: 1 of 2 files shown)


vonnieda:
Very nice @jacoburge  - this really improves the setup for 0402s! Thanks for keeping an eye on backwards compatibility too!

etp123:
I am trying interface ReferenceDragFeeder with Openpnp and Liteplacer for 0402,
I have placed my PnP Material Stack Block on the x axis at the back edge of the machine,
feeding 0402 package -> pitch is set 2 -> dragging the strip works fine 

could you please advise how to set the direction of strip, 
because nozzle is trying to pick the part from 2mm LEFT side of pickup location rather than taking from 2mm BACK of pickup location

markmaker:
Hi etp123

If I read this correctly, you can't. It's hardcoded to X left.

https://github.com/openpnp/openpnp/blob/9f9e37c88d8a46598f969ec6e6a49d9b69358eef/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java#L94

However, you should be able to use the ReferencePushPullFeeder for dragging. It adds many additional benefits, like OCR, robust semantic vision working out-of-the-box, auto-setup and auto-learning for extremely quick setup of multiple similar feeders, etc. Don't let the 3D-printed example feeder in the documentation videos and images fool you. The software side of the feeder is very flexible and should cover any mechanical feeding, including drag feeding. All tape orientations, part pitches etc. are supported through the GUI.

https://github.com/openpnp/openpnp/wiki/ReferencePushPullFeeder

If you were to be the first to try it for drag feeding, I would be grateful.  ;-)

I'd help fix any setup issues. Should there be anything missing (not that I expect this), I will quickly fix it. However, in this case you'd have to jump on the testing version. 

https://openpnp.org/test-downloads/

There is a testing-round on-going with multiple goodies like camera lights and many bug-fixes.

https://github.com/openpnp/openpnp/pulls?q=is%3Apr+is%3Aclosed+label%3Atesting-round

_Mark

etp123:
Hi markmaker, 
	Thanks for your reply
         I have tried ReferencePushPullFeeder as DragFeeder it works perfectly fine, I have tried on 0402,0603 and 0805 looks good and fine.

Thanks for you support
Eugine

PR #1090, opened 2020-12-18
Removed Scripting from global configuration object

Jnesselr:
# Description
The Configuration class is a sort of god object at the moment. It was holding a singleton instance of the Scripting class to allow other places in the code to access it. I simply made the Scripting class hold a static instance of itself, made the constructor private, and added a "get" method to retrieve the instance.

# Justification
Hopefully the reduction of responsibilities for the Configuration class is justification enough. 🤞 

# Instructions for Use
Anywhere `Configuration.get().getScripting()` was used can now be replaced with `Scripting.get()`. For the end user, nothing should change at all.

# Implementation Details
I tested that the application still started correctly and that the example scripts that I could test still worked. There were a couple of classes in org.openpnp.spi.base that had to be updated as well.


markmaker:
Hi @Jnesselr 
More out of curiosity: is this really just for the purely academic sake of "reduction of responsibilities" or is there an underlying motive?
_Mark

Jnesselr:
I wouldn't call reduction of responsibilities purely academic. ;p

Right now, the Configuration class is intertwined throughout the entire application. Worse, it's a singleton object that holds a lot of state. That makes unit testing extremely hard because it makes it harder to test different scenarios cleanly. I just saw the Scripting class as an easy first step in unwinding a lot of this. I'd like to get to the point where a unit test can just say "here's a machine/part/etc configured exactly as I need it" without having to go through the Configuration object/filesystem.

markmaker:
Hi @Jnesselr 

___
The following is purely in the spirit of constructive critique. English is not my native tongue so if this sounds off-putting in any way, rest assured, it was not meant that way. 
___

Are you referring to the "[Single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)" or the "[Separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns)" principle?

If yes, I would argue that _obtaining_ an object does not fall into the responsibilities of that object. In fact, in my eyes, the opposite is true: in order for the object to be truly separate and self-reponsible, it must be completely agnostic towards its _ownership_.

In my view, true single responsibility would be when the Scripting object was oblivious to the Configuration object. So these lines **inside** `Scripting` should be fixed, **not** the ownership **outside** of it:

https://github.com/openpnp/openpnp/blob/e6217a00f77ad2058d0c22683faa6ed205833406/src/main/java/org/openpnp/Scripting.java#L67-L68

https://github.com/openpnp/openpnp/blob/e6217a00f77ad2058d0c22683faa6ed205833406/src/main/java/org/openpnp/Scripting.java#L292-L293

I.e. you would set the `scriptDirectory` and `Map<String, Object> initialGlobals` in the constructor, so the object can be instantiated without a configuration or machine object present. For unit tests etc. 

I have very strong reservations against Singletons. Sorry, they are **evil**, like every form of global data. The notion that one will _never ever_ want to instantiate more than one object, is IMHO just a sign of lack of humility towards what the future might bring and/or lack of imagination. OpenPnP has two fundamental Singletons (`Configuration` and `MainFrame`) that I'm aware of, and even though I would prefer to abolish those too, the're OK as a compromise in my eyes. But introducing more Singletons would certainly be bad.

All the other objects should have (hierarchic) ownership, so that instances can be obtained _contextually_, typically from `this`. Access to objects should always be _relative_ through parents or qualified children. 

So for instance the frequent `Configuration.get().getMachine()` should be replaced by a member `getMachine()` on each object, cascading to the top of the tree to get the right machine (caching is allowed, of course). If you want to be strict, you do it generically using a `Machine.getFromChild()` method, so the hierarchy objects can be made truly independent of the Machine object (providing a generic `getParent()` method is enough). So if one far-away day we decide that OpenPnP should become a "shop floor controlling software" and be able to control multiple machines, it would be a much easier change. 

In this spirit: removing the scripting getter from Configuration goes 180° into the wrong direction, IMHO. The getter should be made _more_ not _less_ specific to its context. Again, the right thing in my view, would be to get it from `this`, _contextually_. So if one day we decide hat each HeadMountable should have its own Scripting engine for performance/concurrency reasons (with CPU core numbers ever-increasing and Scripting engines often being single-threaded), it would again be easy to change.

Just my 2 cents.

_Mark

Jnesselr:
> Are you referring to the "Single responsibility principle" or the "Separation of concerns" principle?

In a very hand-wavy sort of way, yes. If I had to apply it to a SOLID principle, I guess I'd say it most closely aligns with SRP, yeah. I don't really think about those as principles directly anymore, I treat them more like I treat algorithms and data structures. That is, I know how I want to use them but don't explicitly think "Is this SOLID" or whatever.

> If yes, I would argue that obtaining an object does not fall into the responsibilities of that object. In fact, in my eyes, the opposite is true: in order for the object to be truly separate and self-reponsible, it must be completely agnostic towards its ownership.

I agree with this. I would prefer to do dependency injection from the get-go, but that's definitely not going to be a "snap your fingers, it's done" sort of change. There's several ways we can go about getting there, and I'm flexible as to the approach. Another option is to have a singleton that's like the configuration object that has references to things like the machine, user preferences, etc. The Configuration object becomes just a loader of configuration and wouldn't be required to use in tests and that could make testing faster/easier.

> In my view, true single responsibility would be when the Scripting object was oblivious to the Configuration object. So these lines inside Scripting should be fixed, not the ownership outside of it:

I definitely agree on the scripts directory, but I'd rather not inject a raw map into the Scripting class and instead have a reference to the machine, config, etc. passed in. We may have to think about circular dependencies at some point. If something inside the machine needs reference to the scripting class, but the scripting class needs a reference to the machine, that's a circular dependency issue.

> I have very strong reservations against Singletons. Sorry, they are evil, like every form of global data.

I 100% agree with you, assuming we're talking about Singletons that hold an instance of themselves vs injected Singletons. The only difference this diff makes is that you can't get a new scripting object by calling "load" on the configuration. But given that the Scripting object starts up a file watch thread every time it's instantiated, I think that's actually a good thing for now.

> So for instance the frequent Configuration.get().getMachine() should be replaced...

I agree it should be replaced, I'm not sure I agree on how. And I definitely don't want to try and make it so that getMachine is a method that can be called on every object. I do agree that I want to eventually be able to have OpenPnP be closer to shop controlling software, though.


If you're comfortable with it, I'd like to move this discussion to the discord server to iterate a little faster. It gives us a chance to dive in a little deeper, discuss philosophies, etc. and then we can update this diff with a comment after that.

markmaker:
> If you're comfortable with it, I'd like to move this discussion to the discord server to iterate a little faster. It gives us a chance to dive in a little deeper, discuss philosophies, etc. and then we can update this diff with a comment after that.

If you ask like this, no, I'm not comfortable with it ;-) In fact I find the form awful. No beginning, no end, everybody talking in parallel. No time for proper thinking. And it shows. 

I like it, when discussions such as these are properly related to a topic, as they clearly are in here. I like to be able to re-find it three months from now, when I've got another issue relating to it. I get notified if this discussion goes on (and not the 100 others that I don't care about). I can easily link to code, issues, PRs etc. and the latter will even automatically link back. 

It's quiet and focused in here. I like it.

An btw. I've had rapid exchange discussions in here as well, there's no technical reason this won't work. Hope you're equally OK with that.

_Mark

Jnesselr:
The reason I suggested it was because I wanted to discuss, at a higher level, how do we deal with the configuration object being a god object. I'm not sure this PR is the best place to do that, but I don't have strong opinions on it either way.

I have my own thoughts on how to do this, but I have less experience in the code base. My "where I want to be":
1. Configuration isn't a singleton god object.
2. MainFrame isn't a singleton.
3. There's a class that manages user preferences (easy to adjust those in tests)
4. There's a class that manages loading the machine(s). Machines can be made "manually" for things like tests.
5. There's a class/set of classes that manage parts and parts libraries.

That's just at a higher level, and most of it is around wanting to have better functional unit testing.

markmaker:
For 1: I strongly agree, but it means that each and every access to the Configuration.get() would have to made contextual (as laid out earlier). 

For 2: I'm not sure. Swing is not "multi-head" AFAIK so this is less important. Most of the testing code runs even if the GUI is not initialized. 

For 3, 4: Sorry, I don't know what _exactly_ you mean to be different. The whole Configuration tree is Reflection driven (Attributes). That's the gold standard in my book. Among other substantial benefits, it has complete and utter separation of concerns. Not much to improve there. 

For 5: Again I don't understand what you mean to be different. Just note that the feeders' setup is not independent of parts. 

_Mark

Jnesselr:
I'm mostly trying to establish a "where are we going" point we can agree on so it's easier to then discuss "how do we get there". ie I'm not quite ready to discuss what to do implementation-wise. Most of this is in context of unit tests, but I had talked to @vonnieda about using JavaFX instead of Swing and if that's something we want to pursue, I'd like to move towards making that easier than it is now.

For 2: It basically makes it so that you can't test well around the GUI at all or really anything close to it. It makes it much harder, if nothing else.

For 3, 4, 5: Using a reflection based loader isn't the worst (it does have issues here and there, but I'm not suggesting we change that just yet) but the fact that you have them all part of one ginormous class is bad. I mean on my machine it takes 4 seconds to run each and every test because many of them require you to initialize the configuration based on a test file that has defaults. It's all super fragile. There's no reason it has to be like this. That class needs to be broken up. You have 400+ different calls to `Configuration.get()`, over half of which are just to get the machine. That's a perfect code smell that maybe that needs to be extracted out. I do recognize that feeders' setup is dependent on parts, btw. I've looked through the codebase at a decently high level, but I've also asked a ton of questions to Jason in the discord to solidify my understanding of the project.

markmaker:
Hi @Jnesselr 

I agree with most of that. 

When you say "It's all super fragile" I don't understand what you mean, though. If you refer to the reflection stuff, I disagree, that's super robust. You can serialize and de-serialize any branch of the tree. The fact that unit tests are not run on such partial branches is just that it's easier to configure (in the GUI) and use the whole machine. If you refer to (allegedly missing) unit testing, I'd beg to differ.

I agree that for the very basic "algorithmic" tests, the classes themselves should be testable in isolation. And this is often done. But for many tests, very quickly you have components from throughout the machine involved. I really plowed into OpenPnP with the [Advanced Motion Control](https://github.com/openpnp/openpnp/issues?q=label%3Aadvanced-motion-control+is%3Aclosed) development and the two most important unit test, that caught 99% of any unfinished work or true regression issues were the BasicJobTest and SampleJobTest, both putting many sub-systems together. At least as an experienced developer, you seldom make stupid mistakes that are caught by some single class level knee-jerk test. The relevant errors happen in the synthesis of the system, when the whole becomes so much more than the sum of its parts, both in its functional power but also in its combinatorial complexity. 

My own substantial addition to unit testing was therefore exclusively in a) basic "algorithmic" tests like the new AdvancedMotionTest and b) massively extending the SampleJobTest. 

I actually built in computer vision based success control by looking at the [simulated machine table image](https://raw.githubusercontent.com/openpnp/openpnp/develop/src/main/resources/samples/pnp-test/pnp-test.png) and optically recognizing feeder tape pockets and solder lands to determine whether correct Pick & Place locations were computed. This is completely independent of any job data from OpenPnP and as "physical" and "results oriented" as it can get, therefore it will catch any regression including loading the job, processing placement data, vision-nailing fiducials, feeders etc. right through to finally placing the part.

On top of that I created the SimulationModeMachine with simulated imperfections like nozzle tip runout, non-squareness, post-motion vibrations, camera noise, to stress-test the corresponding compensation features, camera settling etc., all in an "as real-world as possible" combination/integration. 

grafik

Btw. you can try it live by using this machine.xml ...
https://github.com/openpnp/openpnp/blob/develop/src/test/resources/config/SampleJobTest/machine.xml
...and running the [pnp-test job](https://github.com/openpnp/openpnp/tree/develop/samples/pnp-test) installed with OpenPnP. 

So I'm an ardent fan of whole-machine integration tests. Yes, these could still be extended to cover more sub-systems, but I don't see a large unit testing deficit per se. 

If you referred to this: no, it is not "all super fragile" ;-)

_Mark

vonnieda:
Hi @markmaker - @Jnesselr and I discussed this a bit and he's going to go in a different direction, so I will close this PR. The new direction will be essentially similar to what you described above - Scripting will take a directory in it's constructor, there will be no singleton, and Configuration (for now) will pass the directory in when it creates it's Scripting instance. The ultimate goal here is to be able to create a Configuration instance in memory without touching the filesystem for testing purposes, so that's what these changes are working toward.

PR #1092, opened 2020-12-27
Remove GUI dependencies from Scripting

Jnesselr:
# Description
The Scripting class is used to execute scripts (through a menu item) or execute events (triggered in various places in the code). The Scripting class was also handling copying example scripts over from project resources as well as starting a file watcher which updated a menu in the GUI.

This diff:
- Moves the Scripting class to an org.openpnp.scripting package.
- Refactors the scriptsDirectory to be a parameter passed into the Configuration class.
- Adds a ScriptFileWatcher class in the scripting package that handles copying the example scripts over and handles synchronizing the menu in the GUI application.

# Justification
This is mostly to clean up the code for testability. We don't have tests relying on scripting (e.g. verifying that events are sent off properly) just yet, but the scripting engine was created (indirectly through Configuration.initialize) in a lot of tests which meant the file watcher was also kicked off for each of those tests.

# Instructions for Use
This should be a change that only affects developers. End users should notice no change whatsoever.

# Implementation Details
I tested this change by running the application and updating some scripts in the scripts directory as well as running them through the menu. No changes to the spi but the configuration class did get a small change just to inject the scriptsDirectory into the scripting class. Unit tests pass.

### DIFF
--- a/src/main/java/org/openpnp/gui/MainFrame.java
+++ b/src/main/java/org/openpnp/gui/MainFrame.java
@@ -92,6 +92,7 @@ import org.openpnp.gui.support.OSXAdapter;
 import org.openpnp.gui.support.RotationCellValue;
 import org.openpnp.model.Configuration;
 import org.openpnp.model.LengthUnit;
+import org.openpnp.scripting.ScriptFileWatcher;
 import org.pmw.tinylog.Logger;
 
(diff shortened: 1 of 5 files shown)


vonnieda:
Thanks @Jnesselr! For future readers, this is part of a change that will enable in memory Configuration for running unit tests without hitting the disk.

PR #1093, opened 2020-12-29
Drag Pin Safety Checks

c-riegel:
# Description
This change impacts drag feeders.  It adds an optional "Actuator Safe Value".  If this field is populated then the actuator (drag pin) will be read during the feed operation to confirm safe operation of the machine.  The value read will be compared with the value entered in the "Actuator Safe Value" field.  

The actuator will be read to confirm drag pin extension.  The value is compared to "Actuator Safe Value" and the reading should not be equal to the safe value.

The actuator will be read after drag pin retraction.  The value should match the safe value.  If not, then the feeder will be instructed to backoff one more time before an exception is thrown

# Justification
CharmHigh machines have an opto-interruptor that can be used for verifying if the drag pin is up or down.  This check adds safety for machine operation.

# Instructions for Use
Fill in the value that should be read when drag pin is in a state where it is safe to move in the "General Settings" area of a drag feeder.

# Implementation Details
By running the machine and confirming the checks and operation.
Yes
No changes made to these
pass

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceDragFeeder.java
@@ -81,6 +81,8 @@ public class ReferenceDragFeeder extends ReferenceFeeder {
     protected String actuatorName;
     @Attribute(required = false)
     protected String peelOffActuatorName;
+    @Attribute(required = false)
+    protected String actuatorSafeValue;
     @Element(required = false)
     protected Vision vision = new Vision();
     @Element(required = false)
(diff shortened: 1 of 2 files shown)


vonnieda:
Hi @c-riegel - I think this is now covered by the axis interlock actuator that was recently added? See https://github.com/openpnp/openpnp/wiki/Axis-Interlock-Actuator. See https://github.com/openpnp/openpnp/wiki/Axis-Interlock-Actuator#safety-confirmation-sensor-on-a-drag-pin for a specific example.

Thanks,
Jason

c-riegel:
Yes, this would work.  I'll just use that.  Thanks!

PR #1095, opened 2020-12-31
Feature / Actuator Profiles, Camera Lights

markmaker:
# Description

* Actuators can be configured for a specific (writing) value type (**Boolean**, **Double**, **String**, **Profile**).
* Non-Boolean actuators can be configured for specific typed ON and OFF values, so they can be triggered using Boolean semantics.
* EDIT: Actuators can automatically be state-initialized or actuated when Machine States change: Enabled, Homed, Disabled.
* The GUI was reworked to adapt itself to the value types. 
* `HttpActuator` was extended to support all value types.
* `ScriptActuator` was extended to support all value types.
* The new actuator value type **Profile** introduces a "multiple choice" actuator (like a custom Enum type). 
* A **Profile** type actuator can control multiple target actuators (including itself). 
* Each of the choice profiles controls a set of specific values to be written to the target actuators. 
* Added a **Light Actuator** to the Camera. 
* A **Profile** type actuator is the ideal choice to control multi-channel lights such as RGB lights or other selective/angular light sources, or to provide predefined step values for scalar light intensity. 
* Actuator based light triggering is now a part of the `lightSettleAndCapture()` function. 
* The pipeline `ImageCapture` stage can set/override the **Light Actuator** value as a parameter.
* `VisionProvider` functions now universally use `settleAndCapture()` instead of `capture()` to enable unified actuator driven lighting.
* EDIT: Camera implementations have been unified in terms of broadcasting, preview FPS etc.
* EDIT: Full support for 0 fps where only the actual vision captures are displayed.
* EDIT: Improved SwitcherCamera to only broadcast the correctly switched channel. 
* EDIT: Optional Auto-Camera-Preview: the active Camera is automatically switched in single preview. 
* Various related bug-fixes.

# Justification
Justification was discussed here:
https://groups.google.com/g/openpnp/c/j4ucK1JuyH0/m/kR66HElcBQAJ

# Instructions for Use

EDIT: this was updated 2021-01-13

## Actuator Value Type

Each Actuator now has a **Value Type** field:

Actuator value type
Actuator value type help  

**NOTE:** The **Value Type** mostly just controls how values are edited in the GUI. It does not force the actuator function to actually use that value type. If existing functional actuator assignments in the Machine Setup, or Scripts or G-code use different or even mixed actuator value types, it should still work as before.  The **Value Type** will automatically be proposed when an actuator is first used in a functional assignment. 

For the **Double** or **String** value type, a pair of ON and OFF default values appear:
Default ON OFF values

These will be actuated, when Boolean semantics is used to control an actuator e.g. for an actuator that can be set to a specific scalar value (analog control), but also can be switched ON/OFF to specific preset ON/OFF values.

## Actuator Machine States

Actuator Machine States

The **Enabled**, **Homed**, **Disabled** machine state actuation settings can be used to **assume** or **actuate** specific ON/OFF values when these machine states are entered. 

* **LeaveAsIs** leaves the actuator state at its previous state (maybe the unknown state).
* **AssumeUnknown** sets the actuator state to unknown. OpenPnP will not know the state until it has first actuated the actuator.
* **AssumeActuatedOn** or **AssumeActuatedOff** assumes that the machine state has changed the actuator state to ON / OFF respectively (as a definitive side effect of the machine state change or perhaps via custom G-code). 
* **ActuateON** or **ActuateOFF** bring the actuator to a desired defined ON / OFF state when the machine state is entered. 

**NOTE**: **Enabled** and **Homed** actuation happens **after** the principal machine state actions have been completed. However, the **Disabled** actuation happens **before** the principal machine state actions are executed. 

## Actuator with Profiles

The **Profile** actuator value type is special. Once you have selected it, press **Apply**. A new "Profiles" Wizard tab will appear: 

Profiles Wizard tabl

You can assign up to 6 target actuators to be controlled by the profile. Again press **Apply** for new target actuators to appear as columns in the table. 

New profiles can be created using the Plus button, or deleted using the Delete button.

A Profile will become active once you assign a **Name**. 

The **Default ON** and **Default OFF** checkboxes control which profile will be selected when the actuator is driven using Boolean semantics. Obviously, the checks are mutually exclusive. 

**NOTES:** 

If the target Actuator is itself a **Profile** type actuator, it will be driven as a **String** value (to prevent endless recursion). Using the actuator itself as a target actuator is therefore explicitly supported. Using this technique, you can create self-contained multiple choice actuators. One application example would be a scalar actuator e.g. a light actuator with PWM intensity control. Using a limited set of profile entries you can drive it to specific step values, a bit like the f-stops in photography. Usage becomes more reproducible and effective values can be centrally maintained (otherwise they would be littered across pipelines, see the [ImageCapture](#image-capture-stage) stage below). 

Stepped Profile Actuator

Even though a profile target actuator is driven as a String value, you can still effectively drive Double values as well. Just enter them as Strings. It won't make a difference in effective GcodeDriver or HttpActuator encoding, and in ScriptActuators you can convert the values inside the script. 

## Actuator Control Panel

The Actuator Control Panels will now display **Value Type** specific GUI controls:
Double Actuator Dialog Boolean Actuator Dialog  
Profile Actuator Dialog

**NOTE**: if the **Default ON** and **Default OFF** values are defined as explained above, the **On** and **Off** buttons will appear in addition to the typed value field. 

As a take-away, the **With** field has been added to make the **Set with Double** function more independent (formerly the **Read Value** field was reused for that purpose). 

## HttpActuator 

The HttpActuator can now also drive **Double** and **String** type values. For these, the HttpActuator has a new **Parametric URL** field with a placeholder `{val}` that can encode the value into the URL. You can use formatting in the form of `{val:%.4f}` (like in the GcodeDriver): 

HttpActuator

**NOTE:** no URI-escaping is performed, you can therefore enclose complex URI fragments such as multiple parameters. 

If the Actuator has **Boolean** value type, the **On URL** and **Off URL** will be presented (as before):
grafik

 ## Script Actuator

For completeness, the ScriptActuator can now also drive **String** type values. The value is available as the `actuateString` global. 

## Camera Light Actuator

The camera can now be assigned a **Light Actuator**:

Camera Light Actuator

There are various ON and OFF actuation Options:

* **Before Capture ON**: to actuate the Light Actuator before a computer vision operation captures a camera image. The happens before the [Camera Settling](/openpnp/openpnp/wiki/Camera-Settling). See the [ImageCapture Stage](#image-capture-stage) section to control pipeline specific light actuation values and profiles. **NOTE:** This option should almost always be enabled. 
* **After Capture OFF**: to actuate the Light Actuator OFF after a computer vision operation captured a camera image. 
* **User Camera Action ON**: to actuate the Light Actuator ON when a user action is deliberately positioning or otherwise using the camera. 
* **Anti-Glare OFF**: to prevent this camera light from blinding another camera, it is actuated OFF before any other camera is capturing an image. Only cameras looking the other way (up/down) are taken into consideration (see the **Looking** field). 

As soon as a **Light Actuator** is assigned, the Camera View will display a new overlay light symbol in the upper right corner. It shows the camera light status and can be clicked to turn the light on or off:
 
CameraViewLightOnOff

## New Camera Properties

To improve the experience with this new options for all the Camera implementations, and especially for the Switcher Camera, a few other possibilities have been created. 

**Preview FPS** (frames per second) is now universally available for all the camera implementations, including the SwitcherCamera with proper channel preview (at reasonably low FPS). Note, you can set fractional FPS like **0.5** for an update every 2 seconds. The implementation has been unified and now fully supports the **0 fps** setting, where only frames explicitly captured for computer vision or for other deliberate user camera actions are shown in the Camera View. 

**Suspend during tasks?** allows you to use a relatively high **Preview FPS** during manual machine control, while effectively setting it to efficient **0 FPS** during machine tasks, especially during Jobs. This is also the only reasonable setting for a SwitcherCamera with preview. 

**Auto Camera View?** automatically selects the active camera in the Camera View (if another single camera was selected before). This happens on deliberate user camera actions such as positioning, jogging the camera or when computer vision captures frames or displays marked-up result images. 

When both  **Suspend during tasks?** and **Auto Camera View?** are enabled, there will only be clean result images presented during a job. By only displaying one camera at a time, you get a larger preview with much better resolution. **Tip:** make sure to right-click the camera View and choose **High** or **Highest** quality rendering. 

## ImageCapture Stage 

Every computer vision pipeline typically contains the ImageCapture stage to capture an image from the camera:

ImageCapture Stage

As soon as the camera has a **Light Actuator** assigned, the ImageCapture stage can control the lighting. 

If the **defaultLight** property is switched on (default) the **Default ON** value of the actuator will be used to determine the lighting. If it is switched off, the **light** property comes into play.

The **light** property controls the lighting for this specific pipeline (if **defaultLight** is off). The user interface responds to the value type of the actuator:

Light Property Boolean Light Property Double  
Light Property Profile

# Implementation Details
1. The changes have been tested in the three controller lab setup. **PR discussion and testing are still considered incomplete.** 
3. Changes in the `org.openpnp.spi` or `org.openpnp.model` packages concern the addition of the actuator value type, profiles and camera actuator assignment.


markmaker:
Hi @vonnieda 

As you see the PR exploded a bit. ;-)

With the Profile actuator I chose an implementation that is agnostic to its use as a light actuator. It can be used for completely different things, whenever a set of choices is needed and also whenever more than one actuator must be driven as a "gang", across different drivers and/or across mixed Gcode- Http- and ScriptActuators. 

Before I add further effort (close the last loopholes, add more thorough testing), I'd like to be sure you're OK with this. 

Documentation should also go into the Wiki.

Thanks for looking into this.

And Btw: 

"Rütsch guet übere!" as we say in Switzerland, would be something like "slide over well", meaning the New Year of course ;-)

_Mark

vonnieda:
Hey @markmaker  - I think it looks awesome! The Actuator profiles are very clever and that's a great way to handle the camera profiles. My only concern is the `Object` being passed to settleAndCapture() - I'd prefer to see something type safe, but I haven't had a chance to really look closely at it. I'm sure you had a good reason for it.

So, I'd say the approach looks great, and please continue. I might want to make some minor adjustments to the implementation later, once I've looked at it more closely.

Nice work, and thank you! This will be a really nice system!

Jason

markmaker:
Hey @vonnieda 

**A happy new year!**

> My only concern is the Object being passed to settleAndCapture() - I'd prefer to see something type safe

Yeah, I was reluctant about this too. I was starting with a Wrapper object, a [std::variant<>](https://en.cppreference.com/w/cpp/utility/variant) of sorts, then I looked up "Variant + Java" and found this:

https://forums.codeguru.com/showthread.php?36871-What-s-equivalent-of-VARIANT-in-Java

Also found the same advice in a German-speaking forum. It _does_ make sense in a way. 

It was also surprisingly hard to pry my foot into the `PropertyEditor` and `TableCellRenderer` setup for the [`PipelinePropertySheetTable`](https://github.com/openpnp/openpnp/blob/a307bbf247c15036a89205c994fb4926717f7358/src/main/java/org/openpnp/vision/pipeline/ui/PipelinePropertySheetTable.java#L35).  After about four attempts exploding into ridiculous complexity, I found the advice to use Object typing the easiest to implement, so I could use the standard implementation classes `BooleanAsCheckBoxPropertyEditor`, `BooleanCellRenderer`, `DoublePropertyEditor`, `StringPropertyEditor` and I only had to sub-class one new `ComboBoxPropertyEditor` as the [`ActuatorProfileEditor`](https://github.com/openpnp/openpnp/blob/a307bbf247c15036a89205c994fb4926717f7358/src/main/java/org/openpnp/vision/pipeline/ui/PipelinePropertySheetTable.java#L69). With a Wrapper object I would have had to rewrite a whole new Editor and Renderer plus some conversion stuff (at least that is the way I understand it, the l2fprod PropertySheetTable is practically undocumented). 

But I guess it could still be relatively easy to rework this into such a Variant type: I could leave the Object typing in the Stage's **light** property but then wrap it into the Variant before passing it to `settleAndCapture()`. And similar with other GUIs.  

_Mark

markmaker:
Hi

Tied up most loose ends. 

Code that used its own brew of actuator value typing was reworked to use the central actuator valueType. Migration of the old value type to the actuator was added. 

- ReferenceAutoFeeder and Slot classes
- ActuatorWrite vision stage (now supports all value types including Strings and Profile with proper GUI)

Actuators that have a clear functional value type (being driven with a specific data type hardwired in code) are also migrated both initially and when new actuators are used for the first time. Note: these are not hard assignments, the user can always override the valueType manually (e.g. for backwards compatibility with exotic mixed-type applications). 

- SwitcherCamera
- Vacuum Blow Actuator usage 
- SchulzFeeder and Slot classes (see also the PM to Bill, discussing that the former value typing is probably no more needed)

_Mark

markmaker:
Best look at the diff with whitespace ignored.

https://github.com/openpnp/openpnp/pull/1095/commits/7a32bc6381ab35558ec460f35b7e88f5f45b43ca?diff=unified&w=1

I took the liberty to apply standard indentation on the cleaned-up Schultz Wizards as they contained large amounts of inconsistent Tab-formatting that looks bad on Github.

_Mark

markmaker:
Added up/down buttons to reorder the profiles:

Profiles

markmaker:
Hi

* Added Camera Light exposure simulation to the ImageCamera and SimulatedUpCamera while also simulating camera lag.
* Configured light actuators in the SampleJobTest/machine.xml
* Changed the NullDriver to GcodeAsyncDriver with GcodeServer backend in SampleJobTest/machine.xml 
* This effectively tests camera lighting and settling, as vision would fail with the darkened camera view.
* Also effectively tests GcodeAsyncDriver operation and flow control and GcodeServer G-code interpretation. 
* Created Machine.execute() to execute tasks immediately, either directly (when already inside Machine task) or submit() and wait for completion.
* Camera light actuators and also the SwitcherCamera actuator (WIP) are now properly actuated in Machine tasks.
* Any such execute() will timeout if the machine remains busy for more than a second. So if for instance you open the pipeline editor, while a Job is running, the light actuation for ImageCapture is rejected, the Job must not be endangered by fluke light actuation. The ImageCapture stage fails.
* Machine Actuators also need machine coordination options displayed on the Wizard. They may not be moved themselves, but they must still be able to customize coordination with machine motion. E.g. the bottom camera light actuator coordination should be switched **off**, so the light is actuated immediately, before the nozzle move even starts (we talked about this, it reduces blinking).

The following little video shows the new simulation, the uncoordinated light actuation, how fluke actuation through pipeline editing is rejected while a job is running:
https://user-images.githubusercontent.com/9963310/104057464-7f068000-51f2-11eb-8d62-a91d66477722.mp4

_Mark

markmaker:
Hey @vonnieda 

Working on this I am a bit astonished that the SwitcherCamera does not create problems when both preview capture() and machine task (long-running job) vision capture() use the switching actuator from different threads. I guess the actuator itself is synchronized through SwitcherCamera.switchers but still... 

Please have a critical look at the changes I just pushed.

I probably have to undo the changes for the SwitcherCamera, so it can still switch for preview from a different thread, while a job is running... ???

_Mark

markmaker:
Hi @vonnieda 

**Please have a look at this and intervene if necessary.** 

I spent more time analyzing this. Some points I observed:

If I'm analyzing this correctly, the SwitcherCamera is not optimized to be used efficiently in a Job. It has an Actuator delay of 500ms by default and is set to 10 fps (not configurable in the SwitcherCameraConfigurationWizard), so the Camera thread is locking the SwitcherCamera.switchers (synchronized block) for 500ms preventing any other captures, then 100ms sleeping. Consequently, computer vision has only free access 1/6 of the time, the other time it has to wait up to 500ms. With two cameras preview-switching constantly, the chances are 50%, that the wrong camera is switched and this then adds another 500ms wait time. 

https://github.com/openpnp/openpnp/blob/8f290c2ced797bf3c65d585c49e0eacef3c56057/src/main/java/org/openpnp/machine/reference/camera/SwitcherCamera.java#L68-L90

https://github.com/openpnp/openpnp/blob/8f290c2ced797bf3c65d585c49e0eacef3c56057/src/main/java/org/openpnp/machine/reference/camera/SwitcherCamera.java#L128-L138

I guess preview switching must be suspended as soon as a Machine task / Job runs. The off-switched camera should stop switching and broadcasting. 

This prompted me to look at this whole broadcasting system. The comments you made here...

https://github.com/openpnp/openpnp/blob/8f290c2ced797bf3c65d585c49e0eacef3c56057/src/main/java/org/openpnp/machine/reference/camera/OpenPnpCaptureCamera.java#L109-L111

... and here ...

https://github.com/openpnp/openpnp/blob/8f290c2ced797bf3c65d585c49e0eacef3c56057/src/main/java/org/openpnp/machine/reference/camera/OpenPnpCaptureCamera.java#L174-L191

... are very valid. I think its high time to fix this.

I'm in the process of taking the superior design of OpenPnPCaptureCamera as a template and, as you suggested, move it up into a new base class AbstractBroadcastingCamera that handles all that stuff for all camera implementations. The actual camera implementations should only have to deal with opening, closing their devices and capturing an actual image. 

I'm deviating slightly from what you said in the second comment (linked/listed above). Instead of directly broadcasting the captured image from the internalCapture(), I'm just storing the image in transformImage() and notifying the captureNotifier you introduced. The camera thread will then pick up the stored image and broadcast it. This way a machine thread calling capture() is not burdened by that GUI stuff and immediately free to go on. In its place the camera thread can further optimize this i.e. only broadcast some out of many rapidly fired captures (but always the last one). 

If this is not going into the right direction, please shout STOP!
 
Thanks ;-)

_Mark

markmaker:
Ok I was mistaken.

The Camera thread calls captureForPreview() but only a full (vision) capture() is effectively switching. So the camera preview is showing the wrong image. Should have consulted the demo video where this is evident:
https://www.youtube.com/watch?v=gFXqbMJM2wM

But I still think it makes sense to optimize this. CameraView should freeze the last captured image of the off-switched camera and this is achieved simply by not broadcasting images obtained from the wrong camera. 

All the other stuff I said, is still valid. I'm going ahead...

_Mark

markmaker:
Hi

Ok, further work: 

* Completely overhauled and unified the Camera broadcasting system.
* Unified/introduced Preview FPS settings across all Camera implementations. 
* Preview broadcasting can be suspended during machine tasks/jobs.
* Optional Auto-CameraView selection, Auto-Light triggered by camera related user action, computer vision. Introduced across the system.
* Full support for 0 fps configuration. 
* SwitcherCamera will no longer show the wrong source camera's image. 
* Properly associate SwitcherCamera lens calibration and transforms to source analog camera and lens. The digitizer camera is captured raw. 
* Extensive tests, improvements/unification/bug-fixes across Camera implementations.
* **Note**: GUI and log testing but no functional testing for OnvifIPCamera, Neoden4Camera, truly switched SwitcherCamera.
* Made actuator machine coordination illegal outside machine task (needed UnitTest adaptations).

grafik

Good night. :-)

_Mark

markmaker:
Hi 

Reworked the **Instructions for Use** as a prep for the Wiki. Mostly in...

* Camera Light Actuator (reworked)
* New Camera Properties (new)

Tested this thoroughly in the SimulationModeMachine. 

@vonnieda  do you think I could deploy this as a **Testing Version**? 

_Mark

markmaker:
Btw. yes, I made a 180° U-turn on checkboxes. 

After separating the Machine State actuation to be a feature available for all Actuators (please read the updated PR **Instructions for Use**) ...

grafik

... I was able to boil light actuation down to just four Booleans ...

grafik

... and despite my lengthy explanation to the contrary (in the group), we get (almost) full degrees of freedom (well, it does rarely make sense to switch off **Before Capture ON**, I just added it for symmetry). 

I think this is quite easy to use.

_Mark

vonnieda:
Mark - please feel free to push to test whenever you like, including
overwriting my changes, deleting the branch and re-creating it, etc. I'll
let you know otherwise if I'm working on something important.

I'm swamped this week so I haven't had time to look at anything.

Jason

markmaker:
This is a repository for Wiki images. Please do not delete.
___
Camera View Options
Machine Setup
Position Camera
Machine Controls

PR #1097, opened 2021-01-03
Added maxFeedCount feature to ReferenceStripFeeder

desertgreg:
# Description
This change adds a 'maxFeedCount' field to ReferenceStripFeeder.  When the feeder's feedCount exceeds 'maxFeedCount' then it will throw an exception indicating that it is out of parts.  The message shown to the user will display the part and feeder name.  If the user leaves 'maxFeedCount' at the default value of zero, this feature will be ignored so we don't force everyone to immediately set maxFeedCount on all of their feeders.

# Justification
This is a much-needed feature for projects using strip feeders that cannot fit all of a given part into one strip.  For example I have a project which uses 64 leds and there is no way to safely use OpenPNP to populate those parts on my machine because it continues to feed past the end of the strip.

# Instructions for Use
If the user wants to use the new maxFeedCount option on ReferenceStripFeeder, they simply set the desired value in the wizard for the feeder.
image
When a feeder runs out of parts a message like this will be displayed:
image

# Implementation Details
1. I tested the change by running several test jobs using more parts than my 'maxFeedCount' setting.
2. The coding style is followed.
3. No changes to `org.openpnp.spi` or `org.openpnp.model` packages but there is a new optional attribute named 'maxFeedCount' on ReferenceStripFeeder.
4. mvn test runs without errors

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java
@@ -108,6 +108,9 @@ public class ReferenceStripFeeder extends ReferenceFeeder {
     @Attribute
     private int feedCount = 0;
 
+	@Attribute(required = false)
+	private int maxFeedCount = 0;
+
     private Length holeDiameter = new Length(1.5, LengthUnit.Millimeters);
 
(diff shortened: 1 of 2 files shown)


vonnieda:
Looks good to me, thanks @desertgreg!

PR #1098, opened 2021-01-05
Pipeline editor enhancements.

tonyluken:
# Description
1) Added capability to the pipeline editor to display images in the true color space of the image rather than just assuming the color space is BGR.
2) Added a dialog confirmation when the pipeline editor is closed.

# Justification
Makes using the pipeline editor more intuitive for the user when working in alternate color spaces such as HSV.  Also permits the user to accept or reject the changes when closing the pipeline editor - previously there was no way to reject any changes.

# Instructions for Use
1) The pipeline editor now has an icon (in the shape of an eye) in the toolbar at the top of the results panel that permits the user to select either the actual color space (now the default) of the image or the BGR color space (the way it used to work) for the display.  With images displayed in true color, the user will no longer be presented with false color images with misleading RGB and HSV values.  It is important to note the the actual pipeline results are unaffected by this setting - it only affects what is visible in the pipeline editor display.

2) When the pipeline editor is closed, a dialog box now pops-up (if there have been any changes to the pipeline) that give the user the choice to keep the changes, reject the changes, or cancel the close operation.   

# Implementation Details
_1. How did you test the change?_ Edited several different pipelines and observed the results.
_2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?_  Yes.
_3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing._  No changes were made to these packages.
4_. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted._  All Maven tests were run and passed.


tonyluken:
Oh, one thing I forgot to mention in the PR - I added a property to the ImageRead stage that selects which color space should be used for the image (defaults to BGR).  This can be used to have the image display correctly in case the it was written in a color space other than BGR..  This does mean that the machine.xml file gets updated and  will no longer be backward compatible with older versions of OpenPnP -  **be sure to backup your machine.xml file before trying this!**

src/main/java/org/openpnp/vision/FluentCv.java
```diff
@@ -107,6 +107,26 @@
...
+    public enum ColorSpace {
+        Gray(0),
+        Bgr(1),
+        Rgb(2),
+        Hls(3),
+        HlsFull(4),
+        Hsv(5),
+        HsvFull(6);
+        
+        private int code;
```
markmaker, line 119: Hi Tony

It seems this is obsolete, right?

Also, you could have used ordinal()

_Mark

src/main/java/org/openpnp/vision/pipeline/ui/CvPipelineEditor.java
```diff
@@ -1,216 +1,244 @@
-package org.openpnp.vision.pipeline.ui;
```
markmaker, line 1: Hi Tony

Do you know why it does this? Marking everything as changed? Some newline issue? 

I suspect, this will create artificial conflicts with #1106 etc. that are in testing. Could you please resolve this? Thanks!

_Mark

src/main/java/org/openpnp/vision/pipeline/stages/ImageCapture.java
```diff
@@ -1,76 +1,77 @@
-package org.openpnp.vision.pipeline.stages;
-
```
markmaker, line 2: Hi Tony

Do you know why it does this? Marking everything as changed? Some newline issue?

I suspect, this will create artificial conflicts with #1106 etc. that are in testing. Could you please resolve this? Thanks!

_Mark

tonyluken: It appears the current file in the repository has CRLF line endings but my changed file has LF line endings.   I believe LF is the correct line ending.  Looks like the previous change to the file introduced the wrong line endings.

Tony

src/main/java/org/openpnp/vision/pipeline/ui/CvPipelineEditor.java
tonyluken: It appears the current file in the repository has CRLF line endings but my changed file has LF line endings.   I believe LF is the correct line ending.  Looks like the previous change to the file introduced the wrong line endings.

Tony

markmaker:
Hi @tonyluken 

My idea is to create a little testing round with several changes at once, in the hope that we contributors can test each other's code too.

grafik

This has now created the predicted conflicts. I don't know what to do about the line endings, or how they came to be. Looking at the log, it was _you_ who created the last two versions, so I don't understand. Locally I don't see any line ending problems. I have resolved the conflicts with #1095 and merged it. 

Be sure to pull the test branch. 

_Mark

tonyluken:
It's quite possible I introduced the problem with some of my early pull requests.   Someone scolded me (rightly so) for not having my GIT configured to properly handle EOL conversions.  I'd suggest someone more knowledgeable about GIT (Jason?) add something to the Developers section of the Wiki so that everyone knows the proper GIT configuration.

PR #1099, opened 2021-01-06
Improved "MaxFeedCount" for Reference Strip Feeder

desertgreg:
# Description
This adds a few more details to the 'maxFeedCount' feature of ReferenceStripFeeder
- When a feeder feeds its last part, it will automatically disable itself.  This causes OpenPNP to automatically switch to another feeder for that part (if there is one)
- There is a new button for automatically setting the MaxFeedCount based on the configured reference hole locations

# Justification
This further improves the behavior of ReferenceStripFeeder when you need more parts than will fit in one strip.

# Instructions for Use
If you press the 'Auto Set MaxFeedCount' button, the MaxFeedCount field will be set based on the distance between the reference hole locations you have selected and the pitch of the part.  If you configure multiple strip feeders for a given part and one of them runs out of parts, OpenPNP will automatically start using the next strip that it can find in the list.
image

# Implementation Details
1. Tested the change by running jobs which used more parts than fit in one strip.  Also tested the 'Auto Set MaxFeedCount' button on some strip feeders to see if it computed the correct number of parts.
2. This follows the coding style
3. No changes to the model
4. mvn test runs without errors

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java
@@ -223,11 +223,20 @@ public class ReferenceStripFeeder extends ReferenceFeeder {
     public void feed(Nozzle nozzle) throws Exception {
         setFeedCount(getFeedCount() + 1);
 
+        // normally the feeder will automatically disable when it runs out but
+        // if the user re-enables and tries to feed, throw an exception that we are out of parts.
         if ((maxFeedCount > 0) && (feedCount > maxFeedCount)) {
 			throw new Exception("Tried to feed part: " + part.getId() + "  Feeder " + name + " empty.");
 		}
 
         updateVisionOffsets(nozzle);
+        
+        // automatically disable when we reach the last part, openpnp will try
+        // to pick another feeder that provides the same part for the next feed.
+        if ((maxFeedCount > 0) && (feedCount >= maxFeedCount))
+        {
+        	enabled = false;
+        }
     }
 
     private void updateVisionOffsets(Nozzle nozzle) throws Exception {
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java
@@ -223,20 +223,12 @@ public class ReferenceStripFeeder extends ReferenceFeeder {
     public void feed(Nozzle nozzle) throws Exception {
         setFeedCount(getFeedCount() + 1);
 
-        // normally the feeder will automatically disable when it runs out but
-        // if the user re-enables and tries to feed, throw an exception that we are out of parts.
+        // Throw an exception when the feeder runs out of parts
         if ((maxFeedCount > 0) && (feedCount > maxFeedCount)) {
 			throw new Exception("Tried to feed part: " + part.getId() + "  Feeder " + name + " empty.");
 		}
 
         updateVisionOffsets(nozzle);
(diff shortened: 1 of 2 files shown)


src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java
```diff
@@ -223,11 +223,20 @@ public Location getPickLocation() throws Exception {
...
 
+        // normally the feeder will automatically disable when it runs out but
+        // if the user re-enables and tries to feed, throw an exception that we are out of parts.
         if ((maxFeedCount > 0) && (feedCount > maxFeedCount)) {
 			throw new Exception("Tried to feed part: " + part.getId() + "  Feeder " + name + " empty.");
 		}
 
         updateVisionOffsets(nozzle);
+        
+        // automatically disable when we reach the last part, openpnp will try
```
vonnieda, line 234: Hi @desertgreg - with my recent fixes to the job processor this is handled automatically. The feeder should just throw an exception if it can't feed. Is there a need for this additional change?

desertgreg:
Ok then I just need the new button to automatically compute the
maxfeedcount.  I can make a new pull request.

Greg

src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java
desertgreg: Upon further thought, if I only throw the exception, will a message box pop up and wait for user input?  Thats what was happening for me.  With this change the system seamlessly continues placing parts until the last feeder runs out.  So it might be better to keep this change but I will defer to your opinion :-)

desertgreg: Hi Jason, what do you think?  I tested it again without my auto-disabling code and I think the user experience is worse.  Without this code, the job pauses and puts a message box up that the user has to click through and then they have to press 'play' on the job again to get it going.  With this code in, there is no pause or user interaction needed, openpnp just moves to the next feeder.  I suppose you could implement this on your side though.  You could only show the exception message if you can't find another feeder for the given part.

vonnieda: @desertgreg You need to set the pick retry count for the part in the Parts panel. This activates the new setting. The reason it should be done this way is so that it acts the same for all feeders - and this leaves it up to the user as to whether they want to auto advance to a second feeder.

markmaker:
Hi @desertgreg 

My idea is to create a little testing round with several changes at once, in the hope that we contributors can test each other's code easily.

grafik

_Mark

desertgreg:
Thanks! I seem to have an issue where eclipse is reformatting entire source
files or the source code that is checked in has different formatting from
the recommended "auto-format" because when I went to work on this again, my
next pull request changed the whole file.

markmaker:
Hi Greg

I think the Auto-Format should *not *be applied automatically. I usually 
only apply it to lines I actually changed (select + CTRL-I). There are 
exception like when I (or Windows Builder) changed 80% of the file 
anyway and the indentation is particularly messy (like mixed tabs and 
spaces).

https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style 
<https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style>

_Mark


Am 18.01.2021 um 17:23 schrieb desertgreg:

markmaker:
Github Mailing gateway lost the image, I wanted to say:

> I think the Auto-Format should **not** be applied automatically.
grafik

_Mark

PR #1103, opened 2021-01-15
Fix / Check and Diagnose Pipeline Results

markmaker:
# Description

Pipelines and stages would often generate `java.lang.NullPointerException` when named stages were not found or `java.lang.ClassCastException` when result models had the wrong type, including when Exceptions are already propagated through results. Only few pipeline callers pulled all the tricks to diagnose all possible failure modes correctly. The non-descript Java Exceptions are quite unhelpful for the average user, stack traces do not point to the _cause_ of the failure, but to some obsure _symptoms_ later on. 

* Implemented unified pipeline stage result lookup with the new `CvPipeline.getExpectedResult()` method (in addition to the existing method). If the named stage is missing, an Exception is thrown. 
* Created type safe `CvStage.Result.getExpectedModel(Class<T>)` and `CvStage.Result.getExpectedListModel(Class<T>, Exception)` methods that provide diagnostic model class mismatch and optionally custom empty list Exceptions. Any Exception propagated as a result model is also properly re-thrown.
* Reworked all callers outside and inside the pipeline (CvStages).
* Some usages have been made stricter: if a stage name is given, it must now be found. In CvStages this was often ignored, with the stage silently doing nothing, which was quite hard to diagnose. 

# Justification
See one example here:
https://groups.google.com/g/openpnp/c/pNNzyTCfXRs/m/8d9dtuW1AQAJ

# Instructions for Use
The usage remains the same. Only the disagnostics were improved:

Result stage missing

Wrong result model type

Stage input stage missing

# Implementation Details
1. Tested all pipelines in the simulation machine as far as they are defined there. Provoked all failure modes. Exhaustive tests across all us cases and stages are almost impossible, I therefore propose to merge this and fix issues quickly. I'm ready to act fast. 

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java
@@ -800,16 +800,7 @@ public class ReferenceNozzleTipCalibration extends AbstractModelObject {
             List<Location> locations = new ArrayList<>();
 
             String stageName = VisionUtils.PIPELINE_RESULTS_NAME;
-            Result pipelineResult = pipeline.getResult(stageName);
-            if (pipelineResult == null) {
-                throw new Exception(String.format("There should be a \"%s\" stage in the pipeline.", stageName));
-            }
-
-            Object results = pipelineResult.model;
-
-            if (results instanceof Exception) {
-                throw (Exception)results;
-            }
+            List results = pipeline.getExpectedResult(stageName).getExpectedModel(List.class);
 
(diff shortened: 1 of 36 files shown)


markmaker:
Hi @vonnieda 

I know you are swamped, just a quick assessment of this:

<blockquote>Exhaustive tests across all us cases and stages are almost impossible, I therefore propose to merge this and fix issues quickly. I'm ready to act fast.</blockquote>

Shall I?

_Mark

markmaker:

My idea is to create a little testing round with several changes at once, in the hope that we contributors can test each other's code too.

grafik

_Mark

PR #1112, opened 2021-01-26
Add support for units per pixel calibration at different heights

tonyluken:
# Description

1. Cameras can now be calibrated at two different heights which permits more accurate measurements and camera jogging when the height of an object is known to be at something other than the height at which the camera was calibrated.  When upgrading from earlier versions of OpenPnP, the existing calibration remains in effect until calibration is re-run at two different heights.
2. The camera views are enhanced to allow the operator to adjust the height of the reticle so that it is scaled appropriately for any desired height.  This height is also used to scale camera drag and shift-click jogging so that it is correct for that height.
3. The camera views are enhanced to allow the operator to estimate object heights based on two different perspectives of the object.  This feature is mainly proof-of-concept but is included in case someone finds it useful.
4. Bottom camera drag and shift-click jogging are changed to be more consistent with top camera jogging.  Bottom camera jogging now moves the machine such that the point at the end of the drag (or the point shift-clicked) is moved to the center of the camera view (exactly the same end result as with top camera jogging).  A separate pop-up menu option is added to bottom camera views to allow the operator to quickly move the selected nozzle to the bottom camera much as the original bottom camera jogging worked. 

# Justification
Item 1 above is mostly a stepping stone for future upgrades.  It adds properties and methods to cameras which will allow future upgrades to obtain better estimates of object sizes and distances from the vision system.  For instance, since the height of the top of a strip feeder is known, the vision processing associated with a strip feeder can now be upgraded (but not included with this PR) to accurately scale the images of the strip at that height to get more precise hole locations and hence more accurate pick locations.

Items 2 and 3 above are enhancements that allow the operator to take advantage of the new calibration data when performing manual operations with the camera.

Item 4 above makes bottom camera jogging more intuitive for the user.

# Instructions for Use
1.  The Issues and Solutions system detects if a camera has not been calibrated at two different heights and shows a recommendation to calibrate the camera a two different heights.  The General Configuration tab of the Camera wizard provides instructions on how to perform the new calibrations. Unique instructions are provided to the operator for bottom and top cameras and should be self explanatory.  For top cameras only, a new field called Default Working Height needs to be populated with the height at which objects will be assumed to be if no explicit height information for the object is known .  This should be set to the height where the legacy Units Per Pixel was calibrated which was typically the height of the top surface of the circuit board.  For bottom cameras, the default working height is simply the Z height of the camera so no additional field is needed there.  Only once calibration is completed at two different heights, will Items 2 and 3 become available to the operator.
2. Ctrl-scroll wheel (1 unit per tick) and ctrl-alt-scroll wheel (0.1 unit per tick) change the height at which the reticle is correctly scaled.  If the reticle height is changed from its default, the lower right corner of the camera's view will display the reticle's height along with its default height. A pop-up menu selection "Reset Reticle Height to Default" is also available to quickly reset the reticle height to its default.   Camera drag and shift-click jogging is performed using the same scaling as the reticle.  Example for use: Suppose a board's top surface is at Z=-35mm and a 10mm tall part is sitting on the board.  With the top camera over the part, the reticle height can be changed to -35+10 = -25mm and the reticle will be correctly scaled for the top surface of the part.   Drag and/or Shift-Click jogging can then be used to accurately jog around on the top surface of the part.
3. A pop-up menu option "Estimate Z Coordinate of Object" allows the operator to use the camera to estimate the height of a feature.  When the menu option is selected, instructions will appear guiding the operator on how to make the measurement.  They should be self explanatory (basically click on the feature, jog to a different location, and click on the feature again).
4.  Bottom camera jogging now works similar to the top camera jogging.  Grab the drag handle with the mouse and the location where the mouse is released will move to the center of the camera view.  Shift-click on a location and it will move to the center of the camera view.  These are best observed with a part on the nozzle over the bottom camera.

# Implementation Details
1. **How did you test the change?** Started with a legacy machine.xml file and verified camera vision still worked as it had previously.  Checked the machine Issues and Solutions tab and verified both cameras had a recommendation to calibrate at two different heights.  Performed the calibrations and verified the the issues cleared.  Places a ruler over/under the camera at different heights to verify that the reticle scaling was correct at those heights and also that jogging was accurate at those heights.  Selected the "Estimate Z Coordinate of Object" menu selection and followed to the instructions that appeared - verified that the estimated height was close (within a millimeter) to the height measured by touching the nozzle to the object.  Tested the bottom camera jogging changes by picking a part with the nozzle and moving it over the bottom camera.  Tried jogging to various pins of the part and they moved to the center of the camera view as expected.
2. **Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?**  Yes.
3. **If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.**  Changes were made to the Camera interface to add new interfaces to allow calibration at two different heights and allow that calibration information to be used.  No existing interfaces were changed.  The new methods were implemented in AbstractCamera and only one existing method (getUnitsPerPixel()) was modified so that any legacy code that calls getUnitsPerPixel() works correctly both before and after calibration is performed at two different heights.
4. **Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.**  The Maven tests were run and all passed.

### DIFF
--- a/src/main/java/org/openpnp/spi/base/AbstractCamera.java
+++ b/src/main/java/org/openpnp/spi/base/AbstractCamera.java
@@ -303,7 +303,7 @@ public abstract class AbstractCamera extends AbstractHeadMountable implements Ca
      * @param zCoordinate
      * @return the height above the camera in the same units as zCoordinate
      */
-    protected Length getHeightAboveCamera(Length zCoordinate) {
+    public Length getHeightAboveCamera(Length zCoordinate) {
         Location cameraLocation = getLocation();
         try {
             //Replace virtual axis coordinates, if any, with the head offset
@@ -321,7 +321,7 @@ public abstract class AbstractCamera extends AbstractHeadMountable implements Ca
      * @param location - the location whose height is being determined
      * @return the height above the camera in the same units as location
      */
-    protected Length getHeightAboveCamera(Location location) {
+    public Length getHeightAboveCamera(Location location) {
         return getHeightAboveCamera(location.getLengthZ());
     }
     
@@ -384,12 +384,24 @@ public abstract class AbstractCamera extends AbstractHeadMountable implements Ca
         this.unitsPerPixel = unitsPerPixel;
     }
 
-    @Override
+    /**
+     * Gets the secondary units per pixel
+     * 
+     * @return a location whose x and y coordinates are the measured pixels per unit for those axis
+     *         respectively and the z coordinate is the height above the camera at which the
+     *         measurements were made.
+     */
     public Location getUnitsPerPixelSecondary() {
         return unitsPerPixelSecondary;
     }
 
-    @Override
+    /**
+     * Sets the secondary units per pixel
+     * 
+     * @param unitsPerPixelSecondary - a location whose x and y coordinates are the measured pixels
+     * per unit for those axis respectively and the z coordinate is the height above the camera at
+     * which the measurements were made.
+     */
     public void setUnitsPerPixelSecondary(Location unitsPerPixelSecondary) {
         this.unitsPerPixelSecondary = unitsPerPixelSecondary;
     }
@@ -404,7 +416,16 @@ public abstract class AbstractCamera extends AbstractHeadMountable implements Ca
         this.defaultZ = defaultZ;
     }
 
-    @Override
+    /**
+     * Estimates the Z height of an object based upon the observed units per pixel for the
+     * object. This is typically found by capturing images of a feature of the object from two
+     * different camera positions. The observed units per pixel is then computed by dividing the
+     * actual change in camera position (in machine units) by the apparent change in position of the
+     * feature (in pixels) between the two images.
+     *
+     * @param observedUnitsPerPixel - the observed units per pixel for the object
+     * @return - the estimated Z height of the object
+     */
     public Length estimateZCoordinateOfObject(Location observedUnitsPerPixel) throws Exception {
         LengthUnit units = observedUnitsPerPixel.getUnits();
         double uppX = Math.abs(observedUnitsPerPixel.getX());
--- a/src/main/java/org/openpnp/spi/base/AbstractCamera.java
+++ b/src/main/java/org/openpnp/spi/base/AbstractCamera.java
@@ -7,6 +7,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
@@ -29,14 +30,22 @@ import org.openpnp.CameraListener;
 import org.openpnp.ConfigurationListener;
 import org.openpnp.gui.MainFrame;
 import org.openpnp.gui.support.Icons;
+import org.openpnp.machine.reference.axis.ReferenceVirtualAxis;
+import org.openpnp.machine.reference.axis.ReferenceControllerAxis.BacklashCompensationMethod;
 import org.openpnp.model.Configuration;
+import org.openpnp.model.Length;
 import org.openpnp.model.LengthUnit;
 import org.openpnp.model.Location;
+import org.openpnp.model.Solutions;
+import org.openpnp.model.Solutions.Issue;
+import org.openpnp.model.Solutions.Severity;
+import org.openpnp.spi.Axis;
 import org.openpnp.spi.Camera;
 import org.openpnp.spi.Head;
 import org.openpnp.spi.HeadMountable;
 import org.openpnp.spi.MotionPlanner.CompletionType;
 import org.openpnp.spi.VisionProvider;
+import org.openpnp.spi.Axis.Type;
 import org.openpnp.util.OpenCvUtils;
 import org.openpnp.util.SimpleGraph;
 import org.pmw.tinylog.Logger;
@@ -54,8 +63,31 @@ public abstract class AbstractCamera extends AbstractHeadMountable implements Ca
     @Attribute
     protected Looking looking = Looking.Down;
 
+    /**
+     * The primary units per pixel for this camera. This is used to convert the apparent size (in
+     * pixels) of an object's image to an estimate of its physical size (in units). Note that this
+     * conversion is only valid for objects at the same height above the camera at which the
+     * calibration of the units per pixel was performed. The units per pixel z-coordinate contains
+     * the height above the camera at which the conversion is valid. Also see unitsPerPixelSecondary
+     * below.
+     */
     @Element
     protected Location unitsPerPixel = new Location(LengthUnit.Millimeters);
+
+    /**
+     * The secondary units per pixel for this camera. This is typically calibrated at a different
+     * height above the camera than the primary units per pixel (above) so that the two together
+     * can be used compute an object's true size (in units) assuming its actual height above the
+     * camera is known.
+     */
+    @Element(required = false)
+    protected Location unitsPerPixelSecondary = null;
+
+    /**
+     * The Z coordinate at which objects are assumed to be if their true height is unknown. 
+     */
+    @Element(required = false)
+    protected Length defaultZ = null;
 
     @Element(required = false)
     protected VisionProvider visionProvider;
@@ -171,6 +203,17 @@ public abstract class AbstractCamera extends AbstractHeadMountable implements Ca
                 settleMethod = SettleMethod.FixedTime;
             }
         }
+        
+        // If the secondary units per pixel doesn't exist, we're migrating from the old single
+        // height calibration to the new dual height calibration.  Since a second calibration hasn't
+        // actually been performed yet, the secondary units per pixel is just set to the same as the
+        // primary setting.  Since they are then the same, height above the camera is irrelevant so
+        // we just initialize the default Z to zero.  This will ensure legacy units per pixel
+        // calibrations will be correctly handled.
+        if (unitsPerPixelSecondary == null) {
+            unitsPerPixelSecondary = unitsPerPixel.derive(null, null, null, null);
+            defaultZ = new Length(0.0, LengthUnit.Millimeters);
+        }
     }
 
 
@@ -254,13 +297,169 @@ public abstract class AbstractCamera extends AbstractHeadMountable implements Ca
     }
 
     @Override
+    public Location getActualLocation() {
+        Location location = getLocation();
+        LengthUnit units = location.getUnits();
+        //Check for any virtual axis and replace them with their home coordinate
+        Axis axis = getAxisX();
+        if ((axis != null) && (axis.getClass() == ReferenceVirtualAxis.class)) {
+            Length home = ((ReferenceVirtualAxis) axis).getHomeCoordinate().convertToUnits(units);
+            location = location.derive(home.getValue(), null, null, null);
+        }
+        axis = getAxisY();
+        if ((axis != null) && (axis.getClass() == ReferenceVirtualAxis.class)) {
+            Length home = ((ReferenceVirtualAxis) axis).getHomeCoordinate().convertToUnits(units);
+            location = location.derive(null, home.getValue(), null, null);
+        }
+        axis = getAxisZ();
+        if ((axis != null) && (axis.getClass() == ReferenceVirtualAxis.class)) {
+            Length home = ((ReferenceVirtualAxis) axis).getHomeCoordinate().convertToUnits(units);
+            location = location.derive(null, null, home.getValue(), null);
+        }
+        axis = getAxisRotation();
+        if ((axis != null) && (axis.getClass() == ReferenceVirtualAxis.class)) {
+            Length home = ((ReferenceVirtualAxis) axis).getHomeCoordinate().convertToUnits(units);
+            location = location.derive(null, null, null, home.getValue());
+        }
+        return location;
+    }
+    
+    @Override
+    public Location getActualLocation(HeadMountable tool) {
+        if (tool != null) {
+            return getActualLocation().subtract(tool.getCameraToolCalibratedOffset(this));
+        }
+        return getActualLocation();
+    }
+
+    /**
+     * Gets the height above the camera of the specified z coordinate
+     * 
+     * @param zCoordinate
+     * @return the height above the camera in the same units as zCoordinate
+     */
+    protected Length getHeightAboveCamera(Length zCoordinate) {
+        return zCoordinate.subtract(getActualLocation().getLengthZ());
+    }
+    
+    /**
+     * Gets the height above the camera of the specified location
+     * 
+     * @param location - the location whose height is being determined
+     * @return the height above the camera in the same units as location
+     */
+    protected Length getHeightAboveCamera(Location location) {
+        return getHeightAboveCamera(location.getLengthZ());
+    }
+    
+    /**
+     * Gets the Z coordinate that is a given height above the camera
+     * 
+     * @param heightAboveCamera
+     * @return the Z coordinate that is the given height above the camera
+     */
+    protected Length getZCoordinateAboveCamera(Length heightAboveCamera) {
+        return heightAboveCamera.add(getActualLocation().getLengthZ());
+    }
+    
+    @Override
     public Location getUnitsPerPixel() {
-        return unitsPerPixel;
+        return getUnitsPerPixel(defaultZ);
+    }
+
+    @Override
+    public Location getUnitsPerPixel(Length atHeight) {
+        if (atHeight == null) {
+            atHeight = defaultZ;
+        }
+        LengthUnit units = unitsPerPixel.getUnits();
+        Location uppCal1 = unitsPerPixel;
+        Location uppCal2 = unitsPerPixelSecondary.convertToUnits(units);
+        if (uppCal1.getZ() == uppCal2.getZ()) {
+            // Calibration wasn't performed at two different heights above the camera so just
+            // return the primary units per pixels
+            return unitsPerPixel;
+        }
+
+        double heightAboveCamera = getHeightAboveCamera(atHeight).convertToUnits(units).getValue();
+
+        // Linearly interpolate between the two calibration points
+        double k = (heightAboveCamera - uppCal2.getZ()) / (uppCal1.getZ() - uppCal2.getZ());
+        return new Location(units, k * (uppCal1.getX() - uppCal2.getX()) + uppCal2.getX(),
+                k * (uppCal1.getY() - uppCal2.getY()) + uppCal2.getY(), heightAboveCamera, 0.0);
+    }
+
+    @Override
+    public Location getUnitsPerPixel(Location location) {
+        if (location == null) {
+            return getUnitsPerPixel(defaultZ);
+        }
+        return getUnitsPerPixel(location.getLengthZ());
     }
 
     @Override
     public void setUnitsPerPixel(Location unitsPerPixel) {
         this.unitsPerPixel = unitsPerPixel;
+    }
+
+    @Override
+    public Location getUnitsPerPixelSecondary() {
+        return unitsPerPixelSecondary;
+    }
+
+    @Override
+    public void setUnitsPerPixelSecondary(Location unitsPerPixelSecondary) {
+        this.unitsPerPixelSecondary = unitsPerPixelSecondary;
+    }
+
+    @Override
+    public Length getDefaultZ() {
+        return defaultZ;
+    }
+
+    @Override
+    public void setDefaultZ(Length defaultZ) {
+        this.defaultZ = defaultZ;
+    }
+
+    @Override
+    public Length estimateZCoordinateOfObject(Location observedUnitsPerPixel) throws Exception {
+        LengthUnit units = observedUnitsPerPixel.getUnits();
+        double uppX = Math.abs(observedUnitsPerPixel.getX());
+        double uppY = Math.abs(observedUnitsPerPixel.getY());
+
+        Location uppCal1 = unitsPerPixel.convertToUnits(units);
+        Location uppCal2 = unitsPerPixelSecondary.convertToUnits(units);
+
+        if (uppCal1.getZ() == uppCal2.getZ()) {
+            throw new Exception("Camera Units Per Pixel has not been calibrated at two different " +
+                    "heights.");
+        }
+        
+        if (!Double.isFinite(uppX) && !Double.isFinite(uppY)) {
+            throw new Exception("Apparent change in position or apparent size of object feature " +
+                    "is too small to estimate object Z coordinate.");
+        }
+
+        if (uppX == 0 && uppY == 0) {
+            throw new Exception("Actual change in camera position or actual feature size too " +
+                    "small to estimate object Z coordinate.");
+        }
+
+        // Compute the ratio of where the measurement falls between the two cal points using
+        // whichever measurement is larger for better accuracy
+        double k;
+        if (!Double.isFinite(uppY) || (uppX > uppY)) {
+            k = (uppX - uppCal2.getX()) / (uppCal1.getX() - uppCal2.getX());
+        }
+        else {
+            k = (uppY - uppCal2.getY()) / (uppCal1.getY() - uppCal2.getY());
+        }
+        
+        // Compute the Z offset relative to the camera
+        double heightAboveCamera = k * (uppCal1.getZ() - uppCal2.getZ()) + uppCal2.getZ();
+        
+        return getZCoordinateAboveCamera(new Length(heightAboveCamera, units));
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


src/main/java/org/openpnp/spi/base/AbstractCamera.java
```diff
@@ -253,16 +296,172 @@ public Location getLocation(HeadMountable tool) {
...
+
+        Location uppCal1 = unitsPerPixel.convertToUnits(units);
+        Location uppCal2 = unitsPerPixelSecondary.convertToUnits(units);
+
+        if (uppCal1.getZ() == uppCal2.getZ()) {
+            throw new Exception("Camera Units Per Pixel has not been calibrated at two different " +
+                    "heights.");
+        }
+        
+        if (!Double.isFinite(uppX) && !Double.isFinite(uppY)) {
```
markmaker, line 439: Should this be:
 if (!(Double.isFinite(uppX) && Double.isFinite(uppY))) {

markmaker:
Hi @tonyluken 

This looks great! I'll be testing that as soon as I can. 

Some questions/suggestions re the code:

**Get Actual Location**

1. First, to know we're on the same page, I assume you need `Camera.getActualLocation()` only as an arbitrary relative reference, in case the camera is physically moved in Z. As we discussed earlier, you interpolate between two measurements, and you could calculate the true optical image plane from there, but `Camera.getActualLocation()` does not need to give the true image plane. I ask, because I see no way to know where the camera's true image plane really is. Multi-lens optics will completely obfuscate that, I believe. 
2. In `Camera.getActualLocation()` you filter out virtual axes and use their home (=Safe Z) coordinate as the true camera position. I think this is wrong in the general case (it is on my machine). The virtual axis is a proxy for the nozzle Z axis, it stores the Z when switching back and forth between tools. The home coordinate doubles as Safe Z and should generally match the one on the nozzle. E.g. on my machine it is currently at -22mm and it is subject to change, if one day I have a taller part or feeder. 
3. I believe you should instead take the camera's Head Offsets coordinates. That's the closest to the real Z location of a fixed-in-Z camera. It is typically simply at 0 but that's OK if (1) is true.
4. Furthermore, it would be nice, if you could integrate the `getActualLocation()` function into the `AbstractHeadMountable.getApproximativeLocation(Location, Location, LocationOption...)` and make a new `LocationOption.NonVirtual` or similar. It could then filter out the virtual axes coordinates and replace them with 0.0 in the `axis` loop. The 0.0 would then be offset by the Head Offset in the `toHeadMountableLocation()`. 
5. This would also keep all the axis knowledge out of `AbtractCamera`, where IMHO it does not belong. The many new imports could then be removed. 

**Camera interface**

Jason seemed to keep the spi interfaces as terse as possible. Therefore, you should probably move `estimateZCoordinateOfObject(), setDefaultZ(), getDefaultZ(), setUnitsPerPixelSecondary(), getUnitsPerPixelSecondary()` down to the AbstractCamera i.e. to the actual implementation. `getActualLocation()` can be removed, see above. `VisionUtilsTest` is then also not burdened by it. 

**Use Cases**

I wonder how this will be applied (other than in the camera view). I assume every call to `Camera.getUnitsPerPixel()` will have to be reviewed and if the Z is known (e.g. feeder pick location Z in a ReferenceStripFeeder, BlindsFeeder, ReferencePushPullFeeder and others), it should be passed, right?

And in the testing version with the Auto Camera View, we could optionally tell it to `setViewingPlaneZ()` to the semantic Z. This would probably need to be reset back to the default, when we move the camera away some distance. Similar to the Virtual Z to Safe Z reset, I implemented. 

**Branch to merge**

I assume I could then merge this into **test** as well, right?

_Mark

src/main/java/org/openpnp/spi/base/AbstractCamera.java
```diff
@@ -253,16 +296,172 @@ public Location getLocation(HeadMountable tool) {
         return super.getLocation();
     }
 
+    @Override
+    public Location getActualLocation() {
+        Location location = getLocation();
+        LengthUnit units = location.getUnits();
+        //Check for any virtual axis and replace them with their home coordinate
+        Axis axis = getAxisX();
+        if ((axis != null) && (axis.getClass() == ReferenceVirtualAxis.class)) {
```
markmaker, line 305: Checks such as these

` if ((axis != null) && (axis.getClass() == ReferenceVirtualAxis.class)) {`

should be

` if (axis instanceof ReferenceVirtualAxis) {`

Note, `instanceof` includes the `null` test.

src/main/java/org/openpnp/spi/base/AbstractCamera.java
tonyluken: I don't think so - as long as we have one finite non-zero value we can calculate the height.  For example, if the machine is only moved in one axis the other axis may have a zero answer (or Nan if the pixels in that direction also didn't change).  The other case is the machine barely moved in one axis but the pixels in that direction didn't change so you get infinity for an answer on that axis.  I'll add a comment to the code to make that clear.

src/main/java/org/openpnp/spi/base/AbstractCamera.java
tonyluken: **Get Actual Location**

1. Yes, ActualLocation() does not need to know anything about the true image plane (that is where the image is in perfect focus) but it does need to be fixed in relation to the physical location of the camera.
2. Through 5. Yes, that all makes sense to me, I'll see if I can make those changes.

**Camera interface**
Ok, I can do that.

**Use Cases**
Yes, eventually all the places where getUnitsPerPixel() is called should be reviewed to see if a known height can be passed (that part of the future upgrades I mentioned).  Before then, they should continue to work as they have in the past.

As far as the Auto Camera View I'll let you decide what is best there since I haven't tried it yet.

**Branch to merge**
Yes, I soon as I make the suggested updates I think it could be merged into **test**

src/main/java/org/openpnp/spi/base/AbstractCamera.java
markmaker: Ah, I see it now. Thanks.

`if (!(Double.isFinite(uppX) || Double.isFinite(uppY))) {`

being the ideal ;-)

src/main/java/org/openpnp/spi/base/AbstractCamera.java
tonyluken: @markmaker,

I'm running into problems moving the new methods from the Camera interface to AbstractCamera.  There seems to be several places where Camera is used where I need access to at least some of those methods.  Should I change as many as I need to AbstractCamera (which may lead to a big chain of changes) or should I leave them as I originally had?  Also you didn't mention getUnitsPerPixel(Length z) and getUnitsPerPixel(Location location) should those be moved as well?

Oh, and I forgot to previously mention - yes to instanceof.

Tony

markmaker: As far as I've understood, `getUnitsPerPixel(Length z)` and `getUnitsPerPixel(Location location)` are the real interface. They are valid additions to `Camera`. The others should be internal. 

I've seen the use in `isReticleHeightChangable()` but this can equally be expressed in comparing `getUnitsPerPixel(Length z)` at two arbitrary `z`, right?

Other uses?

_Mark

tonyluken: If I use your suggestion to move things out of Camera into AbstractCamera then don't I need to change CameraView.setCamera(Camera camera) to CameraView.setCamera(AbstractCamera camera) because at least some of the new methods are used in CameraView?  And then I think there are several of places in CameraPanel which will need to change from Camera to AbstractCamera.  Is that what you really want or am I missing something?

Tony

markmaker: I guess `getDefaultZ()` must be added to `Camera` too. Other than that, what are you missing?

_Mark

tonyluken:
Ok Mark, I think I made all the requested changes.  Please be sure to carefully review what I did in getApproximativeLocation because I'm not positive what I did there is totally correct (although it appears to work for me).

markmaker:
Big thanks, @tonyluken 

Just one thing: it flipped line-endings on `AbstractHeadMountable.java`:

https://github.com/openpnp/openpnp/pull/1112/files#diff-c45b0f533875899da5ad3f8fd85d6c26e8aa2145db945cee8772174486b6be9dL1

Have you investigated? (Have you got the mails I sent you at all?)

_Mark

tonyluken:
@markmaker,

It appears the current version of AbstractHeadMountable.java in the repository has the wrong line endings (CRLF) rather than LF like it should.  I suspect someone's git configuration is not setup correctly.  Like I said previously, someone with more git knowledge than me should update the developer's section of the wiki to ensure everyone sets-up their git configuration correctly.  I think it amounts to setting core.autocrlf=true but I'm not positive about that.

I probably got your emails but just haven't seen them because I get my email on another computer than where I'm typing right now.

markmaker:
You're right, that's from the initial commit when I created that file:
```
git show  f6713b6ae01891fad9628bb19062e715d1b6cb6a:src/main/java/org/openpnp/spi/base/AbstractHeadMountable.java | file -
/dev/stdin: ASCII text, with CRLF line terminators

```

I read all about these settings (see the mails I sent you for details) and I have definitely set them correctly. **But they obviously don't work reliably.** 

Good night,
Mark

tonyluken:
@markmaker,

I checked my email and the only ones I see from you are those generated by github.  If you send me email directly, I did not see it - maybe I accidentally deleted it thinking it was a copy of what is posted here.

markmaker:
This is a re-post of said E-Mail (slightly edited): 
___
`2021-01-21 19:52`

Hi Tony

As you might have seen, I suddenly had issues with line-endings too. I almost gave up trying to find a solution... git is one hell of a source of frustration.

Finally, I found that the files actually had the wrong CR+LF line endings **inside the repo.**

First I found no straight-forward way to find out what is actually stored inside the repo (binary). All the tools seemingly operate on the checked-out files and those are fine, because the automatic core.autocrlf conversion takes place back and forth. That's also why I thought the conflicts went away when I merged your last PR (while resolving true conflicts too). But it turned out the line-ending conflicts were merely invisible while checked out, they immediately reappeared when commited ... which I overlooked ;-(.

So I backtracked some files (as an example, FeedersPanel.java ) and I saw that it already flipped all lines here:

  ```
  Revision: 690919aae3bdbc4e29bd7b590684ffbc6cc4f786
    Author: tonyluken <tonyluken@att.net>
    Date: 19.12.2019 04:21:01
    Message:
    Automatically update the feeder table when the wizard's Apply button is
    pressed
    ----
    Modified: src/main/java/org/openpnp/gui/FeedersPanel.java
    Modified: src/main/java/org/openpnp/gui/support/PropertySheetWizardAdapter.java
    Modified: src/main/java/org/openpnp/spi/Feeder.java
    Modified: src/main/java/org/openpnp/spi/base/AbstractFeeder.java
```

I then investigated some more and found that you can use `git show` to cat a file straight from the repo without applying the `core.autocrlf `conversion:

```
> git show 690919aae3bdbc4e29bd7b590684ffbc6cc4f786:src/main/java/org/openpnp/gui/FeedersPanel.java | file -
/dev/stdin: ASCII text, with CRLF line terminators
```

https://stackoverflow.com/a/64215694

It seems that some actions with git save all the lines back with new line-endings. Others seem to preserve the line-ending convention already present in the repo file. This explains why some of my commits came after yours and still didn't flip the line-endings back. This time I must have done something differently, perhaps I used a different git command or tool, i.e. the line-endings were not preserved this time and the files flipped back.

Whatever, it seems, these originate from your commits. **[EDIT: and now also from mine]** I've just spent hours experiencing how intransparent and frustrating the whole "git experience" can be, so no blame at all here. In fact, I fully blame the tool, it should really prevent such a thing! 

Have you checked your `core.autocrlf` settings in the openpnp repo?

```
cd openpnp
git config --get core.autocrlf
```

But don't fix it just yet, see below. **[EDIT: after my own experience now, I presume the setting is already OK and it's just git not adhering to it]**

I'm investigating adding a `.gitattributes` to (hopefully) enforce this in the future, i.e. regardless of user's settings:

https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings#per-repository-settings

I would be grateful, if you could second-guess my analysis, and perhaps test a `.gitattributes` on your side. 

Ideally, adding it would immediately remedy the situation, despite a verified wrong `core.autocrlf `!
Thanks!

_Mark

tonyluken:
I know I am at least partially guilty as I didn't have git properly setup during some of my earlier commits but I fixed that quite some time ago (probably more than a year ago although I can't remember exactly when).  I know I now have core.autocrlf set to true:

Tony@CAD-1 MINGW64 ~/git/openpnp (feature/UnitPerPixelCalibrationAtTwoZHeights)
$ git config --get core.autocrlf
true

I also have this setting with a comment "convert line endings during a merge (like autocrlf does during a pull/commit) to avoid line-ending-only conflicts"
Tony@CAD-1 MINGW64 ~/git/openpnp (feature/UnitPerPixelCalibrationAtTwoZHeights)
$ git config --get merge.renormalize
true

markmaker:
Hi @tonyluken 

I'm in the process of merging your PR into `test`, it's a hell of conflicts.

So I am going through it line by line and I saw this:

https://github.com/openpnp/openpnp/blob/9fd863ca30f10d1f7d612b6f5f57cb681b2ed27b/src/main/java/org/openpnp/gui/components/CameraView.java#L684-L685

"Reticle Height"? Wouldn't it be more understandable if this was "Z Focus" or "Focus Z"? 
(I know it is not strictly speaking the optical focus, having a fixed lens, but it is the "effective focus plane" for computer vision.) 

And instead of showing the "Default" height, wouldn't it be more useful to show the difference?

I'll make these changes, we can always change it back while testing.

_Mark

markmaker:
Hi @tonyluken

I'm trying to test your function to see if I haven't botched it in the gruesome conflict resolution. 

I seemed too stupid to use it, so I went to the code to figure it out. It's more or less clear now, and all "correct" mathematically, but it was not transparent enough for noobs like me to understand. At least not if you **want** to understand for the purpose of this test. Also had trouble with the LocationButtonsPanel, because X, Y fields are not present. 

So I'll propose a more "kindergarten" like GUI. 

There's one technical thing I'd like to change, to simplify things: Currently the measurements are relative to the "Height above Camera".  Part of my confusion came from the fact that the two reference planes are completely different, one time it's the nominal focal plane (bottom camera), one time a plane at head offset (top camera). As we discussed before, this is an arbitrary relative distance, with no optical significance (an artifact of nozzle homing, actually).

I agree that a camera position _relative_ reference is relevant in _theory_. But it has IMHO no _practical_ use in OpenPnP:

1. Using it with a Z-movable camera is non-sensical. If the Z information is available, you would move the camera to focus the subject. So the Units per Pixel would actually remain constant. AFAIK, this is the very purpose of a movable camera. 
2. If you change the top camera mount Z, you still don't change the head offset, you'd typically just leave it at 0, otherwise the whole meaning of the OpenPnP coordinate system changes. So you can't reuse the Upp and must recalibrate. 
3. If you change the bottom camera's mount, chances are 99% you also change the focal plane, so again you can't reuse the Upp and must recalibrate. 
4. Currently, the OpenPnP coordinate systems usually has Z=0 at the nozzle home-to max or nozzle balance point, relative to nozzle tip points. If we were to change things one day, with Z=0 at PCB height, Z=0 would actually become a very useful reference point for Upp stuff. 

All this leads to my proposal to just use absolute Z as the reference. It would make it so much easier on the GUI. 

Veto?

_Mark

tonyluken:
Hi @markmaker,

First I'm a Git/GitHub noob so I don't understand what it means when is says there are conflicting files when it also says that they can be automatically merged:

image

> "Reticle Height"? Wouldn't it be more understandable if this was "Z Focus" or "Focus Z"
I struggled with the correct terminology there but I still prefer "Reticle Height" as the word focus has different connotations to me.  What the operator sees when they change the setting is the reticle changing so that is why I chose that.   

> And instead of showing the "Default" height, wouldn't it be more useful to show the difference?
I'm fine with that.

What in particular did you find difficult to use?  The gui instructions or the code or both?  If the instructions are difficult to follow correctly, then they need to be fixed.  I do plan on updating the wiki to add more more detail on how this all works.  Also keep in mind that once we change to a more automated method of calibration, this part of the gui will most likely go away entirely.

> Using it with a Z-movable camera is non-sensical.
What about if the Z height is unknown and you want to use the movable camera to estimate the height?  To do that you still need two upp at two different distances from the camera.

markmaker:
> First I'm a Git/GitHub noob so I don't understand what it means when is says there are conflicting files when it also says that they can be automatically merged

I changed it to merge into `test` as discussed. Don't know why that is not updated on your side.  

> I struggled with the correct terminology there but I still prefer "Reticle Height" as the word focus has different connotations to me. What the operator sees when they change the setting is the reticle changing so that is why I chose that.

First I find "Height" etc. always difficult, because we're looking at a 2D camera image and 2D reticle, "height" is ambiguous. At least in the code width and height are often used for X and Y. Even if it isn't ambiguous with proper math, English background and university schooling, for a mixed international audience, I prefer the unambiguous Z. 

I see that Reticle ruler is involved, but I don't think the Reticle is actually the important thing, here. You did not solely develop it, so the Reticle ruler works better, right? You did develop it, so Computer Vision can eventually make better Upp based measurements. E.g. my BlindsFeeder detecting the blinds always suffered from Viewing Plane errors, because the feeders obviously have varying heights. I'm sure it will benefit a lot, once I plug in the tape surface Z, that I know. 

So if you don't like "Focus Z", what about "Viewing Plane Z"? 

> What in particular did you find difficult to use? The gui instructions or the code or both? If the instructions are difficult to follow correctly, then they need to be fixed. I do plan on updating the wiki to add more more detail on how this all works. Also keep in mind that once we change to a more automated method of calibration, this part of the gui will most likely go away entirely.

It is difficult to explain. There are many small things that are not dramatic in themselves, but put together they threw me. Some of it was already strange in the previous version of the calibration, it just got stranger by duplicating it. I made notes, but when I tried to report them to you, I seemed unable not to make them sound very offending, so I'll skip that. More constructive to make a new proposal. I've committed the merge after conflict resolution, so we can go back.

> What about if the Z height is unknown and you want to use the movable camera to estimate the height? To do that you still need two upp at two different distances from the camera.

If we really ever implement a movable camera, this could be revisited. Hopefully, the automatic calibration is ready by then. Unlike the current calibration, the movable camera does no longer need two objects with different surfaces, instead it can move itself. It does therefore not make sense, IMHO, to make today's calibration compatible with such a camera. 

With an automatic calibration, everything can be automated using a single fiducial (movable camera) or two fiducials at different height (fixed camera). Unless I'm missing something you'd always need to determine the true fiducial Z's first. 

I'm working on an extension of the ContactProbeNozzle that supports Vacuum sniffling for Nozzles without switch and automatic calibration of Nozzle Tip Z on a reference surface. So even the probing could then be automated. 

_Mark

markmaker:
@tonyluken 

One more thing. If you think about developing more stuff, please base it on `test`. Otherwise, this will become impossible to untangle later. 

Thanks,
_Mark

tonyluken:
Hi Mark,

> First I find "Height" etc. always difficult, because we're looking at a 2D camera image and 2D reticle, "height" is ambiguous. At least in the code width and height are often used for X and Y. Even if it isn't ambiguous with proper math, English background and university schooling, for a mixed international audience, I prefer the unambiguous Z.

I agree that Z is certainly less ambiguous than height so I'm ok with changing from height to Z.

> I see that Reticle ruler is involved, but I don't think the Reticle is actually the important thing, here. You did not solely develop it, so the Reticle ruler works better, right? You did develop it, so Computer Vision can eventually make better Upp based measurements. E.g. my BlindsFeeder detecting the blinds always suffered from Viewing Plane errors, because the feeders obviously have varying heights. I'm sure it will benefit a lot, once I plug in the tape surface Z, that I know.

But the section of code you are talking about here only affects what the operator sees in the camera view during manual operations and the reticle is the one thing that the operator can see change when that setting is changed.  Nothing else visible to the operator changes with that setting.  And the reticle height (now Z) has nothing to do with normal machine operation and in none of the other places you mentioned would it come into play.  In all those places, you would simply replace calls to getUnitsPerPixel() with getUnitsPerPixel(z).

> So if you don't like "Focus Z", what about "Viewing Plane Z"?

How about "Reticle Z Coordinate"?

> If we really ever implement a movable camera, this could be revisited. Hopefully, the automatic calibration is ready by then. Unlike the current calibration, the movable camera does no longer need two objects with different surfaces, instead it can move itself. It does therefore not make sense, IMHO, to make today's calibration compatible with such a camera.

Originally, I hadn't even thought about cameras movable in Z but I added support for it because Jason asked about it on the OpenPnP forum post.  The calibration instructions do allow for moving the camera in Z rather than moving the object in Z: "Place a spacer (about as thick as the tallest part that may ever be expected to be placed) under the object (**or for machines that can physically move the top camera in Z, jog the camera up or down by about that much**) and repeat...."

markmaker:
Hi @tonyluken 

This is way too much discussion, but I like the insights that come out of it. So thanks for bearing with me. :-D

> How about "Reticle Z Coordinate"?

First I should admit that "Reticle" is not a word I knew until I looked it up for your PR. I always shrugged it away when using OpenPnP,  I did understand "Crosshairs" and "Ruler". 
reticle

Does this count as a reason? For an international audience? ;-)

> But the section of code you are talking about here only affects what the operator sees in the camera view during manual operations 

I should explain: in the test branch there is this new `MovableUtils.fireTargetedUserAction()` function that is called whenever the user deliberately targets  a tool towards a location. I've revised all of the OpenPnP GUI actions to use this function where appropriate. This is used to automatically switch the `CameraView` to the right camera, switch on the light, capture a frame for a `0 fps` camera and/or select the right tool in `MachineControls` panel (all these are optional). 

https://github.com/openpnp/openpnp/blob/3f3af9e4dd5f12ece9ce3f244e378e9d6c36ec8a/src/main/java/org/openpnp/util/MovableUtils.java#L50

Obviously this should be extended to include automatic adjustment of the viewing plane Z in the CameraView. When I press the "Move Camera To Pick Location" button, the function would adjust the viewing plane Z to the pick location Z. I automatically see the pockets and sprocket holes correctly spaced in the reticle ruler. The camera is effectively "targeted" at the Z. If we ever have a camera that can move in Z, it would now be moved. 

In fact, as I'm writing this, I wonder under which conditions the Virtual Z should be bound to the Viewing Plane Z and vice versa? I guess it could be whenever it is not at Safe Z.

Btw, what does a movable camera do, when it is told to move at Safe Z? Will it be out of focus? I guess so. 

For fixed cameras, binding Upp to Virtual Z (except at Safe Z) would have the benefit that the user can alternatively use Z Machine Controls instead of the obscure Ctrl-Scroll. And it would clarify the role and state of the Virtual Z axis. It would also reset to "Default Working Plane" if the user presses the Z **P** button (Safe Z). If we move the camera away from a location it would also automatically do that (it was your input that a relocated Virtual Z was unsafe, remember?). 

The longer I think about it, we should call it "Target Z". Is this enough of a hint at "Reticle"? ;-) 

Or even just "Z"?
 
Regarding the movable camera and the Z reference point: I'll try out an idea...

_Mark

tonyluken:
Hi Mark,

Yeah, I think we've discussed this one enough, I hadn't realized you were automating the whole thing - in that case I'll defer to you as to how it should be named.

Tony

markmaker:
Hi @tonyluken,

Yesterday, I tested this on my machine, where I'm in the process of changing my table height, so this was quite instructive. There were some chicken and egg problems, the Units per Pixel being an elementary requirement i.e. you might not yet have your head offset, Safe Z etc. set up properly when you need to set the Upp. So I felt the old 2D mode without automatic machine motion should still be possible. So I made it switchable:

2D

This also helps simplify the UI for the beginner. 

Only later you can then switch to the 3D mode:

3D

And it will still warn the eager user before motion:

Warn Movement

Despite my meddling on the surface, the "inner values" are practically unchanged. I think this is a great addition, and it will become more valuable when plugged into the known Z heights during vision. And those Z heights will become more with the contact Z probing I'm also working on. 

I'll try to merge it now... git is not my friend, so cross your fingers (or whatever the right saying is).

Thanks,
_Mark

markmaker:
@tonyluken please have a look and if OK, announce it on the group. Thanks!

The testing version is now ready.
https://openpnp.org/test-downloads/

_Mark

markmaker:
Oh, and in case you wonder what happened to the movable camera. An additional column will appear, as soon as you assign a non-virtual Z axis:

grafik

The distance between the two Z is internally used for the formula ("Height above camera"). I know this is mathematically "over-defined" (or whatever the right term is), but I believe it is much easier to understand for the user. Potentially it could be used to servo the camera during future automatic (re-) calibration.

_Mark

PR #1113, opened 2021-01-28
Refactored actuator read to template method

Jnesselr:
# Description
For the Index feeders, I needed to have an actuator read method that accepted Strings. I decided to just make this a generic function that can take any type parameter. It substitutes the "Value" parameter in the ACTUATOR_READ_COMMAND with whatever the parameter passed in is, so we can pass in a string, double, int, boolean, etc. As long as String.format can understand it and use it, it can be used in the actuator read method. I did make sure it was backwards compatible so if a double is passed in, it will substitute the DoubleValue and IntegerValue fields in the gcode as well.

# Justification
Like I said, I needed a String method in actuator Read for the Index feeders. I figured that instead of adding a ton of other methods, we could use generic types in Java and have one method to rule them all. I think we should also make the actuate method be generic templated as well, but that's something that can be discussed.

# The Issue I ran Into
ACTUATOR_READ_WITH_DOUBLE_COMMAND should be able to take a "Value" parameter now, as opposed to just the DoubleValue and IntegerValue. So I'm thinking it should just be ACTUATOR_READ_WITH_PARAMETER_COMMAND. But now I'm wondering why it's even a separate command from ACTUATOR_READ_COMMAND. Each of these commands define the variable names that can use it, but it doesn't seem like the variable names are used anywhere at all. Similarly instead of ACTUATE_type_COMMAND, why not just have an ACTUATE_COMMAND that takes the generic type Value?

So after changing the ACTUATOR_READ_WITH_DOUBLE_COMMAND to ACTUATOR_READ_WITH_PARAMETER_COMMAND, I realized it actually breaks backwards compatibility since the command type name changes. And that's where I could use some help/discussion.

# Instructions for Use
Wiki page [here](https://github.com/openpnp/openpnp/wiki/GcodeDriver%3A-Command-Reference#actuator_read_command) and [here](https://github.com/openpnp/openpnp/wiki/GcodeDriver:-Command-Reference#actuator_read_with_double_command) updated with relevant documentation.

# Implementation Details
Lots of manual testing. Most of my tests were done with a machine.xml file I made that had the ACTUATOR_READ_COMMAND have a gcode value of "foo" and ACTUATOR_READ_WITH_DOUBLE_COMMAND have a value of "bar".
- I removed the read command and made sure the read with double migrated over automatically.
- I removed the read with double and made sure the read command stayed put.
- I made sure that if both were defined, there was an issue in the Issues & Solutions tab.
- I made sure that accepting the issue replaced the read command with the read with double command. Then verified that the driver showed "bar" under the read command and the read with double command no longer existed.
- I made sure that dismissing the issue removed the read with double command and that "foo" showed in the read command.

Yes, I did
I changed the spi classes for Driver and Actuator. In Actuator, it was just to replace double with Object for actuator read. For Driver, I made a similar change for the actuatorRead call, but I also found a few javadoc issues I just tidied up real quick.
I ran the unit tests. Also, is this question as needed anymore now that github actions also runs tests?


Jnesselr:
@vonnieda I think code wise, this is mostly good. The remaining things I know of are:
- There's no wiki link in the issues & solutions
- Does the ACTUATOR_READ_COMMAND now need to have Value, DoubleValue, and IntegerValue in its "variableNames" list? Those variable names seem to be unused.
- I need to fill out Instructions for use. Is that what goes in the wiki page? I'm assuming the wiki page explains the deprecation and what to do.
- I need to fill out implementation details (Which I'm about to do now)

vonnieda:
@Jnesselr This looks really good to me. Only question/concern I have is it looks like the backwards compatibility is handled either by accepting a solution in Issues and Solutions or in the code during each call a check is made. I think it would be better if it just migrated the configuration automatically on the first startup. This would remove the old setting from the config, and it would reduce the complexity of the code going forward so we don't have to check for the old config on every call.

I'm willing to merge it as it is, but wanted to bring that up to get your thoughts.

Thanks,
Jason

Jnesselr:
> it looks like the backwards compatibility is handled either by accepting a solution in Issues and Solutions or in the code during each call a check is made

There's code in the Commit method that does it. I don't check the read with double command on each actual call.

> I think it would be better if it just migrated the configuration automatically on the first startup.

It does, unless both are defined.

Did you see my comment above [here](https://github.com/openpnp/openpnp/pull/1113#issuecomment-787520619)? Those are the major questions I still have about this diff.

vonnieda:
> > it looks like the backwards compatibility is handled either by accepting a solution in Issues and Solutions or in the code during each call a check is made
> 
> There's code in the Commit method that does it. I don't check the read with double command on each actual call.
> 
> > I think it would be better if it just migrated the configuration automatically on the first startup.
> 
> It does, unless both are defined.

Thanks, I guess that's what I get for only looking at diffs :) Nice work :)


> @vonnieda I think code wise, this is mostly good. The remaining things I know of are:
> 
>     * [ ]  There's no wiki link in the issues & solutions

You can probably just update https://github.com/openpnp/openpnp/wiki/GcodeDriver%3A-Command-Reference#actuator_read_command and use that as a link.

> 
>     * [ ]  Does the ACTUATOR_READ_COMMAND now need to have Value, DoubleValue, and IntegerValue in its "variableNames" list? Those variable names seem to be unused.

Go ahead and add them. I put them there because I thought it would be handy at some point to be able to select them in the UI instead of having to remember each name, and I still think that's a good idea. No harm in adding them.

> 
>     * [ ]  I need to fill out Instructions for use. Is that what goes in the wiki page? I'm assuming the wiki page explains the deprecation and what to do.

See the above. It doesn't have to be verbose - just list the new variables and briefly explain how to use it.

> 
>     * [x]  I need to fill out implementation details (Which I'm about to do now)

Jnesselr:
All checks good on my side. I think we're good to go.

vonnieda:
This looks good to me. @markmaker - any concerns?

markmaker:
@vonnieda, @Jnesselr,

No concerns from my side. 

Cosmetics: Diffs are indicating some white space artifacts (at line-endings), but I guess those are flipping back to the correct LF-only state, right? Or are these spaces?

grafik

_Mark

Jnesselr:
Correct, this is going back to the correct format with no trailing whitespace.

markmaker:
@Jnesselr 

Usually we only change formatting in code that is really touched, even if other code does not strictly conform to the canonical format. The idea, I believe, is to better track code changes into the past, i.e. support `git blame`. 

Sometimes CR+LF line-endings seeped in from Windows, these are always ok to fix, but trailing spaces I think not. May @vonnieda can clarify how strict this should be handled. 

_Mark

Jnesselr:
Honestly, I didn't realize it changed that many lines. I think the formatter just did it because that was the file I was in. I can revert those if need be.

markmaker:
I guess there are options in IDEs/Editors as to how formatting is applied:

Eclipse Settings

_Mark

vonnieda:
Generally I'll let whitespace stuff slide if it's not entire files getting reformatted as part of a feature. I'm not gonna worry too much about this one :)

markmaker:
@Jnesselr 

I just now found an issue, sorry I didn't catch it right away:

https://github.com/Jnesselr/openpnp/blob/5e63b1ccf39e63ab806bf5c0ad78b3580d06eb65/src/main/java/org/openpnp/machine/reference/driver/GcodeDriverSolutions.java#L713-L722

- Undo is missing: it should restore the removed command. See the many other solutions for guidance. Thanks to Java's capturing it is usually not hard to keep around the old state for Undo. 
- Dismiss should not do anything, i.e. it is not an _alternative_ solution but the _suppression_ of the solution. 
- I suggest you leave the less likely alternative for the user to fix manually. If there are truly multiple choices, you should open multiple Solutions and then make sure, only one is accepted (throw when the user tries to accept a second one). The alternatives should then vanish when the user uses "Find Issues & Solutions" again.

The Issues & Solutions system should be very consistent IMHO, as it is intended to make things safer and simpler for the user. If simple solutions (and their undoing) are not possible, I would rather not add the Solution at all. 

_Mark

PR #1115, opened 2021-01-31
AdvancedLoosePartFeeder: Support for the async gcode driver

doppelgrau:
# Description
Fixed synchronization so that it works with the async gcode driver:

- In the default pipelines made sure the "settle-first" is enabled
- Added explicit synchronization in the code to make sure the next call to the vision pipeline is called after movement finished

Added a safety feature, to make sure the vision pipeline can not "run away" during the location. Ensures the pick location stays inside the camera field of view if the camera is above the feeder location.

# Justification
fixed bug and added safety feature.

# Instructions for Use
no changes

# Implementation Details
1. How did you test the change? Pipelines not tested, but tested with own pipelines, works normal again

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/AdvancedLoosePartFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/AdvancedLoosePartFeeder.java
@@ -54,10 +54,7 @@ public class AdvancedLoosePartFeeder extends ReferenceFeeder {
 
     @Override
     public Location getPickLocation() throws Exception {
-        if (pickLocation == null) {
-            throw new Exception("Feeder " + getName() + ": No parts found.");
-        }
-        return pickLocation;
+        return pickLocation == null ? location : pickLocation;
     }
 
     @Override
@@ -68,6 +65,10 @@ public class AdvancedLoosePartFeeder extends ReferenceFeeder {
         // if there is a part, get a precise location
         for (int i = 0; i < 3 && pickLocation != null; i++) {
             pickLocation = locateFeederPart(nozzle, pickLocation);
+        }
+        // throw if feed results in no parts
+        if (pickLocation == null) {
+            throw new Exception("Feeder " + getName() + ": No parts found.");
         }
     }
 
--- a/src/main/java/org/openpnp/machine/reference/feeder/AdvancedLoosePartFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/AdvancedLoosePartFeeder.java
@@ -54,7 +54,10 @@ public class AdvancedLoosePartFeeder extends ReferenceFeeder {
 
     @Override
     public Location getPickLocation() throws Exception {
-        return pickLocation == null ? location : pickLocation;
+        if (pickLocation == null) {
+            throw new Exception("Feeder " + getName() + ": No parts found.");
+        }
+        return pickLocation;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


markmaker:
Hi @doppelgrau,

Are you sure this is just a bug fix?

There seem to be additional changes showing up. Maybe the wrong branch base? 

Or am I missing something? 

(perhaps better go to bed now ;-)

_Mark

doppelgrau:
Oh, the checkIfInInitialView was an additional (local) feature to make it a bit more foolproof (had one "runaway" with a bad pipeline)
Didn't pay attention to that, can remove it.

But if the wait for completion is optional (e.g. bottom vision), it can be argued, that only the XML files for the default pipelines should be updated to include a settle first. => all changes in the java file can be removed.

doppelgrau:
Which solution is preferred:

- mention the "in initial view" check in the pull request
- remove the check
- only change the pipelines?

markmaker:
Hi @doppelgrau ,

I see automatic nozzle tip changing in the `feed()` method. That is a no-go, IMHO and I believe what you wanted to achieve has since been integrated in the FeederPanel's `pickFeeder()` method, where it belongs:

https://github.com/openpnp/openpnp/blob/c613059bef22cd10628d04bc278d2b573570c4c0/src/main/java/org/openpnp/gui/FeedersPanel.java#L576-L603

The "in initial view" could be in the same PR and described properly, as it is an update to the _one_ feeder. As this is intended to go into test I see no problem there.

@vonnieda, correct me if I'm wrong. 

_Mark

doppelgrau:
Cleaned up the redundant nozzleChange operation (was logic error, needed it on the HeapFeeder, since there I do picks during feed...) and edited the pull request.

vonnieda:
re: in initial view - sounds fine to me.

markmaker:
Thanks @doppelgrau !

markmaker:
Hi @doppelgrau,

I'm in the course of implementing 3D Units per Pixel usage inside computer vision. I assume you have followed the development of that, otherwise tell me, and I'll quickly explain it.

As a prerequisite for that, the subject Z needs to be known. I intend to use the camera's _virtual_ Z axis to set the wanted subject Z. For feeders, that's usually the pick location Z. Moving the camera to the pick location will physically move it in X and Y and _virtually_ in Z. Units per Pixel can then easily be calculated by the camera. That seems to be a "natural" way to go about it. Even if somebody builds a machine with the camera on a _physical_ Z axis (i.e. physically shifting the focal plane), this comes out exactly right.

Now I've seen you have added this:

`pickLocation = location.derive(null, null, Double.NaN, 0.0);` 

i.e. you are removing Z (and rotation) from the location before moving the camera to that location. This spoils that idea.

**The question is, why did you do it?** And more importantly: **Can I remove that?**

IMHO, unless the machine is configured wrong, there is no need to do that. Moving the camera virtual Z is not dangerous. Even with the camera on a _physical_ axis, it is the _focal plane_ that is moved down to that Z coordinate, not the camera itself, i.e. it will still be safely at a distance. 

Thanks for your help,
_Mark

doppelgrau:
Followed it "with one eye" (my machine has cameras with a narrow focal plane and try to bring everything to that height => not used), but was pixel/mm changes with height?!

Why: Guess z was "cargo cult", see no point why that can not be changed. Rotation on the other hand I think was used to get the rotation later of the part correct, so if you change that, you need to subtract the orientation later from the resulting orientation after vision. (from memory).

markmaker:
Thanks, will do.

Just learned a new idiom 🙃
https://en.wikipedia.org/wiki/Cargo_cult_programming

> but was pixel/mm changes with height?!

Yep.

PR #1128, opened 2021-02-22
Fix VisionUtilsTest by creating a local configuration

TWilmer:
# Description
The change creates the Configuration object before the unit test is executed.
# Justification
The Unit test would fail if executed alone or before other component tests which create the static configuration object
# Instructions for Use
none
# Implementation Details
Obvious change.
mvn test works even in directory where unit test is executed first.

### DIFF
--- a/src/test/java/VisionUtilsTest.java
+++ b/src/test/java/VisionUtilsTest.java
@@ -1,13 +1,17 @@
 import java.awt.image.BufferedImage;
+import java.io.File;
 import java.io.IOException;
+
 
 import javax.swing.Action;
 import javax.swing.Icon;
 
 import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
 import org.openpnp.CameraListener;
 import org.openpnp.gui.support.Wizard;
+import org.openpnp.model.Configuration;
 import org.openpnp.model.Length;
 import org.openpnp.model.LengthUnit;
 import org.openpnp.model.Location;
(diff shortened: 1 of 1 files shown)


markmaker:
Hi @TWilmer 

Is this the same issue as in this report by Clemens?
https://groups.google.com/g/openpnp/c/xBe02vW-kzo/m/3fKnpkzqAAAJ

And the reusing of VMs explains why it did not happen in the Github Actions and in our own `mvn test` runs?

_Mark

markmaker:
Note I received this by E-Mail

> @TWilmer pushed 1 commit.
> 
>     3e986da Fork for each test a new VM

But it seems not to be in here (?)

(Sorry if this is obvious, I'm not a git/Github enthusiast)

_Mark

TWilmer:
Hi
I think it's a different problem as it's not
Only throwing errors for the configuration in VisionUtilsTest.

The root cause for my problem was that the unit test needed a configuration
object, but didn't create it. Instead it relied on any earlier unit test to
have it created.

I have created two pr, one which fixes this unit test and the other which
forces a new vm on each test, which should find such problems in the
future. But I don't know yet if the 2nd one is the right way to do and I
don't know if it has other side effects, which are unwanted.

Kind Regards
 Thorsten


markmaker <notifications@github.com> schrieb am Mo., 22. Feb. 2021, 20:58:

TWilmer:
Hi,
The first problem without -e -X is likely the same problem. One would have
to see the complete trace.

Kind Regards
 Thorsten

markmaker <notifications@github.com> schrieb am Mo., 22. Feb. 2021, 20:58:

vonnieda:
Thanks @TWilmer! 

Context: We discussed and diagnosed this on Discord. At least three people seem to be having this issue although I am not able to reproduce it. I tested on a fresh Mac build and a fresh Ubuntu 20.04 VM. It seems like it could be related to order in which tests run. Regardless, at some point the AbstractCamera came to rely on a valid config being available, which seems to have broken this test in in practice for some, and in theory for all, and this change seems to fix it.

PR #1129, opened 2021-02-22
Fork for each test a new VM

TWilmer:
# Description
This change changes the configuration of the plugin to execute the unit tests to recreate
a vm on each test
# Justification
This is a good idea as static variables are used which may be not initailized accidentally.

# Instructions for Use
Nothing
# Implementation Details
Got this information from 
https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -363,6 +363,7 @@
 				<artifactId>maven-surefire-plugin</artifactId>
 				<version>2.19</version>
 				<configuration>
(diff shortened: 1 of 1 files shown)


vonnieda:
Thank you for sending this over @TWilmer , but I think we will skip this one, as it causes the tests to take far too long.

PR #1130, opened 2021-02-22
Add read for http actuator

TWilmer:
# Description
This change adds a read URL to read values from an http actuator.

# Justification
I have a small c++ Program on my Rpi which implements a simple http server to actuate GPIO and
read Preassure (and possibly Distance ) sensor via http. The request looks like this:
http://192.168.0.42:8001/msr the response looks like this:
pump:11583

The idea is that the Pattern Matching is used to read then the preasure value.
Sadly the Null Driver has no pattern matching - it is thus not easy to test or should this be part of the HTTP actuator?

# Instructions for Use
The regex to match the result is given in parallel to the URL to read. It's only available for non boolean
actuators.

# Implementation Details
1. How did you test the change? I can read the value with the read button, which should be the first MVP.
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? I hope so.

I hope this didn't affect anything.

done, I even added a unit test particually for this item.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/HttpActuator.java
+++ b/src/main/java/org/openpnp/machine/reference/HttpActuator.java
@@ -54,6 +54,9 @@ public class HttpActuator extends ReferenceActuator {
     // configuration changes. 
     protected String lastActuationUrl = null;
 
+    @Element(required = false)
+    protected String readUrl = "";
+
     public HttpActuator() {}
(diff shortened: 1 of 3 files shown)


vonnieda:
Looks good to me, thank you @TWilmer!

PR #1135, opened 2021-02-28
ReferenceBottomVision - Part size check option for detecting bad part pick

crashmatt:
# Description
Option to check package size during bottom camera part alignment.  Compares measured vs expected dimension with a part configurable tolerance.

# Justification
Enables optical detection of badly picked parts which are not sitting flat on the nozzle.

# Instructions for Use
1. Set the expected body dimension in the package (for the part) under the vision tab
2. Enable "Check size" checkbox under the Part ReferenceBottomVision tab
3. Adjust size tolerance under the Part ReferenceBottomVision tab.  Default 20%.

Only works when the bottom vision pipeline returns a single rectangle for the whole pacakge.  Is not expected to operate if individual pads are measured.

# Implementation Details
1. Tested on 0603 with a pick offset for high failure rate.  With a run of 20 parts found every instance of bad pick and rejected no instances of good picks.
One instance of a part flat on the nozzle but rotated at 45 deg.  This passed the size check.

2. Needs review on following:
	Where this funciton is implemented.  It is implemented in findOffsetxxRotate which is the only place where the part outline rect is avaialable. This method does not return measured part size.

	Comparison of lengths is in pixel units.  Would prefer human units for readadabilitu and debugging.

	Does not attempt to take account for part height.  Only uses camera calibrated pixel units.

	Compares longest side vs longest measurement to avoid problems with dimensions being swapped in the package.

	No check against users entering bad values into the tolerance field.

3. No changes to`org.openpnp.spi` or `org.openpnp.model`.  Makes changes to PartSetting in machine.xml with the check size option and tolerance.

4. `mvn test` passed

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -295,8 +295,8 @@ public class ReferenceBottomVision implements PartAlignment {
 
         double widthTolerance = pxWidth * 0.01 * (double) partSettings.getCheckSizeTolerancePercent();
         double heightTolerance = pxHeight * 0.01 * (double) partSettings.getCheckSizeTolerancePercent();
-        double pxMaxWidth = pxWidth + heightTolerance;
-        double pxMinWidth = pxWidth - heightTolerance;
+        double pxMaxWidth = pxWidth + widthTolerance;
+        double pxMinWidth = pxWidth - widthTolerance;
         double pxMaxHeight = pxHeight + heightTolerance;
         double pxMinHeight = pxHeight - heightTolerance;
 
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -247,84 +247,77 @@ public class ReferenceBottomVision implements PartAlignment {
     
     
     private boolean partSizeCheck(Part part, PartSettings partSettings, RotatedRect partRect, Camera camera) {
-    	//Check if this test needs to be done
-    	PartSettings.PartSizeCheckMethod partSizeCheckMethod = partSettings.getCheckPartSizeMethod();
+        // Check if this test needs to be done
+        PartSettings.PartSizeCheckMethod partSizeCheckMethod = partSettings.getCheckPartSizeMethod();
 
-		double checkWidth = 0.0; 
-		double checkHeight = 0.0;
+        double checkWidth = 0.0;
+        double checkHeight = 0.0;
 
-    	if(partSizeCheckMethod == PartSettings.PartSizeCheckMethod.PartSizeCheckNone) {
-			return true;
-		}
-		
-    	Footprint footprint = part.getPackage().getFootprint();
-		LengthUnit footprintLengthUnit = footprint.getUnits();
-		
-		//Get the part footprint body dimensions to compare to
-		switch(partSizeCheckMethod) {
-		case PartSizeCheckNone:
-			return true;
-		case PartSizeCheckBodySize:
-	    	checkWidth = footprint.getBodyWidth();
-	    	checkHeight = footprint.getBodyHeight();
-			break;
-		case PartSizeCheckPadExtents:
-			Rectangle bounds = footprint.getPadsShape().getBounds();
-			checkWidth = bounds.getWidth();
-			checkHeight = bounds.getHeight();
-			break;
-		}
+        Footprint footprint = part.getPackage().getFootprint();
+        LengthUnit footprintLengthUnit = footprint.getUnits();
 
-    	//Make sure width is the longest dimension
-    	if (checkHeight > checkWidth) {
-    		double height = checkHeight;
-    		double width = checkHeight;
-    		checkWidth = height;
-    		checkHeight = width;
-    		}
-    	
-    	Length width = new Length(checkWidth, footprintLengthUnit);
-    	Length height = new Length(checkHeight, footprintLengthUnit);
-    	double pxWidth = VisionUtils.toPixels(width, camera);	
-    	double pxHeight = VisionUtils.toPixels(height, camera);
-    	
-    	//Make sure width is the longest dimension
-    	Size measuredSize = partRect.size;
-    	if (measuredSize.height > measuredSize.width) {
-    		double mHeight = measuredSize.height;
-    		double mWidth = measuredSize.width;    		
-    		measuredSize.height = mWidth;
-    		measuredSize.width = mHeight;
-    	}
-    	
-    	double size_tolerance_multiplier = 1.0 + (partSettings.getCheckSizeTolerancePercent() * 0.01);
-    	double pxMaxWidth = pxWidth * size_tolerance_multiplier;
-    	double pxMinWidth = pxWidth / size_tolerance_multiplier;
-    	double pxMaxHeight = pxHeight * size_tolerance_multiplier;
-    	double pxMinHeight = pxHeight / size_tolerance_multiplier;
-    	
-    	if (measuredSize.width > pxMaxWidth) {
-    		Logger.debug("Package pixel width {} : limit {} : measured {}", 
-    				pxWidth, pxMaxWidth, measuredSize.width);                		
-    	} else if (measuredSize.width < pxMinWidth) {
-    		Logger.debug("Package pixel width {} : limit {} : measured {}", 
-    				pxWidth, pxMinWidth, measuredSize.width);        
-    		return false;
-    	} else if (measuredSize.height > pxMaxHeight) {
-    		Logger.debug("Package pixel height {} : limit {} : measured {}", 
-    				pxHeight, pxMaxHeight, measuredSize.height);                		
-    		return false;
-    	} else if (measuredSize.height < pxMinHeight) {
-    		Logger.debug("Package pixel height {} : limit {} : measured {}", 
-    				pxHeight, pxMinHeight, measuredSize.height);                		
-    		return false;
-    	}
-    	
-    	Logger.debug("Package pixel size ok. Width {}, Height {}", 
-				measuredSize.width, measuredSize.height); 
-    	return true;
+        // Get the part footprint body dimensions to compare to
+        switch (partSizeCheckMethod) {
+        case Disabled:
+            return true;
+        case BodySize:
+            checkWidth = footprint.getBodyWidth();
+            checkHeight = footprint.getBodyHeight();
+            break;
+        case PadExtents:
+            Rectangle bounds = footprint.getPadsShape().getBounds();
+            checkWidth = bounds.getWidth();
+            checkHeight = bounds.getHeight();
+            break;
+        }
+
+        // Make sure width is the longest dimension
+        if (checkHeight > checkWidth) {
+            double height = checkHeight;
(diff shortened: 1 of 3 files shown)


doppelgrau:
A question: why don't you implement that as a pipeline stage?
E.g. a FilterPartSize a bit like the FilterRects, just that the dimensions are taken from the part loaded on the nozzle tips and you might add some parameters for allowed tolerances.
Less changes in the GUI needed, and can be more flexible combined with other approaches (e.g. check size before using a template match)

markmaker:
> A question: why don't you implement that as a pipeline stage?

@doppelgrau it already exists as a stage (CheckSize) but the problem is you then have to create a custom pipeline per part. Best read this https://groups.google.com/g/openpnp/c/_9Q9FBOT9y8/m/zHfrs8SxAQAJ

_Mark

markmaker:
@crashmatt 

I'm not 100% sure about this (I guess @vonnieda must say) but I think the body dimensions are meant as a different thing. For parts with real pins (not pads), the body is only the inner thing. For instance, if you look at the TQFP32 that comes as an example with OpenPnP, it has the body as 7x7mm i.e. the inner (plastic) body.  

TQFP32
TQFP32 simulated


The body dimensions are currently only used for drawing the part in the SimulatedUpCamera AFAIK, but there is perhaps potential for future usages. 

I guess it would be better to use separate dimensions i.e. specifically as a "visual size". It could then be agnostic to whether the visible bright contacts are "sprouting" out from a body or whether they are receded from the edge of a body (e.g. ball grid) or whatever visual "erosion" happens through bevels on the contacts, median filters and thresholds. This would also open it up for auto-learning: ReferenceBottomVision could check if the visual size is not yet set and then take the first vision result it encounters to automatically set it. 

In fact, I think auto-learning would be quite a small thing to add right now, it would improve usability a lot! 
;-)

_Mark

doppelgrau:
@markmaker but if you get the dimensions from the footprint, you do not need an own pipeline for each part.

markmaker:
>  if you get the dimensions from the footprint

@doppelgrau in that case I agree, that could be another possibility. 

However, when you read my [second comment](https://github.com/openpnp/openpnp/pull/1135#issuecomment-787451017) here, it will become apparent that additional GUI elements need to be added anyway. And it would be easier to enable/disable from the outside i.e. the feature can be used by users that don't know how to edit pipelines. 

_Mark

vonnieda:
I've had the question of where to get the data on my mind for a while. Didn't expect someone to implement this so fast, so thank you @crashmatt for pushing the issue!

I think that ideally the data should be based on what would be present in the datasheet for a part. This will make it easy to manually enter data at first, and then later we can automatically import it from CAD.

Let's take an [ATmega328p](https://ww1.microchip.com/downloads/en/DeviceDoc/ATmega48A-PA-88A-PA-168A-PA-328-P-DS-DS40002061B.pdf) as an example:

Screen Shot 2021-02-28 at 11 20 53 AM

Dimensions D1 and E1 are the body width and length, and D and E are the extents of the pins. D and E *should* be exactly what vision sees using the default pipeline. This should hold true even in the case of things like QFNs:

Screen Shot 2021-02-28 at 11 25 55 AM

And in my experience all datasheets have a dimension that includes the extents of the pins.

So, with that, I think the right answer is to calculate the extents of the pins in the footprint and use that for the size check, and not be concerned with the body width and length. This means more data entry, at first, because we need to enter the pins but ideally these can from CAD and be shared by users. Additionally, if you just want to get a part working without entering every pin, you can just enter one giant pin that matches D and E from the datasheet.

Thoughts?

Jason

crashmatt:
Thanks everyone for your comments on the PR.

My interest is in enabling this feature for parts that might not be picked flat.  Maybe anything with a 2mm dimension or less?  That rules out most leggy packages.  Maybe there are use cases I have not thought of.

To keep things simple and the feature creep minimized I suggest starting with the simple feature set and getting it reviewed.
In the long term an option to use body size, footprint pad extents or captured size could be added.  For now the tolerance value can be modified to help with leggy packages.

I am setting my package pads and body size based on PCB land pattern from KiCAD.  Not ideal but the best tool I have.  At least it gets me in the right ballpark.

Writing this code was much more entertaining than the thought of hand tweaking pipelines.

doppelgrau:
> Writing this code was much more entertaining than the thought of hand tweaking pipelines.

Yes, hand Tweaking should not be necessary.
My suggestion was more directed to a pipeline-stage where you enter the allowed deviation (would thing in percent and absolute pixel), and calculate the required size using the length and width multiplied with "camera.getUnitsPerPixel().convertToUnits(LengthUnit.Millimeters).getX()" ...

markmaker:
@crashmatt 

> My interest is in enabling this feature for parts that might not be picked flat

I think you should think about how your contribution is useful for a large portion of users and for very likely use cases. 

Consider a tray feeder where a part is missing, a tube feeder, where the next part is stuck, or a reeled feeder where the cover tape was not properly pulled back. I believe this PR should cover these cases as they are very likely to happen and users will expect the size check to catch these. 

https://github.com/openpnp/openpnp/wiki/Developers-Guide#general-usefulness

Also, I think users should not be led into entering persistent data that is semantically "wrong" and hard to migrate to a better data design later. 

Implementing  @vonnieda 's suggestion would make sure that data is valid. 

@vonnieda, I would still argue, that implementing my suggestion (auto-learning from the first bottom vision) would be more comfy to use as it requires no additional manual data entry (or ECAD footprint import support), and it would automatically account for any vision detection artifact such as partial contact reflection, median blur erosion etc. And in case of ECAD-imported footprints, it would not be bothered by the fact that contacts (as seen and masked in bottom vision) are usually smaller and often center-shifted when compared to imported footprints, those typically being solder lands. 

Maybe you remember, I implemented a visual "place check" for the SampleJobTest, where the ImageCamera's view of the PCB is visually analyzed to see whether a part is properly placed to validate that the machine has the correct position (as a test to see if AMC generated proper Gcode, if it is interpreted correctly by the GcodeServer, while also simulating bad run-out, bad non-squareness etc. being compensated). 

As the PCB view is an Eagle board rendering, this essentially extracts the footprints (more specifically the solder lands/"cream" layer) from the PCB image (by HSV masking). In my tests I then used Gaussian blurring to create a template match out of the part drawing. The gradients from blurring act a bit like surface tension in reflow, so the size and position mismatch of the solder lands does not matter:

Place Check

IMHO, a future new/alternative bottom vision class using (ECAD-imported) footprint data should work the same way, allowing for solder lands to be used. Consequently, I argue that footprint data semantics should remain unconstrained in that respect by this PR. 

_Mark

vonnieda:
@crashmatt The change I'm suggesting will work for the future, should work with your existing data import, and should be a one or two liner change. I can help with if it needed, but the gist is just grab the Shape instead of the body width and length, and use the width and length of the Shape instead.

I think this is a really nice feature, but as Mark said, we need to make sure it's using the right data or users will spend a lot of time setting up data that in the future will not be correct.

Body width and length are by definition incorrect for this task because our vision system does not "see" the body - it only sees the pins.

@markmaker Agree on auto learning, but I don't think that precludes merging this. This is a good feature that people can start using right away, and can use in the future if they don't prefer auto learning. Once someone decides to write auto learning, I think the data it creates can just be stored alongside the existing data. The footprint data can be human verified "reference" data, and the auto learned data can be used if available.

So, my suggestion is @crashmatt make the small change to use pin extents instead of body and we merge this.

Jason

markmaker:
Agree!

crashmatt:
I quickly answered my own question about other use cases. Just mis-picked a 4x5mm TQFN and had the bottom vision mistake the nozzle for the part.

@vonnieda If you would prefer it to be done then I can make the change.
Unless you think otherwise I will put these values in PartSettings.  Then it keeps it separated from the package body dimension settings and the menus make more sense.  The only down side is that each individual part needs to be configured.
I will also attempt to translate the measured pixel size to Length.  Then the setting make more sense to humans and can survive changes to cameras.

/Matt

vonnieda:
@crashmatt Sorry if I wasn't clear - what I am saying is there is no need to add new settings. Just use the extents of the defined pins, which you can get from `Packages.getFootprint().getPadsShape().getBounds()`.

crashmatt:
@vonnieda That breaks with my use case where the pads are defined by the cad footprint so the result is too large.
Will try to find an elegant way to meet all requirements.

/Matt

vonnieda:
@crashmatt What do you use the CAD imported pads for?

One thing worth mentioning is that the Footprint and Pads terminology in the Packages panel are misnomers. All of this data should be from the Part datasheet package information, not from the solder footprint that one would use on a PCB. There is no value in defining the pads with respect to the PCB footprint because we don't and can't use them as they are covered in solder paste.

crashmatt:
A simple fix.  Changed the checkbox to a combo box with more options.

part_size_check_menu

Now the user does not know what they are getting.  Would be good to show the expected shape on the bottom camera.

My pads are imported from the CAD footprint.  Not accurate but much easier than doing it from the datasheet.  It also adds a pin mark making it easy to sanity check orientation.

/Matt

markmaker:
Hi @crashmatt 

That's a good and extensible solution, IMHO. The auto-learn can then be easily added later, nice.

Some minor issues:

- Code seems to contain tabs, see also the [Eclipse formatter](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style) 
   grafik
- Just my 2¢: enum constants should not repeat the enum scope name:
   `PartSizeCheckMethod.None` instead of `PartSizeCheckMethod.PartSizeCheckNone` 
   Plus a label in front on the GUI.

_Mark

crashmatt:
Code cleanup done.  Change to the panel layout with the change in combo box naming.
Also changed the percent tolerance from double to int.  Nobody needs that tolerance in less than 1% steps.

part_size_check_menu2

markmaker:
Thanks, @crashmatt, this now looks very good to me. The fact that it is stored on the PartSettings rather than the Package or Part itself, is in my opinion an extension of the existing architecture and therefore ok (i.e. compatible with future plans).

@vonnieda  what do you think?

_Mark

crashmatt:
@markmaker @vonnieda I just noticed that I should have had my branch named as /Feature/part_check_size.  Is this a problem?
Do I need to make a feature branch?  Maybe I can consolidate to a single commit at the same time.

/Matt

vonnieda:
Great work on this @crashmatt - thank you for this nice feature!

crashmatt:
Thanks @vonnieda <https://github.com/vonnieda> for making it possible to
contribute.

I ran this feature on a recent build and it made a significant improvement
to placement reliability.
The only problem I had was parts that were picked in the wrong rotation
from the tape.  The size check only checks longest and shortest sides. It
does not care about orientation.  That is something to improve on.

Will be submitting a follow-up PR with usability features.  Being able to
copy settings to all components with the same package is the first one I
have in mind.

/Matt

markmaker:
@crashmatt,

> Will be submitting a follow-up PR with usability features.  Being able to
copy settings to all components with the same package is the first one I
have in mind.

Please read the following discussion first. The discussion takes some detours and there is some controversy, but I believe it has to evolve somewhere in that direction. If this is implemented (and I intend to) then your usability features would become obsolete. The intended system has dynamic inheritance, so new parts imported from ECAD will automatically get the settings from the package level. 

Your features could still be useful in the meantime, but I thought you should know, before you put too much energy into it:

https://groups.google.com/g/openpnp/c/7DeSdX4cFUE/m/VYDG6x6-AAAJ

_Mark

PR #1138, opened 2021-03-08
Implement "Snap" option for OrientRotatedRets

MatSpy:
Hello,

# Description
Some of code that works with pipelines expect unambiguous orientation of oriented rectangle at the end of pipeline.
That are e.g. LoosePartFeeder and FullRotate option. Stage MinAreaRect in fact returns random 1 of 4 angles of
found rectangle. Sometimes it is required for pipeline to output 1 certain angle from those 4. When option SnapToAngle
will be selected, pipeline will return certain angle. This will work more like Align option for bottom vision. 

# Justification
Where it is needed:
1. LoosePartFeeder - when I place small component that requires Down-looking camera alignmet before pick, and
this component has defined orientation. I can place it with known orientation (e.g. about 0 degrees). Then use Snap
option with SnapAngle set to 0 degree. Pipeline then will output angle in range of -45 +45 degrees, and pick will be 
successful. 
2. When using Full rotate option for part (in my machine it is useful for bigger parts, since I have poor C accuracy).
I can force pipeline to output angle in certain range. This prevent random rotation when aligning with Full option. 
With this method I managed to achieve way better orientation accuracy for problematic part (big with fine pith). 

When I started to play with LoosePartFeeder I cannot get it to work without this option. It occured to be very
practical, so I created this Pull-request.

# Instructions for Use
Just select SnapToAngle option in OrientRotatedRets, and put angle to which output angle will be snapped. 
Supported angles are 0, 90, 180, 270. 

# Implementation Details
Rotated rect has already angle, but in fact it is random one from 4 possible angles of rectangle. 
I bring that angle to normalized range -45 to 45 degrees. Then i rotate it to desired range, where
0, 90, 180, 270 are centers of that range.

Yes

No

4. Tests passed

Regards,
Mateusz

### DIFF
--- a/src/main/java/org/openpnp/vision/pipeline/stages/OrientRotatedRects.java
+++ b/src/main/java/org/openpnp/vision/pipeline/stages/OrientRotatedRects.java
@@ -16,7 +16,8 @@ import org.simpleframework.xml.Attribute;
 public class OrientRotatedRects extends CvStage {
     public enum Orientation {
         Landscape,
-        Portrait
+        Portrait,
+        SnapToAngle,
     }
     
     @Attribute(required = false)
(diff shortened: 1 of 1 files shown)


vonnieda:
Looks good to me, thank you @MatSpy!

PR #1143, opened 2021-03-12
Recycling Feature: Core

doppelgrau:
# Description
Changes to the "core" to allow to recycle (put back into feeder) the part on the current active nozzle.
Added a "Recycle" button next to the "Discard" button on the JogControlsPanel. The button is only set active, if there is a part loaded on the active nozzle and for this part there is at least one enabled feeder that signals, that it can take back a part.
Also by accident some indentations have been corrected (tab/spaces).
Needed Change on the SPI classes Feeder, AbtractFeeder and the GUI class JogControlsPanel.

A feeder wanting to support this needs to implement two methods: 

- boolean canTakeBackPart() which signals if the feeder can currently take back a part (can be decided dynamically, e.g. only one part taken back before the next pick, or only until feedCount = 0 ...)
- void takeBackPart(Nozzle nozzle) is responsible for putting the part on the nozzle back into the feeder.

This pull request only contains the changes to the openpnp core. Will be accompanied by other pull request for the different feeders, that will support this feature.

# Justification
During machine setup often parts are picked to check the setup (e.g. to check pick locations, or bottom vision). In many cases these parts should not be discarded, so currently they need to be placed back in a feeder, the feed count must be changed. This takes time and is error prone (e.g. wrong rotation).
So for feeder that allow "taking back" parts, that can now be done easily.

# Instructions for Use
If "recycling" is possible, the button is active. In that case it can be pushed, and the part on the nozzle will be put back in one enabled feeder. (A bit like discard, but putting it back into a feeder.)

# Implementation Details
1. How did you test the change? Tested it in the simulator. More extensive tests will follow with the feeder implementations.
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. Changes were done in Feeder, AbtractFeeder.


doppelgrau:
One little "bug" is still in the GUI, if the "canTakeBack"-State of a feeder changes (e.g. enabled/disabled or reset the feedCount) while the part is picked, it is not reflected in the GUI. (if the button was enabled before and the user click on it, nothing will happen, since no matching feeder can be found).
But I guess this is a too minor corner case to justify the number of property change listeners needed (for all feeders all attributes that could enable/disable the feature). Usually the process is (I guess) pick the part, maybe check bottom vision, put back. Not changing the whole configuration while part on the nozzle.

doppelgrau:
Added pull requests for all feeders, where I can currently see a working recycle feature.
During testing of the feeders, GUI/core got more tests, no problems found so far.

vonnieda:
Looks good to me, thanks @doppelgrau!

PR #1142, opened 2021-03-11
NgUI - Themeable User Interface

SG-O:
# Description
Replaced the default system dependent UI look and feel with a customizable one. This uses Darklaf by weisJ (MIT License).

# Justification
The default LAF is different on many operating systems. This makes things consistent and flexible. The user can choose a style that suits him. It also provides accessible themes for people that are visually impaired.

# Instructions for Use
The user can change the theme in Window -> Change Appearance 

# Implementation Details
1. How did you test the change?
I checked the visual appearance manually for all themes. I did not touch any non visual code.
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?
Yes

PR #1144, opened 2021-03-12
Revert "NgUI - Themeable User Interface"

vonnieda:
Reverts openpnp/openpnp#1142

Users reported and I confirmed that there seems to be some UI problems. I am unable to set fields in tables such as Part Height and and it seems to not be possible to select some items in the Machine Setup tree such as GcodeAsyncDriver.


vonnieda:
Hi @SG-O - I had to revert this PR unfortunately. See my comments in the description above. If you could have a look I'd love to re-merge this as I think the changes look very good.

SG-O:
I can confirm this and am working on the resolution of this problem.

PR #1151, opened 2021-03-13
NgUI - Themeable User Interface

SG-O:
# Description
Replaced the default system dependent UI look and feel with a customizable one. This uses FlatLaf by JFormDesigner
(Apache-2 License).

# Justification
The default LAF is different on many operating systems. This makes things consistent and flexible. The user can choose a style that suits him. It also provides accessible themes for people that are visually impaired.

# Instructions for Use
The user can change the theme in Window -> Change Appearance 

# Implementation Details
I checked the visual appearance manually for all themes. I did not touch any non visual code. Previous problems when trying to edit table cells are fixed. 
Non the less someone else besides me should take a look at this. A lot of files have been changed.
Yes


vonnieda:
@SG-O Great work on this, again :) I think it looks very nice. Two questions:

1. Would it be possible to offer to, and default to, the default look and feel that existed before this change? If I can merge this without it actually changing anything for anyone, and instead just offering a new option then it's a no-brainer to merge.

2. In the previous PR the table rows had alternating background colors and I thought that looked really nice, and added a nice breakup of the white space. Is that available in these LAFs?

Thanks,
Jason

SG-O:
@vonnieda 
About 1. This should not be too hard to implement.

About 2. I'll have to look into it. It may take a few hours..

vonnieda:
Thank you @SG-O! This looks great!

markmaker:
Hi @SG-O 

Cool feature!

Some usage questions:

Is switching back to "Windows" the same as the default right after the upgrade? I.e. each OS has a different default theme, right?

grafik

Would it make sense to somehow mark it in the list? Or provide a "Revert to Default" dialog button?

As to the extra themes: are these completely identical on different Operating Systems? 

I'm asking because it could become an option to document Wiki screenshots in one of the common themes in the future.

_Mark

markmaker:
Btw. I was looking for the setting in the machine.xml and for the first time realized these GUI settings are not stored there. Do you know where these are stored?

I.e. this is not all there is...
https://github.com/openpnp/openpnp/wiki/FAQ#where-are-configuration-and-log-files-located

_Mark

markmaker:
Hi @SG-O 

Is it OK if I report some missing spots here? I guess you are much faster in fixing those properly:

grafik

grafik

_Mark

SG-O:
@markmaker Hi Mark,

The Default theme is different on different OSs. It may be hard to implement highlighting the default theme. I'll look into it.
The Extra themes are identical on every machine.

The settings are stored in the Java User Preferences (e.g. on Windows in the registry) This is the place where OpenPnP saves settings like Language selection.

Thanks for reporting the places I missed. I'll get those looking correctly.

\-SG-O

PR #1155, opened 2021-03-21
Actuator: Support all kind of value type (WIP)

diorcety:
# Description
- This allows to all any object type with actuator
- Remove all special cases with multiple function signatures (boolean, double, ...)
- Remove default(On/Off)(Double/String) for handling double/string conversion when internal need boolean value
- Remove any (suspicious and maybe incorrect) at runtime (only used for user input with JTextField string conversion)
- Add new ReferenceActuatorMapper:
  - Replace the limited profile
  - Still allow desired simple type conversion (like defaultOn/defaultOff)
TODO:
- Migrate Profile to ReferenceActuatorMapper
- Migrate default(On/Off)(Double/String)
- Migrate Actuator valueType to valueClass

# Justification
Form my point of view the all actuator value handling is quiet overcomplicated and we should avoid risky implicit type conversion.


markmaker:
Hi @diorcety,

Welcome! 

Whew... I'm a bit overwhelmed by the sheer amount of changes here! The justification is very thin, compared to that. 

As the [Pull Request Template](https://github.com/openpnp/openpnp/blob/develop/PULL_REQUEST_TEMPLATE.md) states, you should first discuss such changes in the discussion group.

> If you would like to add a new feature to OpenPnP that changes core or reference functionality, please create a topic on the discussion group to discuss the feature first. Also consider if your feature can be implemented as a new subclass of a Reference class, or an entirely new class, instead of changing or adding to the Reference class.

Please explain your goals in much more detail before continuing. Why did you start it? It can't just be a bunch of overloaded methods that bothered you in terms of design beauty, right? What new types of Actuator values do you want to support? Why have you renamed "Profiles" to "Mapper"? Do you really need more than 6 actuators? 

Knowing this is still WIP, I see several changes in your code, where I don't see how the original function is preserved in what you are replacing. Especially when it comes to Default On/Off, Actuator Profiles and Light actuation in the Pipeline. 

Have you read this, including all the follow-up?
https://github.com/openpnp/openpnp/pull/1095

Some code relies on overloaded interpretation of `actuate()` i.e. Boolean semantics as well as Value semantics. This is not just the Java `isAssignableFrom()` semantics that I see in your code. The same dual Boolean and value interface should be available in the GUI, but you seem to have deleted all that. I'm not saying it is impossible to rewrite in a more polymorphic way, but again, the justification would need to match the effort and risk. 

I see how the Actuator _interface_ could be beautified, but I doubt that you will have much simpler code _internally_, providing you guarantee the same functionality and full version migration, actuator value persistence, etc. taking into account that many users use Actuators from scripts and those should also continue to work. 

It seems you have applied global text search & replace, so you replaced "profile" with "mapping" in code that handles `MotionProfile`s, which is obviously wrong. Frankly, for such broad and intrusive PR, the expectation is to go through it with a fine-tooth comb, before submitting. 

Please also merge the newest `develop` into your branch, so only this PR's changes are shown.

Thanks for explaining before continuing. 

_Mark

diorcety:
Hi,
Sorry I didn't follow the process. Initially I made such modification(and simplier) more than 6 months ago only for my self. All was broken when I try to rebase my work on the last dev (notably due too #1035, overwhelmed I was too).
My goal is to try to use the openpnp base work for my own machine using completly different driver for feeder and motor (not using gcode), and this not so straightforward due too few "overcomplicated" aspects of the software. Just for an example: 
I need to use a int as activator value, this is douable? Yes using a float and doing conversion, but seems more a hack that a real good solution.

There is lot of changes just in the goal to remove the ambiguity about the actuator values, and allow code simplification (like duplicated code https://github.com/openpnp/openpnp/blob/develop/src/main/java/org/openpnp/machine/reference/ReferenceActuatorProfiles.java#L325)

> Some code relies on overloaded interpretation of actuate()
Didn't say I cover 100% of the cases. I'm opened to discuss about that

> Why have you renamed "Profiles" to "Mapper"

Not needed indeed but closer to the function of the object: Mapping one value of the actuator type to actuate other actuators with other values (and types)

>  Do you really need more than 6 actuators?

Do you really need more than 5 actuators? Sorry for the sarcasm...

> The same dual Boolean and value interface should be available in the GUI, but you seem to have deleted all that.

This is still doable but in another way, but creating another actuator that from a boolean "value" will map to other value for example. This allow more flexibility and removing duplicated codes: 
https://github.com/openpnp/openpnp/blob/0826ee7a826725b57cf3af680a008d0143176f87/src/main/java/org/openpnp/spi/base/AbstractActuator.java#L31

https://github.com/openpnp/openpnp/blob/0826ee7a826725b57cf3af680a008d0143176f87/src/main/java/org/openpnp/spi/base/AbstractActuator.java#L36

https://github.com/openpnp/openpnp/blob/develop/src/main/java/org/openpnp/machine/reference/ReferenceActuatorProfiles.java#L240

>It seems you have applied global text search & replace, so you replaced "profile" with "mapping" in code that handles 

Sorry indeed my IDE changes also other stuffs I will recheck that. That why I put a WIP, I didn't expect to be merge soon, didn't expect at all now.

markmaker:
@diorcety 

I understand these concerns about code duplication and about the limit in the number of actuators in the profiles. I'm happy to concede that some of these are simply lack of knowledge on my side, of how to do these both simply _and_ dynamically in Java and more importantly inside WindowBuilder, while keeping the GUI complexity and usability under control. 

The limit for 6 actuators came from discussions about how to control multi-channel lights. From the possibilities discussed there, I concluded that 5 were enough, and I added one spare, while keeping everything straightforward i.e. adding more would be easy. 

But you're welcome to improve these aspects. Ideally one by one in separate PRs. 

As far as I understand how OpenPnP `machine.xml` persistence works, one should probably try to avoid renaming stuff. Also, IMHO "mapping" is not really better than "profile". [Mapping ](https://en.wikipedia.org/wiki/Map_(mathematics))is usually 1:1 not 1:n, otherwise it is called [multimapping](https://en.wikipedia.org/wiki/Multimap). I chose "profile" both for how it is used in computer science (as a set of certain properties) and for the geometric meaning, with the (light) actuator channels forming a profile, like the sliders on an audio equalizer. 

_Mark

diorcety:
The profile things is really a good part but looks really like a explicit type conversion (from one actuator to many others), that sound to me a good way to replace the implicit conversion. In this PR I clearly limit the actuator combox box to the wanted type: Light of the camera is either on or off from the camera POV. The string value seems strange.
At the end the result is quiet similar: https://imgur.com/a/rKMXBzo
Excluding that doesn't introduce another type to the activator

>  IMHO "mapping" is not really better than "profile"
I'm not good at naming, I agreed. The mapping was more like mapping 1 activator state to 1 state of multiple activators.

markmaker:
@diorcety 

Thanks for your answers.

>  that sound to me a good way to replace the implicit conversion

I see, but do you agree, that this is much more work to create and then much less clear to use and maintain for the average user? In comparison, Default ON/OFF is an easy-to-understand concept, I believe. I don't think it is worthwhile to make the code simpler, but then burden the user with the complexity. 

The double actuator solution also does not easily solve hybrid use cases such as controlling the light both ON/OFF _and_ with specific values. For instance, see the `ImageCapture` pipeline stage. Your solution to put the overriding actuator into the pipeline is not as versatile, because pipelines should remain as machine and camera agnostic as possible, so OpenPnP can ship with good stock pipelines, and they can be shared between users. Future development might involve multi-camera setups where you want to share the pipeline too. Again, a fixed actuator is a hindrance there. Of course, you could _dynamically_ resolve the profiled/mapped target actuator instead, but I still think this is much more complicated for the user to setup. 

So I hope you understand that I would not like to part with Default ON/OFF. Of course @vonnieda has the final say. 

>  Light of the camera is either on or off from the camera POV. The string value seems strange.

The idea is to control any type of controller. If you have a controller that takes hexadecimal RGB values, then you need to send `"FFFF00"` for yellow. Perhaps read more of #1095. 

> In this PR I clearly limit the actuator combox box to the wanted type:

Yes, that would be a valuable addition. There is some migration concern there, because Actuators were untyped in older versions of OpenPnP. So it would have to be tolerant when the type is not yet determined. See `AbstractActuator.valueTypeConfirmed`. 

As for the profiles, I really believe that the current tabular overview has better usability in comparison to the tabbed view:
grafik

But please don't misunderstand: your criticism about the 6 actuator limit and code duplication is valid. The possibility to add as many actuators as wanted and to reduce code duplication would be very welcome. If it could be integrated into something that gives the same easy overview to the user, it would be a great addition. 

Maybe you're concerned that the table will become too wide? Personally, I don't think you will ever need many more actuators/channels per profile, because if you think about reasonable _technical_ solutions in the back-end, you would at some point put the channel logic into a dedicated controller and then control it using a single String code per profile, perhaps a hexcode as explained above. Note that String values can include multiple params i.e. something like `"R=255&G=255&B=0"` for a HttpActuator is possible. This is also faster than sending many separate commands to many actuators. 

_Mark

PR #1156, opened 2021-03-23
Use interfaces in spi instead of Reference implementations

diorcety:
Simply remove Reference* class import in spi package. That kind break the abstraction pattern.


markmaker:
Looks good to me, and seeing it was my mistake, I think I can merge it right away. 

For reference: #1035 
> Remodeled ReferenceDriver interface into spi Driver interface and class hierarchy similar to other machine objects.
I missed that then.

Thanks @diorcety !

PR #1159, opened 2021-03-27
Feature: HeapFeeder

doppelgrau:
# Description
Implements a new type of feeder.
The HeapFeeder is a high density feeder (around 3cm² for each part), that stores the parts loose in bins.

# Justification
I think this feeder can be useful for other machines too. It is very slow, but with the high density so many parts can be stored. (For example a full E24 resistor collection and a nice collection of capacitors needs less space than a sheet of paper).
So for machines that are used for prototypes and very small batches, where the setup time is the larges factor it can be a nice addition.

# Instructions for Use

## Overview
The HeapFeeder is a high density feeder implementation.

HeapFeeder_Overview

The core concept is, that the parts are stored in heaps, a random number is picked using the nozzle and thrown in a drop box. There vision is used to determine which parts are upside up and the orientation. If no parts are upside up, parts are picked up and dropped again, until it is by chance upside up.

A vacuum sensor is required, blow of functionality (to avoid parts sticking to the nozzle) and good top vision system is highly recommended.

Since it heavily utilizes vision, makes multiple moves for each part it is a very slow feeder, but over 200 feeder can be placed in the space required for a single A4 paper sheet.

Example 3D printed parts can be found [here](https://github.com/doppelgrau/PnP/tree/master/HeapFeeder).

## Heap

Each heap is a simple „bin“, where the parts are stored in 3 dimensions in a chaotic order. From the OpenPnP side the Heap is defined by the top location and the depth. The size is assumed, that around the top location there is at least 1.25mm play in each direction, at any (valid) nozzle height.

HeapFeeder_FeederConfig

The other settings are:

- The used DropBox
- Three way-points used to move the parts from or to the heap. So it can be avoided that parts not safely attached to the nozzle drops into an other heap, by choosing a safe way (See below).
- FeedRetry Count like in any feeder, the number of feed retries.
- PickRetryCount like in any feeder, the number of feed retries.
- Max flip attempts says how many times a part is picked up and thrown away without finding any part upside up, before the content of the DropBox is purged and tried again with new parts. Total attempts is currently hard coded at 12 tries. After that there is a feed failed exception raised.
- LastFeedDepth the depth relative to the center top, where the last time parts were found. Usually the „searching“ for a part is limited to 3 times (to avoid going down forever if something is wrong with the vacuum), the part height, except the Last Feed Depth is zero.
- The next start for part searching is a bit higher, so all parts on the same level can be caught before the nozzle dives further down.
- Vacuum Difference the required difference in the vacuum sensor to assume a part sticking on the nozzle. Since parts are often only partly sticking to the nozzle, the „partOn“ from the nozzleTip can not be used.
- Part like in all feeders the part in that feeder. I would not recommend anything larger that 1206 or anything with many pins that could bend. Non polar parts like resistors and capacitors are the easiest, but with good top- and bottom vision also polar parts like diodes can be used.
- DetectionPipeline is the vision pipeline that is used to detect the parts, upright up. Should have a very low false positive rate. Used template matching. Reset gives a good default for DropBoxes with the color green, white or black.
- TemplatePipeline is used to create a template image. The template should show the part upside up in the desired 0° orientation. Reset gives a good default for the current selected drop box, with the color/name green, white or black. If a good template is generated, it has to be moved from the „templates/new“ folder inside the openpnp configuration in the „templates“ folder.
- Get Samples get some parts from that heap and drops them in the dropBox. Can be used for creating a template.

HeapFeeder_Move

## Drop Box

The drop box has two functions:

- Deliver a good background for the vision. So depending on the part color, choose a matching drop box.
- Contain the parts when dropped into the drop box, but at the same time increase the change of touring the part while dropping.

Drop boxes are shared between the heaps.

HeapFeeder-DropBoxSettings

The settings for the drop boxes are way simpler:

- Center Bottom marks the center point of the drop box. Used to position the camera for the vison pipelines and to determine the pick height in combination with the part height.
- Drop Location determines where the parts are dropped. I use the sidewalls as drop location, to increase the chance that the part bounces a bit in the drop box, so it hopefully flips over.
- Part Pipeline is used to detect any part in the drop box. Used to make sure the drop box is empty and to find parts that can be thrown in again. Should have a very low false negative rate.
- Dummy Part is the part, if unknown what part is currently in the drop box. Usually the last used heap is known and that part is used, but in some circumstances (e.g. after start up) the part is not known.
- CleanDropBox removes the parts from the drop box. If the last heap is known, the parts are put back in the heap. Else the parts are discarded.

# Implementation Details
1. How did you test the change? Did run some PCBs on my machine.
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. - No changes necessary there.


doppelgrau:
Pictures for the description/wiki

vonnieda:
Looks good to me @doppelgrau! Congratulations on getting this released, and nicely done on all your hard work!

markmaker:
Hi @doppelgrau,

For a new little PR, I need to associate pipelines in the code with their default pipelines. Remember, these are stored here in these directories:

https://github.com/openpnp/openpnp/tree/develop/src/main/resources/org/openpnp/machine/reference

Now, I need to make sure your `ReferenceHeapFeeder` default pipelines are not mixed up. It seems that way, because "Part" is reset to "DropBox", "Feeder" is reset to "Part". "Training" is reset to "Training" (OK).

https://github.com/openpnp/openpnp/blob/810146ee712f85f7a78af9f9fea02287a1a30eff/src/main/java/org/openpnp/machine/reference/feeder/ReferenceHeapFeeder.java#L896

https://github.com/openpnp/openpnp/blob/810146ee712f85f7a78af9f9fea02287a1a30eff/src/main/java/org/openpnp/machine/reference/feeder/ReferenceHeapFeeder.java#L551

https://github.com/openpnp/openpnp/blob/810146ee712f85f7a78af9f9fea02287a1a30eff/src/main/java/org/openpnp/machine/reference/feeder/ReferenceHeapFeeder.java#L555

So my question is this: is this a bug (i.e. they are really mixed up) or is this just unfortunate naming?

_Mark

doppelgrau:
Naming:

- Part/DropBox should detect ANY part in the dropBox (low false negatives) (the call is also in the DropBox class)
- Training/Training should create a good template for the parts in the Heap(Feeder)
- Feder/Part should detect a part that can be picked (low false negatives)

markmaker:
Hmmm.. that does not really answer my question 😁 

Note, I'm not trying to _use_ or _understand_ the Heap Feeder. I'm working on a generic function across all the vision use cases. Then I stumbled over this naming inconsistency, and just wanted to alert you to a _potential_ bug. 

l'll just leave things as they are.

_Mark

doppelgrau:
I thought the first line made it clear, just naming, that is not ideal from a "big picture" point of view.

PR #1161, opened 2021-03-30
Fixed issues in PartSizeCheck feature in ReferenceBottomVision

my-3d-cnc-pnp:
# Description
Fixed issues in partSizeCheck code found while testing this feature.

# Justification
I tried the new version of size checking using the parts definition height and width added a few weeks ago.
I tried the partSizeCheck on a 1206 resistor and encountered a few problems.

(1) The partSizeCheck in findOffsetsPreRotate never returns an error even if I intentionally define wrong part size.
I found that a false result from the size check code only causes the "for(int pass = 0;;) " loop to reiterate until
maxVisionPasses is reached. The part is placed even if size check fails. I did not see any code to display an error message.
Also if the center offset or corner offset or angle offset is still to large after maxVisionPasses then partSizeCheck is never even called. 
I added the Exception Message display code from findOffsetsPostRotate below the loop and changed the scope of "rect"

(2) Contrary to never displaying size error in PreRotate code, The partSizeCheck in PostRotate always displayed
the part Size Error message. I found this error in partSizeCheck code which sets the wrong size for width.
`
double width = checkHeight; <<<<<< SHOULD BE checkWidth >>>>>>>>>>>
`

(3) After fixing (1) and (2),  I flipped the resistor on its side and the size check did not fail and placed the part.
I found that the MaxWidth check is missing the "return false"
```
if (measuredSize.width > pxMaxWidth) {
    Logger.debug("Package pixel width {} : limit {} : measured {}", pxWidth, pxMaxWidth, measuredSize.width);
    <<<<<<<<<< missing "return false" >>>>>>>>>>>>>>>
} else if (measuredSize.width < pxMinWidth) {
    Logger.debug("Package pixel width {} : limit {} : measured {}", pxWidth, pxMinWidth, measuredSize.width);
    return false;
```
# Instructions for Use
As before

# Implementation Details
On my machine with 1206 resistors changing part size to invoke fails
Yes (I hope so!)
 No
 I ran the test

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -119,14 +119,15 @@ public class ReferenceBottomVision implements PartAlignment {
         Location nozzleLocation = wantedLocation;
         MovableUtils.moveToLocationAtSafeZ(nozzle, nozzleLocation);
         final Location center = new Location(maxLinearOffset.getUnits());
-
+        RotatedRect rect;
+        
         try (CvPipeline pipeline = partSettings.getPipeline()) {
 
             // The running, iterative offset.
             Location offsets = new Location(nozzleLocation.getUnits());
             // Try getting a good fix on the part in multiple passes.
             for(int pass = 0;;) {
-                RotatedRect rect = processPipelineAndGetResult(pipeline, camera, part, nozzle);
+                rect = processPipelineAndGetResult(pipeline, camera, part, nozzle);
                 camera=(Camera)pipeline.getProperty("camera");
 
                 Logger.debug("Bottom vision part {} result rect {}", part.getId(), rect);
@@ -179,14 +180,17 @@ public class ReferenceBottomVision implements PartAlignment {
                             offsets, Math.abs(angleOffset), getMaxAngularOffset());
                 }
                 else {
-                	if(partSizeCheck(part, partSettings, rect, camera)) {
-                        // We have a good enough fix - go on with that.
                         break;                		
-                	}
                 }
 
                 // Not a good enough fix - try again with corrected position.
                 nozzle.moveTo(nozzleLocation);
+            }
+            if (!partSizeCheck(part, partSettings, rect, camera) ) {
+                throw new Exception(String.format(
+                        "ReferenceBottomVision (%s): Incorrect part size.",
+                        part.getId() 
+                        )); 
             }
             Logger.debug("Offsets accepted {}", offsets);
             // Calculate cumulative offsets over all the passes.  
@@ -274,7 +278,7 @@ public class ReferenceBottomVision implements PartAlignment {
         // Make sure width is the longest dimension
         if (checkHeight > checkWidth) {
             double height = checkHeight;
-            double width = checkHeight;
+            double width = checkWidth;
             checkWidth = height;
             checkHeight = width;
         }
@@ -302,6 +306,7 @@ public class ReferenceBottomVision implements PartAlignment {
 
         if (measuredSize.width > pxMaxWidth) {
             Logger.debug("Package pixel width {} : limit {} : measured {}", pxWidth, pxMaxWidth, measuredSize.width);
+            return false;
         } else if (measuredSize.width < pxMinWidth) {
             Logger.debug("Package pixel width {} : limit {} : measured {}", pxWidth, pxMinWidth, measuredSize.width);
             return false;
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -126,7 +126,7 @@ public class ReferenceBottomVision implements PartAlignment {
             Location offsets = new Location(nozzleLocation.getUnits());
             // Try getting a good fix on the part in multiple passes.
             for(int pass = 0;;) {
-            	RotatedRect rect = processPipelineAndGetResult(pipeline, camera, part, nozzle);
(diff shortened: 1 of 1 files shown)


markmaker:
Hi @my-3d-cnc-pnp 

Thank you for making the extra effort for the PR!

This already looks good, and I would be OK to merge it. However, I wonder if it could even be slightly improved for pre-rotate.

The size check is now only done at the end of the multi-pass loop. I wonder if it should instead throw immediately inside the multi-pass loop. The rationale is that when the size check fails, then the part is (as far as vision can tell us) missing or tombstoned or whatever, and any attempt at centering it in further passes is doomed or even misdirected, sending the machine to some random speck in the image, mistaken as the part. 
 
### Cosmetics

- While removing the `if ()`, you should keep the comment "We have a good enough fix..." here:
https://github.com/openpnp/openpnp/blob/edd1bb236f6ea07f5c4e7a8183d0104fc3a36327/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java#L181-L184

- Consider fixing the PR title and description text formatting. You might be surprised how instructive these can be for the community and future contributors.  :-D

Note it is easy to amend a PR: just commit new changes and then push again. 

Thanks again.

_Mark

my-3d-cnc-pnp:
Thank you for your feedback Mark!

I fixed up the Description and Justification sections a bit.  Sorry about that.
I must have done something wrong in the copy and pastes and the deletes that I did. (I was trying to edit out some of the extra text and damaged the font sizes).  I will work on the formatting and also on my grammar some more.

I do not understand what is wrong with the PR title. (I shortened it a bit)

I left the size check as the last check because that was how it was in the original code. I thought it was that way so that the part would be centered and squared so the size check could get the best possible measurement on a non distorted rectangle.

I see your point though and will move the size check to be the first check if you think it is best.

So far I have only tested the new size check with 1206 parts.  
In the next weeks I will be running a board with smaller and larger parts and see if other issues show up.

Best Regards
Peter

markmaker:
Cool, thanks @my-3d-cnc-pnp 

> I left the size check as the last check because that was how it was in the original code. I thought it was that way so that the part would be centered and squared so the size check could get the best possible measurement on a non distorted rectangle.

Hmmm... I thought about this too. But I assumed that the part will never be _that_ much distorted so close to the camera center. I expect the pick offsets to be something in the order of half a millimeter max, right? I.e. you will never see the side of the part (which could in deed enlarge the detected outline). Or am I mistaken? My practical experience _is_ limited though!
 
_Mark

my-3d-cnc-pnp:
I modified the code to do the Size Check first. 
I did some testing with my 1206 and when the part is centered on the nozzle it will pass the size check at 1% tolerance. It fails at 0%. 
If I slide the 1206 to the end of the nozzle it fails and I have to up the tolerance to 2%.
If I rotate the 1206 to a little less than 45 degrees it fails at 2% and I have to up the tolerance to 4%.
I don't think anyone will run at these low tolerances anyway but it shows that it makes some difference (at least on my setup)

_Peter

markmaker:
Hi @my-3d-cnc-pnp 

Thanks. 

I would expect the angle to be relevant in _absolute_ terms rather than _relative_ i.e. alignment offset's terms. If you have 45° pre-rotate / placement angle you will probably get the same size discrepancy even if the angle was picked correctly i.e. has little alignment offset, right? And this is also easily explained when you think about the pixel raster and it being "jagged" at 45°, rather than smooth at 0°.  

The 4% tolerance doesn't bother me and I expect it needs to be even larger for smaller parts. The important thing is to detect a tombstoned or missing part and I guess this will for most parts still work at 20% or 30%. 

Can you push it?
 
Thanks!

_Mark

my-3d-cnc-pnp:
Done!  Thank you.

Peter

markmaker:
Hi @my-3d-cnc-pnp 

Brilliant! 

Just some cosmetics remain. As this is your first PR and assuming you are like me and want to know these things ;-), you can apply automatic formatting over the changed sections. From the [Coding Style section](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style):

> There is an Eclipse formatter definition [here](https://github.com/openpnp/openpnp/blob/develop/OpenPnP_Eclipse_Formatter.xml). You can import this into Eclipse and use it to format your code to be sure it's following the standard.

Just select the section and press `Ctrl-I`

This will make sure the whitespace  (all spaces) and other formatting is correct. This helps for viewing files in Github because tabs are indented differently there, and it reduces change artifacts through reformatting in the future, improving `git blame`: 

grafik

_Mark

my-3d-cnc-pnp:
Formatted and committed and pushed!

Peter

markmaker:
Hi @my-3d-cnc-pnp 

Seems the rebuild is a special privilege. I triggered it, and it passed. 

@vonnieda if you're reading this, do you know why these can intermittently fail?

_Mark

markmaker:
Thank you @my-3d-cnc-pnp!

 :-)

vonnieda:
> Hi @my-3d-cnc-pnp
> 
> Seems the rebuild is a special privilege. I triggered it, and it passed.
> 
> @vonnieda if you're reading this, do you know why these can intermittently fail?
> 
> _Mark

It's possible it failed due to being cancelled by a newer commit. I'm not sure though, and since it was re-run I can't see the history. Next time it happens, before re-running, click on the Job and you can see the logs. The logs should show why it failed. Or just let me know and I'll look.

my-3d-cnc-pnp:
@vonnieda    First commit on day 1 was ok.  The commit on day 2 and commit on day 3 had this error in the job log..

    Tests in error:
    SampleJobTest.lambda$testSampleJob$1:109 » Pipeline stage "results" returned ...

    Tests run: 27, Failures: 0, Errors: 1, Skipped: 0

Peter

PR #1107, opened 2021-01-17
Feature / Nozzle Tip Tool Changer Goodies

markmaker:
# Description

Nozzle tip tool changer location capturing is a delicate matter. It typically takes a lot of tuning to get it just right. Setting up six or more nozzle tips can be very hard work. A new **Clone Tool Changer Settings** function can make this simpler. The locations, speeds and actuators are cloned from a selected template nozzle tip to the current nozzle tip. The locations are translated relative to the **First Location**.

For those that do _not_ have an automatic tool changer, a new **Manual Change Location** can now be set centrally on the nozzle. Formerly, the nozzle tip **End Location** was re-used as the manual changing location. This was often unknown to users, created seemingly erratic movement and crashes (default zero location, often offset into the negative by the nozzle offset). It also required users to duplicate the location for each nozzle tip.  

The new manual location is now also independent, switching the automatic tool changer on/off while tweaking the machine is now safe.  

# Justification
Re manual changing, see here:
https://groups.google.com/g/openpnp/c/wjW_ilQvdxE/m/QSjlg4DEAQAJ

# Instructions for Use

Cloning from another nozzle tip is easy. 

1. Capture the **First Location** as a reference for this nozzle tip. 
2. Select the nozzle tip you want to use as a template.
3. Press the **Clone Tool Changer Settings** button.
4. The locations will be cloned, but translated _relative_ to the **First Location**.

Clone

If you disable the automatic tool changer, a new **Manual Change Location** will appear:

Manual

For existing `machine.xml`, this location will be migrated from the first nozzle tip **Last Location**.

# Implementation Details
1. Tested on the simulation machine.
2. Did follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style).
3. No changes in the `org.openpnp.spi` or `org.openpnp.model` packages.
4. Successful `mvn test` before submitting the Pull Request.

PR #1164, opened 2021-04-03
Feature/Nozzle Tip Changer XYZ Calibration, Part Height, Feeder, Placement contact probing, Auto-Focus

markmaker:
# Description
This adds Nozzle tip changer XY and Z calibration using vision and probing. Used to calibrate the nozzle to be able to precisely contact probe part heights, feeder heights and placements. Also adds Auto-Focus based part height sensing for parts with enabled bottom vision. 

Watch the video:
https://youtu.be/9uFxV1-vnXw

## Main Features

* Vision calibrated nozzle changer slot locations (XY) based on template images. Also detects nozzle changer slot empty/occupied status and prevents collisions.
* Manual and automatic nozzle/nozzle tip/slot Z calibration for precision multi-nozzle leveling and nozzle tip length calibration.
* Adds Focus sensing method to cameras (FocusProvider interface). AutoFocusProvider default implementation.
* Bottom vision uses Focus sensing to determine the part height, if unknown 
* ContactProbeNozzle provides on-the-fly part height probing/learning, feeder Z calibration, placement height calibration. 

Nozzle Tip Changer Calibration, Height Sensing

## Feature Details

* Nozzle tip specific touch location for faster after-loading Z calibration and varied tip lengths. Configurable triggers/options. Touch locations can be "Z-leveled" to single reference. 
* Adds _experimental_ "sniffle" method for contact probing _without_ dedicated sensor. May be useful for nozzle/nozzle tip Z calibration but obviously not feasible with part on the nozzle, i.e. pick and place height sensing is not possible.
* Nozzle tips have (practical) maximum part height and maximum part diameter settings to enable proper operation with unknown part dimensions.
* Dynamic Safe Z uses the maximum part height as long as a part height is unknown.
* Auto-Focus servos from maximum part height down and masks itself to maximum part diameter. Future usage with CircleMask possible.
* Job processor allows running jobs with part heights not yet set, it will only complain when no part height sensing method is available for parts without height. 
* Nozzle contact probing capability becomes part compatibility criteria in the planner. Multi-nozzle machines might only have one nozzle with probing sensor, so parts with unknown height will have to be loaded on that nozzle. Only relevant if Auto-Focus does not apply. 
* The placementLocation in the job processor is now the base location (i.e. underneath the part), in order to allow part height sensing on the fly (backwards-compatible Scripting, i.e. parameter works as before, new parameter placementLocationBase). 
* Feeders have new isPartHeightAbovePickLocation() method to determine if part height is above the pick location and must be sensed when part height is unknown (loose part feeder classes).
* As predicted long ago, the Nozzle now has overridable moveToPickLocation() and moveToPlacementLocation() methods. Plain moveToLocationAtSafeZ() calls in job processor and GUI actions were replaced.
* ContactProbeNozzle now overrides these methods to provide on-the-fly probing. Can be done each time or only once and learned (faster Z moves without probing).
* Auto focus of a Z calibrated nozzle tip can be used to set precise focal plane of bottom camera,
* Calibrations triggered by machine homing now have an option to enable/disable failing the homing cycle.
* UiUtils.messageBoxOnExceptionLater allows for deferred/multiple GUI related actions. Needed in the homing cycle when multiple calibrations can fail (and failing the homing cycle is disabled).
* Some improved diagnostic messages on errors.
* Test Alignment GUI action now uses the current nozzle rotation angle (to test pre-rotate + Auto-focus).
* Visual homing fiducial locator vision Test button.
* Made LocationsButtonsPanel more configurable to use in situations when some buttons are unwanted. Removed duplicate buttons/actions definitions (in ReferenceHeadConfigurationWizard). Added proper setEnabled() handling.
* Simple depth of focus (blur) simulated in the SimulatedUpCamera to test Auto-Focus. Checkbox on the Wizard (slow!)
* Issues&Solutions for ContactProbing was improved/adapted.
* Issues&Solutions can convert ReferenceNozzles into ContactProbeNozzles while preserving configuration.

## Bugfixes during testing

These bugs hindered testing, therefore I presume inclusion is justified.

* Edit pipeline on ReferencePushPullFeeder did not wait for motion completion, therefore camera capture lighting failed.
* Polymorphic Object type property sheet handlers need resetting when switching stages.
* Recycle part to feeder feature: Missing null handler when feeder has no part.
* Pipeline stages now at least record exceptions in the log. Issue of swallowed machine exceptions is still not solved.
* various.

## Breaking Changes

* ContactProbeNozzle retraction (Actuator `false` actuation) now happens immediately after probing. Previously it was `false`-actuated after picking/placing. Default Gcode for `{False}` is now empty, retraction is controlled by GUI and done by normal Z move.
* moveToLocationAtSafeZ() will only move to safe Z if the location changes in X, Y or C. This allows for quick Z only relocation. Happens when bottom vision follows auto-focus or when pick follows BlindsFeeder push-cover opening etc.

# Justification
To justify that these very different parts belong together into one PR, one must look at the dependencies.  

This started out as a better-contact-probing PR. But it soon became evident, that nozzle tip Z needs to be calibrated for the required precision (0402 parts are very thin). 

Hence, probing Z calibration was also needed. Different nozzle tips may have different lengths (certainly the "unloaded" nozzle tip (stand-in for the naked nozzle) needs a different touch reference Z). And the extra move to a common touch location also seemed quite slow. Hence, it was integrated in the changer locations series and into automatic cloning/translating (see also #1107). 

But now that Z was calibrated it seemed odd not to calibrate X, Y as well. This was emphasized by a few nasty collisions when setting things up and testing. Hence, the template image based X, Y slot calibration. As a bonus, it can also detect if the slot is empty or occupied and prevent the painful collisions I had. 

When I finally started to implement part height auto-learning, I realized that for parts with bottom vision, the part height is required _before_ it can be probed in placement (except for the loose part feeders, where it can be probed in the pick). Hence, the Auto-Focus was needed. This also immediately made this PR much more widely useful, i.e. part height sensing/learning is now available for all parts with vision enabled, even if you don't have a contact probing nozzle. 

But you still need Z calibrated tips for Auto-focus to be precise enough, hence "sniffle" probing using the vacuum system. Though that turned out "experimental" on my machine (using a stiff vacuum reservoir and no flow restriction, the tests do not work reliably on the finest nozzle tips). Maybe it will work on direct pumps.

# Instructions for Use
In preparation for the Wiki. I'm partially using Wiki markup links here, they won't work until in the Wiki. 

### What is it?
The ContactProbeNozzle can probe the surface of things in Z. This can simply be used to perform probing picks and places all the time, i.e. Z coordinates do not need to be very precise, the nozzle will automatically stop when contact is sensed. 

But this does not cover other applications where Z coordinates must be quite precise without making contact. Examples are sideways nozzle tip tool changing, special feeder movements like the [[BlindsFeeder]] cover actuation or even part picking where spring-loaded/hard contact is not desired ([suspended tape feeders](https://docs.mgrl.de/maschine:pickandplace:feeder:manualfeeders)). 

Machines that home the Z axis with easily shifted Zmax switches ([e.g.](https://liteplacer.com/the-machine/assembly-instructions/pnp-head-step-11-attach-z-high-limit-switch/) Liteplacer), hard to balance dual-nozzle mid-axis sensors or even by just letting springs retract unpowered Z motors, may not provide a precise and repeatable Z reference eternally.  

A ContactProbeNozzle can then be used to calibrate the Nozzle in Z, more specifically the point of the loaded Nozzle Tip. By probing a known reference Z touch surface, it can then apply the obtained calibration offset to all future Z movements. 

For multi-nozzle machines, the nozzles can therefore be perfectly harmonized in Z. This can (optionally) be extended to individual nozzle tips if these have varying heights, or if the coupling to the nozzle is not very consistent. 

Having a calibrated Z not only establishes precision in movements, but also in capturing or probing other locations. 

### Enable the ContactProbeNozzle 
If you don't have a ContactProbeNozzle yet, let [[Issues and Solutions]] replace the existing ReferenceNozzle for you:
 
Replace nozzle

This extends the Nozzle with a new tab **Contact Probe**.

### Method Selection

Contact Probe Method

**Method** lets you select the probing method. 

- **None** switches contact probing off. The nozzle behaves as if it were a regular ReferenceNozzle. 
- **Vacuum Sense** uses the Vacuum Sensing setup of the nozzle to perform "sniffle probing". More specifically, the [Part Off](https://github.com/openpnp/openpnp/wiki/Setup-and-Calibration%3A-Vacuum-Sensing#measurement-method) detection method is used to sense if the nozzle tip has touched the surface. 
- **ContactSensingActuator** uses a separate Actuator to probe and sense for contact. This is a setup made popular by the Liteplacer kit. 

Notes: The Wizard will show different fields according to the selection. If at least one nozzle has an active **Method** selected, other parts of the OpenPnP GUI will also show additional elements for contact probing. 

### Vacuum Sense Method

Vacuum Sense

**Start Offset** determines how high above the nominal probing location, the sensing should start. For the **Vacuum Sense** method, this must be high enough to get a clear "Part Off" result on the first sniffle.

**Probe Depth** determines how far the probing should go. 

**Sniffle Increment** determines the step in Z. 

**Sniffle Dwell Time [ms]** sets the waiting time between two sniffles. This is needed to clear out any under-pressure from the previous sniffle.

**Final Adjustment** after contact was sensed, add this adjustment to the Z position. Positive values retract the nozzle, negative values add spring loading. You should aim for a light touch with barely any spring loading.  This way the tip Z reference can also be used to push things sideways such as with the [[BlindsFeeder]] cover actuation. 

**Calibration Z Offset** shows the last calibration result. 

**Calibrate Now** performs the Z calibration. You must first setup a **Touch Location** on all the noozle tip's **Tool Changer** tab. 

### Contact Sense Method

ContactSenseActuator

**Start Offset** determines how high above the nominal probing location, the sensing should start. For the **Vacuum Sense** method, this must be high enough to get a clear "Part Off" result on the first sniffle.

**Probe Depth** determines how far the probing should go. In fact it may go further but if the probing result is beyond the given depth an exception is thrown. 

**Final Adjustment** after contact was sensed, add this adjustment to the Z position. Positive values retract the nozzle, negative values add spring loading. You should aim for a light touch with barely any spring loading. This way the tip Z reference can also be used to push things sideways such as with the [[BlindsFeeder]] cover actuation. 

**Feeder Height Probing** determines when feeders should be probed for pick location Z offsets. 

**Part Height Probing** determines when parts should be probed for placement location Z offsets. 

Probing Triggers

* **Off** switches off any probing.
* **Once** probes for a Z offset only once. The offset is stored in the configuration (machine.xml). 
* **AfterHoming** probes for a Z offset once after homing the machine.  
* **Each Time** probes for contact in all picks and placements. This is the most tolerant setting.  

**Calibration Z Offset** shows the last Z calibration result for this nozzle with the current nozzle tip attached. See also the Nozzle Tip Probing Configuration.

**Calibrate Now** performs the Z calibration. You must first setup the Nozzle Tip Probing Configuration.

### Part Height Auto-Learning

With a ContactProbeNozzle and enabled feeder and part height probing, OpenPnP can now also learn part heights automatically. Whenever a part with unknown height is picked or placed for the first time, the probed height will be stored on the part. 

* Part height auto-learning works for feeders that have the part height _above_ the pick location Z (ReferenceLoosePartFeeder and AdvancedLoosePartFeeeder). 
* Part height auto-learning also works in all placements, as obviously all parts are above the known board location Z. 

The job processor knows when a ContactProbeNozzle has this capability and will in this case allow starting a Job with unknown part heights. If only one nozzle in a multi-nozzle setup has probing capabilities, the planner will automatically restrict parts with unknown heights to this nozzle. As an alternative, part heights can also be auto-learned using auto focus, if bottom vision is enabled for the part. 

Part height unknown

### Part Dimensions

The first pre-requisite for probing are the maximum **Part Dimensions**, defined in the **Configuration** tab:

Part Dimensions

**Max. Part Diameter** determines the diameter (or diagonal) of the largest part you ever expect to pick with this nozzle tip. Add a few millimeters on top of data sheet dimension for tolerances. This property will be used in present and future up-looking camera operations to restrict the region of interest (Auto-Focus, MaskCircle etc.). 

**Max. Part Height** determines the height of the tallest part you ever expect to pick with this nozzle tip. Add a few tenths of a millimeter on top of data sheet dimension for tolerances. This property will be used to start probing above a part with unknown height. Furthermore, it will be used for dynamic Safe Z, again as long as a part height is unknown. 

### Tool Changer 

As soon as at least one ContactProbeNozzle has an active probing method, the Tool Changer tab will show extra elements:

Tool Changer

**Touch Location** lets you define a special Z probing touch location. It will be the reference for all other Z coordinates on your machine, therefore it should represent a fundamental and "eternal" reference surface for your machine.  At the same time it should be very close to the changer locations so the extra moves for Z calibration do not take much time.  

Nozzle Tip Change and Z Calibrate

After capturing the X, Y coordinates with the camera, you must initially probe the reference Z coordinate with the middle button:

Probe 

___
**CAUTION**: do not do this again later. You will lose your "eternal" Z reference! 
___

**Auto Z Calibration** determines when the Z calibration happens. 

Auto Z Calibration

* **Manual** switches automatic Z calibration off. You can manually calibrate and the obtained Z offset will be stored in the configuration (machine.xml).
* **MachineHome** performs automatic Z calibration if this nozzle tip is loaded when the machine is homed. When other nozzles are later loaded to the same nozzle, no additional Z calibration is triggered. **Z calibration is per nozzle**. This can handle uneven nozzles in multi-nozzle machines. 
* **NozzleTipChange** performs automatic Z calibration whenever this nozzle tip is loaded to a nozzle. In addition, Z calibration also happens if this nozzle tip is loaded when the machine is homed. **Z calibration is per nozzle tip**.  This can handle uneven nozzle tips with different heights or inconsistent coupling. 

**Fail Homing?** If enabled, aborts the homing cycle when Z Calibration fails as part of it. If disabled, continues the homing cycle and only displays an error message. 

**Z** (behind the **Auto Z Calibration**) indicates the last obtained Z calibration offset. 

**Reset** removes the current Z calibration offset. 

**Calibrate now** performs manual Z calibration. 

**Note:** the Z calibration affects all the nozzle Z motion as well as the capturing of coordinates from the current nozzle position. But it does not affect Safe Z, as Safe Z is relative to the raw axis coordinate system, not the calibrated nozzle Z coordinate system. To balance a multi-nozzle machine's nozzles, use the [Axis capture buttons](https://github.com/openpnp/openpnp/wiki/Machine-Axes#kinematic-settings--axis-limits) instead.  

### Cloning Settings

Cloning Settings

Among the other settings, the **Z Calibration** settings can be cloned to/from other nozzle tips, so you need to configure them only once. The **Touch Location** is automatically translated according to the **First Location** (along with the other locations). 

Marking this nozzle tip as the **Template** makes the **Touch Location Z** the global machine reference Z. 

Use the **Clone Tool Changer Settings to all Nozzle Tips** button to distribute the Z calibration settings to all the other nozzles. 

Use the  **Callibrate all Touch Locations' Z to Template** to reference them to the global machine reference Z.

## Vision Calibration

Using computer vision Nozzle Tip Changer locations can be calibrated in X and Y. The process uses two template images, of the empty and occupied changer slot respectively. It can therefore also detect if the slot is empty/occupied as expected.

**Vision Location** determines the changer location at which the template images are taken. Chose a location where the nozzle tip is visible when it occupies the slot. 
 
Vision Location

**Calibration Trigger** determines when the vision calibration takes places:

Calibration Trigger

* **Manual** allows only manual calibration. The calibration will be stored in the configuration (machine.xml)
* **Machine Home** invalidates the calibration when the machine is homed. As soon as nozzle tip load/unload is requested, the slot will first be vision calibrated. 
* **NozzleTipChange** recalibrates the the locations on each nozzle tip change. 

Vision Calibration

**Template Width** and **Template Height** determine the dimensions of the template image. Use a crop that includes relevant horizontal and vertical edges of the changer slot surrounded by an extra millimeter or so. The nozzle tip must be visible, when it occupies the slot. You can use the **Capture** buttons to optimize.  

**Tolerance** determines by how far from the nominal location the vision detected location may be. 

**Wanted Precision** determines at what distance the obtained template image match location is close enough. Otherwise, the camera is centered to the detected location and another vision pass is made (drill down).

**Max. passes** limits the maximum number of vision passes. 

**Minimum Score** determines how similar the camera image must be in comparison with the template image (0.0 ... 1.0). If the score is not reached, the calibration fails. 

**Last Score** shows the template match score obtained from last live or test calibration. 

Use the **Test** button to test the calibration. 

Use the **Capture** button to capture the respective **Template Empty** or **Template Occupied** image. The camera will automatically move to the selected **Vision Location**. You can hover over the captured template thumbnail to see it enlarged in the camera view. 

Use the **Reset** button to remove templates images. 

The two images are both used and the better match tells OpenPnP if the slot in question is empty or occupied. If this does not correspond to the state internally stored by OpenPnP, it will abort the calibration and therefore the nozzle tip load/unload operation. Nasty collisions can be safely prevented. 

Found occupied

## Up-looking Camera Auto-Focus

Unknown part heights can automatically be detected by auto focussing its underside. The nozzle will servo up and down to find the optimum. 

As a pre-requisite, the **Part Dimensions** must be defined on each nozzle tip (Configuration tab). Both the **Max. Part Diameter** and **Max. Part Height** are used to determine the autofocus range and senstive area. 

Part Dimensions

To enable the auto focus on the camera, select the **Focus Sensing Method**. It is only available on up-looking cameras: 

Enable Auto-Focus

Once you press **Apply**, a new **Auto Focus** tab appears:

grafik

**Focal Resolution** determines the maximum step size for the Auto Focus servo. 

**Averages Frames** lets you run the autofocus for multiple frames and average the results. This will improve the accuracy by effectively averaging out image noise etc. 

**Focus Speed** sets the relative motion speed for the focus stepping. 
 
**Show Diagnostics?** switches on auto focus edge highlighting and status info in the camera view.  

**Last Focus Distance** indicates the last live or test focus distance measured. If there is a part on the nozzle, this is the part height. 

Press the red dot button to test the auto focus.  

Auto Focus

### Calibrating the Up-looking Camera Z

The auto focus can help you calibrate the precise up-looking camera Z position. This will make sure your bottom vision happens at optimum distance. 

___
**CAUTION**: this procedure will change the subject to camera distance, which will change the scale of subjects. You will need to readjust the **Units per Pixel** after performing it. 
___

1. Use a Z calibrated nozzle tip with no part on. 
2. Auto focus it, using the red dot button in the **Auto Focus** camera tab. . 
3. Press the **Adjust Camera Z** button to set the camera Z Position to the detected distance. 
4. Go to the **General Configuration** tab to adjust **Units per Pixel**.

### Part Height Auto-Learning

Whenever Alignment (a.k.a. Bottom Vision) encounters a part with unknown part height, it will employ the auto focus. The detected part height is then stored on the part. This will take some additional time (but hardly longer than manually measuring and entering the data). All subsequent vision and placement operations will reuse the height directly, with no further speed penalty incurred. 

The job processor knows when a camera has auto focus capability and bottom vision is enabled for the part. It will then allow starting a Job with unknown part heights.  As an alternative, part heights can also be auto-learned using a ContactProbeNozzle.  

Part height unknown

# Implementation Details
1. Code was first tested on the simulation machine, hence the simulation of depth of focus. Then it was tested and tuned on my machine. Since I was also mechanically changing my machine (tool changer relocated), this turned out to be quite instructive. 
3. Made changes in the `org.openpnp.spi` or `org.openpnp.model` packages, as described, i.e. see Feeder.isPartHeightAbovePickLocation(), Nozzle.moveToPickLocation(), Nozzle.moveToPlacementLocation().

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,25 @@
 This file lists major or notable changes to OpenPnP in chronological order. This is not
 a complete change list, only those that may directly interest or affect users.
 
+# 2021-04-05 
+
+## Nozzle Tip Changer XYZ Calibration, Auto-Focus, Part Height, Feeder, Placement sensing 
+
+This adds Nozzle tip changer XY and Z calibration using vision and probing. Used 
+to calibrate the nozzle to be able to precisely contact probe part heights, feeder 
+heights and placements. Also adds Auto-Focus based part height sensing for parts 
+with enabled bottom vision. 
+
+* Vision calibrated nozzle changer slot locations (XY) based on template images. 
+  Also detects nozzle changer slot empty/occupied status and prevents collisions.
+* Manual and automatic nozzle/nozzle tip/slot Z calibration for precision multi-
+  nozzle leveling and nozzle tip length calibration.
+* Adds Focus sensing method to cameras (FocusProvider interface). 
+  AutoFocusProvider default implementation.
+* Bottom vision uses Focus sensing to determine the part height, if unknown. 
+* ContactProbeNozzle provides on-the-fly part height probing/learning, feeder Z 
+  calibration, placement height calibration. 
+
 # 2021-03-20
 
 ## (Another) OpenCV Upgrade and Arm7 (32 Bit)
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -52,15 +52,14 @@ import javax.swing.event.DocumentEvent;
 import javax.swing.event.DocumentListener;
 import javax.swing.event.ListSelectionEvent;
 import javax.swing.event.ListSelectionListener;
-import javax.swing.table.DefaultTableCellRenderer;
 import javax.swing.table.TableRowSorter;
 
 import org.openpnp.events.FeederSelectedEvent;
 import org.openpnp.gui.components.AutoSelectTextTable;
 import org.openpnp.gui.components.ClassSelectionDialog;
-import org.openpnp.gui.support.CustomBooleanRenderer;
 import org.openpnp.gui.support.AbstractConfigurationWizard;
 import org.openpnp.gui.support.ActionGroup;
+import org.openpnp.gui.support.CustomBooleanRenderer;
 import org.openpnp.gui.support.Helpers;
 import org.openpnp.gui.support.Icons;
 import org.openpnp.gui.support.MessageBoxes;
@@ -70,6 +69,7 @@ import org.openpnp.gui.tablemodel.FeedersTableModel;
 import org.openpnp.model.BoardLocation;
 import org.openpnp.model.Configuration;
 import org.openpnp.model.Job;
+import org.openpnp.model.Length;
 import org.openpnp.model.Location;
 import org.openpnp.model.Part;
 import org.openpnp.model.Placement;
@@ -493,6 +493,15 @@ public class FeedersPanel extends JPanel implements WizardContainer {
     	tableModel.fireTableChanged(null);
     }
 
+    protected Location preliminaryPickLocation(Feeder feeder, Nozzle nozzle) throws Exception {
+        Location pickLocation = feeder.getPickLocation();
+        if (feeder.isPartHeightAbovePickLocation()) {
+            Length partHeight = nozzle.getSafePartHeight(feeder.getPart());
+            pickLocation = pickLocation.add(new Location(partHeight.getUnits(), 0, 0, partHeight.getValue(), 0));
+        }
+        return pickLocation;
+    }
+
     public Action newFeederAction = new AbstractAction() {
         {
             putValue(SMALL_ICON, Icons.add);
@@ -552,7 +561,10 @@ public class FeedersPanel extends JPanel implements WizardContainer {
 
                 nozzle.moveToSafeZ();
                 feeder.feed(nozzle);
-                Location pickLocation = feeder.getPickLocation();
+                // Note, we do not use nozzle.moveToPickLocation(feeder) as this might involve 
+                // probing, which we don't want to happen here. Instead we need to use a preliminary 
+                // pick location. 
+                Location pickLocation = preliminaryPickLocation(feeder, nozzle);
                 MovableUtils.moveToLocationAtSafeZ(nozzle, pickLocation);
                 MovableUtils.fireTargetedUserAction(nozzle);
             });
@@ -586,34 +598,7 @@ public class FeedersPanel extends JPanel implements WizardContainer {
         }
         feeder.prepareForJob(false);
 
-        // Check the nozzle tip package compatibility.
-        Nozzle nozzle = MainFrame.get().getMachineControls().getSelectedNozzle();
-        org.openpnp.model.Package packag = feeder.getPart().getPackage();
-        if (nozzle.getNozzleTip() == null || 
-                !packag.getCompatibleNozzleTips().contains(nozzle.getNozzleTip())) {
-            // Wrong nozzle tip, try find one that works.
-            boolean resolved = false;
-            if (nozzle.isNozzleTipChangedOnManualFeed()) {
-                for (NozzleTip nozzleTip : packag.getCompatibleNozzleTips()) {
-                    if (nozzle.getCompatibleNozzleTips().contains(nozzleTip)) {
-                        // Found a compatible one. Unload and load like the JobProcessor.
-                        nozzle.unloadNozzleTip();
-                        nozzle.loadNozzleTip(nozzleTip);
-                        resolved = true;
-                        break;
-                    }
-                }
-            }
-            if (nozzle.getNozzleTip() == null) {
-                throw new Exception("Can't pick, no nozzle tip loaded on nozzle "+nozzle.getName()+". "
-                        +"You may want to enable automatic nozzle tip change on manual pick on the Nozzle / Tool Changer.");
-            }
-            else if (! resolved) {
-                throw new Exception("Can't pick, loaded nozzle tip "+
-                        nozzle.getNozzleTip().getName()+" is not compatible with package "+packag.getId()+". "
-                        +"You may want to enable automatic nozzle tip change on manual pick on the Nozzle / Tool Changer.");
-            }
-        }
+        Nozzle nozzle = getCompatibleNozzleAndTip(feeder, true);
 
         // Like in the JobProcessor, make sure it is calibrated.
         if (!nozzle.isCalibrated()) {
@@ -625,8 +610,7 @@ public class FeedersPanel extends JPanel implements WizardContainer {
         feeder.feed(nozzle);
 
         // Go to the pick location and pick.
-        Location pickLocation = feeder.getPickLocation();
-        MovableUtils.moveToLocationAtSafeZ(nozzle, pickLocation);
+        nozzle.moveToPickLocation(feeder);
         nozzle.pick(feeder.getPart());
         nozzle.moveToSafeZ();
 
@@ -643,6 +627,40 @@ public class FeedersPanel extends JPanel implements WizardContainer {
         MovableUtils.fireTargetedUserAction(nozzle);
     }
 
+    protected static Nozzle getCompatibleNozzleAndTip(Feeder feeder, boolean allowNozzleTipChange) throws Exception {
+        // Check the nozzle tip package compatibility.
+        Nozzle nozzle = MainFrame.get().getMachineControls().getSelectedNozzle();
+        org.openpnp.model.Package packag = feeder.getPart().getPackage();
+        if (nozzle.getNozzleTip() == null || 
+                !packag.getCompatibleNozzleTips().contains(nozzle.getNozzleTip())) {
+            // Wrong nozzle tip, try find one that works.
+            boolean resolved = false;
+            if (nozzle.isNozzleTipChangedOnManualFeed() && allowNozzleTipChange) {
+                for (NozzleTip nozzleTip : packag.getCompatibleNozzleTips()) {
+                    if (nozzle.getCompatibleNozzleTips().contains(nozzleTip)) {
+                        // Found a compatible one. Unload and load like the JobProcessor.
+                        nozzle.unloadNozzleTip();
+                        nozzle.loadNozzleTip(nozzleTip);
+                        resolved = true;
+                        break;
+                    }
+                }
+            }
+            if (nozzle.getNozzleTip() == null) {
+                throw new Exception("No nozzle tip loaded on nozzle "+nozzle.getName()+". "
+                        +"You may want to enable automatic nozzle tip change on manual pick on the Nozzle / Tool Changer.");
+            }
+            else if (! resolved) {
+                throw new Exception("Loaded nozzle tip "+
+                        nozzle.getNozzleTip().getName()+" is not compatible with package "+packag.getId()+". "
+                        +(allowNozzleTipChange ? 
+                                "You may want to enable automatic nozzle tip change on manual pick on the Nozzle / Tool Changer." 
+                                : ""));
+            }
+        }
+        return nozzle;
+    }
+
     public Action moveCameraToPickLocation = new AbstractAction() {
         {
             putValue(SMALL_ICON, Icons.centerCameraOnFeeder);
@@ -657,7 +675,8 @@ public class FeedersPanel extends JPanel implements WizardContainer {
                 Feeder feeder = getSelection();
                 Camera camera = MainFrame.get().getMachineControls().getSelectedTool().getHead()
                         .getDefaultCamera();
-                Location pickLocation = feeder.getPickLocation();
+                Nozzle nozzle = getCompatibleNozzleAndTip(feeder, false);
+                Location pickLocation = preliminaryPickLocation(feeder, nozzle);
                 MovableUtils.moveToLocationAtSafeZ(camera, pickLocation);
                 MovableUtils.fireTargetedUserAction(camera);
             });
@@ -676,9 +695,9 @@ public class FeedersPanel extends JPanel implements WizardContainer {
         public void actionPerformed(ActionEvent arg0) {
             UiUtils.submitUiMachineTask(() -> {
                 Feeder feeder = getSelection();
-                Nozzle nozzle = MainFrame.get().getMachineControls().getSelectedNozzle();
+                Nozzle nozzle = getCompatibleNozzleAndTip(feeder, false);
 
-                Location pickLocation = feeder.getPickLocation();
+                Location pickLocation = preliminaryPickLocation(feeder, nozzle);
                 MovableUtils.moveToLocationAtSafeZ(nozzle, pickLocation);
                 MovableUtils.fireTargetedUserAction(nozzle);
             });
--- a/src/main/java/org/openpnp/gui/JogControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/JogControlsPanel.java
@@ -26,7 +26,6 @@ import java.awt.FocusTraversalPolicy;
 import java.awt.Font;
 import java.awt.Window;
 import java.awt.event.ActionEvent;
-import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.util.HashMap;
 import java.util.Hashtable;
@@ -809,16 +808,21 @@ public class JogControlsPanel extends JPanel {
             // add property listener for recycle button
             // enable recycle only if part on current head
             PropertyChangeListener recyclePropertyListener = (e) -> {
-                    boolean canTakeBack = false;
-                    Part part = machineControlsPanel.getSelectedNozzle().getPart();
-                    if (part != null) {
-                        for (Feeder feeder : Configuration.get().getMachine().getFeeders()) {
-                            if (feeder.isEnabled() && feeder.getPart().equals(part) && feeder.canTakeBackPart()) {
-                                canTakeBack = true;
+                    Nozzle selectedNozzle = machineControlsPanel.getSelectedNozzle();
+                    if (selectedNozzle != null) {
+                        boolean canTakeBack = false;
+                        Part part = selectedNozzle.getPart();
+                        if (part != null) {
+                            for (Feeder feeder : Configuration.get().getMachine().getFeeders()) {
+                                if (feeder.isEnabled() 
+                                        && feeder.getPart() == part
+                                        && feeder.canTakeBackPart()) {
+                                    canTakeBack = true;
+                                }
                             }
                         }
+                        recycleAction.setEnabled(canTakeBack);
                     }
-                    recycleAction.setEnabled(canTakeBack);
                 };
             // add to all nozzles
             for (Head head : Configuration.get().getMachine().getHeads()) {
--- a/src/main/java/org/openpnp/gui/components/LocationButtonsPanel.java
+++ b/src/main/java/org/openpnp/gui/components/LocationButtonsPanel.java
@@ -36,6 +36,7 @@ import javax.swing.SwingUtilities;
 import org.openpnp.gui.MainFrame;
 import org.openpnp.gui.support.Helpers;
 import org.openpnp.gui.support.Icons;
+import org.openpnp.machine.reference.ContactProbeNozzle;
 import org.openpnp.model.Configuration;
 import org.openpnp.model.Length;
 import org.openpnp.model.Location;
@@ -62,8 +63,13 @@ public class LocationButtonsPanel extends JPanel {
     private JButton buttonCenterTool;
     private JButton buttonCaptureCamera;
     private JButton buttonCaptureTool;
-    
+    private JButton buttonCenterToolNoSafeZ;
+    private JButton buttonContactProbeTool;
+    private JSeparator separator;
+    private JButton buttonCenterCamera;
+
     private Location baseLocation;
+    private boolean contactProbeReference;
 
     public LocationButtonsPanel(JTextField textFieldX, JTextField textFieldY, JTextField textFieldZ,
             JTextField textFieldC) {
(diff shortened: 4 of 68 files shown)


markmaker:
Prepared Wiki documentation. See the **Instruction for Use** section in the PR description. 
_Mark

vonnieda:
@markmaker Does this use the existing z-probe system at all? Or replace it? If not, do you have any plans to combine the two?

markmaker:
Hi @vonnieda 

These are two distinct systems. 

1. Laser probe on head.
2. Contact probe on nozzle with the ability to stop motion in Z when contact is made.

We talked about it previously, here:
https://github.com/openpnp/openpnp/pull/855#issuecomment-500987727

I'm sure the laser is great for setup. To capture any surface Z including delicate locations that cannot be probed with a poking nozzle. Like small, easy to flip parts in tapes. 

However, the laser cannot in general be used for part height detection. In the pick, the part height can only be measured when the part is _above_ feeder pick location Z. That is only the case for the ReferenceLoosePartFeeder and AdvancedLoosePartFeeder classes i.e. very rare. In placement the laser is too late, i.e. the part height must already be known for the placement location Z, measuring afterwards does not make much sense (maybe for very thin parts, that are smaller than the spring travel). 

A contact probe nozzle can be used during placement, with the part on the nozzle, it will stop in real-time (controller side) when contact is made. Plus the contact probe can be used to Z calibrate nozzles and tips, as described in the instructions. Again the laser can't do that. 

We would need a second laser, _pointing up_, near the camera perhaps  (ok, dangerous) ;-).

There could even be more options: some systems use touch plates (or whatever they are called) with a stationary contact sensor. I tried to make the system extensible, so we can add these later.

On the other hand, auto-focus works surprisingly well too. And everybody already has a camera. 

I know the terminology is confusing. Any suggestions of how to untangle these?

_Mark

vonnieda:
Hi @markmaker - I have to say, much like your comment [here](https://github.com/openpnp/openpnp/pull/1155#issuecomment-803603190) I am just completely overwhelmed by the size of this PR. There are so many changes across so many systems I don't see how I can reasonably review it aside from just saying "Oh well, I give up, merge it." or "No".

At the very least, I think this should be broken up into:

- Nozzle Tip Changer vision
- Camera Auto Focus
- Z Probing - with an effort to merge the two systems (see below)
- Job Processor Part Height changes
- Nozzle Tip physical dimension additions
- Concept of pick location being above the part. This is a significant change to the feeder API and I think it needs discussion.

And probably several more - really just based on the feature list in your description.

I understand you feel these are all closely tied together, but that worries me even more. It appears there are some nice features here, but I think it's important that we prove these features stand on their own lest we make significant changes to OpenPnP's core that are only useful for LitePlacer.

Regarding the Z probing:

> These are two distinct systems.
>     1. Laser probe on head.
>     2. Contact probe on nozzle with the ability to stop motion in Z when contact is made.

I disagree. The Z Probe actuator does not dictate a laser. It is just intended as a lightweight interface to capturing a Z coordinate. A contact probe nozzle fits into this concept as well.

In the comment thread you linked I discussed expansion of the Z Probe actuator into it's own first class object, and I still think that needs to happen. A Nozzle, then, could have a Z Probe assigned, and in the case of a ContactProbingNozzle, perhaps assigned to itself. The point being that Z probing is a well defined feature and API on its own, rather than as part of a specific type of nozzle.

Thanks,
Jason

hydra:
> There are so many changes across so many systems I don't see how I can reasonably review it aside from just saying "Oh well, I give up, merge it." or "No".
> 
> At the very least, I think this should be broken up.

I concur, i would start by pulling out some code changes from this PR into several smaller PRs and periodically rebasing this one after each of the smaller PRs are merged other wise the review effort is very large indeed.  

There's some really cool stuff in here that i would here to see go to waste.

Start with the bug fixes and smaller less contentious changes so the ball starts rolling.

Just my two cents.

markmaker:
Hi @vonnieda 

Ok, I knew the PR looks ghastly from the amount of changes. I tried to explain in the Justification section, but I guess this isn't enough. I'm in no hurry, so take a quiet moment to read this, even if it means waiting a week. 

## Out of the way stuff...

With some effort (that I'm guilty of sneakily having hoped to avoid), I can break out _some_ things:

- Nozzle Tip Changer vision (entangled with _half_ the nozzle tip template cloning stuff)
- Visual homing wizard changes including fiducial locator vision Test button (entangled with _half_ of the LocationButtonsPanel changes, i.e. button options: yes, contact probing: no).
- Bugfix of #1143 et al (not mine btw.).
- Improved error messages and various other minor changes.

## Now for the other stuff...

I have a feeling that my explanations must not have been good enough to build a common ground of understanding of the technical challenge. Equally, I must say that I seem to continue misunderstanding your [concept here ](https://github.com/openpnp/openpnp/pull/855#issuecomment-501282321), i.e. the bridge from concept to the real world implementation. 

So the following is my attempt to sort this out. 

Note, when I ask _"Do you agree?"_ at the end of each section then I mean _with the cold analysis_, not the valuation or conclusions (yet). I don't want to sound nagging, but I would like to be on the same page analysis-wise, section by section. As these assumptions build on top of each other, it is valuable to know if and where we start to disagree.

## What we want to do / Use cases

1. Sense/probe the surface of feeder/board/whatever, today exclusively for setup but potentially also in production (e.g. board "Z leveling"/fiducial Z sensing). 
2. Sense part heights, as these seldom come imported from the E-CAD. Part heights are used... 
  (a) to compute the pick location in loose part feeders,
  (b) to compute the nozzle location for alignment, to have the underside of the part in the focal plane,
  (c) to compute the placement location,
  (d) to compute optimized dynamic Safe Z.
3. Sense the point of nozzle tips for Z calibration. Can also be thought of as "precision nozzle Z homing", including to equalize (shared) Z axes on multi-nozzle machines. 
4. Building "chains of references" between these. 

Do you agree?

## Feeders and Part Heights

Most feeders expose parts flush to their surface Z, i.e. the height of the part is obscured underneath that surface. Examples are plastic tapes suspended on the sides and the part hanging underneath in an embossed pocket. Or a paper tape being pushed up with a spring-loaded bed in an auto feeder. Or the typical grid containers, again having the parts aligned flush with the surface, i.e. having custom-depth pockets for each type/height of IC. 

Therefore, probing these feeders _does_ help you set it up, but it _does not_ give you the part height. 

The exception is the loose part feeder, where the stored feeder location Z is the table surface, i.e. the _underside_ of (mixed) parts. If you want to probe part heights there, you need the addition of the `Feeder.isPartHeightAbovePickLocation()` method or an equivalent distinction and/or method to get the underlying Z. This could perhaps be extended to the strip feeder if we use table Z and have known and consistent enough bottom-of-pocket and double-sided tape thicknesses. 

Do you agree?

## Physical means for height measuring

In order to address these use cases, there are four distinct categories of measuring means that I can think of: 

1. Sensing displacement from a fixed/known Z above: the head mounted laser displacement sensor is one example. Could probably be a stereoscopic algorithm with the down-looking camera too. 
2. Sensing displacement from a fixed/known Z below: the auto-focus method is one example. Could be a laser pointing up. 
3. Sensing contact from above. It is hardly sensible to make a dedicated physical probe by itself. Therefore, it is typically done with the nozzle. This has the added benefit, that probing can be done with the part on the nozzle i.e. the part can be sandwiched between the probing nozzle tip Z and a fixed/known surface Z below. This gives it the unique advantage to work on-the-job both in picks and placements. 
4. Sensing contact from a fixed/known surface below: the touch surface in [this video](https://youtu.be/3kkf6bktIHA) is an example. 

Do you agree?

## Job Processor 

Let's assume (for the moment) that we want to cover the whole range of these means. From the job processor perspective, the part height determination can happen at various points in time: 

1. For loose part feeders: **before the pick** with the head-mounted displacement sensors (missing in this PR, but can be added later).
2. Or else, again for loose part feeders: **during the pick** with the contact probe nozzle. 
3. Or else: **after the pick** with the touch surface (missing in this PR, but can easily be added later).
4. Or else, for parts with alignment: **during the alignment** with the auto-focus.
5. Or else, for parts without alignment: **during the placement** with the contact probe nozzle.
6. Or else: **not at all**. The job must not be allowed to start in the first place. User must add the part height manually.

Do you agree? (again assuming, **for the moment**, that we want to cover the whole range of these means)

## Nastiness Factor

This means that...

1. Height sensing is **all over the place**, in the machine.
2. Height sensing is **all over the place**, in the job process. 
3. For a comprehensive and flexible solution, the PR _necessarily_ looks and feels equally **all over the place**. 

And I haven't even talked about the necessary preparatory steps, like nozzle and nozzle tip Z calibration. 

Do you agree? (again assuming, **for the moment**, that we want to cover the whole range of these means)

## Conclusions

Finally, as to the assumption **for the moment**, that we want to cover the whole range of these means, this **moment** is now over.

These are the possible conclusions, that I see, out of the above:

1. **Forget part height sensing**: I agree it is only crucial in DIY/prototyping scenarios, when you may place not more than a handful of one particular part. It isn't necessary in large volume production, as manually entering a part height _once_ is a small matter if you produce a gazillion boards and reuse the parts in a gazillion jobs. 
2. **Simplify it**: Restrict part height sensing to "after pick/before alignment" only, i.e. make it a new processor step in between the two. Support auto-focus and touch surface only. Incurs a small detour for parts without alignment but who cares. Forget the laser, the contact probe nozzle, stereoscopic algorithms etc. Forget picking from loose part feeders with unknown part heights. Still needs the nozzle/nozzle tip Z calibration, but that can be done by the same means i.e. auto-focus or touch surface. Still not a simple PR, but much more orderly. 
3.  **Go for it**: that was the intention of this PR. I don't see how I can break it apart, without tripling the workload, the artificially incremental community testing, documentation and user learning marathon (and the bugs). 

I hope this helps as a basis for further discussion. 

Cheers, 
_Mark

P.S.  I had hoped to have a little credit from previous PRs as to how high the reviewing "scrutiny factor" must be. If this is a mistaken impression, then please tell me openly or privately. Appreciate.

markmaker:
Slightly edited my last comment to reflect the insight that the [machine in the video](https://youtu.be/3kkf6bktIHA) actually has a contact probe nozzle (you see it probing the feeders too). Please read the [online-version](https://github.com/openpnp/openpnp/pull/1164#issuecomment-814137785).

_Mark

markmaker:
Closing this, as being superseded by #1169, #1170, #1171, #1172, #1173 and #1174. 

Note, the [discussion here](https://github.com/openpnp/openpnp/pull/1164#issuecomment-814137785) is still on.going.

PR #1166, opened 2021-04-06
fix graying feeders line in feeders panel does not work when table is…

JulienDevillers:
Trivial, look at the code.


vonnieda:
Thanks @JulienDevillers!

PR #1169, opened 2021-04-11
Fix/Round-up of Bug-Fixes and some Missing Utility Features

markmaker:
# Description
This is a round-up of bug-fixes and conceptually missing utility features. These have popped up while working on #1164 (which will likely be re-submitted as a new one later). 

## Bug-fixes
* Bug-fix for #1143 (null pointer exception when part is missing in feeder)
* Parts table does not update on part property changes.
* Inconsistent FIRMWARE detection in `GcodeDriverSolutions`.
* Edit Pipeline does not wait for machine in `ReferencePushPullFeederConfigurationWizard`, resulting in the lighting actuator failing.
* Test Alignment in `ReferenceBottomVisionPartConfigurationWizard` did not conserve rotation of selected nozzle.
* Nozzle tip calibration failed the homing cycle if enabled on homing, creating a chicken egg situation during machine setup. Added an option to switch off. Improved error messages. This required  `UiUtils.messageBoxOnExceptionLater()` for deferred GUI tasks (in this case deferred throws/error messages).
* `null` check missing in `AbstractHeadMountable.toHeadLocation()`, refactored/extracted `AbstractHeadMountable.applyConfiguration()` for future use.
* Pipeline stages that throw exceptions are logged for easier diagnostics. The underlying issue that some exceptions should fail the pipeline immediately is still open.
* In the Pipeline Editor: custom `Object` type property sheet handlers need resetting when switching stages.

## Missing conceptual/utility features
* Missing `NozzleTip.home()` for Machine Setup tree topological homing.
* Missing `Length.compareTo()` 
* Missing `Location.getXyzLengthTo()` i.e. length including Z coordinate.
* Added `Location.origin` constant as the (0, 0, 0, 0) coordinate, for easy use in the form `vector.getXyzLengthTo(Location.origin)` to measure vector length.
* Missing `Location.multiply()` similar to the other operations.
* Missing `Head.permutateNozzle()` same as for cameras, Also fixed parameter name in `permutateCamera()`
* Missing `ImageUtils.clone()` to deep copy a `BufferedImage`.

# Justification
This is the first break-out of #1164 trying to implement [this input by @vonnieda](https://github.com/openpnp/openpnp/pull/1164#issuecomment-813675599).

I found it acceptable to put these in one PR as these are almost entirely **one class = one fix** and small (counting class+Wizards as one). 

# Instructions for Use
The only visible change is the new option **Fail Homing?** on the nozzle tip calibration.

Fail Homing

# Implementation Details
[...]

PR #1170, opened 2021-04-11
Feature/Nozzle Tip Maximum Part Dimensions

markmaker:
# Description
Based on #1169 i.e. includes all the changes in addition. Look at https://github.com/openpnp/openpnp/pull/1170/commits/090a5e2f9ea7b5f837c67f7a4c802dbe4631e20f to just see the changes in this PR. 

- Adds maximum part height/part diameter configuration on the `NozzleTip`. Used for Dynamic Safe Z (and in upcoming PRs).
- The `Part.isPartHeightUnknown()` method is introduced so a future implementation may add a formal boolean instead of just testing partHeight <= 0.0.
- `Nozzle.getSafePartHeight()` is added to allow for consistent nozzle positioning when part height is unknown (Safe Z, probing start height etc.)

# Justification
This is the second break-out of #1164 trying to implement [this input by @vonnieda](https://github.com/openpnp/openpnp/pull/1164#issuecomment-813675599).

This is not strictly self-contained i.e. **Max. Part Diameter** will only be used in upcoming PRs.

# Instructions for Use

Part Dimentions

**Max. Part Diameter** determines the diameter (or diagonal) of the largest part you ever expect to pick with this nozzle tip. Add a few millimeters on top of data sheet dimension for tolerances. This property will be used in future up-looking camera operations to restrict the region of interest (Auto-Focus, MaskCircle etc.). 

**Max. Part Height** determines the height of the tallest part you ever expect to pick with this nozzle tip. Add a few tenths of a millimeter on top of data sheet dimension for tolerances. This property will be used for dynamic Safe Z, as long as the part height is unknown. In future PRs this will determine the Z probing start offset. 

Note **Dynamic Safe Z** is enabled on the nozzle:

Dynamic Safe Z

# Implementation Details
1. Tested **Dynamic Safe Z** in simulation. 
3. Changes in the `org.openpnp.spi` are: `Part.isPartHeightUnknown()` and `Nozzle.getSafePartHeight()` as described.


markmaker:
Look at https://github.com/openpnp/openpnp/pull/1170/commits/090a5e2f9ea7b5f837c67f7a4c802dbe4631e20f to just see the changes in this PR.

vonnieda:
Hi @markmaker - I think these features should be reconsidered, but I'm open to discussion:

- For Max Part Diameter: Couldn't this be derived from the (currently incorrectly named) footprint data?
- For Max Part Height: I think we should just assume that a Part has a valid height or we won't pick it. Nozzle.pick() could even be modified to throw an Exception. We already enforce this running jobs so I don't see any reason to not enforce it globally. I'd rather do this than add another special case that has to be checked everywhere.

I think both of these are hard to guess when setting up the machine, and are hard to remember to update when circumstances change. I think it's fine to say that if the user wants the benefit of dynamic safe-z they'll need to set up their data correctly.

Jason

markmaker:
Hi @vonnieda 

Before answering here, I read all your other comments, so I will answer this with the whole picture in mind and also assuming you'll be on the same page in this respect. 

> * For Max Part Diameter: Couldn't this be derived from the (currently incorrectly named) footprint data?

> * For Max Part Height: I think we should just assume that a Part has a valid height or we won't pick it. Nozzle.pick() could even be modified to throw an Exception. We already enforce this running jobs so I don't see any reason to not enforce it globally. I'd rather do this than add another special case that has to be checked everywhere.

Ok, I should have been more outspoken about the middle/long term intentions. It is all connected, and it is very difficult to justify with just the broken out PR, or maybe even the bunch of PRs now lined up.  

My middle/long term intention is to help make OpenPnP extremely automatic towards repetitive tasks. **Import from E-CAD. Place. Done.** I'm convinced we can make it smart enough to do that for > 95% of the parts. All Vision should work out of the box, with either no manual tuning required, _or_ automatic interactive tuning supported. 

Any additional manual data setup after E-CAD-import is too much work, and too error-prone, IMHO. That does include setting part heights, and most certainly footprints. From a "saved-work-per-placed-part" stand-point this is most important for DIY/prototyping, but I don't think large volume shops will be mad if their job is made a little easier too. 

One might be able to import footprints from the E-CAD, but then this _has_ to work with data available there, i.e. with solder pads, despite them not being ideal for vision (my experiments with blurred template matching are promising, though). But for E-CAD and/or import-interfaces that don't supply footprints, we still need another solution. 

We discussed this in the group: I intend to make a bottom vision thingy that can auto-learn footprints (it might be a different alignment class or just a new pipeline stage, I'm not yet sure). On the first go per part/package, the stage tries to detect the part using classic vision operations (mostly like today) and then presents the result to the user in some interactive way. The user can adjust the alignment if needed, etc. and then "teach-in" the part/package as a calibrated template image. The next time the part/package is encountered, the template image is used to align the part. This can also determine pin one orientation and detect bent pins etc. by match score (the threshold of which can also be auto-learned, as I've [done with the **Precision Confidence Limit** here](https://github.com/openpnp/openpnp/wiki/ReferencePushPullFeeder#vision--calibration)). 

With that in mind, I'll go back to the two fields we're discussing: 

The **Max Part Height** is needed for Safe Z when the part height is not yet known. Sensing the part height (almost) always includes transporting the part to the sensor/touch plate/camera. It is also needed for the Auto-Focus method to determine (and optimize) the search depth. 

The **Max Part Diameter** is currently used for Auto-Focus, which needs some rough input for the size of the feature, otherwise it will for instance focus on the air bore instead of the rim of a cup nozzle. Internally this is used in a heuristic to control the significant fractile of the sharpest pixel contrasts. The diameter is also intended for later use with the alignment auto-learn feature.  Judging from experience with #1171 and #1179, I expect to need at least _some_ rough input as to the relevant area for adequate preliminary detection, i.e. a 0402 is so much smaller than a 144 pin IC monster, I don't think I can do without. 

> I think both of these are hard to guess when setting up the machine, and are hard to remember to update when circumstances change. 

That's true. On the other hand, these are easy to understand and constrained physical properties. The **Max Part Height** can only ever be half the Safe Z when not dynamic, or the Safe Z to Z+ soft limit when dynamic with single nozzle, or (half) the Safe Z Zone when dynamic with dual-nozzle. So if in doubt just set it to that value. As for the **Max Part Diameter**, the idea is for them to input into automatisms that in their absence would otherwise require a user to set mask diameters etc. in very obscure pipeline stages. IMHO much harder to guess and update when circumstances change. 

Having said that, I will keep this in mind. For instance the interactive alignment template image learning dialog could also expose the **Max Part Diameter** and let the user adjust it on-the-go as needed. 

> I think it's fine to say that if the user wants the benefit of dynamic safe-z they'll need to set up their data correctly. 

It would be a pity, not to support the best settings in combination. ;-) Dynamic Safe Z can effectively almost halve your Safe Z travel for the typical majority of small parts and this brings a lot of speed, at least to DIY machines. 

_Mark

PR #1173, opened 2021-04-12
Feature/Visual Homing Consolidation

markmaker:
# Description

Based on #1169, #1170, #1171 and #1172. Look at https://github.com/openpnp/openpnp/commit/79a3ece22174e3bb939dd3f1f03a4216f187949d to see just the changes of this PR.

* Visual Homing can now be tested and performed as such in the head wizard. 
* `LocationButtonsPanel` has been extended to support tool or camera buttons only.
* `LocationButtonsPanel.setEnabled()` has been implemented to enable/disable all buttons. 
* Therefore, the head wizard can now use `LocationButtonsPanel` for the **Homing Fiducial** and the **Park Location** instead of duplicating buttons and code. 

# Justification

This is the fifth break-out of #1164 trying to implement [this input by @vonnieda](https://github.com/openpnp/openpnp/pull/1164#issuecomment-813675599).

## Visual Homing Test
Almost all vision setup wizards now have Test buttons. Visual homing was sorely missing it, as I quickly noticed when setting up a new homing fiducial for my machine. 

## LocationButtonsPanel
The rationale is to unify location capturing and positioning controls, so future functionality added or changed in the `LocationButtonsPanel` will automatically be established throughout.  

The incentive was created while developing #1164. Originally, a single touch location was defined on the head. Therefore, I wanted to unify usage of `LocationButtonsPanel` on that wizard. Later the touch locations were actually moved to the nozzle tip tool changer wizard, for fast calibration while changing a tool. But I think the rationale is still valid.  

# Instructions for Use

Head Wizard

**Visual Test** performs the homing fiducial vision only. It does not affect the machine coordinate system. 

**Visual Home** performs the visual homing per se. It will recalibrate the machine coordinate system. During setup this could for instance be used to quickly recalibrate the machine after having lost steps on a X/Y stepper, e.g. due to acceleration limit testing. Visual homing is usually much quicker than a full homing cycle. 

# Implementation Details
1. The feature was fully tested on the machine in #1164. As a breakout it was tested on a simulation machine, with GcodeServer driver. 
2. Did follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style).
3. No changes in the `org.openpnp.spi` or `org.openpnp.model` packages.
4. Successful `mvn test` before submitting the Pull Request.

PR #1172, opened 2021-04-11
Feature/Nozzle Tip Changer Vision

markmaker:
# Description
Based on #1169, #1170 and #1171. Look at https://github.com/openpnp/openpnp/commit/f26a18e3f028d0b485b5e4abda2ae90e251d3084 to see just the changes of this PR.

## Vision X, Y Calibration

This PR adds computer vision based calibration of nozzle changer slot locations (in X, Y). Vision is based on two template images, allowing to detect nozzle changer slot empty/occupied status and preventing collisions.

## Template Images

This PR introduces `TemplateImage` and `TemplateImageControl` to manage template images linked to the `Configuration` efficiently. 

`TemplateImageControl` represents the template image as a thumbnail control in the Wizard. Hovering over the control, the full size template is shown in the associated CameraView. The control allows for trivial property binding, directly to the `TemplateImage`.

Image data is hashed to generate unique image references. Duplicated template image files are only saved/loaded once if shared between multiple instances (e.g. the template images of multiple identical nozzle tip changer slots). Even if multiple configurations i.e. `machine.xml`s are used back and forth, the correct template images will always be loaded, as they are referenced by their unique image data hashes. Unless images are manually removed from the `.openpnp2/org.openpnp.vision.TemplateImage` directory, no image is ever lost or overwritten. 

# Justification

This is the fourth break-out of #1164 trying to implement [this input by @vonnieda](https://github.com/openpnp/openpnp/pull/1164#issuecomment-813675599).

As a caveat of this breakout, the PR does not yet introduce cloning the vision settings between nozzle tips. This will be reintroduced, when X, Y calibration is complemented by Z calibration. 

# Instructions for Use

## Vision Calibration

Using computer vision, the Nozzle Tip Changer locations can be calibrated in X and Y. The process uses two template images, of the empty and occupied changer slot respectively. It can therefore also detect if the slot is empty/occupied as expected.

**Vision Location** determines the changer location at which the template images are taken. Chose a location where the nozzle tip is visible when it occupies the slot. 
 
Vision Location

**Calibration Trigger** determines when the vision calibration takes places:

Calibration Trigger

* **Manual** allows only manual calibration. The calibration will be stored in the configuration (machine.xml)
* **Machine Home** invalidates the calibration when the machine is homed. As soon as nozzle tip load/unload is requested, the slot will first be vision calibrated. 
* **NozzleTipChange** recalibrates the the locations on each nozzle tip change. 

Vision Calibration

[...]

PR #1171, opened 2021-04-11
Feature/Auto-Focus

markmaker:
# Description
Based on #1169 and #1170. Look at  https://github.com/openpnp/openpnp/commit/b9e4eedf40bb4a2ae7fa4a4280021054b71b7080 to see just the changes of this PR.

This adds Auto-Focus to OpenPnP. As a standalone feature, this can be used to precisely set the up-looking camera's location Z. In upcoming PRs it can be used to auto-learn part heights.

## Details

* Added new `spi` `FocusProvider` interface.
* Added `AutoFocusProvider` implementation.
* Added `Camera.getFocusProvider()` implementation in the `ReferenceCamera`
* Added method to `CameraConfigurationWizard`.
* Simulate depth of focus blurring in the `SimulatedUpCamera` for testing. Optional (it is slow!)
* Added option in the `SimulatedUpCameraConfigurationWizard`.

# Justification
This is the third break-out of #1164 trying to implement [this input by @vonnieda](https://github.com/openpnp/openpnp/pull/1164#issuecomment-813675599).

# Instructions for Use

## Up-looking Camera Auto-Focus

Unknown part heights can automatically be detected by auto focussing its underside. The nozzle will servo up and down to find the optimum. 

As a pre-requisite, the **Part Dimensions** must be defined on each nozzle tip (Configuration tab). Both the **Max. Part Diameter** and **Max. Part Height** are used to determine the autofocus range and sensitive area. 

Part Dimensions

To enable the auto focus on the camera, select the **Focus Sensing Method**. It is only available on up-looking cameras: 

Enable Auto-Focus

Once you press **Apply**, a new **Auto Focus** tab appears:

Auto Focus tab

**Focal Resolution** determines the maximum step size for the Auto Focus servo. 

**Averages Frames** lets you run the autofocus for multiple frames and average the results. This will improve the accuracy by effectively averaging out image noise etc. 

**Focus Speed** sets the relative motion speed for the focus stepping. 
 
**Show Diagnostics?** switches on auto focus edge highlighting and status info in the camera view.  

**Last Focus Distance** indicates the last live or test focus distance measured. If there is a part on the nozzle, this is the part height. 

Press the red dot button to test the auto focus.  

Auto Focus

### Calibrating the Up-looking Camera Z

The auto focus can help you calibrate the precise up-looking camera Z position. This will make sure your bottom vision happens at optimum distance. 

___
[...]

PR #1174, opened 2021-04-12
Feature/Z Calibration, Probing, Part Height Auto-Learning

markmaker:
# Description

Based on #1169, #1170, #1171, #1172 and #1173. Look at https://github.com/openpnp/openpnp/commit/c3bb012e92e31bd8dfdc23ce77e175e1b86ecee2 to see just the changes of this PR. This is the residue and **re-instantiation of #1164.**

Consolidates Z calibration, contact probing, height sensing, part height auto-learning. Also complements nozzle tip changer slot cloning of vision X, Y and probing Z calibration settings. 

## Open Questions

Please refer to https://github.com/openpnp/openpnp/pull/1164#issuecomment-814137785 for the issues that now need discussing. 

## Main Features

* Manual and automatic nozzle/nozzle tip/slot Z calibration for precision multi-nozzle leveling and nozzle tip length calibration.
* ContactProbeNozzle provides on-the-fly part height probing/learning, feeder Z calibration, placement height calibration. 

Nozzle Tip Changer Calibration, Height Sensing

## Feature Details

* Nozzle tip specific touch location for faster after-loading Z calibration and varied tip lengths. Configurable triggers/options. Touch locations can be "Z-leveled" to single reference. 
* Adds _experimental_ "sniffle" method for contact probing _without_ dedicated sensor. May be useful for nozzle/nozzle tip Z calibration but obviously not feasible with part on the nozzle, i.e. pick and place height sensing is not possible.
* Job processor allows running jobs with part heights not yet set, it will only complain when no part height sensing method is available for parts without height. 
* Integration of Auto-Focus (#1171) into part height auto-learning.
* Nozzle contact probing capability becomes compatibility criteria in the planner. Multi-nozzle machines might only have one nozzle with probing sensor, so parts with unknown height will have to be loaded on that nozzle. Only relevant if Auto-Focus does not apply. 
* The placementLocation in the job processor is now the base location (i.e. underneath the part), in order to allow part height sensing on the fly (backwards-compatible Scripting, i.e. parameter works as before, new parameter placementLocationBase). 
* Feeders have new isPartHeightAbovePickLocation() method to determine if part height is above the pick location and must be sensed when part height is unknown (loose part feeder classes).
* As predicted long ago, the Nozzle now has overridable moveToPickLocation() and moveToPlacementLocation() methods. Plain moveToLocationAtSafeZ() calls in job processor and GUI actions were replaced.
* ContactProbeNozzle now overrides these methods to provide on-the-fly probing. Can be done each time or only once and learned (faster Z moves without probing).
* Z-Calibrations triggered by machine homing has option to enable/disable failing the homing cycle.
* Issues&Solutions for ContactProbing was improved/adapted.
* Issues&Solutions can convert ReferenceNozzles into ContactProbeNozzles while preserving configuration.

## Breaking Changes

* ContactProbeNozzle retraction (Actuator `false` actuation) now happens immediately after probing. Previously it was `false`-actuated after picking/placing. Default Gcode for `{False}` is now empty, retraction is controlled by GUI and done by normal Z move.
* moveToLocationAtSafeZ() will only move to safe Z if the location changes in X, Y or C. This allows for quick Z only relocation. Happens when bottom vision follows auto-focus or when pick follows BlindsFeeder push-cover opening etc.

# Justification

This is the sixth and final break-out of #1164 trying to implement [this input by @vonnieda](https://github.com/openpnp/openpnp/pull/1164#issuecomment-813675599).

# Instructions for Use

In preparation for the Wiki. I'm partially using Wiki markup links here, they won't work until in the Wiki. This the consolidated doc of #1169, #1170, #1171, #1172, #1173 and this one.

### What is it?
The ContactProbeNozzle can probe the surface of things in Z. This can simply be used to perform probing picks and places all the time, i.e. Z coordinates do not need to be very precise, the nozzle will automatically stop when contact is sensed. 

But this does not cover other applications where Z coordinates must be quite precise without making contact. Examples are sideways nozzle tip tool changing, special feeder movements like the [[BlindsFeeder]] cover actuation or even part picking where spring-loaded/hard contact is not desired ([suspended tape feeders](https://docs.mgrl.de/maschine:pickandplace:feeder:manualfeeders)). 

Machines that home the Z axis with easily shifted Zmax switches ([e.g.](https://liteplacer.com/the-machine/assembly-instructions/pnp-head-step-11-attach-z-high-limit-switch/) Liteplacer), hard to balance dual-nozzle mid-axis sensors or even by just letting springs retract unpowered Z motors, may not provide a precise and repeatable Z reference eternally.  

A ContactProbeNozzle can then be used to calibrate the Nozzle in Z, more specifically the point of the loaded Nozzle Tip. By probing a known reference Z touch surface, it can then apply the obtained calibration offset to all future Z movements. 

For multi-nozzle machines, the nozzles can therefore be perfectly harmonized in Z. This can (optionally) be extended to individual nozzle tips if these have varying heights, or if the coupling to the nozzle is not very consistent. 

Having a calibrated Z not only establishes precision in movements, but also in capturing or probing other locations. 

### Enable the ContactProbeNozzle 
If you don't have a ContactProbeNozzle yet, let [[Issues and Solutions]] replace the existing ReferenceNozzle for you:
 
Replace nozzle

This extends the Nozzle with a new tab **Contact Probe**.

### Method Selection

Contact Probe Method

**Method** lets you select the probing method. 

- **None** switches contact probing off. The nozzle behaves as if it were a regular ReferenceNozzle. 
- **Vacuum Sense** uses the Vacuum Sensing setup of the nozzle to perform "sniffle probing". More specifically, the [Part Off](https://github.com/openpnp/openpnp/wiki/Setup-and-Calibration%3A-Vacuum-Sensing#measurement-method) detection method is used to sense if the nozzle tip has touched the surface. 
- **ContactSensingActuator** uses a separate Actuator to probe and sense for contact. This is a setup made popular by the Liteplacer kit. 

Notes: The Wizard will show different fields according to the selection. If at least one nozzle has an active **Method** selected, other parts of the OpenPnP GUI will also show additional elements for contact probing. 

### Vacuum Sense Method

Vacuum Sense

**Start Offset** determines how high above the nominal probing location, the sensing should start. For the **Vacuum Sense** method, this must be high enough to get a clear "Part Off" result on the first sniffle.

**Probe Depth** determines how far the probing should go. 

**Sniffle Increment** determines the step in Z. 

**Sniffle Dwell Time [ms]** sets the waiting time between two sniffles. This is needed to clear out any under-pressure from the previous sniffle.

**Final Adjustment** after contact was sensed, add this adjustment to the Z position. Positive values retract the nozzle, negative values add spring loading. You should aim for a light touch with barely any spring loading.  This way the tip Z reference can also be used to push things sideways such as with the [[BlindsFeeder]] cover actuation. 

**Calibration Z Offset** shows the last calibration result. 

**Calibrate Now** performs the Z calibration. You must first setup a **Touch Location** on all the noozle tip's **Tool Changer** tab. 

### Contact Sense Method

ContactSenseActuator

**Start Offset** determines how high above the nominal probing location, the sensing should start. For the **Vacuum Sense** method, this must be high enough to get a clear "Part Off" result on the first sniffle.

**Probe Depth** determines how far the probing should go. In fact it may go further but if the probing result is beyond the given depth an exception is thrown. 

**Final Adjustment** after contact was sensed, add this adjustment to the Z position. Positive values retract the nozzle, negative values add spring loading. You should aim for a light touch with barely any spring loading. This way the tip Z reference can also be used to push things sideways such as with the [[BlindsFeeder]] cover actuation. 

**Feeder Height Probing** determines when feeders should be probed for pick location Z offsets. 

**Part Height Probing** determines when parts should be probed for placement location Z offsets. 

Probing Triggers

* **Off** switches off any probing.
* **Once** probes for a Z offset only once. The offset is stored in the configuration (machine.xml). 
* **AfterHoming** probes for a Z offset once after homing the machine.  
* **Each Time** probes for contact in all picks and placements. This is the most tolerant setting.  

**Calibration Z Offset** shows the last Z calibration result for this nozzle with the current nozzle tip attached. See also the Nozzle Tip Probing Configuration.

**Calibrate Now** performs the Z calibration. You must first setup the Nozzle Tip Probing Configuration.

### Part Height Auto-Learning

With a ContactProbeNozzle and enabled feeder and part height probing, OpenPnP can now also learn part heights automatically. Whenever a part with unknown height is picked or placed for the first time, the probed height will be stored on the part. 

* Part height auto-learning works for feeders that have the part height _above_ the pick location Z (ReferenceLoosePartFeeder and AdvancedLoosePartFeeeder). 
* Part height auto-learning also works in all placements, as obviously all parts are above the known board location Z. 

The job processor knows when a ContactProbeNozzle has this capability and will in this case allow starting a Job with unknown part heights. If only one nozzle in a multi-nozzle setup has probing capabilities, the planner will automatically restrict parts with unknown heights to this nozzle. As an alternative, part heights can also be auto-learned using auto focus, if bottom vision is enabled for the part. 

Part height unknown

### Part Dimensions

The first pre-requisite for probing are the maximum **Part Dimensions**, defined in the **Configuration** tab:

Part Dimensions

**Max. Part Diameter** determines the diameter (or diagonal) of the largest part you ever expect to pick with this nozzle tip. Add a few millimeters on top of data sheet dimension for tolerances. This property will be used in present and future up-looking camera operations to restrict the region of interest (Auto-Focus, MaskCircle etc.). 

**Max. Part Height** determines the height of the tallest part you ever expect to pick with this nozzle tip. Add a few tenths of a millimeter on top of data sheet dimension for tolerances. This property will be used to start probing above a part with unknown height. Furthermore, it will be used for dynamic Safe Z, again as long as a part height is unknown. 

### Tool Changer 

As soon as at least one ContactProbeNozzle has an active probing method, the Tool Changer tab will show extra elements:

Tool Changer

**Touch Location** lets you define a special Z probing touch location. It will be the reference for all other Z coordinates on your machine, therefore it should represent a fundamental and "eternal" reference surface for your machine.  At the same time it should be very close to the changer locations so the extra moves for Z calibration do not take much time.  

Nozzle Tip Change and Z Calibrate

After capturing the X, Y coordinates with the camera, you must initially probe the reference Z coordinate with the middle button:

Probe 

___
**CAUTION**: do not do this again later. You will lose your "eternal" Z reference! 
___

**Auto Z Calibration** determines when the Z calibration happens. 

Auto Z Calibration

* **Manual** switches automatic Z calibration off. You can manually calibrate and the obtained Z offset will be stored in the configuration (machine.xml).
* **MachineHome** performs automatic Z calibration if this nozzle tip is loaded when the machine is homed. When other nozzles are later loaded to the same nozzle, no additional Z calibration is triggered. **Z calibration is per nozzle**. This can handle uneven nozzles in multi-nozzle machines. 
* **NozzleTipChange** performs automatic Z calibration whenever this nozzle tip is loaded to a nozzle. In addition, Z calibration also happens if this nozzle tip is loaded when the machine is homed. **Z calibration is per nozzle tip**.  This can handle uneven nozzle tips with different heights or inconsistent coupling. 

**Fail Homing?** If enabled, aborts the homing cycle when Z Calibration fails as part of it. If disabled, continues the homing cycle and only displays an error message. 

**Z** (behind the **Auto Z Calibration**) indicates the last obtained Z calibration offset. 

**Reset** removes the current Z calibration offset. 

**Calibrate now** performs manual Z calibration. 

**Note:** the Z calibration affects all the nozzle Z motion as well as the capturing of coordinates from the current nozzle position. But it does not affect Safe Z, as Safe Z is relative to the raw axis coordinate system, not the calibrated nozzle Z coordinate system. To balance a multi-nozzle machine's nozzles, use the [Axis capture buttons](https://github.com/openpnp/openpnp/wiki/Machine-Axes#kinematic-settings--axis-limits) instead.  

### Cloning Settings

Cloning Settings

Among the other settings, the **Z Calibration** settings can be cloned to/from other nozzle tips, so you need to configure them only once. The **Touch Location** is automatically translated according to the **First Location** (along with the other locations). 

Marking this nozzle tip as the **Template** makes the **Touch Location Z** the global machine reference Z. 

Use the **Clone Tool Changer Settings to all Nozzle Tips** button to distribute the Z calibration settings to all the other nozzles. 

Use the  **Callibrate all Touch Locations' Z to Template** to reference them to the global machine reference Z.

## Vision Calibration

Using computer vision Nozzle Tip Changer locations can be calibrated in X and Y. The process uses two template images, of the empty and occupied changer slot respectively. It can therefore also detect if the slot is empty/occupied as expected.

**Vision Location** determines the changer location at which the template images are taken. Chose a location where the nozzle tip is visible when it occupies the slot. 
 
Vision Location

**Calibration Trigger** determines when the vision calibration takes places:

Calibration Trigger

* **Manual** allows only manual calibration. The calibration will be stored in the configuration (machine.xml)
* **Machine Home** invalidates the calibration when the machine is homed. As soon as nozzle tip load/unload is requested, the slot will first be vision calibrated. 
* **NozzleTipChange** recalibrates the the locations on each nozzle tip change. 

Vision Calibration

**Template Width** and **Template Height** determine the dimensions of the template image. Use a crop that includes relevant horizontal and vertical edges of the changer slot surrounded by an extra millimeter or so. The nozzle tip must be visible, when it occupies the slot. You can use the **Capture** buttons to optimize.  

**Tolerance** determines by how far from the nominal location the vision detected location may be. 

**Wanted Precision** determines at what distance the obtained template image match location is close enough. Otherwise, the camera is centered to the detected location and another vision pass is made (drill down).

**Max. passes** limits the maximum number of vision passes. 

**Minimum Score** determines how similar the camera image must be in comparison with the template image (0.0 ... 1.0). If the score is not reached, the calibration fails. 

**Last Score** shows the template match score obtained from last live or test calibration. 

Use the **Test** button to test the calibration. 

Use the **Capture** button to capture the respective **Template Empty** or **Template Occupied** image. The camera will automatically move to the selected **Vision Location**. You can hover over the captured template thumbnail to see it enlarged in the camera view. 

Use the **Reset** button to remove templates images. 

The two images are both used and the better match tells OpenPnP if the slot in question is empty or occupied. If this does not correspond to the state internally stored by OpenPnP, it will abort the calibration and therefore the nozzle tip load/unload operation. Nasty collisions can be safely prevented. 

Found occupied

## Up-looking Camera Auto-Focus

Unknown part heights can automatically be detected by auto focussing its underside. The nozzle will servo up and down to find the optimum. 

As a pre-requisite, the **Part Dimensions** must be defined on each nozzle tip (Configuration tab). Both the **Max. Part Diameter** and **Max. Part Height** are used to determine the autofocus range and senstive area. 

Part Dimensions

To enable the auto focus on the camera, select the **Focus Sensing Method**. It is only available on up-looking cameras: 

Enable Auto-Focus

Once you press **Apply**, a new **Auto Focus** tab appears:

grafik

**Focal Resolution** determines the maximum step size for the Auto Focus servo. 

**Averages Frames** lets you run the autofocus for multiple frames and average the results. This will improve the accuracy by effectively averaging out image noise etc. 

**Focus Speed** sets the relative motion speed for the focus stepping. 
 
**Show Diagnostics?** switches on auto focus edge highlighting and status info in the camera view.  

**Last Focus Distance** indicates the last live or test focus distance measured. If there is a part on the nozzle, this is the part height. 

Press the red dot button to test the auto focus.  

Auto Focus

### Calibrating the Up-looking Camera Z

The auto focus can help you calibrate the precise up-looking camera Z position. This will make sure your bottom vision happens at optimum distance. 

___
**CAUTION**: this procedure will change the subject to camera distance, which will change the scale of subjects. You will need to readjust the **Units per Pixel** after performing it. 
___

1. Use a Z calibrated nozzle tip with no part on. 
2. Auto focus it, using the red dot button in the **Auto Focus** camera tab. . 
3. Press the **Adjust Camera Z** button to set the camera Z Position to the detected distance. 
4. Go to the **General Configuration** tab to adjust **Units per Pixel**.

### Part Height Auto-Learning

Whenever Alignment (a.k.a. Bottom Vision) encounters a part with unknown part height, it will employ the auto focus. The detected part height is then stored on the part. This will take some additional time (but hardly longer than manually measuring and entering the data). All subsequent vision and placement operations will reuse the height directly, with no further speed penalty incurred. 

The job processor knows when a camera has auto focus capability and bottom vision is enabled for the part. It will then allow starting a Job with unknown part heights.  As an alternative, part heights can also be auto-learned using a ContactProbeNozzle.  

Part height unknown

# Implementation Details

1. Code was first tested on the simulation machine, hence the simulation of depth of focus. Then it was tested and tuned on my machine (strictly speaking this was done with #1164, which should be identical except for some cosmetic changes). Since I was also mechanically changing my machine (tool changer relocated), this turned out to be quite instructive. 
3. Made changes in the `org.openpnp.spi` or `org.openpnp.model` packages, as described, i.e. see Feeder.isPartHeightAbovePickLocation(), Nozzle.moveToPickLocation(), Nozzle.moveToPlacementLocation().


vonnieda:
@markmaker Apologies for not getting back to this quickly. I'm not ignoring it. Recently got my COVID shot and have been mostly sleeping when I'm not working. Hopefully back in full operation next week :)

markmaker:
Speedy recovery!

vonnieda:
HI @markmaker - so here we are at last :)

First, going back to your [questions](https://github.com/openpnp/openpnp/pull/1164#issuecomment-814137785), I agree with all of the points and most emphatically with the "Nastiness Factor". I think we should combine 1 and 2 from your summary:

- I think part height sensing is a great feature to add.
- I think adding it so many places adds more code complexity than the feature is worth.
- I wonder if you have tested contact probing the placement when solder paste is on the board? Surely this affects the measured "height"? I wonder if this really only leaves the auto focus and touch plate methods as valid?
- With my comment on another PR that we should assume the part height is always correct I think it leaves one obvious solution to where to do it, and that is to leave it to the user:
  
  Just add a button to the Parts panel that lets the user perform the part height detect for one or more parts. If the user tries to start the job with invalid part heights they get an error and can perform the 2 or 3 clicks more to run the check and resolve it.

This will necessitate discarding the part in some cases, but we can use the recently added "recycling" functionality to put the part back in the feeder for feeders that support it.

> P.S. I had hoped to have a little credit from previous PRs as to how high the "reviewing scrutiny factor" must be. If this is a mistaken impression, then please tell me openly or privately. Appreciate.

Mark, you have an enormous amount of credit with me, and I have a great respect for your work. I know that I barely even need to glance at the code itself, because it will be of extremely high quality. But these are a bunch of very large features that have come out of the blue without any prior discussion and that add quite a bit of code complexity. I think it's important we discuss whether these features are worth the complexity added.

I wanted these split up because it was very hard to tell by looking at the code what feature went with what changes, and to judge how much complexity that feature added. After now reviewing all of them I see that the majority of the features are relatively self contained but the auto part height sensing is all over the place. It was very hard to see that as one huge PR.

Thanks,
Jason

markmaker:
Hi @vonnieda 

This answer builds on top of [this one](https://github.com/openpnp/openpnp/pull/1170#issuecomment-825017560) and most specifically my "dream" that we can **Import. Place. Done**. 

If I'm understanding you correctly, then it boils down to this point:

> I think adding it so many places adds more code complexity than the feature is worth.

I would humbly ask, _does it?_ 

If I look at the remainder diff here https://github.com/openpnp/openpnp/commit/c3bb012e92e31bd8dfdc23ce77e175e1b86ecee2 the changes/complexities do not seem so big to me. 

First, I am assuming that complexity added to the ContactProbeNozzle sub-class does not count. I'm glad you made me create a sub-class back then! Now nobody is forced to use it. It has a separate Wizard tab that does not "pollute" the GUI if you don't want it. But its added value is great, the [video](https://youtu.be/3kkf6bktIHA) you featured back in #98 shows exactly that. You do agree that this little machine in the video features a contact probing nozzle, right? I believe a probing nozzle should become a more common feature for OpenPnP machines.

Second, I'm willing to sacrifice `ReferenceLoosePartFeeder` and `AdvancedLoosePartFeeder` support. This will remove the need for `Feeder.isPartHeightAbovePickLocation()` and surrounding code changes (FeederPanel, JobProcessor, ReferenceNozzle, etc.). I would still - at the moment - argue to keep it in and later expand it to support strip feeders, but it's the first thing that can go, if it needs to.

Then If I go through the https://github.com/openpnp/openpnp/commit/c3bb012e92e31bd8dfdc23ce77e175e1b86ecee2 (mainly from top to bottom) I see the following (I suggest you open a parallel [window](https://github.com/openpnp/openpnp/commit/c3bb012e92e31bd8dfdc23ce77e175e1b86ecee2) and go through it as discussed below):

```
        Location pickLocation = feeder.getPickLocation();
        MovableUtils.moveToLocationAtSafeZ(nozzle, pickLocation);
```

was replaced everywhere with:

```
        nozzle.moveToPickLocation(feeder);
```

And the analog for `moveToPlacementLocation()`. 

In my book that's not adding complexity, just benign refactoring and encapsulation. And it enabled a Nozzle sub-class to override. 

Then there is some refactoring in the FeederPanel with `getCompatibleNozzleAndTip()` that removes some inconsistencies, i.e. the old code did not always use the nozzle with a feeder part compatible nozzle tip loaded, now it does. These changes are completely benign i.e. they would make sense even outside this PR. 

Still inside FeederPanel, the business with `preliminaryPickLocation()` could go, if we ditch part height sensing for loose part feeders. 

The `LocationsButtonsPanel` adds a new contact probing button, similar to your (laser) Z probe. Needed for Z calibration which is a must, even when this PR is restricted to use a touch plate.

Then it adds a new icon for that button.

In the ReferenceNozzle we have a little artifact to _only_ reset all the nozzle tip calibrations when X and Y in head offsets change (and not if only Z changes). That's a left-over from an earlier incarnation of how Z calibration worked. It still makes sense, semantically, but I could remove it, if it bothers you. 

Then still in ReferenceNozzle we have the default implementations of `Nozzle.moveToPickLocation()` and `Nozzle.moveToPlacementLocation()` (see above). If we ditch loose part feeder support, these become one-liners. 

Then still in ReferenceNozzle I unified the way a LocationOption is tested (`contains()` call instead of `for` loop) and removed a potential bug in runout-compensation, i.e. the `super` call was in the wrong place, topologically. These changes are completely benign i.e. they would make sense even outside this PR. 

Then I added four calls where Z calibration must be ensured during nozzle tip change. Again a must. 

Then in the NozzleTip I added Z calibration support. Again a must. 

Most other changes in NozzleTip are Template/Cloning support, a convenience function for the user to clone settings from one nozzle tip to the other, now including Z calibration and the #1172 settings and an option to lock a nozzle tip against cloning, if it is special. Some of these changes have been refactored from the Wizard to the NozzleTip according to MVC. 

In JobProcessor part heights > 0 are no longer a strict prerequisite. In its place, there is a small change in how compatible nozzle tips are determined. Instead of determining the cross-section between package-to-nozzle-tip and nozzle-to-nozzle-tip compatibilities right there in the JobProcessor, this is delegated to the nozzle. Again, in my book this is not adding complexity,  just benign refactoring and encapsulation. 

The reason is that a nozzle sub-class (e.g. the ContactProbeNozzle) can now override and include part height probing capability into the compatibility equation. There might be dual nozzle machines, where only one nozzle can probe, so the planner must assign parts with unknown heights to that nozzle. 

Still in JobProcessor (and are as already discussed above) there are the `Nozzle.moveToPickLocation()` and `Nozzle.moveToPlacementLocation()` refactorings. 

One diff in JobProcessor concerns the change of semantics of `getPlacementLocation()` to not yet include the part height as it is subject to probing on-the-go. The rest of the JobProcessor changes provide backward compatibility to scripts (I was perhaps overly "correct" there).

Frankly, looking at the overall changes in JobProcessor, I would argue that the refactoring has a very light touch, considering the huge benefits and considering the "all over the place" nastiness factor explained [here](https://github.com/openpnp/openpnp/pull/1164#issuecomment-814137785). 

Then there is the AdvancedLoosePartFeeder and ReferenceLoosePartFeeder part height sensing support. These changes are small, considering the benefit. But as said, they could go, if needed.

In ReferenceBottomVision we have auto-focus based part height sensing. These changes are small, considering the benefit and if I understand you correctly, you're not against this in principle.

In ReferenceNozzleTipToolChangerWizard we see the largest changes mostly due to how Window Builder regenerates everything, I have grouped all fields at the end, and I might then at one point also have reformated white-spaces as Window Builder sometimes goes crazy with indentation. Other than that, what I said about NozzleTip and Template/Cloning support also applies here. Then there is Z calibration which again cannot be avoided, even if we limit the scope of the PR to just the touch plate and manually triggered part height sensing. 

In the Feeder interface we see the added `Feeder.isPartHeightAbovePickLocation()`. This can go, if we ditch loose part feeder support. 

In the Nozzle interface we see the already discussed  `Nozzle.moveToPickLocation()` and `Nozzle.moveToPlacementLocation()` refactoring. 

Still in Nozzle, and the implementation in AbstractNozzle we have the `Nozzle.getCompatibleNozzleTips(Part part)` overload to encapsulate the already discussed cross-section between package-to-nozzle-tip and nozzle-to-nozzle-tip compatibilities that was formerly done in the JobProcessor planner. 

In AbstractPnpJobProcessor we have an additional `Nozzle.moveToPlacementLocation()` refactoring for discard. Again as discussed.

MovableUtils has the optimization to not move to Safe Z when X, Y, C are already at the target location. This is a useful in this PR to optimize auto-focus → alignment, probing → pick, probing → place moves, but also outside for similar moves such as cover actuation → pick in the BlindsFeeder. As stated in the PR description, this is open for discussion. 

That's it. 

All in all, I'm almost surprised myself how light it turned out after breaking out the other PRs. What do you think?

> * I wonder if you have tested contact probing the placement when solder paste is on the board? Surely this affects the measured "height"? I wonder if this really only leaves the auto focus and touch plate methods as valid?

No I haven't, and I do agree that the "part height" we get will effectively be "part and squished solder paste height". 

But if you think about it, this is all we need. If you look back at the ["Job Processor" section here](https://github.com/openpnp/openpnp/pull/1164#issuecomment-814137785) the placement (step 5) is the first point in time inside the process where the part height is needed. And it is exactly the "part and squished solder paste height" that we need there. 

Note that for other parts (those that have known part height at the time of placement), the ContactProbeNozzle subclass will typically still do a first-time probing in the placement i.e. it will account for the "squished solder paste height" offset with an extra probing. On machines without probing nozzle, this offset is ignored, i.e. parts are pressed down as if there was no solder paste! 

> Just add a button to the Parts panel that lets the user perform the part height detect for one or more parts. If the user tries to start the job with invalid part heights they get an error ...

We could add such a button. But I _sincerely_ hope it will not be restricted to this! This would really defy the very purpose of this whole PR. 

> But these are a bunch of very large features that have come out of the blue without any prior discussion and that add quite a bit of code complexity. I think it's important we discuss whether these features are worth the complexity added.

Yes they come out of the blue for myself too. Originally, I only wanted to add this to the ContactProbeNozzle (and I discussed this in the group and in the original PR). But then I realized step by step what it really meant. I thought I could make it worthwhile by adding generally available features such as part height sensing using Auto Focus and then it just kept growing. 

I now agree, that breaking it up was essential, but I also hope the work was worthwhile and that by now looking at the remainder in detail, you see it's not so bad after all and I was able to shift the weight a bit here, more in favor of making this feature worth-while. 

_Mark

vonnieda:
@markmaker Given the changes we're discussing with regards to I&S I am comfortable now with this, and its dependent PRs being merged. I'll leave the merging to you so that I don't screw up the order.

Thanks,
Jason

markmaker:
Thank you, @vonnieda, I appreciate. 

I'll push the I&S PR first and go from there. 

_Mark

markmaker:
Updated the Wiki:

markmaker:
Images for the Wiki, please do not delete.
See the new page here:
https://github.com/openpnp/openpnp/wiki/Nozzle-Tip-Changer
Linked from here:
https://github.com/openpnp/openpnp/wiki/Setup-and-Calibration:-Nozzle-Setup#nozzle-tip-changer
___
Nozzle Tip Changer

markmaker:
Another Wiki page added:
https://github.com/openpnp/openpnp/wiki/Up-looking-Camera-Auto-Focus

Linked from:
https://github.com/openpnp/openpnp/wiki/Setup-and-Calibration%3A-Bottom-Camera-Setup#auto-focus
https://github.com/openpnp/openpnp/wiki/Setup-and-Calibration:-General-Camera-Setup#general-configuration

PR #1178, opened 2021-04-19
Fixes use of JPanel in property sheet in FeedersPanel

Jnesselr:
# Description
I noticed when working on the IndexFeeder's configuration panel that a PropertySheet has a getPropertySheetPanel method which returns a JPanel. The FeedersPanel always assumes that the return from that method can be cast to AbstractConfigurationWizard which is not 100% true for all tabs in my case. I have a "search" tab that will search for feeders, and therefore doesn't need revert/apply buttons. I decided to return just a raw JPanel which had the effect that my tabs showed up under the machine configuration panel, but not under the feeders tab. This caused an exception in the FeedersPanel class.

# Justification
I'd say this just cleans up a mismatched interface. Since a developer can return just a JPanel (which I think is the correct behavior for property sheets), making FeedersPanel more intelligently handle the case when the JPanel isn't secretly an AbstractConfigurationWizard seemed like the appropriate behavior.

# Instructions for Use
There is no user facing change here.

# Implementation Details
I originally tested this in my index feeder branch where I have a property sheet that just uses a JPanel directly. Then I ported it back to the develop branch.

✅ I did
🙅 No changes there
✅ I did

### DIFF
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -275,9 +275,12 @@ public class FeedersPanel extends JPanel implements WizardContainer {
                         priorFeederId = feeder.getId();
                         PropertySheet[] propertySheets = feeder.getPropertySheets();
                         for (PropertySheet ps : propertySheets) {
-                            AbstractConfigurationWizard wizard = (AbstractConfigurationWizard) ps.getPropertySheetPanel();
-                            wizard.setWizardContainer(FeedersPanel.this);
-                            configurationPanel.addTab(ps.getPropertySheetTitle(), wizard);
+                            JPanel panel = ps.getPropertySheetPanel();
+                            if(panel instanceof AbstractConfigurationWizard) {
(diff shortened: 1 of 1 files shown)


Jnesselr:
Here's the terminal output of me running mvn test: [OpenPnP mvn test.txt](https://github.com/openpnp/openpnp/files/6332787/OpenPnP.mvn.test.txt)
Looks like I'm running Java 10, but it all passed with no errors.

The error in the Java 13 failed build is:
```
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 4.7 sec <<< FAILURE! - in SampleJobTest
testSampleJob(SampleJobTest)  Time elapsed: 4.699 sec  <<< ERROR!
java.lang.Exception: Pipeline stage "results" returned no List.
	at SampleJobTest.lambda$testSampleJob$1(SampleJobTest.java:109)
```

Jnesselr:
I decided to just run commit --amend and force push to get a new commit and have it re-run the checks. Looks like it might have been some test instability.

vonnieda:
Looks good to me - nice catch @Jnesselr!

PR #1179, opened 2021-04-19
Feature/Vision Stage Circular Symmetry

markmaker:
# Description
This is a new vision stage called `DetectCircularSymmetry`. It searches the image for [Circular Symmetry](https://en.wikipedia.org/wiki/Circular_symmetry) i.e. for center points where the pixels of individual surrounding concentric rings show little change (variance) when compared to the overall variance in the same area. 

As an example for Circular Symmetry, consider an archery target:

Archery target 
Source:  [Wikipedia about Circular Symmetry](https://en.wikipedia.org/wiki/Circular_symmetry)

The stage can be used to detect concentric things, like fiducials, sprocket holes or nozzle tips in runout calibration. 

As an input it takes the expected diameter (minimum and maximum) of the feature to be detected. This is often a property that is well known or can easily be measured (using a caliper) without special knowledge about computer vision.

Unlike other stages like `DetectCirclesHough`, the `Countours` family of stages, it does not require sharp edges, thresholds, filtering, footprint colors etc. internally or as a prerequisite. Instead, it simply **compares the image to itself**, so **it is entirely self-tuning**, and it can robustly work with very weak contrasts, very soft images and changing ambient light, any colors.  

## Example images
Here is a series of images before/after detection. All these are with the same **standard pipeline**, only adjusted for nozzle tip/fiducial diameter, **no tuning**:

Runout calibration with dark Samsung CP40 nozzle tips:
circular-symmetry-nozzle-tip

Same pipeline with a very low contrast, very soft/out of focus nozzle tip:
circular-symmetry-blur

Off-center view for camera position/rotation calibration. The nozzle tip is seen slightly from the side which means the various features of the nozzle are no longer concentric. The air bore is also slightly asymmetric. By simply selecting the wanted min/max diameter range it automatically snaps to the right ring. Observe how the blue center point jumps:
circular-symmetry-diameter-selection

The same exact pipeline used for a fiducial:
circular-symmetry-fiducial

## Diagnostics

For diagnostic purposes you can enable a heat map, indicating the degree of symmetry:
circular-symmetry-diagnostics

# Justification
All current pipelines need tuning and/or they are not very stable when factors such as ambient light change. In most cases they are also completely machine specific i.e. they need to be tuned to the camera lighting, exposure settings, gamma, resolution, distance, etc. so OpenPnP can't ship with out-of-the-box usable pipelines. Users can't share them.  

This stage aims to change this for the above-mentioned use cases. The idea is to make default pipelines some day use the `DetectCircularSymmetry`, if practice shows it works across many machines. 

# Instructions for Use
The standard pipeline is very simple (nozzle tip calibration example):
Pipeline Editor

The `DetectCircularSymmetry` stage is configured as follows:

### General Properties

**diagnostics**: Switches on an overlaid heat map of [Circular Symmetry](https://en.wikipedia.org/wiki/Circular_symmetry). 

**minSymmetry**: Minimum relative circular symmetry. Defaults to `1.2`. Any match below this ratio is considered invalid. The score value is calculated as the ratio of overall area pixel variance divided by the sum of pixel ring variances. These are variances weighed by area i.e. by pixel count. Because the image is compared against itself, this ratio is very stable, i.e. it should not be necessary to tune this.  

**subSampling**: To improve performance, only one pixel out of a block of subSampling × subSampling pixels is sampled. The preliminary winner is then locally searched using iteration with smaller and smaller subSampling size. The subSampling value will automatically be reduced for small diameters. Use BlurGaussian before this stage if subSampling suffers from interference/moiré effects. 

### Properties used when controlled by Vision Operations

**propertyName**: Sets the property name, as set by vision operations using this pipeline. Currently, this is `nozzleTip` for the nozzle tip calibration and `fiducial` for the fiducial locator. As soon as the named property is set by the vision operation, the stage is dynamically controlled and no longer uses the fixed properties in the stage itself. Proper camera Units Per Pixel scaling makes sure the dynamic properties work across machines/cameras/resolutions. 

**innerMargin**: The inner margin, relative to the diameter. Only used when the **propertyName** is set i.e. when the nominal diameter is supplied by vision operations. The **minDiameter** is then computed as `diameter*(1 - innerMargin)`.

**outerMargin**: The outer margin, relative to the diameter. Only used when the **propertyName** is set i.e. when the nominal diameter is supplied by vision operations. The **maxDiameter** is then computed as `diameter*(1 + outerMargin)`.

### Properties used when controlled by the Stage 

**minDiameter**: Minimum diameter, in pixels, of the wanted circular feature. This should always be a bit smaller than the actual diameter, because the symmetry analysis needs some margin area to detect circular image gradients. 

**maxDiameter**: Maximum diameter, in pixels, of the wanted circular feature. This should always be a bit larger than the actual diameter, because the symmetry analysis needs some margin area to detect radial image gradients. 

**maxDistance**: Maximum search distance, in pixels, from the nominal location of the feature (usually from the center of the camera view). 

## Vision Operations Control

To supply the wanted diameter from the vision operations, some new GUI properties have been added: 

### Nozzle Tip Calibration 
Nozzle Tip Calibration

**Nozzle Tip Diameter**: Must be set to the physical diameter of the feature you want to detect. Typically, a feature at the very point of the nozzle tip. Either the outer diameter or the air bore. 

The **maxDistance** property is automatically derived from the already present **Offset Threshold**.

Standard pipeline (Edit the pipeline and paste this using the Paste button):

```
<cv-pipeline>
   <stages>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageCapture" name="0" enabled="true" default-light="false" settle-first="true" count="1">
         <light class="java.lang.Double">128.0</light>
      </cv-stage>
      <cv-stage class="org.openpnp.vision.pipeline.stages.BlurGaussian" name="13" enabled="false" kernel-size="7"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DetectCircularSymmetry" name="results" enabled="true" min-diameter="30" max-diameter="50" max-distance="200" min-symmetry="1.2" property-name="nozzleTip" outer-margin="0.2" inner-margin="0.4" sub-sampling="8" diagnostics="false"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DrawCircles" name="6" enabled="true" circles-stage-name="results" thickness="2">
         <center-color r="0" g="204" b="255" a="255"/>
      </cv-stage>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageWriteDebug" name="7" enabled="false" prefix="runoutCalibration" suffix=".png"/>
   </stages>
</cv-pipeline>
```

### Fiducial Locator
Fiducial Locator

The **diameter** of the fiducial is automatically derived from the footprint. Therefore, the same underlying data can be used both for the classic `MatchPartTemplate` based pipeline and the new pipeline (obviously this only works if the ficucial is actually circular). 

**Max. Distance**: Maximum allowed distance between nominal fiducial location and detected fiducial location. If this is exceeded, the locator will throw. 

Standard pipeline (Edit the pipeline and paste this using the Paste button):

```
<cv-pipeline>
   <stages>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageCapture" name="image" enabled="true" default-light="true" settle-first="true" count="1"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.BlurGaussian" name="blur" enabled="false" kernel-size="3"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DetectCircularSymmetry" name="circular" enabled="true" min-diameter="10" max-diameter="100" max-distance="200" min-symmetry="1.2" property-name="fiducial" outer-margin="0.2" inner-margin="0.4" sub-sampling="8" diagnostics="false"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ConvertModelToKeyPoints" name="results" enabled="true" model-stage-name="circular"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DrawCircles" name="draw" enabled="true" circles-stage-name="circular" thickness="1">
         <color r="255" g="255" b="0" a="255"/>
         <center-color r="255" g="153" b="0" a="255"/>
      </cv-stage>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageWriteDebug" name="debug_results" enabled="false" prefix="fidloc_results_" suffix=".png"/>
   </stages>
</cv-pipeline>
```
 
# Implementation Details
1. Tested in simulation and with various source images. 


doppelgrau:
Personally not sure about the integration inside the GUI, but great pipeline, I think it can really help in some cases where currently a very tight control camera setting, light and pipeline-parameters must match.
Can't wait to use that stage.

markmaker:
Just made another test. Transparent tape, varying reflections. Back when I made [that feeder](https://makr.zone/openpnp-new-referencepushpullfeeder/393/), I was entirely unsuccessful to create a robust pipeline. 

With the new stage, again, **no tuning**:

circular-symmetry-transparent

Note, the stage is not yet capable to return more than one hit, i.e. it is not yet usable for lines of sprocket holes. But that's easy to add later. The model is already a `List<Circle>`.

_Mark

src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java
```diff
@@ -1017,8 +1028,13 @@ public void setEnabled(boolean enabled) {
         this.enabled = enabled;
     }
 
-    public CvPipeline getPipeline() throws Exception {
-        pipeline.setProperty("camera", VisionUtils.getBottomVisionCamera());
+    public CvPipeline getPipeline(Camera camera, Location measureLocation) throws Exception {
+        pipeline.setProperty("camera", camera);
+        pipeline.setProperty("nozzleTip.diameter", getCalibrationTipDiameter());
+        pipeline.setProperty("nozzleTip.tolerance", getOffsetThresholdLength().multiply(1.4));
```
vonnieda, line 1034: Why is this multiplied by 1.4?

vonnieda (changes requested):
Hi @markmaker, this looks like a really great addition! This might finally solve all the problems with HASL fiducials. Nice work!

I have a couple small comments:

- Nozzle Tip Diameter: Can we change the name and be explicit? If this is specifically a vision property, call it Vision Diameter, or something like that. If it is specifically the outer diameter that we want, call it Outer Diameter.

Thanks,
Jason

src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java
markmaker: >     * Nozzle Tip Diameter: Can we change the name and be explicit? If this is specifically a vision property, call it Vision Diameter, or something like that. If it is specifically the outer diameter that we want, call it Outer Diameter.

I'll call it Vision Diameter. It is not necessarily the outer diameter, i.e. from the Instruction for Use:

> **Vision Diameter**: Must be set to the physical diameter of the feature you want to detect. Typically, a feature at the very point of the nozzle tip. Either the outer diameter or the air bore. ...

_Mark

markmaker: > Why is this multiplied by 1.4?

I must give it some margin to robustly detect the nozzle tip even outside the threshold, then causing an exception. Otherwise, it would possible "detect" a random speck of the background as the nozzle tip, falsely assuming it is inside the threshold. 

This typically happens in camera position/rotation calibration with the nozzle tip being off-center, if your Units per Pixel are off e.g. because the focal plane for the bottom camera Z has changed or when using a different video mode etc. Better to get "positive falses" than "false positives" (hope this translates to English) ;-) 

_Mark

vonnieda: I should have been more direct: Can you add a comment in the code explaining this magic number? :)

markmaker: You're right. Made it configurable through `machine.xml` if absolutely needed and gave it some comments.

markmaker: Are these changes ok?

markmaker: The mailer does not represent this well. What I meant are 29183eb and 2a28ff8 (hope this will be hyperlinked as usual).

markmaker:
Repository for Wiki images (do not delete)
___________
Out of Focus

Issue #1176, opened 2021-04-14
Import Board on KiCAD .pos  does not assign every part


RRudi92:
# Problem Report
## Describe the Issue
I have encountered several problems with importing placement files. Using .pos (KiCAD), some parts are not assigned, using Eagles .mnt files, none part is assigned.

**Using Eagle .mnt files:**
None of the parts are assigned. Even if the part-ID are exact the same as already existing parts.
=> See issue <<<<will be created in a few minutes>>>>>

**Using KiCad .pos files:**
Some of the parts are not assigned.

## Steps to Reproduce
Import file with contents (both parts exist in parts library/tab):
```
# Ref Val Package PosX PosY Rot Side
C7 1µ rcl-C0805 -22.25 -14.25 270.00 top
C8 10n rcl-C0603 -3.75 -17.00 90.00 top
```

## Expected Result
Correct assignment of both parts

## Actual Result
only C8, the 100n-part is assigned. The assignment of 1µ is missing.
![image](https://user-images.githubusercontent.com/64029642/114740676-5ab6c400-9d4a-11eb-9a4f-63a5fde4a7d1.png)
![image](https://user-images.githubusercontent.com/64029642/114740821-7e7a0a00-9d4a-11eb-92f4-fbc6695bd3fc.png)

If using "Use only Value as Part ID", no part is assigned.

If using "Create missing parts", a new part is created for the 1µ:
```
<part id="rcl-C0805-1Âµ" height-units="Millimeters" height="0.0" package-id="rcl-C0805" speed="1.0" pick-retry-count="0"/>
```

Image:
First two parts: Assignment
Parts 3&4: Assignemt using only value
Parts 5&6: Assignem using "Create missing parts"
Last two parts: Assignment after new parts were created.
![image](https://user-images.githubusercontent.com/64029642/114745193-b8e5a600-9d4e-11eb-83b9-31efddd60cbb.png)

## Notes, Log Files, Screen Captures, Videos, etc. to Show the Issue
The Char 'µ' seems to be wrongly assigned to two Chars.

doppelgrau:
I'm quite sure at least for the kicad importer: It's an encoding problem.
Just to be sure: only non-ASCII-characters are affected in your setup?

Additional information that might help: Which encoding is default on your system and in which encoding the .pos is saved?

RRudi92:
The file is in UTF-8.
As far as I'm aware only the µ is affected. I will chek on wednsday if sth. else could be affected. 

[...]

PR #1182, opened 2021-04-21
Kicad importer: Assume UTF8 as charset

doppelgrau:
# Description
Assume UTF8 encoding when importing a kicad .pos file.

# Justification
If non-ASCII character like "µ" are used, importing fails. See bug #1176 

# Instructions for Use
No change in the behaviour. Only assumption is, that the file is now in UTF8, and not in an undefined character set.

# Implementation Details
1. How did you test the change? Tested from the bug reporting person
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. yes


vonnieda:
Looks good, thanks @doppelgrau. And thank you for helping diagnose this issue!

PR #1185, opened 2021-04-28
Adding a chinese translation to OpenPnP

mccj:
Adding a chinese translation to OpenPnP


vonnieda:
Nice work @mccj - thank you!

Issue #1184, opened 2021-04-26
The latest version resets the Z axis position


bacrua:
The latest version resets the Z axis position. This was also observed in the version that I downloaded on 02/20/2021. But only now in the work I was able to make sure that there was an error.
1. Go to any strip feeder. Click on autosetup.
2. after the first autosetting, the storing does not change.
3. Select the second available strip feeder and click Autosetup.
4. Z axis has zeroed values.
![PNP1](https://user-images.githubusercontent.com/48532007/116163332-3969b680-a700-11eb-8b99-3f89db3d015f.png)
![PNP2](https://user-images.githubusercontent.com/48532007/116163333-3a9ae380-a700-11eb-8186-c92dea12f0a8.png)

doppelgrau:
Not sure if that's intended behavior.
If not a fix would be quite simple, in the Wizzard lines 600ff just keep the old height instead setting it to zero.

`                                 final Location referenceHole1 = referenceHoles.get(0)
                                                                               .derive(null, null,
                                                                                       null, 0d);
                                 final Location referenceHole2 = referenceHoles.get(1)
                                                                               .derive(null, null,
                                                                                       null, 0d);
`

markmaker:
Yep.

```
diff --git a/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java b/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
index df574b7e84..5ea7996df0 100644
--- a/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
@@ -599,11 +599,15 @@ public class ReferenceStripFeederConfigurationWizard extends AbstractConfigurati
                                          firstPartLocation, secondPartLocation,
                                          part1HoleLocations, part2HoleLocations);
                                  final Location referenceHole1 = referenceHoles.get(0)
-                                                                               .derive(null, null,
-                                                                                       null, 0d);
+                                         .derive(null, null,
+                                                 null, 0d)
+                                         .derive(feeder.getReferenceHoleLocation(), 
+                                                 false,  false,  true, false);
                                  final Location referenceHole2 = referenceHoles.get(1)
-                                                                               .derive(null, null,
[...]

PR #1186, opened 2021-04-29
Fix/Keep the Z in ReferenceStripFeederConfigurationWizard Auto Setup.

markmaker:
# Description
This keeps the Z coordinate of the two reference hole locations intact through **Auto Setup**.

The bug appears to haven been brought in by adding property change firing here:

diff
https://github.com/openpnp/openpnp/commit/ee7ea6fd1b402c4a9fe6fc7eca180ca5b8004f0b#diff-b348cb0fe5f0f395b42e79116b72ea0253f5343f07f653fd186850362bc83f25

Which is benign, of course. But Auto-Setup uses mixed direct model _and_ Wizard controls modifications here (see the `invokeLater`):

https://github.com/openpnp/openpnp/blob/e4f983f76f716593b7fedf72c9b75c68c9c95fc5/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java#L608-L629

... which in earlier versions caused overwriting the zeroed Z with the old coordinate when Apply was pressed. Today the property change firing causes the zeroed Z to be updated into the Wizard immediately.

It is still inconsistent now, i.e. pressing **Reset** on the Wizard will only undo half of the Auto-Setup. So this should be revisited in the future to either do all modifications only on the Wizard, or only on the model. 

# Justification
See #1184.

This PR intends to be minimal, so it does not address the mixed modification.

# Instructions for Use
See #1184.

# Implementation Details
1. Tested in simulation.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceStripFeederConfigurationWizard.java
@@ -599,11 +599,15 @@ public class ReferenceStripFeederConfigurationWizard extends AbstractConfigurati
                                          firstPartLocation, secondPartLocation,
                                          part1HoleLocations, part2HoleLocations);
                                  final Location referenceHole1 = referenceHoles.get(0)
-                                                                               .derive(null, null,
-                                                                                       null, 0d);
+                                         .derive(null, null,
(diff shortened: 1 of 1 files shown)


markmaker:
I'm merging this, as it is a very small change and confirmed bug.

brainstorm:
Thanks soo much for this, it was driving me crazy! :)

PR #1188, opened 2021-05-02
Feature/Better Issues and Solutions

markmaker:
# Description
Aside from many advantages, Issues & Solution has also shown some down-sides:

1. It lures users into adopting advanced features and optimization too early in the machine build process. Instead of focussing on basic machine motion first, users are bogged down with [premature optimization](https://en.wikipedia.org/wiki/Program_optimization#When_to_optimize).
2. There is no guidance as to what should be done first.  
3. Sometimes it is too easy to just accept all solutions (even supported by multi-select). The user forgets she/he is still responsible to check each proposed solution. 
4. Going back from an advanced and perhaps problematic configuration to a more conservative and stable configuration is not possible.

This PR aims to remedy all that by employing a "milestones" pattern for structured progress and guidance in the machine build process. At the same time it adds some powerful new features and solutions to complement the experience.

The UI has been reworked to move Issues & Solutions to a top level tab. The tab even has an indicator to show when issues have been found (automatically searched when OpenPnP starts). There is a new pane underneath the Issues & Solutions list, where each selected issue is described. Issues can be accepted, dismissed, undone right there (the popup message box has gone). 

Issues can now have multiple choice and other inputs to control more elaborate solutions.

# Justification
See the intro.

# Instructions for Use

## Tracking Progress

Issues & Solutions now has **Milestones** to track the progress when building and configuring a machine. Only the Issues & Solution relevant for the current milestone target will be shown. Some solutions will be adaptive to the target milestone, proposing simpler solutions for earlier milestones. These will be revisited when you proceed to later milestones. 

For troubleshooting, you can also go back to earlier milestones. Issues & Solutions will then sometimes propose taking back more advanced choices you already made. This way you can try to solve a problem by starting from a safer base. 

The last issue per milestone will always be the **Milestone issue**, giving you the option to proceed or go back:

Milestone issue

### Welcome Milestone 
The **Welcome** milestone allows you to get to know OpenPnP with a simulation machine. Furthermore, you can use Issues & Solutions to choose between common head / nozzle topologies: standalone nozzles, pairs of nozzles with shared Z and negation, pairs with shared Z and cam. Using a multiplier you can tell OpenPnP how many standalone nozzles or nozzle pairs you want. Issues & Solutions then creates all the required axes and nozzles and wires them up. 

Nozzle Solutions

Even complex nozzle solutions are created with a few clicks:

Complex nozzle solution

You can even revisit the solution i.e. change the number and type again and again. Issues & Solutions will carefully re-use any nozzles and axes it finds and only add new ones as needed. On reused components the detail configuration is preserved (except for names).  

### Connect Milestone 
The **Connect** milestone handles the conversion from the simulated machine to a real connected machine with motion controller and USB cameras. It also tests the driver connection and tries to discover the firmware of your controller. If the firmware cannot be detected, there is the option to use generic G-code.

Connect Firmware

### Basics Milestone 
The **Basics**  milestone handles moving around the machine and switching vacuum and lights. It makes sure all axes have letters and are mapped to drivers. It generates G-code etc. 

### Vision Milestone 
The **Vision** milestone handles some Computer Vision issues. This is currently very thin, more is planned in the future. 

### Calibration Milestone 
The **Calibration** milestone handles calibration issues. This is currently very thin, more is planned in the future. 

### Production Milestone 
The **Production** milestone handles recurrent configuration tasks such as feeder setup as well as production issues such as board, placement and job handling. This is currently empty, but feeder issues are already in the pipeline. More is planned in the future. 

### Advanced Milestone
The **Advanced** milestone, while also serving as a production milestone, finally exposes all the advanced features, such as advanced driver settings, motion planner, asynchronous driver, contact probing nozzle etc. For the most problematic of these settings, Issues & Solutions will propose taking them back when going back to earlier milestones.

# Implementation Details
1. Tested with the lab bench three controller rig.
3. Changes in the `org.openpnp.spi` or `org.openpnp.model` packages are in `Solution` (obviously) and one minor method in the `Axis.Type`.

Note that the number of changed files is large due to the distributed character of the `findIssues()` override that has been refactored. It helps to ignore whitespace chnages when looking at it (indentation changes a lot due to milestone conditions).


Jnesselr:
I kind of like this approach, but I do worry the "Advanced" section will end up being like an "other" category. I have one for the IndexFeeder in my dev branch currently that adds an issue when a slot doesn't have a configured location. I'm also not sure about "dismiss" in some of these contexts. If you dismiss the firmware not detected issue, what happens? Without dismiss, this UI gives a really clear indication that you have to do one or the other option and move forward which feels like a better UX.

markmaker:
Hi @Jnesselr,

Thanks for looking into this.

> I do worry the "Advanced" section will end up being like an "other" category. I have one for the IndexFeeder in my dev branch currently that adds an issue when a slot doesn't have a configured location. 

Each Issue you provide can determine in what milestone(s) it appears. A simple example:

https://github.com/openpnp/openpnp/blob/1367ab1051d70a3fe9f3449e995b5d5c5848bbbf/src/main/java/org/openpnp/spi/base/AbstractCamera.java#L1334-L1357

For now, I assumed that each sub-system of OpenPnP will evolve along the same milestones. So if your feeder issue is "Basics" or "Vision" or "Calibration" **topically**, just condition it to the right milestone. 

Or maybe we should add a new "Feeders" milestone after "Calibration"? Or maybe a more general "Production" milestone? What do you think?

> I'm also not sure about "dismiss" in some of these contexts. If you dismiss the firmware not detected issue, what happens? 

Then Issues & Solutions cannot provide solutions based on firmware detection. It will not try and generate G-code for the GcodeDriver. Any solution that is marked as "Fundamental" signals to the user that it is a foundation for _more solutions_. 

I see that the **Issue** text could be improved to explain what happens, if you don't provide (or dismiss) the solution. 

Perhaps: _"Firmware not detected (machine is disabled). Without known firmware, Issues & Solutions cannot generate any G-code configuration solutions for you."_ 

From a design perspective, I wanted Issues & Solutions to be strictly "opt-in". The user is still the boss! Manual configuration is not off the table, far from it. If users know more about the matter at hand, they can go and provide the solution manually and the Issue will disappear. Or they can dismiss the solution, and if "Fundamental", all the dependent solutions with it. 

There is a certain nag aspect to it. I don't want this to become Clippy!

Clippy
The Science Elf, YouTube

But I'm open to other considerations. Do you have something in mind?

_Mark

Jnesselr:
Calling it "Production" might be a good way to handle it. For the only issue I've added so far for these feeders, I marked it as fundamental. Not because that's how you get more information on what to do next, but because your job will literally not start or work until you fix that issue. I just don't want to give the user the idea of "Oh, yeah I'll just dismiss this and then it won't be a problem". No, it'll still be a problem and if you fix the problem, the issue will go away.

markmaker:
**@Jnesselr**, I just added the Production milestone. So you could target your feeder to it.

However, I would set the **Error** severity, not the **Fundamental**. The latter should be reserved for issues that block others. 

The **Error** severity is actually marked in red i.e. much more visible, and it is very clear that a solution is needed. 

_Mark

Jnesselr:
Awesome, will do! Thank you!

markmaker:
@vonnieda 

**Just to be clear:** with your [comment here](https://github.com/openpnp/openpnp/pull/1174#issuecomment-830851458)...

> @markmaker Given the changes we're discussing with regards to I&S I am comfortable now with this, and its dependent PRs being merged. I'll leave the merging to you so that I don't screw up the order.

... you're on board to merge this one too, right? 

It will hide all the `ContactProbingNozzle` and Z probing suggestions and all the other complexities and premature optimization until a user targets the **Advanced** milestone. 

_Mark

markmaker:
@vonnieda 

**Just to be clear:** with your [comment here](https://github.com/openpnp/openpnp/pull/1174#issuecomment-830851458)...

> @markmaker Given the changes we're discussing with regards to I&S I am comfortable now with this, and its dependent PRs being merged. I'll leave the merging to you so that I don't screw up the order.

... you're on board to merge this one too, right? 

It will hide all the `ContactProbingNozzle` and Z probing suggestions and all the other complexities and premature optimization until a user targets the **Advanced** milestone. 

_Mark

vonnieda:
@markmaker Yep, have at it! :)

markmaker:
Wiki images, please don't delete.
See https://github.com/openpnp/openpnp/wiki/Issues-and-Solutions
____
Info

Jnesselr:
I just found out that this diff broke functionality I was relying on in unit testing. It makes it much much harder to test succinctly and without involving other pieces of the code base. Before, I could do this:

```java
    @Test
    public void findIssuesAddsIssueIfSlotHasNoLocationSet() {
        feeder.setHardwareId(hardwareId);
        feeder.setOffset(feederOffset);
        feeder.setSlotAddress(feederAddress);
        assertNull(feeder.getSlot().getLocation());

        List<Solutions.Issue> issues = new ArrayList<>();
        feeder.findIssues(issues);

        assertEquals(1, issues.size());
        Solutions.Issue issue = issues.get(0);

        assertEquals(feeder, issue.getSubject());
        assertEquals("Feeder slot has no configured location", issue.getIssue());
        assertEquals(Solutions.Severity.Error, issue.getSeverity());
        assertEquals(Solutions.State.Open, issue.getState());
    }
```

But now, I have to do this:
 ```java
     @Test
    public void findIssuesAddsIssueIfSlotHasNoLocationSet() {
        feeder.setHardwareId(hardwareId);
        feeder.setOffset(feederOffset);
        feeder.setSlotAddress(feederAddress);
        assertNull(feeder.getSlot().getLocation());

        Solutions solutions = new Solutions();
        solutions.findIssues();
        solutions.publishIssues();

        List<Solutions.Issue> issues = solutions.getIssues();
        Optional<Solutions.Issue> maybeIssue = issues.stream().filter(i -> i.getSubject() == feeder).findFirst();

        assertTrue(maybeIssue.isPresent());
        Solutions.Issue issue = maybeIssue.get();

        assertEquals(feeder, issue.getSubject());
        assertEquals("Feeder slot has no configured location", issue.getIssue());
        assertEquals(Solutions.Severity.Error, issue.getSeverity());
        assertEquals(Solutions.State.Open, issue.getState());
    }
```
    
Overall, I think this diff might be better UI wise, but it really complicates any unit testing around issues. Is there anything that can be done about this?

markmaker:
Frankly, I don't know a lot about Java Unit Testing. I guess there would have to be some central bootstrap piece (is "test fixture" the right term?) for all future Issues & Solutions tests. Then just instantiate one in each individual test. If you know how to write such a fixture, I would appreciate very much if you did. 

As to why I changed it: formerly the `issues` were just a dumb array. Now it's an object with extensible logic. One important property of the passed `solutions` object is the target Milestone. We [talked about it](https://github.com/openpnp/openpnp/pull/1188#issuecomment-831495384), remember? It is used to filter and influence the solutions. And there are other properties now tucked into the Solutions object rather than polluting the Machine object. 

_Mark

Jnesselr:
I wouldn't modify this in a test fixture. I remember talking about it, for sure, and I still support it. I just didn't realize the code would cause this problem until I rebased and ran the tests. I assumed I could still do `feeder.findIssues(solutions)` on the solutions object, but I cannot.

The reason why you have to call solutions.findIssues() is because pendingIssues is null at the start. I don't understand why you have a separate pending issues and issues, requiring the call to publish issues. It's only used in one place and it's right after find issues is called.

If you didn't have pending vs issues and had that array list already instantiated, then everything would work with the much cleaner test I posted earlier.

markmaker:
Pending issues double buffering is done for various reasons:

1. Avoid race conditions between the issues being built, and the ones still being displayed at the moment. 
2. Ability to roll back if the find fails.
3. Before I switch the milestone to the next, I want to be able to find solutions once more to check for incomplete issues. But I need to do this without overwriting the displayed ones, because these might still contain "undo" state that I don't want to throw away until the user confirms. 

_Mark

markmaker:
Having said that: we could perhaps change `org.openpnp.model.Solutions.add(Issue)` so it will create the `pendingIssues` on demand if `null`. Remove if from `org.openpnp.model.Solutions.findIssues()` (first line) instead.

_Mark

PR #1189, opened 2021-05-03
Updating to JUnit 5

Jnesselr:
# Description
Update from JUnit 4 to JUnit 5

# Justification
Because JUnit 5 is one better than JUnit 4. I was going to upgrade to JUnit 4.13 and then decided to upgrade to JUnit 5. There's some differences here and there between the two, but JUnit 5 has more useful functionality.

# Instructions for Use
Dev only feature.

# Implementation Details
Unit Testing ++
Yes
No, only in tests.
I ran the unit tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -35,7 +35,7 @@
 		</dependency>
 		<dependency>
 			<groupId>org.junit.jupiter</groupId>
-			<artifactId>junit-jupiter-api</artifactId>
+			<artifactId>junit-jupiter-engine</artifactId>
 			<version>5.7.1</version>
 			<scope>test</scope>
 		</dependency>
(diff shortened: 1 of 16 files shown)


markmaker:
Hi @Jnesselr 

How can I switch on unit test output again?

It runs silently now:

terse

Thanks,
_Mark

Jnesselr:
I added the option to redirect test output to files. Writing to system out isn't always consistent when running tests I've found, but writing to files always is. We can remove that option if you and @vonnieda want.

The path for the GcodeDriverTest, for example, would be in `target/surefire-reports/GcodeDriverTest-output.txt`

markmaker:
Thanks, @Jnesselr 

I think it would be better, if the output was visible again. I never observed any inconsistencies. 

Furthermore, I guess the GitHub Actions that run the tests as well would be hard to diagnose otherwise. 

@vonnieda, what do you think?

_Mark

vonnieda:
Agreed - it should work the way it did before.

Jason

Jnesselr:
Done!

PR #1193, opened 2021-05-07
Fix/Exceptions from Pipeline Stages, Scripts not propagated 

markmaker:
# Description
Exceptions from pipeline stages and from script calls are not propagated. See also ##1191.

In case of pipeline stages this is benign/wanted behavior in most cases i.e. when it signals failure of the vision stage itself, but some exceptions, such as those originating from machine motion control (most notably timeout exceptions) must not be ignored and should be propagated. 

In case of scripting, I guess it was never correct behavior to swallow exceptions. 

# Justification
See a real case and discussion:
https://groups.google.com/g/openpnp/c/jc8u42qINM8/m/7lLk5GSLAAAJ

# Instructions for Use

Context in which the issue has increasingly popped up:
1. Use advanced motion control, most notably the ReferenceAdvancedMotionPlanner with continuous motion enabled. 
2. This will only queue motion commands until OpenPnP actually requires the machine to really be at the target (lazy eval). 
3. This is typically the case when the camera wants to capture an image. 
4. Capturing occurs from withing a pipeline stage. If the motion now generates an exception (such as a timeout), it will be swallowed by the stage. 
5. The user never sees an error message, until recently it wasn't even logged. 
6. OpenPnP could get out of sync with the machine without the user/machine task noticing.  

A similar scenario is possible when motion is triggered (lazy evaluated) by script action. 

# Implementation Details
1. Tested/provoked in simulation before/after the fixes. This now correctly aborts the machine task if critical exceptions occur in lazily evaluated motion control inside pipeline stages. In the Pipeline Editor it is indicated by message box. 
   **NOTE**: I did **not** test/provoke scripting exceptions. I must leave that to somebody with existing scripts / scripting experience. **Therefore, I propose merging to `test`, for somebody else to test this.** 
3. Changes in the `org.openpnp.spi` concern `capture()` and `autoSettleAndCapture()` methods now throwing Exceptions from scripting.

Please note that the changes are structured in three commits for easier review etc..

- Pipeline stages
- Scripting
- ScriptAction (unused?)

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -70,12 +70,12 @@ import org.openpnp.events.BoardLocationSelectedEvent;
 import org.openpnp.events.JobLoadedEvent;
 import org.openpnp.events.PlacementSelectedEvent;
 import org.openpnp.gui.components.AutoSelectTextTable;
-import org.openpnp.gui.support.CustomBooleanRenderer;
 import org.openpnp.gui.importer.BoardImporter;
 import org.openpnp.gui.panelization.DlgAutoPanelize;
 import org.openpnp.gui.panelization.DlgPanelXOut;
 import org.openpnp.gui.processes.MultiPlacementBoardLocationProcess;
 import org.openpnp.gui.support.ActionGroup;
+import org.openpnp.gui.support.CustomBooleanRenderer;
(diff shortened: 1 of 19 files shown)


markmaker:
Use diff without whitespace to better see the changes (indentation of `try {}` blocks):

https://github.com/openpnp/openpnp/pull/1193/files?diff=unified&w=1

PR #1194, opened 2021-05-11
Fix/Axis must not fire live Coordinate Property Changes

markmaker:
# Description
Users reported sporadic GUI related exception. The issue was traced back to mullti-threading. 

`AbstractCoordinateAxis.setCoordinate(double)` and `AbstractControllerAxis.setDriverCoordinate(double)` must not `firePropertyChange()` as these changes are live from the machine thread. Coordinate changes must instead be handled through `MachineListener.machineHeadActivity(Machine, Head)`. 

A search revealed that these property names are not used in the source. These were introduced by me, copying the pattern over from other (GUI visible) properties on the axis, purely for completeness. 

# Justification
Bug-fix.

See user case:
https://groups.google.com/g/openpnp/c/ULaw9j_t8-k/m/AvAd1iWSAgAJ

# Instructions for Use
No visible changes.

# Implementation Details
1. Tested with extended simulation machine.
3. Changes in the `org.openpnp.spi` `base` package.

### DIFF
--- a/src/main/java/org/openpnp/gui/MachineSetupPanel.java
+++ b/src/main/java/org/openpnp/gui/MachineSetupPanel.java
@@ -44,6 +44,7 @@ import javax.swing.JTabbedPane;
 import javax.swing.JTextField;
 import javax.swing.JToolBar;
 import javax.swing.JTree;
+import javax.swing.SwingUtilities;
 import javax.swing.event.ChangeEvent;
 import javax.swing.event.ChangeListener;
 import javax.swing.event.DocumentEvent;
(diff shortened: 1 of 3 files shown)


markmaker:
> I think it's fine to merge this, but I want to point out that it doesn't actually "fix" the bug. The bug is in https://github.com/openpnp/openpnp/blob/develop/src/main/java/org/openpnp/gui/MachineSetupPanel.java#L291 which should be calling SwingUtilities.invokeLater.

Agree, I added it.

Unless I'm missing something, `SwingUtilities.invokeLater()` does queue up i.e. if multiple events happen, there will be multiple refreshes, correct? Ideally, it would only flag the contents as dirty and then asynchronously (after a certain refresh grace period) refresh once. But I guess that's just _not_ how Swing does it.  ;-)

_Mark

markmaker:
Just to be clear: this PR still removes the coordinate property change firing. I guess that's the best combination for now.

PR #1196, opened 2021-05-13
Feature: bottom vision offset

doppelgrau:
# Description
Introduces a "visionOffset" in the BottomVision PartSettings.
This offset is relative to the desired pick location, and defines the (expected) center of the bottom vision. After using bottom vision, this offset is subtracted from the determined offset, to get the real center/pick location of the part.

# Justification
There are some parts where the desired pick location is not the center of the bottom vision.
E.g. commonly used footprints for the ESP32-WROOM modules have a center that is around 0.65mm on the y axis moved compared to the bottom vision results. An other example are SMD connectors, the connectors in the back move the center in one direction.

# Instructions for Use
The vision offset is measured from the center/pick location to the (expected) center of the rotated rectangle of the bottom vision. E.g. if the center of the bottom vision is above the part center (like in the example image), the offset for Y is positive.
There are two ways to use this feature:

- manually determine the offsets and enter them
- center the part, then use the "Detect" button. It will run a bottomVision run and center the part, and measure the difference.

Attention: In both cases you will need to "Apply" the changes, before they will me used for "TestAlignment"

# Implementation Details
1. How did you test the change? Tested on my machine and in simulator. Also added test cases.


doppelgrau:
Up_2021-05-14_01 08 46 439
Example image for the wiki if accepted

markmaker:
Hi @doppelgrau 

From my perspective the concept looks good. 

@vonnieda must say regarding added UI complexity and versus his visual training alternative: 
https://youtu.be/zyXNehJzrYk

Back to the technical scope:

I see you add `offsets.getRotation()` here:
https://github.com/openpnp/openpnp/blob/58298c4f1bbb0ffd0324c47e8bf2ad85782ae29e/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java#L218
and analogously `angleOffset` here:
https://github.com/openpnp/openpnp/blob/58298c4f1bbb0ffd0324c47e8bf2ad85782ae29e/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java#L268

Are you sure this is needed/correct? Aren't your `visionOffsets` applied on top of the already justified part?

Just a gut feeling without thinking all too hard or following the code path all the way through to placement. ;-)

_Mark

doppelgrau:
If I understood it right, the part is not necessarily aligned (depending on the configuration of allowed tolerances and whether pre rotate is enabled or not), so that rotational offset can be significant. But usually that will be only a few degrees, so probably only really relevant with allow full rotation and without pre rotation (only one vision pass?!)

markmaker:
@doppelgrau, I did not explain precisely enough. 

You are right, the part may not be physically aligned all the way, but the `offsets` variable contains the whole compensation from "wanted" to "observed". It includes the detected correction from the last vision result, even in multi-pass.

https://github.com/openpnp/openpnp/blob/58298c4f1bbb0ffd0324c47e8bf2ad85782ae29e/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java#L214-L218

So we got the transformation that fully aligns the part already in `offsets`, including rotation. 

Now you want to translate it by your `visionOffset` on top of that. So I just don't think you need to include the rotation offset there (only the placement rotation in pre-rotate).

My head hurts. [When I worked on bottom vision](https://makr.zone/extending-bottom-vision-in-openpnp/341/) (doing the multi-pass), I made tests with a large part and I deliberately picked it with large offsets and the feeder angle off by ~15°. So any mistake would be clearly visible. See here:
https://github.com/openpnp/openpnp/pull/815#issuecomment-484665152

If I remember correctly I got the sign wrong in one of the calculations, this test flushed it out. Better than any amount of thinking ;-)

Unfortunately, your UnitTests can't possibly expose that, because the SimulatedUpCamera can't simulate bad picks. The part on the nozzle is always drawn with perfect zero linear offset, perfect zero angular offset. 

https://github.com/openpnp/openpnp/blob/52dfa2944397426b70e5ea3c4eeaecd407d30e38/src/main/java/org/openpnp/machine/reference/camera/SimulatedUpCamera.java#L148

I always wanted to add bad picks to the imperfection modes of the [SimulationModeMachine](https://github.com/openpnp/openpnp/blob/develop/src/main/java/org/openpnp/machine/reference/SimulationModeMachine.java) that I made to test Advanced Motion Control. It would be very easy. 

grafik


_Mark

markmaker:
Another question: I see why you need the custom `parts.xml` and `package.xml`. But why a custom `machine.xml`?

I just ask, because some development sometimes requires updating all the `machine.xml` resources, so the fewer, the better.

_Mark

doppelgrau:
IMG_20210516_173313_1

Here a bit exaggerated why I think the offset need to be taken into account. without rotation the new center would be more on the left side.

@machine.xml: Just saw the example at an other test, will try it without.

markmaker:
I think we are not talking about the same thing ;-) 

Of course, you need to rotate the vision offset by the **placement angle**. But **not** by the **offset angle**. 

Does this help explain it?

Vision Offset Illustration


_Mark

doppelgrau:
Why is the offset angle not needed?
An example (without prerotation to keep is simpler).
Bottom vision offset of 1mm/1mm
We have a really bad pick (for the numbers) 45° rotation offset, otherwise centered.
So the resulting offset is (1.4mm/0mm/45°)
If I add the offset without rotation, I do not get back to the center, but to (0.4mm, -1mm, 45°), with rotation it's (0mm,0mm,45°)

markmaker:
Well I don't know how to explain differently, does the illustration not help?  :-/ 

From the code flow perspective, it is stored here:
https://github.com/openpnp/openpnp/blob/5b6b43450387e4a7c2978c170adf00a428ca3a88/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L783

and applied here:
https://github.com/openpnp/openpnp/blob/5b6b43450387e4a7c2978c170adf00a428ca3a88/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L976-L979

and here:
https://github.com/openpnp/openpnp/blob/5b6b43450387e4a7c2978c170adf00a428ca3a88/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L990-L1003

Used transforms:

https://github.com/openpnp/openpnp/blob/fc5efe612db302582dbaf887fde9bc46049f77f2/src/main/java/org/openpnp/model/Location.java#L367-L381

https://github.com/openpnp/openpnp/blob/5b6b43450387e4a7c2978c170adf00a428ca3a88/src/main/java/org/openpnp/model/Location.java#L207-L211

I guess a definitive test is needed.

_Mark

doppelgrau:
I understood the graphic in which order the offset should be applied. But I manipulate the offset directly during vision, in my 45° example the offset should be zero, since the part is centered, only rotated. And to reach that, I need to rotate the vision offset by the offset rotation (At least that is my understanding)

markmaker:
Let's make it simpler. Step by step, so you can yell where you disagree.

1. Let's assume we have 0° placement angle so pre-rotate and non-pre-rotate are the same, i.e. pre-rotate (as the name implies) does nothing when the angle is 0°.
2. Let's assume the alignment process would physically move the nozzle after the last vision detection, as a result the part is then physically centered and in the correct angle after alignment. It does not do that in reality, but it does calculate the resulting nozzle location virtually, I hope you see this is equivalent. 
3. We have not yet applied the vision offset, i.e. the part is now centered on the pads. 
4. To achieve that alignment result, the nozzle had to move a certain distance and angle and is now off-center and rotated. We don't care how.
5. We subtract that quirky nozzle location&rotation _from_ the nominal/wanted location&rotation in order to obtain the alignment offsets. That's nicely seen below, and though this is for pre-rotate, it also logically applies for non-pre-rotate if the placement angle is 0° (see 1.).
   https://github.com/openpnp/openpnp/blob/58298c4f1bbb0ffd0324c47e8bf2ad85782ae29e/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java#L215
6. Then we need to shift the center of the part, because we have the part still only centered on the pads (see 3.)
7. Because we have the part aligned already i.e. it is already rotated nicely to the correct 0° angle, the vision offset can simply be added to the nozzle location, to move the part to its correctly centered position in the camera view. This is a strictly parallel shift, no additional rotation needed! I hope you agree. 
8. We could now repeat step 5 to obtain the _final_ offsets. 
9. But instead of moving the nozzle, we can just subtract the same (straight) vision offset directly from the offsets, as obviously this is mathematically equal (see 5.)
10. Finally, in the placement (job processor) the stored offsets are in turn subtracted from the wanted location&rotation. Again, this is for pre-rotate but equivalent when placement angle = 0°.
    https://github.com/openpnp/openpnp/blob/5b6b43450387e4a7c2978c170adf00a428ca3a88/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L977-L978

I explained it for pre-rotate, because as you see, it's actually simpler, especially for step 10, because the offsets are already final and need not be rotated. 

As a formula and always including rotation, we have:
`offsets = wantedLocation@cam - alignedLocation - visionOffset`
Then in placement:
`placementLocation = wantedLocation@pcb - offsets`
which is equivalent to:
`placementLocation = wantedLocation@pcb - (wantedLocation@cam - alignedLocation -  visionOffset)`
which is equivalent to:
`placementLocation = wantedLocation@pcb - wantedLocation@cam + alignedLocation + visionOffset`

As you see **the minus-minus cancels itself out** and the `visionOffset` goes into the final PCB location result without any further rotation or whatever. So if you were to rotate your visual offsets by the alignment offset angle, you would effectively apply it to the PCB location, which I hope you can see is wrong.

If this does not convince you, I strongly suggest you test it with bad picks. 

Prove me wrong, I will gladly admit it, it would not be the first time! ;-)

_Mark

doppelgrau:
1. perfect, makes life simple :) Would also add "vision passes" to 1, to make pre/post more equal (post has no loop)
2. The resulting virtual nozzle location is stored in the offset (relative to the nozzle). But lets assume we move the nozzle.
3. yes
4. yes, for example lets assume vision offset (1,1), bottom camera position (100,100) and pick offset (-1,0,45°) so the nozzle would have moved to center to the vision result x = 0 (pick offset and vision offsets equalizes), y=1 (vision offset) and rotates -45° (remove rotation) => location of the nozzle (100,101,-45°)
5. yes, with wanted location = nozzle location at the start of the bottom vision => with my numbers the result would be (0,-1,45°)
6. yes
7. yes, result would be nozzle location (101,102,-45°)
8. ok, (-1, -2, 45°)

...
I found my logic hole, didn't take into account, the "off center" rotation.
In this case my fault, but yes, I can share the "headache-problem" with bottom vison. Thanks for your patience to guide me seeing my fault.

markmaker:
Cool, @doppelgrau 

Btw, the `angleNorm()` there is not needed. Trig functions work with any angle.

I guess this is still open:

> @vonnieda must say regarding added UI complexity and versus his visual training alternative:
> https://youtu.be/zyXNehJzrYk

If I understand the original visual training correctly, it could still be added later and actually set this new visual offset permanently on the part instead of adjusting the one placement (if this is what it did). 

_Mark

doppelgrau:
Yes, he mentioned once on discord, that he liked the idea of the offsets, but if he'll like this implementation ...

markmaker:
Hi @doppelgrau,
I just realized that what I said earlier is not true. Imperfect picks _can_ be simulated using the error offsets on the camera.

Error Offsets

See also:
https://github.com/openpnp/openpnp/blob/46a32acb453dc9a3f5e15ec88f949582ccf29570/src/main/java/org/openpnp/machine/reference/camera/SimulatedUpCamera.java#L266

So you _could_ actually add those to your Unit Tests, like this one:
https://github.com/openpnp/openpnp/blob/46a32acb453dc9a3f5e15ec88f949582ccf29570/src/test/java/ReferenceBottomVisionTest.java#L50

_Mark

P.S. Sorry about creating conflicts through #1203. That's just inevitable with Swing's non-incremental layout system.

vonnieda:
@doppelgrau This looks good to me. I don't have time at the moment to do a thorough review, but since @markmaker has already done so I think we should merge this. I see there is a conflict. If you'll please resolve the conflict I'll hit merge as soon as you do. Thanks!

Jason

doppelgrau:
@markmaker I tried adding tests with error. But with pre rotation the test always fails.

`    public void testAsymetricPartWithOffsetWithPreRotationWithError() throws Exception {

        Machine machine = Configuration.get().getMachine();
        ReferenceBottomVision bottomVision = (ReferenceBottomVision) machine.getPartAlignments().get(0);
        Nozzle nozzle = machine.getDefaultHead().getDefaultNozzle();
       
        Part part = Configuration.get().getPart("DoubleAsymPart"); 
        PartSettings partSettings = bottomVision.getPartSettings(part);
        
        BoardLocation boardLocation = new BoardLocation(new Board());
        boardLocation.setLocation(new Location(LengthUnit.Millimeters, 0, 0, -10, 0));
        boardLocation.setSide(Side.Top);

        // 1mm offset in both directions
        partSettings.setVisionOffset(new Location(LengthUnit.Millimeters, 0.5, -0.5, 0.0, 0.0));
        
        // No pre rotate
        partSettings.setPreRotateUsage(PreRotateUsage.AlwaysOn);
        
        // test data
        Location[][] testData = {
                {new Location(LengthUnit.Millimeters, 10.0, 10.0, 0.0, 000.0), new Location(LengthUnit.Millimeters, 0.0001, 0.0, 0.0, 0.0)},
                {new Location(LengthUnit.Millimeters, 10.0, 10.0, 0.0, 045.0), new Location(LengthUnit.Millimeters, 0.0002, 0.0, 0.0, 0.0)},
                {new Location(LengthUnit.Millimeters, 10.0, 10.0, 0.0, 090.0), new Location(LengthUnit.Millimeters, 0.0003, 0.0, 0.0, 0.0)},
                {new Location(LengthUnit.Millimeters, 10.0, 10.0, 0.0, 180.0), new Location(LengthUnit.Millimeters, 0.0003, 0.0, 0.0, 0.0)}
        };
        
        Location error = new Location(LengthUnit.Millimeters, -0.6, 0.7, 0.0, 0.0);
        SimulatedUpCamera camera = (SimulatedUpCamera) VisionUtils.getBottomVisionCamera();
        camera.setErrorOffsets(error);

        machine.setEnabled(true);
        Location maxError = new Location(LengthUnit.Millimeters, 0.1, 0.1, 0, 0.03);
        for (Location[] testPair: testData)
            machine.execute(() -> {
                nozzle.pick(part);
                PartAlignmentOffset offset = bottomVision.findOffsets(part, boardLocation, testPair[0], nozzle);
                Location offsets = offset.getLocation();
                assertMaxDelta(offsets.getX(), testPair[1].addWithRotation(error).getX(), maxError.getX());
                assertMaxDelta(offsets.getY(), testPair[1].addWithRotation(error).getY(), maxError.getY());
                assertMaxDelta(offsets.getRotation(), testPair[1].addWithRotation(error).getRotation(), maxError.getRotation());
                return true;
            });
    }
`
But not sure if a problem in my code or the simulated cam. See the attached pictures. Camera offsets counter the part offsets. (0.5mm/-0.5mm part offsets, camera offsets -0.5mm/0.5mm, 22.5°). Without the rotation everything if fine, but once I add rotation the result is off center. (in the example center is to much up and right)
Since we agreed the code should work right, I'm confused why the test failed (same for post rotate works fine).

only moved above bottom camera
after bottom vision

markmaker:
Hi @doppelgrau 

This always seems to happen, no matter how hard one thinks about such a problem. ;-) 

Did you test this interactively too, on the simulation machine? It often helps understand it. 

Some hints:

1. Assuming these are "debug write" images: Make sure the image _would_ be centered at that moment at all, **I don't think so!** This is the last detected image and AFAIK the last pass' correction and the visual offsets would **not yet have been applied**. So it is perfectly normal to not be centered yet. The second image seems also not to be centered, if you look closely, and again, that's fine, because it is the not yet finally aligned shot. 
2. Check if the test actually accounts for the `error` correctly. In the simulation camera, the `error` is applied at nozzle 0° degrees, but you seem to apply it at whatever the nozzle rotation is _after_ alignment. I guess this will be slightly more complicated, i.e. you will have to rotate the `error` back by the nozzle angle (which this time is +/- placement angle +/- offsets angle, I never get the signs right).

Once you figured it out, also test with the `error` rotation != 0.

You might want to choose an `error` that is not almost the same as the negative visual offsets. So you will see any mistakes better, in the debug write images. 

And perhaps make the vision offset asymmetric (not both at 0.5), so you would catch any coordinate mixup (them having different signs is not safe with the pixel-to-machine coordinate system Y flip in play). 

Also note that this comment does not match the code:

```
    // No pre rotate
    partSettings.setPreRotateUsage(PreRotateUsage.AlwaysOn);

```   

_Mark

doppelgrau:
Yes, after the test failed again and again I started to test it interactively, and those pictures resulted.
Same problem with an offset of 1.5mm/1.5mm.

Bottom_2021-05-27_18 51 34 330
Bottom_2021-05-27_18 51 22 763

Stupid question: Since the part is not symmetrical, the rotation in the simulation results in a larger offset that expected? (in my example the bottom right part moves further away compared to the center of the part ...

So if you do not disagree, I'd say I "only" need to adjust the test-cases, not the code (and fix the copy & paste error in the comment)

doppelgrau:
Update: the image was with "prerotate" off, assumed wrongly my defaults in that configuration. So codepath works fine, but since the rotation changes the offset I need to fix the tests

doppelgrau:
ok, after some tests, in addition to changing the test case, I also need so rotate the part-offset in the post-rotate case?!
At least interactive that eliminated the problems with the wrong centering.

But know I'm back on the headache train. Guess I'll take a look on the test cases later or tomorrow.

markmaker:
I only ever thought it through with pre-rotate. Post-rotate is just unnecessarily complicated. I guess you have to create a formula like I did [here](https://github.com/openpnp/openpnp/pull/1196#issuecomment-841865695) for pre-rotate. Or test it with a ton of angles and large error offsets until you're confident. It's all linear math, so once you nailed a few representative cases, you can be confident it is correct in general. 

Frankly, I would remove post-rotate from OpenPnP if it was for me to decide. :-)

_Mark

doppelgrau:
I have one use case where post rotate is needed for me: some ICs are so large, they only fit in one direction into my camera :)

markmaker:
> 
> 
> I have one use case where post rotate is needed for me: some ICs are so large, they only fit in one direction into my camera :)

Yeah, but slightly larger and your stuck. We better create alternative alignment classes that can do multi-shot alignment on corners of the IC. 

_Mark

doppelgrau:
Yes, corner detection sound like a great idea.
And I think I fixed the test-cases.

markmaker:
Hi @doppelgrau 

I'm unsure what the status of this is. Looking at the code it seems all issues are addressed?

> And I think I fixed the test-cases.

You think or you are sure? ;-) 

Should I merge this?

_Mark

doppelgrau:
I'm quite confident, but not yet tested on a real world pcb.
If you want "real world tested" wait, but in the simulator and the test cases it looks really nice now.

markmaker:
I just rechecked in the code: this does not change behavior, if the vision offsets are zero. Therefore, it is safe to merge, even if not yet tested on a real machine.

Thanks, @doppelgrau!!

doppelgrau:
Hi Mark,

if I remember correctly that was after some discussions on the discord.
I think that was necessary since the offset was somewhere else just added, so to get the real center you neded to move it in the other direction.
But if the logic is changed in both places...

Hope that helps, if not enough, I can take a closer look in the next days.

PR #1198, opened 2021-05-15
Fixed duplicate property panel creation in some circumstances

Jnesselr:
For the configuration wizard property sheet wrapper, it'd return the same instance every time getPropertySheetPanel was called. However, in my case, I was returning "new X" every time which was causing some really strange behavior because one instance had all the bindings and the other was what was displayed on the screen.


markmaker:
Makes sense, thanks.

Issue #1199, opened 2021-05-20
Log Tab is replaced by Issues & Solutions


fxframes:
# Read This First
To submit an issue for OpenPnP you must use this template or it will be deleted. Please fill out either the Problem Report or Feature Request section below and delete the other section.

-----------------------------------------------------------------------

# Problem Report
## Describe the Issue
When accessing the Issues & Solutions tab, the Log tab right next to it also changes name to "Issues & Solutions".
Please see image below.

## Steps to Reproduce
How can a developer reproduce the issue you are experiencing? Use a step by step format like this:
1. Start OpenPnP
2. Go to the Issues & Solutions tab.

## Expected Result
The Log tab should have retained its name.

## Actual Result
The Log tab's name has been replaced.

## Notes, Log Files, Screen Captures, Videos, etc. to Show the Issue

<img width="561" alt="Screenshot 2021-05-20 at 18 49 50" src="https://user-images.githubusercontent.com/68516637/119026284-e3a1ca80-b99c-11eb-943e-2974b8f43cfd.png">

Whenever possible, please use the Menu -> Help -> Submit Diagnostics feature and include the URL here. Submit Diagnostics makes it easy to include log files, screen capture, debug images, configuration files, etc.

-----------------------------------------------------------------------

# Feature Request
## Describe the Feature
What do you want it to do? Be as descriptive as possible.

## How Will It Look?
If the feature will include changes to the user interface, how would they look? Include drawings or mockups if possible.

## Why Do We Need It?
1. What problem does it solve?
2. Is it useful for everyone, or does it only solve a problem for your machine?

## Provide Examples
Is this something that exists in other pick and place software? Provide examples and references if possible.

markmaker:
Hi @fxframes,

Thanks for reporting this. 

I hope I fixed it, see #1201. 

**Note:** As I cannot reproduce the issue, this may or may not work.

Can you please upgrade and confirm?

_Mark

fxframes:
Hi @markmaker,

Thank you this has been fixed.

fxframes:
This issue is back after the latest update.

markmaker:
Could you give me the version (from Help/About)?

markmaker:
I just checked the source and the fix is in. 

Just to be sure: you did not download the Testing version, right? 

The `test` branch is behind.

_Mark

fxframes:
I updated it from within the program, I got the (weekly, in my case) update notice.

[...]

PR #1201, opened 2021-05-21
Fix/

markmaker:
# Description
Addresses #1199.

grafik

Not beeing able to reproduce the error, this is only an attempt to solve the issue.

Using the special `JTabbedPane.indexOfComponent(Component)` method, instead of `indexOf()` on `getComponents()`.

# Justification
See #1199 

# Instructions for Use
No change.

# Implementation Details
1. No conclusive test possible (could not reproduce the error before).


markmaker:
Confirmed fixed in #1199.

PR #1202, opened 2021-05-22
Fixing empty location in location buttons panel

Jnesselr:
# Description
If the text field for a location is empty, it'll return a null value when you parse the length. This fixes that.

# Justification
See Description

# Instructions for Use
This should have no impact the user has to worry about. It essentially treats an empty text field as a 0 for the location.

# Implementation Details
Manual testing.
Yes
Nope, just the GUI.
👍


vonnieda:
Looks good to me, thanks @Jnesselr!

PR #1205, opened 2021-05-27
Fix/ContactProbeNozzle Height Maps must be persisted with IDs, not Objects

markmaker:
# Description
This refers to automatic feeder and part height probing using the ContactProbeNozzle, see #1174 and the Wiki:

Bug: The ContactProbeNozzle persisted height maps with objects as map keys instead of their IDs. 

# Justification
Plain stupid bug.

# Instructions for Use
If users used the ContactProbeNozzle and have in-deed probed heights on parts and feeders, their `machine.xml` structure is wrong. It will look similar to this:

```XML
                  <probed-feeder-height-offsets>
                     <entry>
                        <feeder class="org.openpnp.machine.reference.feeder.ReferenceTubeFeeder" version="1.1" id="FDR1682b249cd43b798" name="ReferenceTubeFeeder" enabled="false" part-id="TEST-OBJECT" feed-retry-count="3" pick-retry-count="0">
                           <location units="Millimeters" x="253.39191389757997" y="260.32085144101774" z="-26.60000000000001" rotation="150.0"/>
                        </feeder>
                        <length value="-0.16299999999999315" units="Millimeters"/>
                     </entry>
                     ...
                  </probed-feeder-height-offsets>
```

```XML
                  <probed-part-height-offsets>
                        <part id="TEST-OBJECT" height-units="Millimeters" height="0.01" package-id="TEST-OBJECT-PACKAGE" speed="1.0" pick-retry-count="0"/>
                        <length value="-0.1929999999999943" units="Millimeters"/>
                     </entry>
                     ...
                  </probed-part-height-offsets>
```
After upgrading to this version they will get an error like this:
XML Error

Unfortunately, the user must remove these sections from the `machine.xml` manually. The height calibration information will be lost, but it will just be probed again on next use.

# Implementation Details
1. Testing on machine pending. 

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ContactProbeNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ContactProbeNozzle.java
@@ -125,10 +125,10 @@ public class ContactProbeNozzle extends ReferenceNozzle {
     private double maxZOffsetMm = 2.0;
 
     @ElementMap(required = false)
-    private HashMap<Feeder, Length> probedFeederHeightOffsets = new HashMap<>();
+    private HashMap<String, Length> probedFeederHeightOffsets = new HashMap<>();
 
     @ElementMap(required = false)
-    private HashMap<Part, Length> probedPartHeightOffsets = new HashMap<>();
+    private HashMap<String, Length> probedPartHeightOffsets = new HashMap<>();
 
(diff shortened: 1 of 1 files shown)


vonnieda:
@markmaker I think this is probably fine as it should be low impact. Feel free to merge if you are ready.

PR #1208, opened 2021-06-02
Bugfix: Only call "partOff" if configured during recyling

doppelgrau:
# Description
It was wrongfully assumed, that partOff() always works. (Since Simulator and test machine are equipped with vacuum sensor it didn't show during tests). This results in exceptions after recycling the part as on the mailing list reported.

# Justification
Fixed the behavior to the expected-

# Instructions for Use
No change.

# Implementation Details
1. How did you test the change? Tested in simulator. No feedback (yet) from the reporter of the problem.
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. no changes, only in the feeders


markmaker:
Looks good, thanks @doppelgrau!

PR #1210, opened 2021-06-04
Feature/blinds feeder named group 

crashmatt:
# Description
Named group option for BlindsFeeder which acts in addition to fiducial location to determine connected feeders.

# Justification
If multiple tiles of Blindfeeders are used in the same physical location for different jobs/builds the feeders configuration will be confused.
This benefits everyone using BlindsFeeders a significant amount.  Default behavior is by location.

# Instructions for Use

- In the BlindsFeeder tape settings there is a new editable combo box. Either choose an existing group name or enter a new one to override location based grouping.
- Leave the group name as "Default" to get the normal location based behavior.
- Entering an existing group name for a feeder with default fiducial locations or a feeder with no other connected feeders will join it to that group and give it the correct fiducials.
- Each group name can only be used for a single, co-located feeder group. Entering an existing group name for a group of more than one feeders will be rejected.
- A future suggestion for arranging multiple feeder groups on a common tile is TileName:FeederGroupName.  Please do not use the colon delimiter.
- Setting fiducial1 to the same location as an existing group will not automatically join it with the group.  This protects the name group from having uncontrolled feeders join it from co-located feeder groups.
- Renaming a feeder in a group will rename all of the connected feeders.  It will not separate that feeder from the others.
- Some keywords are treated as default.  These include Location, default, none and "".  Entering any of the location keywords results in reset back to "Default"
- A single feeder can leave a group and be assigned back to "Default".  Any feeders connected to it will be disconnected.

# Implementation Details
Created BlindsFeederTest which performs a series of  BlindsFeeder behavior tests.  The expected fiducials, group name, feeder number and connected feeders are checked for all feeders at every step. 

Basic behavior test including:
  1. Create feeders, set fiducials and check
  2. Connect second feeders using fiducial1
  2. Create third feeder at another location and check that it does not connect to the previous group
  3. Test setting feeder pocket locations operates as expected
  4. Test moving feeder1 to check that conencted feeder2 moves with it.

and a set of group naming tests

1. Create a feeder, set fiducials and change its group name
2. Check that a second feeder set with the same fiducial1 does not join until its group name is set correctly.
3. Check that a feeder at null location can join a group and have its fiducials set correctly
4. Check that a feeder at defined location but no other connected feeders can join a group and be moved to the correct fiducials
5. Check that a group of more than one feeder can't be assigned another existing group name, either at the same or a different location.

Additional tests:
- Feeder group names are persistent through machine.xml
- Feeder group name dropdown contains all of the used feeder group names. Removing all isntances of a group name results in it being excluded from this list when the feeder configuration view is refreshed.

2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?  Yes?
3. No changes to `org.openpnp.spi` or `org.openpnp.model` packages 
4. `mvn test` run and passed

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java
@@ -171,7 +171,8 @@ public class BlindsFeeder extends ReferenceFeeder {
     private boolean calibrating = false;
     private boolean calibrated = false;
 
-    private static final List<String> noGroupNamesList = Arrays.asList(new String[]{"Location", "LOCATION", "location", "", "None", "none", "NONE"});
+    private static final String defaultGroupName = "Location";
+    private static final List<String> noGroupNamesList = Arrays.asList(new String[]{defaultGroupName, "LOCATION", "location", "", "None", "none", "NONE"});
     
     private void checkHomedState(Machine machine) {
         if (!machine.isHomed()) {
@@ -1614,6 +1615,20 @@ public class BlindsFeeder extends ReferenceFeeder {
             }
         }
         return false;
+    }
+    
+    public static List<String> getBlindsFeederGroupNames() {
+    	List<String> list = new ArrayList<>();
+        for (Feeder feeder : Configuration.get().getMachine().getFeeders()) {
+            if (feeder instanceof BlindsFeeder) {
+                BlindsFeeder blindsFeeder = (BlindsFeeder) feeder;
+                String feederGroupName = blindsFeeder.getFeederGroupName();
+            	if (!list.contains(feederGroupName)) {
+            		list.add(feederGroupName);
+            	}
+            }
+        }
+        return list;
     }
 
     public static List<BlindsFeeder> getAllBlindsFeeders() {
--- a/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java
@@ -128,7 +128,7 @@ public class BlindsFeeder extends ReferenceFeeder {
     private int feedersTotal = 0;
 
     @Attribute(required = false)
-    private String feederGroupName = "Location";
+    private String feederGroupName = defaultLocationGroupName;
 
     @Attribute(required = false)
     private int pocketCount = 0;
@@ -171,8 +171,8 @@ public class BlindsFeeder extends ReferenceFeeder {
     private boolean calibrating = false;
     private boolean calibrated = false;
 
-    private static final String defaultGroupName = "Location";
-    private static final List<String> noGroupNamesList = Arrays.asList(new String[]{defaultGroupName, "LOCATION", "location", "", "None", "none", "NONE"});
+    private static final String defaultLocationGroupName = "Location";
+    private static final List<String> locationGroupNamesList = Arrays.asList(new String[]{defaultLocationGroupName, "LOCATION", "location", "", "None", "none", "NONE"});
     
     private void checkHomedState(Machine machine) {
         if (!machine.isHomed()) {
@@ -1642,6 +1642,10 @@ public class BlindsFeeder extends ReferenceFeeder {
         }
         return list;
     }
+    
+    public static boolean groupedByLocation(BlindsFeeder blindsFeeder) {
+    	return locationGroupNamesList.contains(blindsFeeder.feederGroupName);
+    }
 
     public static List<BlindsFeeder> getConnectedFeedersByLocation(Location location, boolean fiducial1MatchOnly) {
         // Get all the feeders with connected by location.
@@ -1649,8 +1653,10 @@ public class BlindsFeeder extends ReferenceFeeder {
         for (Feeder feeder : Configuration.get().getMachine().getFeeders()) {
             if (feeder instanceof BlindsFeeder) {
                 BlindsFeeder blindsFeeder = (BlindsFeeder) feeder;
-                if (blindsFeeder.isLocationInFeeder(location, fiducial1MatchOnly)) {
-                    list.add(blindsFeeder);
+                if(groupedByLocation(blindsFeeder)) {
+                    if (blindsFeeder.isLocationInFeeder(location, fiducial1MatchOnly)) {
+                        list.add(blindsFeeder);
+                    }                	
                 }
             }
         }
@@ -1671,7 +1677,7 @@ public class BlindsFeeder extends ReferenceFeeder {
         for (Feeder feeder : Configuration.get().getMachine().getFeeders()) {
             if (feeder instanceof BlindsFeeder) {
                 BlindsFeeder blindsFeeder = (BlindsFeeder) feeder;
-                if (blindsFeeder.feederGroupName == groupName) {
+                if (blindsFeeder.feederGroupName.equals(groupName)) {
                     list.add(blindsFeeder);
                 }
             }
@@ -1697,7 +1703,7 @@ public class BlindsFeeder extends ReferenceFeeder {
 
     
     public List<BlindsFeeder> getConnectedFeeders(Location location, boolean fiducial1MatchOnly) {
-    	if (noGroupNamesList.contains(this.feederGroupName)) {
+    	if (groupedByLocation(this)) {
             // Get all the feeders with the same fiducial 1 location.
             return getConnectedFeedersByLocation(fiducial1Location, true);
     	} else {
@@ -2086,13 +2092,21 @@ public class BlindsFeeder extends ReferenceFeeder {
     }
 
     public String getFeederGroupName() {
+    	if (locationGroupNamesList.contains(feederGroupName)) {
+            return defaultLocationGroupName;
+    	}
         return feederGroupName;
     }
 
     public void setFeederGroupName(String feederGroupName) {
     	String oldName = this.feederGroupName;
-    	this.feederGroupName = feederGroupName;
-        firePropertyChange("feederGroupName", oldName, feederGroupName);
+    	if (locationGroupNamesList.contains(feederGroupName)) {
+        	this.feederGroupName = defaultLocationGroupName;
+            firePropertyChange("feederGroupName", oldName, defaultLocationGroupName);
+    	} else {
+        	this.feederGroupName = feederGroupName;
+            firePropertyChange("feederGroupName", oldName, feederGroupName);
+    	}
     }
 
     public int getFeedCount() {
--- a/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java
@@ -1619,6 +1619,8 @@ public class BlindsFeeder extends ReferenceFeeder {
 
     public static List<String> getBlindsFeederGroupNames() {
         List<String> list = new ArrayList<>();
+        list.add(defaultGroupName);
+        
         for (Feeder feeder : Configuration.get().getMachine().getFeeders()) {
             if (feeder instanceof BlindsFeeder) {
                 BlindsFeeder blindsFeeder = (BlindsFeeder) feeder;
@@ -2111,7 +2113,7 @@ public class BlindsFeeder extends ReferenceFeeder {
         boolean location_is_null = 
                 fiducial1Location.equals(nullLocation) ||
                 fiducial2Location.equals(nullLocation) ||
-                fiducial3Location.equals(nullLocation); 
+                fiducial3Location.equals(nullLocation);
 
         if (location_is_null) {
             // Set fiducials to existing feeder with new group name if one exits
@@ -2119,7 +2121,7 @@ public class BlindsFeeder extends ReferenceFeeder {
                 BlindsFeeder copyFeeder = feedersWithNewGroupName.get(0);
                 this.updateFromConnectedFeeder(copyFeeder);
             } else {
-                // Do not allow null located feeder to get a non default group name
+                // Do not allow null located feeder to get a non-existent group name
                 proposedGroupName = defaultGroupName;
                 this.feederGroupName = proposedGroupName;
             }
@@ -2132,7 +2134,11 @@ public class BlindsFeeder extends ReferenceFeeder {
                     BlindsFeeder copyFeeder = feedersWithNewGroupName.get(0);
                     this.updateFromConnectedFeeder(copyFeeder);
                 } else {
-                    proposedGroupName = oldName;
+                    if (proposedGroupName.contentEquals(defaultGroupName)) {
+                        this.feederGroupName = proposedGroupName;
+                    } else {
+                        proposedGroupName = oldName;                        
+                    }
                 }
             } else {
                 for (BlindsFeeder feeder : connected_feeders) {
--- a/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java
@@ -2093,68 +2093,58 @@ public class BlindsFeeder extends ReferenceFeeder {
     }
 
     public void setFeederGroupName(String newFeederGroupName) {
-        //Filter out a no change to group name.
-        if(this.feederGroupName.equals(newFeederGroupName)) {
+        // Filter out a no change to group name.
+        if (this.feederGroupName.equals(newFeederGroupName)) {
             return;
         }
-        
+
         String oldName = this.feederGroupName;
         String proposedGroupName = newFeederGroupName;
-        
-        //Check if the group name is one of the location keywords. If so reset name to default location.
+
+        // Check if the group name is one of the location keywords. If so reset name to
+        // default location.
         if (locationGroupNamesList.contains(newFeederGroupName.toUpperCase())) {
             proposedGroupName = defaultGroupName;
         }
-        
+
         List<BlindsFeeder> connected_feeders = getConnectedFeeders();
         List<BlindsFeeder> feedersWithNewGroupName = getBlindsFeedersWithGroupName(proposedGroupName);
         List<String> feederGroupNames = getBlindsFeederGroupNames();
-        
-        boolean proposed_is_default = proposedGroupName.contentEquals(defaultGroupName);
-        boolean proposed_is_existing_group = feederGroupNames.contains(proposedGroupName);
-        
-        boolean location_is_null = 
-                fiducial1Location.equals(nullLocation) ||
-                fiducial2Location.equals(nullLocation) ||
-                fiducial3Location.equals(nullLocation);
-        
-        boolean can_join_named_group =  
-                (location_is_null &&  (feedersWithNewGroupName.size() > 0));
 
-        boolean can_rename_group = (
-                !location_is_null 
-                && !proposed_is_default
-                && !proposed_is_existing_group);
+        boolean proposedIsDefault = proposedGroupName.contentEquals(defaultGroupName);
+        boolean proposedIsExistingGroup = feederGroupNames.contains(proposedGroupName);
 
-        boolean single_can_join_named_group = (
-                !location_is_null 
-                && feederGroupNames.contains(proposedGroupName)
-                && (connected_feeders.size() <= 1) );
-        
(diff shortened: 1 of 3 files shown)


markmaker:
Hi @crashmatt

That's a pragmatic, useful design, thanks!

Some thoughts/questions, please take it as constructive criticism :-)

1. If you change the group name of one feeder, it should check for duplicates in currently (under the old name) unconnected feeders and refuse them. Otherwise, you can create a big mess very easily.
2. If you change the group name of one feeder, it should update the group name of all the other currently (under the old name) connected feeders.  
3. How do you activate/inactivate a feeder group? I.e. how do you make sure OpenPnP always knows which group's "layer" is currently loaded? It would be nice if this was somehow handled per group, i.e. the user is not forced to enable/disable all of them manually. Also, ideally, if one group is activated, all the overlapping feeders of other groups are automatically inactivated. I'm using "activated" as a different term/boolean as "enabled", because the two should be different concepts, i.e. it should be possible to have the group activated, but individual feeders still disabled. And if you inactivate and then again activate a group, the individual feeder disabled state should ideally be preserved. 
4. One of the nice features of the BlindsFeeder, is automatically filling out all the common array settings in Auto Setup (starting from the second feeder). All you needed to do is click on a part and all would be done. This was done through location linking. See here:
    https://youtu.be/dGde59Iv6eY?t=512
    I see how you still do that if all the _other_ feeders have no special name assigned, i.e. one of the `locationGroupNamesList`, but this means you can only assign group names _after_ all the feeders in an array have been defined. And you then have to assign the group name to all the feeders manually. It would be nice to be able to create the group with the first feeder and then also automatically assign/clone it to the others when doing Auto Setup. So I think this should (again) work through the feeder being activated/inactivated (see point 3). 
5. There are some whitespace issues. You probably used tabs. 
   grafik


These are my first thoughts. I have to go now, so this might be incomplete. Again, this is well-meant criticism.

_Mark

crashmatt:
Mark,
I was getting pretty desperate for this solution to get a build done.  I
did almost the minimum possible to separate location and named groups.  The
combo box was an extra but made it very much more usable.

I agree that some extra housekeeping functions would help.  I don't see
that they will change the group name data concept.  I suggest the basic
functionality be pulled into upstream and then work on more advanced
functions later.  Then we avoid mission creep and the whole thing going
stale.

Adding extra functions probably means moving  the controls to a separate
feeder group settings area.
[image: image.png]

Jason had an idea that this kind of grouping would be useful for all
feeders types. I have not got the time/energy right now to cope with that.
Keeping the functionality contained in BlindsFeeder will give me more
freedom to experiment.  What is learned can probably be applied to the
other feeders later.

/Matt

markmaker:
Hi @crashmatt 

> I suggest the basic functionality be pulled into upstream and then work on more advanced functions later.  Then we avoid mission creep and the whole thing going stale.

I would like to agree, but this really changes the behavior of the very core of the BlindsFeeder, by affecting the linking between them. With the current drop-down, it would even be possible to assign a feeder to a group that is not linked by location. The result would likely be a very bad "meltdown" of the fused arrays on the next update that is triggered by any property change. The result is unpredictable and can ruin the setup of dozens of feeders. 

I still feel somewhat responsible for that class and its support on the list, so I do care. I know the BlindsFeeder is not currently foolproof either, but I don't like to worsen the situation. The new field is not hidden away enough, so users won't try to use it before they fully understand the caveats. 

If you like to add it as an interim solution, I would ask for it to be non-destructive to current link guarantees. This means you must not _replace_ the location linking with name linking, but only _qualify_ it with the name. In other words, the name match should be AND-ed to the location match.  You then also don't create a mess if the name is not unique, in fact you could then assign the same group name to all the BlindsFeeder arrays on a slate/mountable unit and make it an even more understandable/usable design. Each mountable unit would then form sort of a "layer" by that name. And that concept could later be upgraded into a more universal feeder grouping design that can be useful for other classes/feeders (e.g. mounted on the same slate/unit). 

The name linking must in essence be _orthogonal_ to the location linking and for good reason: The array linking by location is unique for the BlindsFeeder, other feeder classes don't have to comply with comparable constraints. Unless there is a watertight universal solution to replace the one in BlindsFeeder, I'm not happy to mess with it.

I hope you understand. 

_Mark

crashmatt:
Mark,
I had not thought of ANDing location and group name.  I like it.  Will
modify to your suggested method.

/Matt

crashmatt:
Testing on update with location and group:
Renaming a feeder results in all feeders of the same name and location changing.
Renaming a feeder with two groups of feeders with the same name but different locations results in only the intended group changing name.

Yet to be tested:
New feeder with no location
Multiple feeders with no location and different groups.
Auto setup

Also: Thinking about group based enable/disable. I suspect it needs a specific "group" object to store attributes in.  Otherwise the group enable needs to be stored in every feeder.  Then we are heading towards the feeder groups that Jason suggested.

markmaker:
Thanks, @crashmatt!

- I wonder if it were more appropriate now to set `defaultLocationGroupName` to `"None"` as default to reflect the fact they are always Location linked anyway. Or maybe even the empty String like in other reference semantic drop-downs (like Actuators etc.), so the path to real groups as referenced objects would be straight-forward (in the future).
- Maybe add  `defaultLocationGroupName`  to `getBlindsFeederGroupNames()` as the first array entry, so it is always available as a choice, even if all feeders are currently assigned names, so the user can always "unattach" a feeder willingly without having to remember the correct `defaultLocationGroupName`. 
- I think you should use `getFeederGroupName()` in the `filterFeedersByGroupName()` call to use the canonical `defaultLocationGroupName`. 
- Or even better: use the `locationGroupNamesList` substitution in the setter on `newFeederGroupName`, so the `defaultLocationGroupName` is always enforced _as a class invariant_, and you don't have to substitute on each get. If the new default is always presented in the dropdown, maybe you can also just remove the whole substitution thingy (especially if you choose the empty String). 
- Cosmetics: I think the `compareTo()` diff on line 1660 is only a reformatting artifact. 

_Mark

crashmatt:
I am loosing track of test cases.  Time to look at writing a proper test module.

> Maybe add defaultLocationGroupName to getBlindsFeederGroupNames()
Good idea

"newFeederGroupName" - Was not aware of the "new".  Still learning the features of openpnp.  Thanks for the ideas.

Will change the default name to "None"

markmaker:
> "newFeederGroupName" - Was not aware of the "new". Still learning the features of openpnp. 

Don't understand what you mean ... this is code you wrote ;-)

_Mark

crashmatt:
Sorry. I was reading your comment too quick and not understanding enough.

crashmatt:
Added BlindsFeederTest. It is failing a test which moves a feeder with a connected feeder to a new location.  It does this by moving fiducial1 only.  The modified feeder moves as expected.  The connected feeder does not.
Also tested this on the upstream:develop and had the same result.

Simply using setFiducial1Location which is the same action as the gui.

What am I doing wrong?

Other than that BlindsFeederTest is going quite well.

Still not sure what to do when setting a group to a feeder with no location.  If there are multiple feeder locations with the same feeder group name then we don't know which to use.  Maybe best not to allow a name to be assigned to multiple sets of connected feeders.

crashmatt:
Did something wrong when I used BlindsFeederTest on upstream:develop.  Now it works.

crashmatt:
@markmaker Ready for review again.  Testing is definitely more thorough than it was before.  Have updated the instructions according to the new expected behavior. Any comments?

markmaker:
Hi @crashmatt,

Thanks for the unit test, that's a great addition. 

Some questions/comments (numbers for reference). I'm a bit pressed for time (gotta go) so this might contain mistakes and not sound very diplomatic. Rest assured, it is all well-meant/best effort!

1. I'm still missing/not understanding how the Auto Setup feature is supposed to work? Scenario: Users start to define a new array, they assign a new group name to the first feeder. Now they want to set up the second feeder with the camera on the tape pocket centerline and pressing Auto Setup, as it was possible before. It should now recognize that the camera location is inside the fiducial area of the first feeder, clone the group name along with all the other settings, and recognize the rest using vision. But if there are multiple feeder arrays overlapping the camera location, which one to take for the cloning? By going sequentially through the global feeder list, it will currently take the oldest feeder, but that's almost certainly not what the users want, they most probably want the _newest_ or some better criteria. 
2. If I understand correctly, you can assign a feeder to a different array _if_ it has a location and is single. But it will take just _any_ feeder with that group name and overwrite the very location it just tested before, even if the other feeder array is located "miles" away. See below.
3. Be aware of migration: There are users out there that have multiple arrays already defined. They will all get the default group on upgrade, but with different location linking. This will violate _"Each group name can only be used for a single, co-located feeder group."_ 
4. Imagine what happens, if a user now sets a group name, and reconsiders and sets it back to a default name. Ouch! 
5. I'm unsure if your setter is entirely watertight when silently falling back to oldName (initially the default name). This might again violate the rule _"Each group name can only be used for a single, co-located feeder group"_, not sure though.
6. Default name substitution should be done _before_ testing if the name changes. Otherwise, you can provoke an artificial group reassignment. 
7. What I'm trying to say with the preceding points (and what I said before, when I said it must be "orthogonal"): don't use the group name as the array, **ever!**. **Always** use it as an AND criteria in addition to the location. And instead allow multiple arrays to share the same group name, i.e. your rule _"Each group name can only be used for a single, co-located feeder group"_ should be abolished instead.  
8. Also: I'm still missing the "Default" name always added to the available groups to unassign a feeder from any group. 

Maybe I'm missing something in all this, then just explain, thanks. 

Gotta go.

_Mark

crashmatt:
Mark,
Writing that unit test was strangely satisfying.   Feels good to minimize
the chances of messing up your good work.

Thanks for taking your time to comment.


   1. I'm still missing/not understanding how the Auto Setup feature is
   supposed to work? Scenario: Users start to define a new array, they assign
   a new group name to the first feeder. Now they want to set up the second
   feeder with the camera on the tape pocket centerline and pressing Auto
   Setup, as it was possible before. It should now recognize that the camera
   location is inside the fiducial area of the first feeder, clone the group
   name along with all the other settings, and recognize the rest using
   vision. But if there are multiple feeder arrays overlapping the camera
   location, which one to take for the cloning? By going sequentially through
   the global feeder list, it will currently take the oldest feeder, but
   that's almost certainly not what the users want, they most probably want
   the *newest* or some better criteria.

* I don't see a good way to resolve this so it has the same behavior as
before for feeders with no assigned group name.*

   1. If I understand correctly, you can assign a feeder to a different
   array *if* it has a location and is single. But it will take just
*any* feeder
   with that group name and overwrite the very location it just tested before,
   even if the other feeder array is located "miles" away. See below.

*Might make more sense to bring the group to the new location rather than
the other way around.  I have no preference.   *

   1. Be aware of migration: There are users out there that have multiple
   arrays already defined. They will all get the default group on upgrade, but
   with different location linking. This will violate *"Each group name can
   only be used for a single, co-located feeder group."*

*Should read  "Each non default group name can only be used for a single,
co-located feeder group."*
*and it needs a test*

   1. Imagine what happens, if a user now sets a group name, and
   reconsiders and sets it back to a default name. Ouch!

* Needs a test.*

   1. I'm unsure if your setter is entirely watertight when silently
   falling back to oldName (initially the default name). This might again
   violate the rule *"Each group name can only be used for a single,
   co-located feeder group"*, not sure though.
   2. Default name substitution should be done *before* testing if the name
   changes. Otherwise, you can provoke an artificial group reassignment.

Not sure where this problem is yet. Maybe I can catch it with more test
cases.

   1. What I'm trying to say with the preceding points (and what I said
   before, when I said it must be "orthogonal"): don't use the group name as
   the array, *ever!*. *Always* use it as an AND criteria in addition to
   the location. And instead allow multiple arrays to share the same group
   name, i.e. your rule *"Each group name can only be used for a single,
   co-located feeder group"* should be abolished instead.

*This is the only point I don't agree with. I see a problem with shifting a
tile of multiple feeder sets to a different location. Potential problem
with two sets of feeders of the same name joining together. I don't have a
clear idea of how to resolve this.  That and the additional problem of
assigning a feeder a group name and not knowing which one of the group
locations to join it to.*
*If a good solution is found then joining multiple feeder sets under one
group is easier than trying to separate them later.*

*Regarding the group name setting.  Maybe I can make my implementation
cleaner.  Will think on this.*

   1. Also: I'm still missing the "Default" name always added to the
   available groups to unassign a feeder from any group.

*Agreed.  Will fix.*

/Matt

On Sat, 12 Jun 2021 at 11:44, markmaker ***@***.***> wrote:

> Hi @crashmatt <https://github.com/crashmatt>,
>
> Thanks for the unit test, that's a great addition.
>
> Some questions/comments (numbers for reference). I'm a bit pressed for
> time (gotta go) so this might contain mistakes and not sound very
> diplomatic. Rest assured, it is all well-meant/best effort!
>
>    1. I'm still missing/not understanding how the Auto Setup feature is
>    supposed to work? Scenario: Users start to define a new array, they assign
>    a new group name to the first feeder. Now they want to set up the second
>    feeder with the camera on the tape pocket centerline and pressing Auto
>    Setup, as it was possible before. It should now recognize that the camera
>    location is inside the fiducial area of the first feeder, clone the group
>    name along with all the other settings, and recognize the rest using
>    vision. But if there are multiple feeder arrays overlapping the camera
>    location, which one to take for the cloning? By going sequentially through
>    the global feeder list, it will currently take the oldest feeder, but
>    that's almost certainly not what the users want, they most probably want
>    the *newest* or some better criteria.
>
> * I don't see a good way to resolve this so it has the same behavior as
before for feeders with no assigned group name.*

>
>    1. If I understand correctly, you can assign a feeder to a different
>    array *if* it has a location and is single. But it will take just *any*
>    feeder with that group name and overwrite the very location it just tested
>    before, even if the other feeder array is located "miles" away. See below.
>
> *Might make more sense to bring the group to the new location rather than
the other way around.  I have no preference.   *

>
>    1. Be aware of migration: There are users out there that have multiple
>    arrays already defined. They will all get the default group on upgrade, but
>    with different location linking. This will violate *"Each group name
>    can only be used for a single, co-located feeder group."*
>
> *Should read  "Each non default group name can only be used for a single,
co-located feeder group."*
*and it needs a test*

>
>    1. Imagine what happens, if a user now sets a group name, and
>    reconsiders and sets it back to a default name. Ouch!
>
>  Needs a test.

>
>    1. I'm unsure if your setter is entirely watertight when silently
>    falling back to oldName (initially the default name). This might again
>    violate the rule *"Each group name can only be used for a single,
>    co-located feeder group"*, not sure though.
>    2. Default name substitution should be done *before* testing if the
>    name changes. Otherwise, you can provoke an artificial group reassignment.
>
> *Not sure where this problem is yet. Maybe I can catch it with more test
cases.*

>
>    1. What I'm trying to say with the preceding points (and what I said
>    before, when I said it must be "orthogonal"): don't use the group name as
>    the array, *ever!*. *Always* use it as an AND criteria in addition to
>    the location. And instead allow multiple arrays to share the same group
>    name, i.e. your rule *"Each group name can only be used for a single,
>    co-located feeder group"* should be abolished instead.
>
> *This is the only point I don't agree with. I see a problem with shifting
a tile of multiple feeder sets to a different location. Potential problem
with two sets of feeders of the same name joining together. I don't have a
clear idea of how to resolve this. *
*If a good solution is found then joining multiple feeder sets under one
group is easier than trying to separate them later.*

*Maybe I can make my implementation cleaner.  Will think on this.*

crashmatt:
New idea for group names using format

**TileGroupName:FeederGroupName**

Feeders groups get individual names with the right context and tiles (groups of groups) can be moved/enabled together.  This scheme still enforces individual naming per feeder group.

markmaker:
> *This is the only point I don't agree with. I see a problem with shifting a
tile of multiple feeder sets to a different location. Potential problem
with two sets of feeders of the same name joining together.

You're absolutely right, I was wrong. There is no facility to shift the whole group, only one array at a time. A true feeder slate/mountable unit _object_ would be the right solution, carrying a base location with the feeders only giving the relative location. This is missing, and it shows.

I guess within this simplified (interim) solution using just names, your initial idea to make the groups unique is unavoidable. But it should _not_ apply to the default group name.  By leaving/assigning the default group name, the user is responsible that no location overlap happens.

Thanks for your patience ;-)

_Mark

crashmatt:
One more thing.  Changing a feeder back to default group name only changes that feeder, not the connected feeders.  This enables users to correct unwanted assigned feeders without accidentally breaking the whole group.
I don't see that changing a complete named group back to default will be required often.

markmaker:
> Changing a feeder back to default group name only changes that feeder, not the connected feeders. This enables users to correct unwanted assigned feeders without accidentally breaking the whole group.
I don't see that changing a complete named group back to default will be required often.

Having to unassign feeders one by one is OK. But you need to handle the default group name specially. Currently, if the feeder to be unassigned is a member of an array `(connected_feeders.size() >1)` and as long as there are _other_ arrays present under the default name (after migration perhaps), then this change will be currently rejected. This also happens, once you always add the default name to the list of groups (as you should).

markmaker:
Btw. you wrote regarding Auto Setup:

> I don't see a good way to resolve this so it has the same behavior as
before for feeders with no assigned group name.*

I think it must handle this, otherwise you won't be able to use it once there are co-located feeders defined on other slates/mountable units. 

But it's probably fine to just take the _last_ feeder it finds here: 

https://github.com/openpnp/openpnp/blob/c9640fa186040b88d617fb3fab80a46880437533/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java#L1747

Something like this (written off the top of my head):

```
public boolean updateFromConnectedFeeder(Location location, boolean fiducial1MatchOnly) {
        BlindsFeeder match = null;
        List<Feeder> machineFeeders = Configuration.get().getMachine().getFeeders();
        int latestIndex = -1;
        for (BlindsFeeder feeder : getConnectedFeeders(location, fiducial1MatchOnly)) {
            if (feeder != this) {
                // Get the index of the feeder in the global machine list, knowing this reflects the order 
                // in which feeders were created.
                int index = machineFeeders.indexOf(feeder);
                // Make sure we get the latest.
                if (index > latestIndex) {
                    match = feeder;
                    latestIndex = index;
                }
            }
        }
        if (match != null) {
            updateFromConnectedFeeder(match);
        }
```

_Mark

crashmatt:
About auto setup.  I wish to have well controlled behavior around joining and or leaving a named group.  Otherwise there is a good chance of running into the same mess that the group naming is trying to avoid.  Having auto setup pick the last group at a particular location does not lead to a well defined behavior for the user.
For now, having to explicitly join a group before auto setup is ok by me.  In future there may be a popup list of feeder groups to join at that location or some other selection option.  This is getting beyond the scope of what I can cope with for this PR.

Have tidied up setFeederGroupName so it is easier to read.  Added ability to change feeder back to Default.

markmaker:
> I wish to have well controlled behavior around joining and or leaving a named group. ... For now, having to explicitly join a group before auto setup is ok by me. 

OK, I can follow your argument. 

Just some cosmetics: You've started to use [snake case](https://en.wikipedia.org/wiki/Snake_case) for identifiers (`proposed_is_default` etc). Please see the [OpenPnP coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style) and [specifically here](https://google.github.io/styleguide/javaguide.html#s5.2.7-local-variable-names). 

@vonnieda, I'm a bit confused, I actually thought `mvn test` would reject that (via the maven-checkstyle-plugin)? 

_Mark

crashmatt:
I was treating Booleans as a variable underclass not worthy of capitalization.  Now we have Equality for all variables so I ought to treat them with the respect they deserve.  Have submitted a fix.

Did not find a way for eclipse to enforce camelCase or a way to run maven formatting check.

markmaker:
Cool. But now you've started...

`connected_feeders`, `single_can_join_named_group`, `can_leave_group_for_default`

;-)

vonnieda:
I saw I was pinged: I'm not sure why the Maven Checkstyle didn't catch
that. I'll check on it.

The Java programming convention is to use camelCase regardless of the
project. It's a language thing, rather than a project thing.

Jason

markmaker:
Thanks, @crashmatt 
As far I'm able to tell, this is now ready, right? 
_Mark

crashmatt:
I don't have any more changes to do.

/Matt

markmaker:
Thanks @crashmatt !

PR #1211, opened 2021-06-07
Jnesselr/index feeder

Jnesselr:
# Description
The Index PnP project is a project by Stephen Hawes (@sphawes) to bring a more affordable Pick and Place machine to the world. It's open source (Under the @index-machines organization) and there's even a pretty decent YouTube playlist about it (https://www.youtube.com/playlist?list=PLIeJXmcg1baLBz3x0nCDqkYpKs2IWGHk4)

This PR is the first PR in the org.openpnp.machine.index package, so I wanted to give a little context first.

# Justification
This is specific to the index machines but my hope is that other machines will use this concept and protocol too.

# Instructions for Use
Right now the feeder is disabled in OpenPnP but when we enable it, this wiki page will be fully fleshed out: https://github.com/openpnp/openpnp/wiki/Index-Feeder

# Implementation Details

This feature has more unit tests than the rest of the project. There are 98 tests for the index feeder. I also tested it with a virtual python program that I wrote while using the GUI.

:+1:

Nope, should be all features in the machine category.

Should be good there :)


vonnieda:
This looks good to me. Since the vast majority of the work is in the Index specific classes and the only changes to core are cosmetic I'd say we can merge this. @markmaker - any concerns?

markmaker:
> @markmaker - any concerns?

As far as I can tell, this looks like very high quality code. 

From an admittedly cursory glance, I was wondering some things, though. These are mostly design decision questions, and I'm deliberately wearing the [Advocatus Diaboli](https://en.wikipedia.org/wiki/Devil%27s_advocate) hat here, i.e. it is all well-meant and just intended to poke you into defending your design (and as a result: us knowing why you did things the way you did them).  😁

1. Why is the packet encoding/decoding, feeder scanning, etc. not on the controller side and then exposed as more or less plain (textual) G-code by the controller? This would allow...
2. User manual testing inside/outside OpenPnP using human-readable "command-line" G-code.
3. Keeping the firmware/packet format versioning problem at least constrained to the electronics side (controller & feeders) and not extending it to OpenPnP versioning (don't tell me there will never be a revision to the packet format 🤓).
4. Keeping the high-level G-code stable/backwards compatible and/or easily configurable in OpenPnP if just the syntax were to change.
5. Controlling them with plain Actuators.
6. Finally, in the spirit of Open Source versatility: Using of the feeders from other PnP Software.
7. Related/follows: Why do the feeders talk directly to the controller, i.e. why are no Actuators used as is customary in OpenPnP?  
8. Related/follows: Why does the index machine "re-invent the wheel" of auto-feeder and slots instead of re-using and feature-enhancing or at least sub-classing the existing classes? I know, I know. I just want to hear it 😇
9. Related/follows: this is a lot of code. Do you intend to support it long-term? 

I'm sure, there are good reasons for all of these points. Looking forward to understanding them 🙃

_Mark

P.S. Where are the screen-shots?

markmaker:
Just saw points 5 and 7 are mistakes. You _are_ talking through Actuators. Sorry.

Jnesselr:
> As far as I can tell, this looks like very high quality code.

Thank you, it took me quite a while!

>  i.e. it is all well-meant and just intended to poke you into defending your design (and as a result: us knowing why you did things the way you did them

Ugh Devil's advocates making me explain myself instead of just yeeting my code into the universe. The worst! 😛  But yeah I can answer them. Some of them might be disagreed with and I can dive in more if need be but I think in context of the goals of the project and such, the decisions make sense. I will warn you that I'm kinda sleep deprived so some of the answers my seem snarky.

> Why is the packet encoding/decoding, feeder scanning, etc. not on the controller side and then exposed as more or less plain (textual) G-code by the controller? This would allow...

It's mostly because we already have to modify RS485 to work on Marlin and we decided we didn't want Marlin to have to understand or manage too much of the protocol. Plus it means that protocol updates can be handled inside of OpenPnP instead of having to update Marlin AND the feeders as well. I'm eventually planning on adding functionality to update the feeders over the RS485 bus too if this all works which would be an amazing little trick.

> User manual testing inside/outside OpenPnP using human-readable "command-line" G-code.

It would allow that but honestly, we don't expect a ton of that so we don't really care to have it as a feature.

> Keeping the firmware/packet format versioning problem at least constrained to the electronics side (controller & feeders) and not extending it to OpenPnP versioning (don't tell me there will never be a revision to the packet format 🤓).

We actually figured that updating the Marlin firmware would be harder than updating OpenPnP. And the protocol has a packet version that's just not used in the code yet because we don't need it yet. Ideally the feeders should be able to update sans OpenPnP version updates, but that stuff isn't done yet. The short and sweet of it is that trying to keep it all on the electronics side is mostly just semantics and doesn't appear to offer any tangible benefit. 

> Keeping the high-level G-code stable/backwards compatible and/or easily configurable in OpenPnP if just the syntax were to change.

That's actually a benefit to the approach we're using. The G-code stays the same even if we have 100 new command types.

> Finally, in the spirit of Open Source versatility: Using of the feeders from other PnP Software.

If they want to use them, they can implement the protocol. I get that it's harder than just doing raw G-code command types which might be user configurable, but our goal was to make a powerful intelligent feeder that worked in OpenPnP. If it's Java, they can use all of the packet encoding/decoding classes just by themselves as they're pretty independent of anything else.

> Related/follows: Why does the index machine "re-invent the wheel" of auto-feeder and slots instead of re-using and feature-enhancing or at least sub-classing the existing classes? I know, I know. I just want to hear it 😇

There is no first class concept of a slot in OpenPnP. If there was, I'd probably use it. Trying to shoehorn this into a subclass of the auto feeder is also the wrong approach. It prioritizes the illusion of reuse over actual benefits imho. I tend to "duplicate" things when two things have apparent similarity but are semantically different. One of the HUGE differences out of the gate is that the slot auto feeder has the "feeder" be the slot and internally stores the feeders. The Index Feeder is the opposite. The feeder class itself is the hardware feeder and it has a slot address which refers to the physical slot.

If I were to be so bold, I'd say we're not reinventing the wheel, rather we're doing slotted feeders "right". 😆 But also it might be nice to have slots as a first class concept some day. I'm not sure how that'd affect the reference auto feeder though.

> Related/follows: this is a lot of code. Do you intend to support it long-term?

Being unable to predict the future, I'll say that my hope is that it will be supported, even if not by me directly, by the index project as a whole. But for the time being, yeah I'll be addressing any issues we see with them.

> P.S. Where are the screen-shots?

Coming, once we do a full test and rundown on real hardware with this. The reason the IndexFeeder isn't in the list unless you add it and build from source is because we aren't 100% sure how this will work on real hardware just yet. It's also a big PR and I'd rather have future PR be "Here's the things we're fixing" instead of continually dragging this big PR along. I've run into a few times where I rebase and the API has changed or something and I have to fix merge conflicts.

Once we know this will all work correctly, we will be throwing up screenshots, pictures of the feeders, and configuration information on the wiki.

daveismith:
I would add a few notes as well:
> Why is the packet encoding/decoding, feeder scanning, etc. not on the controller side and then exposed as more or less plain (textual) G-code by the controller? This would allow...

Having it done inside OpenPNP allows for other controllers to use the protocol without needing to implement the specific protocol, instead they only need to implement a mechanism to get it onto the RS485 bus (which could be a standalone UART to 485 converter or built into another controller like smoothieboard)

markmaker:
Thanks, @Jnesselr, @daveismith,

> protocol updates can be handled inside of OpenPnP instead of having to update Marlin AND the feeders 

> Having it done inside OpenPNP allows for other controllers to use the protocol 

OK. You're saying the controller just transparently forwards packets without any logic applied locally? Makes sense. 

> If I were to be so bold, I'd say we're not reinventing the wheel, rather we're doing slotted feeders "right". 😆 

OK. But then I could turn the table on you...   😁

I wonder if you could extract an abstract base class, so other future autofeeder+slot implementations could benefit. I'm just fishing for something that might quell autofeeder+slot implementation "mushrooming" in the long term. Also, it would be nice if some aspect of your work were useful to a larger OpenPnP community base, i.e. even when not using index electronics. What do you think? 

> Coming, once we do a full test and rundown on real hardware with this. The reason the IndexFeeder isn't in the list unless you add it and build from source is because we aren't 100% sure how this will work on real hardware just yet.

Oh! I thought this is working at least "beta" style? How far off are you? Days, Weeks, Months? 

@vonnieda, did you know? What do you think? 

> I've run into a few times where I rebase and the API has changed or something and I have to fix merge conflicts.

To be frank, I believe it is fair for _you_ to bear the burden of resolving merge conflicts as long as your code is not yet finished/usable for other OpenPnP users. I'm doing this all the time on my branches, sometimes for many months. I know it is a pita not having other developers' refactoring steps magically remote-applied to your part of the code, but that's life. On the other hand, if other developers create new stuff, and they can't even instantiate your classes to at least test Wizards and other dry run stuff, that's rather unfair to _them_, don't you think?

In that sense, you should definitely add the feeder class to the list. You can add a warning label on the Wizard to warn users off. 

Waiting for @vonnieda's say regarding the unfinished state. 

_Mark

Jnesselr:
> OK. You're saying the controller just transparently forwards packets without any logic applied locally? Makes sense.

Yep!

> i.e. even when not using index electronics. What do you think?

That's fine. If it follows the same protocol, it doesn't matter to me how the electronics work or if it's on another board. I don't care if it's RS485 or anything really from the OpenPnP side. If you speak the "Index protocol" you can use this feeder type.

Regarding autofeeder + slot mushrooming, honestly that's scope creep to me for this feature. If someone wants to refactor this code at a later date to use some common functionality, that's fine.

> Oh! I thought this is working at least "beta" style? How far off are you? Days, Weeks, Months?

Weeks is the ideal. We were hoping RS485 would be finished by now in Marlin but unfortunately some stuff happened that delayed that so that's what we're blocking on. We've got the feeder side code implemented and the idea is that this OpenPnP code should "just work" assuming all the other pieces are coded as designed.

Plus, this is unit tested to hell and back.

@vonnieda did know about this and signed off on it as long as we didn't have the index feeder in the main list. And I'd rather keep it that way because people will straight up ignore the warning and wonder why nothing works.

> and they can't even instantiate your classes to at least test Wizards and other dry run stuff, that's rather unfair to them, don't you think?

Sure they can, they absolutely can add the feeder class in for testing themselves if they want. That's the whole point, people actively developing can enable this if needed. Personally I don't think they should and they definitely shouldn't have to.

Keeping up with the burden of making sure the API all still works myself vs having others keep up with it is irrelevant in my opinion. It's an extremely unit tested feature and if someone changes the API on something that changes the API on my code they might also have to change the unit tests on it. Whether it's "fully tested with real hardware" is not relevant at all to that fact. And fwiw, I have been keeping this branch up to date with develop, rolling with the changes for over 6 months now.

I think part of this is how we see the state of the code. I consider it fully functional as I've tested it with a virtual driver I wrote that matches the gcode specification and it totally works. If we get to testing it on hardware and something is wrong with it, I'm going to first investigate the hardware and fix the hardware since the OpenPnP side performs to spec. The **only** reason I don't feel comfortable with having the feeder enabled in OpenPnP is because people will think that means it's tested and working with real hardware. But as far as I'm concerned, really, the feature is done and I see no reason not to merge it.

markmaker:
> Regarding autofeeder + slot mushrooming, honestly that's scope creep to me

To stem against everybody reinventing their own wheels? The earlier, the better? That's never "scope creep" in my book. And looking at the quality of your code, I have a hard time believing _you really believe that yourself_ 🙃 

> this is unit tested to hell and back.

> @vonnieda did know about this and signed off on it as long as we didn't have the index feeder in the main list. And I'd rather keep it that way because people will straight up ignore the warning and wonder why nothing works.

In that case, I'm perfectly fine with it. 

_Mark

Jnesselr:
I actually do consider it scope creep for the original task. I looked at the existing auto feeder class and if it had worked similarly (ie had a concept of slots where the feeder was the actual feeder hardware) then I would have probably started with refactoring out something that works for both.

The general problem I have in these sorts of cases is that I'd be modeling it mostly off of my design since there's not enough commonality and it wouldn't simplify the reference slot auto feeder (RSAF from here on) design. Since the "feeder" in the RSAF is actually the slot, I'd be really worried about using it as a basis for refactoring something using that as the other model. Would there be a one to one relationship of slots and feeders in machine configuration? The RSAF uses an internal property to store all of its feeders and make them selectable.

In short, refactoring the RSAF into a design that more matches having the slots as an independent concept and having the feeders themselves be the actual feeders feels like scope creep.



But, if you were to design a slot interface, what would you put in it? For me, I suspect it'd be the base location and the slot number (which for us is also the address). What would the API look like? Can you get a feeder instance from a Slot? Maybe we should make a "SlottableFeeder" interface just for this that has a getSlot method? Because if it's not too complicated, I'm totally fine with refactoring my code to match it but I don't want to touch the RSAF code unless we're changing how that model works.

markmaker:
I guess we're misunderstanding each other. Remember, I was "turning the table on you".

What I suggested was refactoring your concept into two sets of classes: one (abstract base) for **your** generalized concept of feeders and slots and one for the actual index implementation. So if others find your concept of feeders and slots a better fit for their own (non-index) implementation, they can reuse your base classes. Plus the Wizard pages separated or sub-classed too. 

The idea is to facilitate the evolution of OpenPnP as a whole. There are ideas of a generalized feeder grouping around (exchangable/relocatable trolleys, slates etc.) for instance. Or let's take @doppelgrau'r recycle feature that he had to implement for each feeder. These sorts of ideas become harder and harder to do, if the feeder classes mushroom at their bases. 

That's all I suggest(ed). 

_Mark

Jnesselr:
I didn't actually understand what you meant by that phrase re table turning. But I do understand the issue that people might be dealing with re code reuse. The problem with this will always be that I may abstract certain parts of my API and not other parts and get it wrong when it comes time to reuse. So I need to understand what parts you feel will be reused, in detail. The only thing I can think of is having "Slots" be a first class citizen in OpenPnP.

markmaker:
I may be mistaken, but I think these feeders are all very similar in functionality, and I assume it is possible to extract common "truths". Ideally, only the protocol part would be index specific.

vonnieda:
@markmaker I agree that @Jnesselr's code here might be a good basis for a better slot system, but I don't want to block them by asking them to do that work right now. Their goal here is to get their machine and feeder up and running, and I think as long as it's not breaking anything else it's fine to go ahead and merge this. If someone in the future wants to spend the time to revamp the slot system they could use this as a basis, or adapt this to the new system.

Put another way, I consider this machine implementation (org.openpnp.machine.index.*) to be a "plugin" for OpenPnP. In an ideal world it would actually be a self contained plugin, but we don't have a plugin system yet :)

So, if there are no objections on the basis of this breaking other things, I think we can go ahead and merge.

markmaker:
I understand, makes perfect sense. 
_Mark

PR #1216, opened 2021-06-15
Fix/Manual Nozzle Tip Change Motion

markmaker:
# Description
When manual nozzle tip change is enabled, the user is prompted to load/unload a nozzle tip when needed. This is done by throwing an Exception, interrupting the current machine task and ultimately resulting in a message box being displayed: 

Unload

Unload

Bug: When the advanced motion planner is enabled, the move to the manual change location might only be queued and not yet be executed when the Exception is thrown. Instead, the motion commands remained in the queue and would unexpectedly be executed later. 

This PR fixes both problems: 

1. It waits for motion to complete, before throwing the Exception.
2. It clears queued commands from the planner, when the machine task/executor is aborted. Otherwise, these could unexpectedly be executed in the context of the next task.

Note: the second problem is more general and can happen in other exception throwing cases, if continuous planned motion is enabled and motion commands are queued up at the moment. The exception leading to the abortion of the commands happens after the `moveTo()` commands in question. So aborting these queued moves is more conservative than with the `NullPlanner`, where such motion would already have been executed. The whole motion sequence becomes _one transaction_, the pending moves are _rolled back_ on failure (a bit like in a relational database).

# Justification
See #1215 

# Instructions for Use
No change.

# Implementation Details
1. Problem number 2 was discovered earlier (in other ongoing development work), and the fix tested on the machine. Problem number 1 and 2 were also tested in simulation, both inside and outside jobs. The correct behavior was confirmed and also analyzed in the log. 
3. The `MotionPlanner` in `org.openpnp.spi` got the new method `clearQueuedMotion()` to clean up after aborting a task. 

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
@@ -33,6 +33,7 @@ import org.openpnp.spi.Actuator.ActuatorValueType;
 import org.openpnp.spi.Camera;
 import org.openpnp.spi.CoordinateAxis;
 import org.openpnp.spi.Feeder;
+import org.openpnp.spi.MotionPlanner.CompletionType;
 import org.openpnp.spi.Nozzle;
 import org.openpnp.spi.NozzleTip;
 import org.openpnp.spi.PropertySheetHolder;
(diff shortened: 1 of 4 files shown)


markmaker:
@vonnieda, are there any other known "benign" exceptions being thrown to prompt user action? The same fix would have to be applied there. 

Otherwise, I'll merge this if you're OK with it.

_Mark

markmaker:
I re-read everything again, and I am confident this is good, merging it.

PR #1219, opened 2021-06-18
Feature/blinds feeder qrc - QR code part ID check - WIP/Concept/Review

crashmatt:
# Description
Optional check of blinds feeder part ID against QR code placed next to the feeder.
Can generate a png image of QR codes arranged and scaled to be in the correct position for each connected feeder.

# Justification
Prevents bad part or feeder group identification by cross checking expected part numbers expected marked on the feeder and the value set.
Feeder part ID can be read with a phone camera for assist correct component installation.

# Instructions for Use
1. Home the machine
2. Select a BlindsFeeder and calibrate fiducials
2. If the QR code is placed on the left side of the feeder select "ID on feeder left" from the locations section.
3. From the "Tape Settings" section select "Identify part".  The camera will move to the expected QR code location.

If a QRcode image is required select Export QR Code from the "Tape Settings" section.  The image can be found in the folder .openpnp/org.openpnp.machine.reference.feeder.BlindsFeeder

# Implementation Details
1. Only basic testing so far.  More to do after review.
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? Needs tidyup. 
3. No changes to `org.openpnp.spi` or `org.openpnp.model` 
4. `mvn test` ok

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java
@@ -23,6 +23,7 @@ package org.openpnp.machine.reference.feeder;
 
 
 import java.awt.Color;
+import java.awt.Graphics2D;
 import java.awt.geom.AffineTransform;
 import java.awt.geom.NoninvertibleTransformException;
 import java.awt.geom.Point2D;
@@ -31,8 +32,19 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.Hashtable;
+import java.util.Iterator;
 import java.util.List;
 
+import javax.imageio.IIOImage;
+import javax.imageio.ImageIO;
+import javax.imageio.ImageTypeSpecifier;
+import javax.imageio.ImageWriteParam;
+import javax.imageio.ImageWriter;
+import javax.imageio.metadata.IIOInvalidTreeException;
+import javax.imageio.metadata.IIOMetadata;
+import javax.imageio.metadata.IIOMetadataNode;
+import javax.imageio.stream.ImageOutputStream;
 import javax.swing.Action;
 
 import org.apache.commons.io.IOUtils;
@@ -78,8 +90,13 @@ import org.simpleframework.xml.Attribute;
 import org.simpleframework.xml.Element;
 
 import com.google.zxing.BinaryBitmap;
+import com.google.zxing.EncodeHintType;
 import com.google.zxing.MultiFormatReader;
+import com.google.zxing.qrcode.QRCodeWriter;
+import com.google.zxing.qrcode.decoder.ErrorCorrectionLevel;
+import com.google.zxing.BarcodeFormat;
 import com.google.zxing.client.j2se.BufferedImageLuminanceSource;
+import com.google.zxing.client.j2se.MatrixToImageWriter;
 import com.google.zxing.common.HybridBinarizer;
 
 
@@ -978,8 +995,97 @@ public class BlindsFeeder extends ReferenceFeeder {
         MainFrame.get().getCameraViews().getCameraView(camera)
         .showFilteredImage(showResult, 1000);
 
-    } 
+    }
+    
+    public void generatePartNumberQRCodeImage() throws Exception {
+        QRCodeWriter writer = new QRCodeWriter();
+        List<BlindsFeeder> feeders = getConnectedFeeders();
+        
+        List<BufferedImage> qrcImageList = new ArrayList<>();
+        
+        final int printerDpi = 1200;
+        final double srCodeSize_mm = 8.0;
 
+        final double pixelsPerMm = (double) printerDpi / 25.4;
+        final int qrCodeSize = (int) (srCodeSize_mm * pixelsPerMm);
+        
+        final Length extent = getFeederExtent();
+        final Length extentMm = extent.convertToUnits(LengthUnit.Millimeters);
+        
+        final int imageHeightpx =  (int) (extentMm.getValue() * pixelsPerMm);
+        
+        Hashtable<EncodeHintType, ErrorCorrectionLevel> hintMap = new Hashtable<>();
+        hintMap.put(EncodeHintType.ERROR_CORRECTION, ErrorCorrectionLevel.H);
+
+        BufferedImage image = new BufferedImage(qrCodeSize*2, imageHeightpx, BufferedImage.TYPE_BYTE_BINARY );
+        Graphics2D graphics = image.createGraphics();
+
+        //Fill with black border
+        graphics.setBackground(Color.BLACK);
+//        graphics.setPaint ( new Color ( 0, 0, 0 ) );
+//        graphics.fillRect ( 0, 0, image.getWidth(), image.getHeight() );
+        graphics.setPaint ( new Color ( 255, 255, 255 ) );
+        graphics.fillRect ( 5, 5, image.getWidth()-10, image.getHeight()-10 );
+        
+        //Draw marker at top
+        graphics.setPaint ( new Color ( 0, 0, 0 ) );
+        graphics.fillRect ( 10, 10, image.getWidth()-20, 20 );
+        
+        for(BlindsFeeder feeder: feeders) {
+            String partId = feeder.getPart().getId();
+            if(partId != null){
+                com.google.zxing.common.BitMatrix bitMatrix = writer.encode(partId, BarcodeFormat.QR_CODE, qrCodeSize, qrCodeSize, hintMap);
+                BufferedImage qrcImage = MatrixToImageWriter.toBufferedImage(bitMatrix);
+                
+                double centerline = feeder.getPocketCenterline().convertToUnits(LengthUnit.Millimeters).getValue();
+                if(centerline != 0.0) {
+                    int imagePosPx = ((int) (centerline * pixelsPerMm)) - (qrCodeSize/2);
+                    image.getGraphics().drawImage(qrcImage, qrCodeSize/2, imagePosPx, null);
+                }
+            }
+        }
+        
+        if (Logger.getLevel() == org.pmw.tinylog.Level.DEBUG || Logger.getLevel() == org.pmw.tinylog.Level.TRACE) {
+            File file = Configuration.get().createResourceFile(getClass(), "blinds-feeder-qrc", ".png");
+            
+            //Write image together with pixel density metadata so it will print at the correct scale.
+            final String formatName = "png";
+            for (Iterator<ImageWriter> iw = ImageIO.getImageWritersByFormatName(formatName); iw.hasNext();) {
+               ImageWriter imgWriter = iw.next();
+               ImageWriteParam writeParam = imgWriter.getDefaultWriteParam();
+               ImageTypeSpecifier typeSpecifier = ImageTypeSpecifier.createFromBufferedImageType(BufferedImage.TYPE_INT_RGB);
+               IIOMetadata metadata = imgWriter.getDefaultImageMetadata(typeSpecifier, writeParam);
+               if (metadata.isReadOnly() || !metadata.isStandardMetadataFormatSupported()) {
+                  continue;
+               }
+
+               IIOMetadataNode horiz = new IIOMetadataNode("HorizontalPixelSize");
+               horiz.setAttribute("value", Double.toString(pixelsPerMm));
+
+               IIOMetadataNode vert = new IIOMetadataNode("VerticalPixelSize");
+               vert.setAttribute("value", Double.toString(pixelsPerMm));
+
+               IIOMetadataNode dim = new IIOMetadataNode("Dimension");
+               dim.appendChild(horiz);
+               dim.appendChild(vert);
+
+               IIOMetadataNode root = new IIOMetadataNode("javax_imageio_1.0");
+               root.appendChild(dim);
+
+               metadata.mergeTree("javax_imageio_1.0", root);
+
+               final ImageOutputStream stream = ImageIO.createImageOutputStream(file);
+               try {
+                   imgWriter.setOutput(stream);
+                   imgWriter.write(metadata, new IIOImage(image, null, metadata), writeParam);
+               } finally {
+                  stream.close();
+               }
+               break;
+            }
+        }
+        graphics.dispose();
+    }
     
     public void findPocketsAndCenterline(Camera camera) throws Exception {
         // Try to clone some info from another feeder.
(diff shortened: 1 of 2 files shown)


markmaker:
Hi @crashmatt 

First, I appreciate the idea! ... It is only that I've already quite heaviliy invested into a OCR based implementation earlier, so I'm a bit conflicted. My design creates the label directly in OpenSCAD, the idea would be to _discover_ the feeder in OpenPnP with the labels already physically there. There is a header label coded with dimensions of the whole array, so discovery should be possible. 

OCR BlindsFeeder

grafik

I see the advantage of QR-codes to be more compact, on the other hand OCR is human-readable and taking into account that you will likely load the feeder off-machine, and this is all K.I.S.S./hobby oriented, I was favoring OCR. 

As you probably know, I already implemented OCR on the ReferencePushPullFeeder, and the OCR part is actually a stage in the pipeline. 

https://youtu.be/5QcJ2ziIJ14?t=62

https://makr.zone/openpnp-new-referencepushpullfeeder/393/
https://makr.zone/new-all-3d-printed-tapereel-feeder/399/

I was thinking that OCR/QR choice could one day be made available by creating a **QR pipeline stage** with same/similar result type as the OCR stage. This should be easy. It would then be a drop-in replacement for those that prefer QR.

This design also makes it reusable for the next feeder/other OpenPnP applications, e.g. the ReferencePushPullFeeder could be equally used with QR as an alternative. 

What do you think?

_Mark

crashmatt:
I was somewhat aware that you were doing something with OCR. You have mentioned this before.
I am already using QR codes since I am trying to keep the solution as compact as possible.  I don't have much spare space on the machine.  This PR was the result of finding a quick and easy way to verify the QR codes I already have.

I don't see a conflict between your concept and mine.  They could co-exist.  Implementing QRC alternative on top of your OCR should be simple.  Might be best for me to branch from your feature branch rather than have a complete implementation of my own.

One way I use some of my feeders is not compatible with your concept.  I make banks of common part types, say 0603 resistors.  Then I populate the feeders as I need to as projects demand E3/E6 series parts.  Then I don't need to plan exactly what part goes where from the very first project.

The QR codes I have are generated through python from the kicad bom.  These have the part ID text next to the QRC.  This is something I could add to the QRC image generation for openpnp.

I put my QR codes on the left side of the feeder.  In hindsight not the best choice.

markmaker:
Thanks, @crashmatt for being so open-minded. ;-)

> Implementing QRC alternative on top of your OCR should be simple. 

Just to be clear: you agree this should be done as a pipeline stage drop-in replacement? 🤓

>  Might be best for me to branch from your feature branch rather than have a complete implementation of my own.

That would be ideal. I have just merged in the latest develop branch, i.e. including your group names. As you may have noticed, I have made the array common settings a separate Wizard tab, makes the GUI much clearer and also makes it easier to one day make it a separate grouping object.  

Array Wizard

Unfortunately, the feature is still work in progress. It also includes changes to the OpenSCAD part, which I'm not all happy with. 

One example: 

I wanted the OCR (QRC) label on the right side (beginning of tape), because the BlindsFeeder is optimized to be usable on the edge of the machine too, where the camera might not be able to reach the end of the feeder (you might remember, this is explained in the [video](https://youtu.be/dGde59Iv6eY?t=246)). 

Then I made the decision to swap the cover edge alignment, i.e. they are now aligned with the right edge when _closed_. So the OCR (QRC) label can be recognized when the feeder is closed. But for several reasons, I now think that decision is wrong, I want to undo it. The covers should overlap the labels when closed, i.e. they are sort of protected by the label overhang, so when handling a slate etc. one does not accidentally push the covers open. The feeder can then extend right to the edges of the slate, using the whole space. Also, if multiple arrays are located together length-wise, the label of the next one should not be obstructed by the closed covers of the previous one. True, to recognize the OCR (QRC), the cover must be opened first, but that's the lesser evil. And I plan to add Z info to the header label, so at least for arrays with _unified_ tape Z surface (e.g. all 0402 parts, all 0603 parts etc.), the Z needed for opening the covers can be auto-discovered too. 

I guess we would have to work somewhat in parallel, but that should be possible if you concentrate on making the QRC stage first. Once I'm happy with the basics, I'll push to my repo. The auto-discovery will not yet be included, but that's still usable. 

**Future**: I was also thinking about a "provisioning" function in OpenPnP. The part list view would be enhanced to filter for parts in a job that have no feeders assigned yet. A sort-order by package and per-part-placement count would also be added. Then you could multi-select parts and use a new function "Make BlindsFeeder Array" which would create and invoke the OpenSCAD model and instantiate the OpenPnP feeders. The sort-order by per-part-placement count would make sure feeders with high counts are arrayed together and can be located closer to the PCB/bottom camera for better assembly speed. 

In fact, I'm starting to wonder whether auto-discovery is needed at all. 

_Mark

crashmatt:
@markmaker 
Thanks for merging the groups PR and tidying up the gui.

> Just to be clear: you agree this should be done as a pipeline stage drop-in replacement?
I started writing a pipeline QRC reader based on SimpleOcr but switched to the simpler zxing library.  I doubt SimpleQrc will work with the existing greenscreen pipeline.  Regardless of the implementation it would be good to have QRC automatically detected instead of OCR.

My preference  (at the moment)  is to have the QRC on the right so when a push cover is open it does not cover the ID. Then there is no problem with identifying the feeder at startup when the covers are already open.  The QRC can sit a little out from the feeder edge so reading it with closed feeder is no problem.

My main aim for this is to fix accidental bad part assignments and extend it into identifying feeder groups. I don't have a strong desire for auto-config. Auto z-level feature is interesting as it might prevent mistakes that cause damage.  As you say, a future feature.

markmaker:
> I started writing a pipeline QRC reader based on SimpleOcr but switched to the simpler zxing library. I doubt SimpleQrc will work with the existing greenscreen pipeline. 

I don't understand. Why? 

Maybe there's a misunderstanding of how this works? It's best to look at how OCR works in the ReferencePushPullFeeder.

https://youtu.be/5QcJ2ziIJ14?t=62

There will likely be multiple QR codes visible at once, so it is important to extract the right one. You can use the region of interest (label area) to extract the right QR code. Use an AffineWarp stage. The region of interest coordinates can later be semantically passed by the feeder (it already does this in my branch). This is the pipeline in my branch, but that's mostly untested.

```
<cv-pipeline>
   <stages>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRead" name="00" enabled="false" file="test.png"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageCapture" name="0" enabled="true" settle-first="true" count="1"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.BlurGaussian" name="1" enabled="true" kernel-size="3"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.AffineWarp" name="ocr1" enabled="true" length-unit="Millimeters" x-0="0.0" y-0="0.0" x-1="0.0" y-1="0.0" x-2="0.0" y-2="0.0" scale="1.0" rectify="true" region-of-interest-property="regionOfInterest"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ConvertColor" name="ocr2" enabled="false" conversion="Bgr2Gray"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.SimpleOcr" name="OCR" enabled="true" alphabet="0123456789.-+_RCLDQYXJIVAFH%GMKkmuµnp" font-name="Liberation Mono" font-size-pt="7.0" font-max-pixel-size="28" auto-detect-size="false" threshold="0.75" draw-style="OverScaledImage" debug="false"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRecall" name="11" enabled="true" image-stage-name="1"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ConvertColor" name="2" enabled="true" conversion="Bgr2HsvFull"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.Normalize" name="3" enabled="false"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.MaskHsv" name="4" enabled="true" auto="false" fraction-to-mask="0.0" hue-min="65" hue-max="115" saturation-min="50" saturation-max="255" value-min="40" value-max="255" invert="false" binary-mask="true"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.BlurMedian" name="5" enabled="true" kernel-size="13"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.FindContours" name="7" enabled="true" retrieval-mode="List" approximation-method="Simple"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.FilterContours" name="8" enabled="true" contours-stage-name="7" min-area="1000.0" max-area="100000.0"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.MinAreaRectContours" name="results" enabled="true" contours-stage-name="8"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRecall" name="10" enabled="true" image-stage-name="0"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DrawRotatedRects" name="12" enabled="false" rotated-rects-stage-name="results" thickness="3" draw-rect-center="true" rect-center-radius="20" show-orientation="false"/>
   </stages>
</cv-pipeline>
```

From the AffineWarp you get a clean cropped image of the QR code area in the stage. You can apply zxing. This should be super easy.

Why a combined pipeline?

I will try to combine the **CheckOpen** vision operation with the label vision operation, which means the feeder will calculate an optimal camera position that sees both the label and some of the blinds. Having the region of interest passed to the pipeline semantically, will make sure that the right label will always be read.  

For **CheckOpen** see here:
https://github.com/openpnp/openpnp/wiki/BlindsFeeder#choosing-when-the-cover-will-be-opened

> Regardless of the implementation it would be good to have QRC automatically detected instead of OCR.

Again, I don't understand, sorry.

I just like to keep the BlindsFeeder reasonably simple and maintainable. If you got two systems with completely different implementations conflicting with each other, then this is no longer the case. Like I said earlier, I too want to use the right side of the feeder. 

Plus, I would really appreciate the benefit of an independent/reusable QR-Stage implementation so that ReferencePushPullFeeder could also use QR codes instead of OCR, simply as a drop-in replacement. 

**Note**: it would even be possible to _integrate_ the QR variant _into_ the SimpleOCR stage. By offering a special "QR-Code" pseudo font in the **Font** drop-down. This would make it instantly selectable with no pipeline editing. 

Please reconsider. 🧑‍🚀 

_Mark

markmaker:
You know what? I'll just go ahead and integrate QR into SimpleOCR and then go test this on the machine and hopefully push my branch soon for you to try.
_Mark

markmaker:
Works:

grafik

grafik

markmaker:
Btw, I had a hard time printing and the detecting it using an ink jet (8mm code size). Had to use a photo paper. What do you use?

_Mark

crashmatt:
Apologies for my lack of understanding and clarity.  Your implementation is much cleaner.

> I started writing a pipeline QRC reader based on SimpleOcr but switched to the simpler zxing library. I doubt SimpleQrc will work with the existing greenscreen pipeline.
I presumed wrong. Now I see that OCR is done before in its own image process without green screen.

> Regardless of the implementation it would be good to have QRC automatically detected instead of OCR.
Could have read "together with OCR".  Exactly what you just did.

I am using a budget laser printer which is 600-1200 dpi.  Had not thought of inkjets that are closer to 300dpi.  
The python QRC generator generates an image sized to the number of pixels required for the content.  I have not yet found a way to do that with the Java QRC generator.

/Matt

markmaker:
> Apologies for my lack of understanding 

I must not have explained well the first time. 😏

Just tested on the machine and it really works. 

Controllable from the GUI: 
grafik

When set to normal font: 
grafik

When set to [Barcode]:
grafik

Raw images below:
Down looking_2021-06-20_16 31 52 695
Down looking_2021-06-20_16 30 56 647

There might still be bugs, but I'll push my branch now for you to test, if you like. 

**CAUTION**: there has been a change in how the OpenSCAD model works. So _if_ you extract the new files, be sure to adapt the BlindFeeder() calling convention. 

There is a new way to define the tape lanes in the array. This is now better structured (one array of _extensible_ tuples instead of two arrays of single properties). 

grafik


_Mark

markmaker:
See here:
https://github.com/markmaker/openpnp/tree/feature/BlindsFeeder-OCR

Turns out, I don't know how to compare branches inside GitHub without actually creating a Pull Request yet. Do you know this?

_Mark

crashmatt:
> Turns out, I don't know how to compare branches inside GitHub without actually creating a Pull Request yet. Do you know this?

I make a local copy of each branch and then use meld to compare them.

The changes to the tape definition make sense.  Will give that a go.

/Matt

ian-arkver:
Hi Mark,
To compare branches - click on the "38 branches" beside the branch selection pulldown. On that page you should get "Compare" buttons. Once there you also get two pulldowns to select which branches to compare.
Regards,
Ian

markmaker:
I just made the PR. It is marked as "Draft". 
#1220 

_Mark

markmaker:
> To compare branches - click on the "38 branches" beside the branch selection pulldown. On that page you should get "Compare" buttons. Once there you also get two pulldowns to select which branches to compare.
> Regards,
> Ian

Thanks!

crashmatt:
Should we move the conversation over to the new PR and close this one?

I am not good at reading the git diffs.  I find meld easier to read and edit.  Not as easy but it avoids mistakes.

markmaker:
> Should we move the conversation over to the new PR and close this one?
 
Yes, good point.

markmaker:
> Hi Mark,
> To compare branches - click on the "38 branches" beside the branch selection pulldown. On that page you should get "Compare" buttons. Once there you also get two pulldowns to select which branches to compare.

@ian-arkver, 
 
I tried that on this branch and it works:

grafik


But when I'm back on my own repo (or on openpnp), it does not:

grafik

It seems the function is hidden as soon as you've got the right to make a PR. 

I'm quite sure this was possible in the past (I did that repeatedly), so I guess this is one more example of software being "dumbed-down", that infuriates me so. 

_Mark

ian-arkver:
According to the GH docs here...
https://docs.github.com/en/github/committing-changes-to-your-project/viewing-and-comparing-commits/comparing-commits
you need to manually stick /compare at the end of the base repo URL, which seems needlessly tedious and stupid.

I checked with my own fork of openpnp which I obvs have PR/commit permission to. Sticking /compare at the end brings up a comparison of upstream openpnp vs. my own fork with the two pulldowns again.

Dunno why they couldn't have provided a button for this. Sigh.

PR #1220, opened 2021-06-20
Feature/BlindsFeeder OCR

markmaker:
# Description
___
**Work in progress:** be cautious when using this already. Including the new OpenSCAD files. 
___

This is an enhancement of the BlindsFeeder to support OCR and Barcodes to detect parts automatically, similar to the ReferencePushPullFeeder.

1. Most of the OCR functionality is similar to the ReferencePushPullFeeder's so this common base has been moved to OcrUtils and the ReferencePushPullFeeder had to be refactored as well.
2. Barcode detection (QR-codes etc.) has been added to the SimpleOCR stage. It can be switched simply by choosing a special `[Barcode]` font. ReferencePushPullFeeder supports it too.
3. The BlindsFeeder Wizard has been split in two tabs: One _feeder_ specific tab and one _array_ specific tab, i.e. with the settings common to all the array-linked feeders. 
4. Having multiple tabs has made it useful/necessary to make the FeedersPanel remember the last selected tab and re-select it when going to another feeder, similar to Machine Setup. 
5. An inconvenience (introduced by #1174) was fixed: Positioning the _camera_ to the pick location no longer requires a part compatible nozzle tip loaded on a nozzle. However, positioning the _tool_ will now automatically load a compatible nozzle tip if needed and allowed in config. NOTE: this parts may be broken out as a separate PR later.

BlindsFeeder with OCR

# Justification
Hope this is obvious. See also #1219 for a similar initiative.

# Instructions for Use

**Work in progress.**

Once the feeder is setup, the part can be detected by OCR or Barcode: 

OCR Detect Part

Settings (WIP):

grafik

Barcode recognition is treated as a special "Font": 

grafik

When set to a normal font: 

grafik

When set to [Barcode]:

grafik

Raw images below:

Down looking_2021-06-20_16 31 52 695

Down looking_2021-06-20_16 30 56 647

### OpenSCAD

**CAUTION**: there has been a change in how the OpenSCAD model works. So _if_ you extract the new files, be sure to adapt the BlindFeeder() calling convention. 

There is a new way to define the tape lanes in the array. This is now better structured (one array of _extensible_ tuples instead of two arrays of single properties). 

grafik

grafik

# Implementation Details
1. Preliminarily tested on the machine. **Job prep test is pending.**


crashmatt:
@markmaker, please keep the legacy openscad in the distribution so legacy users can access it.

markmaker:
> @markmaker, please keep the legacy openscad in the distribution so legacy users can access it.

Points to consider:

1. The change of the calling convention is really easy to make, you just pivot the arguments around, from this:
   
   grafik
   
   to this (you don't need to provide labels if you don't want them):
   
   grafik

2. I really only made this change, because it was a very bad design to have these "column-wise" instead of properly "table-wise". Rest assured, I will keep this as stable as humanly possible.  

3. This is all file based, so users can just keep their version (the feeder remains geometrically compatible). 

4. If you somehow lose the file version, it is fully versioned in Github. Just go to the file:

    https://github.com/openpnp/openpnp/blob/develop/src/main/resources/org/openpnp/machine/reference/feeder/BlindsFeeder-3DPrinting.scad
   
   press Blame Blame
   press History History

    https://github.com/openpnp/openpnp/commits/2208bea83ec75403fe1c1b6c257d846b978c942b/src/main/resources/org/openpnp/machine/reference/feeder/BlindsFeeder-3DPrinting.scad

_Mark

crashmatt:
@markmaker just started to run your ocr branch.
Since the setting for OCR detect is on a different page/section it is not clear what that button will do. I suggest having a separate button for detect and verify.  This then leaves "OCR Action" to determine what actions are done before running a job.

I also need to implement a left/right setting to fix my feeder construction.  Too late to fix this now.  Everything is glued down and components installed.

Detect OCR/QRC results in an assertion.  detectedOcrModel comes back null.  My pipeline does not have the OCR yet.  Maybe a hint message to users to fix this would help.  More testing to be done.

/Matt

crashmatt:
Correction. I am missing the font and orientation options from the commit I pulled from markmaker:feature/BlindsFeeder-OCR.  Did you push these changes?

image

markmaker:
It is a new file (so there can't be any merge conflicts etc.), and it seems I have not forgotten to add it:

https://github.com/openpnp/openpnp/pull/1220/files#diff-c8ce422cd98a62f9ecd4530ce3cac3a26bbf8a08f1cf6d808463ea1140c7b634

_Mark

crashmatt:
Found it.  Was not expecting those settings to appear depending on the OCR Action setting.

markmaker:
Aha. Yeah, adaptive dialogs are double-edged.

markmaker:
Coming back to these issues: 

> Since the setting for OCR detect is on a different page/section it is not clear what that button will do. I suggest having a separate button for detect and verify. This then leaves "OCR Action" to determine what actions are done before running a job.

Good idea. Will do that. 

> I also need to implement a left/right setting to fix my feeder construction. Too late to fix this now. Everything is glued down and components installed.

OK, I'll add a hack: by setting a _negative_ **OCR Margin** the label will be searched on the other side. 
grafik

However, I will not endorse this "officially" for the reasons I explained earlier. See also here (the yellow overlay and explanation):

https://github.com/openpnp/openpnp/wiki/BlindsFeeder#positioning-and-orienting-the-feeder-on-the-machine-table

_Mark

markmaker:
@crashmatt 

The changes are done:

- Button will always detect, regardless of set OCR Action.
- Negative OCR Margin will put label area to the other side of the feeder.

grafik

Please test again. 

Note, that I never tested job preparation OCR. It would be veeery nice if you could help test that. Deliberately setting wrong parts in the job's feeders and then testing whether they are properly detected on Job prep. With the **ChangePart** action, they should all be changed and reported at once. 

# 😁

_Mark

markmaker:
> 
> 
> Found it. Was not expecting those settings to appear depending on the OCR Action setting.

This is now fixed too.

crashmatt:
Setting OCR margin negative can position the read over the left side.  The value has to be 2x distance so it is a bit awkward to use.  Maybe setting a negative value should offset the camera to fiducial2 offset?

crashmatt:
ah, just read the last comment on your last commit:  

>  Allow OCR region to be on either side of tape (by giving a negative… …

crashmatt:
and the rest of your comments.  Need to speed up to keep up with you.

markmaker:
Images for the Wiki, do not delete:
https://github.com/openpnp/openpnp/wiki/BlindsFeeder
____

grafik

PR #1223, opened 2021-06-23
Fix/Discard Contact Probing

markmaker:
# Description
Fixed null pointer bug for discard placement. Discarding is signalled by passing `part` as `null`. 

- The `NullPointerException` was brought in by #1205 (by accessing the null part id). Fixed.
- Added new option for the ContactProbeNozzle to probe on discard. The part can be brushed/tilted away from the nozzle. Probing will make sure there is no undue pressure even with an unknown part-on-nozzle state (i.e. including when picked badly, tomb-stoned, shifted etc.). On my machine, I'm using soft material to brush parts away from the nozzle (just switching off the valve does not reliably drop 0402 parts). 
- The empty MessageBox on exceptions that have no message is fixed. 
- A ContactProbeNozzle Issues&Solutions bug was fixed: when the firmware was unknown (neither Smoothieware, nor TinyG, where probing G-code can be automatically suggested), it would constantly moan about missing G-code, when in fact it was present.

Discard Probing Option

Exceptions with no message (such as `NullPointerException`s) before and after:

Error message before
Error message after

# Justification
User error report, see:
https://groups.google.com/g/openpnp/c/yea1C90iNlE/m/hFf3QR4BBAAJ

# Instructions for Use
Use the **Discard Probing** function with a ContactProbeNozzle. 

Discard Probing Option

- If **Discard Probing** is enabled, it should now unconditionally probe for the height, as it is not known if the part is on/off/tomb-stoned. But it should never store that probed height as the part height. 
- If **Discard Probing** is not enabled, it should just behave like a normal Nozzle (calls the super class method). 

# Implementation Details
1. Tested in simulation only (machine not functional at the moment). Used "random" relative motion G-code to simulate probing using the GcodeServer. 

Sorry about the many afterthoughts/commits after the initial push.

### DIFF
--- a/src/main/java/org/openpnp/gui/JogControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/JogControlsPanel.java
@@ -63,6 +63,7 @@ import org.openpnp.spi.HeadMountable;
 import org.openpnp.spi.Machine;
 import org.openpnp.spi.Nozzle;
 import org.openpnp.spi.base.AbstractNozzle;
+import org.openpnp.spi.base.AbstractPnpJobProcessor;
 import org.openpnp.util.BeanUtils;
 import org.openpnp.util.MovableUtils;
 import org.openpnp.util.UiUtils;
@@ -630,19 +631,7 @@ public class JogControlsPanel extends JPanel {
         public void actionPerformed(ActionEvent arg0) {
             UiUtils.submitUiMachineTask(() -> {
                 Nozzle nozzle = machineControlsPanel.getSelectedNozzle();
-                // move to the discard location
-                Map<String, Object> globals = new HashMap<>();
-                globals.put("nozzle", nozzle);
-                Configuration.get().getScripting().on("Job.BeforeDiscard", globals);
-
-                MovableUtils.moveToLocationAtSafeZ(nozzle, Configuration.get()
-                        .getMachine()
-                        .getDiscardLocation());
-                // discard the part
-                nozzle.place();
-                nozzle.moveToSafeZ();
-
-                Configuration.get().getScripting().on("Job.AfterDiscard", globals);
+                AbstractPnpJobProcessor.discard(nozzle);
             });
         }
     };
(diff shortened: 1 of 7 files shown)


markmaker:
Hi @vonnieda,

could you please quickly confirm that it is acceptable to reuse the static method `AbstractPnpJobProcessor.discard(Nozzle)` from the machine controls? The former code was (_or rather should have been_) a full duplicate. 

See here:

https://github.com/openpnp/openpnp/blob/4ed8f8a6976122f144e128df991358751f894072/src/main/java/org/openpnp/gui/JogControlsPanel.java#L633-L645

And here:
https://github.com/openpnp/openpnp/blob/a9773de4b918d17b43ca780fa428e2571727f506/src/main/java/org/openpnp/spi/base/AbstractPnpJobProcessor.java#L38-L48

And the proposed change (just this one commit):
https://github.com/openpnp/openpnp/pull/1223/commits/80792413c3d3b606181b4717f7fd5ce890558af2

Thanks!

_Mark

markmaker:
Edited the PR description to reflect the added commits.

vonnieda:
> Hi @vonnieda,
> 
> could you please quickly confirm that it is acceptable to reuse the static method `AbstractPnpJobProcessor.discard(Nozzle)` from the machine controls? The former code was (_or rather should have been_) a full duplicate.
> 
> See here:
> 
> https://github.com/openpnp/openpnp/blob/4ed8f8a6976122f144e128df991358751f894072/src/main/java/org/openpnp/gui/JogControlsPanel.java#L633-L645
> 
> And here:
> 
> https://github.com/openpnp/openpnp/blob/a9773de4b918d17b43ca780fa428e2571727f506/src/main/java/org/openpnp/spi/base/AbstractPnpJobProcessor.java#L38-L48
> 
> And the proposed change (just this one commit):
> [8079241](https://github.com/openpnp/openpnp/commit/80792413c3d3b606181b4717f7fd5ce890558af2)
> 
> Thanks!
> 
> _Mark

Yep, looks fine to me! I personally might have chosen to move it all to the Cycles class, as that class is intended to be a place for reusable or complicated machine motions, but the way you've got it here is totally fine with me too.

Thanks,
Jason

markmaker:
Thanks, @vonnieda 

> I personally might have chosen to move it all to the Cycles class, as that class is intended to be a place for reusable or complicated machine motions

Like so?

https://github.com/openpnp/openpnp/pull/1223/commits/535ca9a9f504a59c9124833ba63c3e848e8487ec

The only thing that looks a bit out of place now, are the script names: `"Job.BeforeDiscard"`, `"Job.AfterDiscard"`. But I guess we need to keep it backward compatible. 

_Mark

markmaker:
After re-checking everything again, I dare merge this.

PR #1237, opened 2021-08-04
Feature / ReferenceCamCounterClockwiseAxis with Arms Angle

markmaker:
# Description
This adds support for an **Arms Angle** in the cam axis transformation. 

grafik

This PR also replaced the old pixel image with a properly colored SVG illustration. Including one for dark mode.

grafik

The `ReferenceCamCounterClockwiseAxis` is a `TransformedAxis` for heads with dual rocker or seesaw driven Z axes powered by one motor. 

The two Z axes are defined as counter-clockwise and clockwise according how the rocker axis rotates when creating positive Z movement. The zero° point is tied to the balance point.

The classical seesaw model is extended by an angle between the two arms. Normally, a 180° angle defines straight-across arms. If the angle is 0°, it describes a one-armed design that always only pushes one side/nozzle at a time. Other "V-shaped" arm angles could be used to describe designs, that eliminate the dead-time, that the 0° one-armed design has.

Designs with an arms angle less than 180° (especially the 0° one-armed design) may move more than 180° i.e more than ± 90° on the arm. This means the sinus of the arm angle would go back down. But we need a strictly monotonic function so the axis transformation can be inverted. Therefore, we pretend the transformed axis will slightly move up further in Z, as the arm rotates further to push down the other nozzle. We create a piece-wise extension slope after the sinus has reached its maximum (or rather its cutoff value). This is barely visible to the user (in decimal representation), but still well-defined in the binary double representation. It is also completely irrelevant mechanically, as there are likely stops, limiting Z travel well before 
that point.

# Justification
Users start building machines with a single arm cam:

grafik
(Image by René)

See the discussion here:
https://groups.google.com/g/openpnp/c/keVhIaR98oE/m/yUNcuEksAAAJ

# Instructions for Use
See the already updated Wiki:
https://github.com/openpnp/openpnp/wiki/Transformed-Axes#referencecamcounterclockwiseaxis

# Implementation Details
1. Tested in simulation.


markmaker:
Seeing that we have deployment problems anyway, I'll go ahead and merge this. Perhaps Réne can still upgrade to this and in the meantime we can sort out the deployment problem.

PR #1238, opened 2021-08-04
Feature / Nozzle Vacuum Actuators by Reference, not by Name (Round 2)

markmaker:
# Description

## Features
* The ReferenceNozzle now stores Actuator object references internally, names are only used for persistence.
* Therefore, changing an actuator name no longer breaks the reference.
* Nozzle actuator setters/getters by name were removed. Code accessing these was refactored.
* The ReferenceNozzleVacuumWizard was reworked to use references instead of names.
* Bonus: The Nozzle adds simple reminders to Issues & Solutions, when vacuum actuators, assigned drivers, and/or their G-code command and/or regex are missing.

## Changes in SPI
* `org.openpnp.spi.Actuator.isDriverless()` was added to declare when an Actuator does not need a driver. `HttpActuator` and `ScriptActuator` are ok to be driverless even though they are derived from `ReferenceActuator` which _needs_ a driver (the distinction is required for Issues & Solutions support). Obviously, a more thorough refactoring would be to create a common abstract base class for  `HttpActuator`, `ScriptActuator`  and `ReferenceActuator`, and then only add the driver to the `ReferenceActuator`. 
* `org.openpnp.spi.Machine.getAllActuators()` was added to get all actuators from all heads and the machine (analog to `getAllCameras()`).

This builds on top of #1231. This turned out to be insufficient for the purpose (which is upcoming Issues & Solutions automatic vacuum actuator creation support). 

# Justification
Same as in #1231.

# Instructions for Use
Same as before. 

https://github.com/openpnp/openpnp/wiki/Setup-and-Calibration%3A-Vacuum-Setup

You can now change actuator names without problems. 

# Implementation Details
1. Tested in simulation. 
3. Changes in the `org.openpnp.spi` package, see "Changes in SPI " section above.

### DIFF
--- a/src/main/java/org/openpnp/machine/neoden4/NeoDen4Driver.java
+++ b/src/main/java/org/openpnp/machine/neoden4/NeoDen4Driver.java
@@ -148,10 +148,8 @@ public class NeoDen4Driver extends AbstractReferenceDriver {
             n = new ReferenceNozzle("N1");
             n.setName("N1");
             head.addNozzle(n);
-            getOrCreateActuatorInHead(head, ACT_N1_VACUUM);
-            getOrCreateActuatorInHead(head, ACT_N1_BLOW);
-            n.setBlowOffActuatorName(ACT_N1_BLOW);
-            n.setVacuumActuatorName(ACT_N1_VACUUM);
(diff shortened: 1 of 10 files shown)


markmaker:
Performed more tests, feel confident to merge. The Blow-Off-Actuator was not tested, though, as I don't have one.

PR #1249, opened 2021-08-15
Feature/multi language

listff:
# Description
Use new multi-language processing functions and modify more than 900 texts to support multi-language

The program will load the json file in the .\i18n folder according to the language code

For example, Chinese is .\i18n\zh.json

Language files are at https://github.com/listff/openpnp-i18n,I don't know how to add the source code,

# Justification
The old multi-language implementation is not suitable, the script generates multi-language

# Instructions for Use

Use I18n.gettxt("your text") in the code, it will automatically find the translation

# Implementation Details
The program displays Chinese normally

yes

Have run mvn test

add file src\main\java\org\openpnp\I18n.java


tonyluken:
Hi @listff,

I appreciate all the work you put into this but...
1. Can you explain more about why this is better than the "old multi-language implementation"?  As far as I can tell, the only difference is that you are directly using the English text as the key for looking-up the translation rather than a special key string (as the "old" implementation does).  While as a native English speaker, I can see the advantage of having the English text appear directly in the source code rather than in a .properties file; I can also see using it directly as the key as a big disadvantage - by using the English text as the key to lookup the translations means any update to the English text (for instance a simple spelling correction) breaks all of the other translations.
2. It appears there are still usages of the old style translations so now there are two different ways of handling translations - not good from a maintenance standpoint.
3. It appears you have changed the line endings of all the files you have touched to CRLF (Windows) when they should be LF (Unix).  That makes looking at the file differences on GitHub near impossible because it shows every line of the files as being changed.

listff:
> Hi @listff,
> 
> I appreciate all the work you put into this but...
> 
> 1. Can you explain more about why this is better than the "old multi-language implementation"?  As far as I can tell, the only difference is that you are directly using the English text as the key for looking-up the translation rather than a special key string (as the "old" implementation does).  While as a native English speaker, I can see the advantage of having the English text appear directly in the source code rather than in a .properties file; I can also see using it directly as the key as a big disadvantage - by using the English text as the key to lookup the translations means any update to the English text (for instance a simple spelling correction) breaks all of the other translations.
> 2. It appears there are still usages of the old style translations so now there are two different ways of handling translations - not good from a maintenance standpoint.
> 3. It appears you have changed the line endings of all the files you have touched to CRLF (Windows) when they should be LF (Unix).  That makes looking at the file differences on GitHub near impossible because it shows every line of the files as being changed.

1. I use a python script to get all the strings, find the text to be translated, and then write it into the file.It is difficult to use the script to automatically replace the .properties file.

The automatically generated .properties name may be too long or difficult to read

3. I can modify the script to use LF.

PR #1251, opened 2021-08-17
Allow pipeline properties to be in pixels

tonyluken:
# Description
Pipeline properties for pipeline stages DetectCircularSymmetry, MaskCircle, and SimpleBlobDetector can now be in pixels or defined system units.

This change also adds support for Area and Volume units, and includes tests for Area, Length, and Volume conversions and their associated arithmetic operations.

openpnp.model.Point was enhanced to allow conversions from/to opencv.core.Point

VisionUtils was updated to add a conversion from Area to pixels and tests were added for it.

Other miscellaneous changes:
- Length conversion to Feet now works correctly (it was erroneously multiplying by 12 when it should have been dividing by 12)
- Length.multiply(Length) now returns Area rather than Length (this method was not used anywhere in the code which is good because it didn't make any sense to return a Length)
- Parsing Length strings now treats the letter "u" the same as the Greek letter mu.
- Line endings in DetectCircularSymmetry.java were changed from CR/LF(Windows) to LF(Unix) line endings to be consistent with the rest of the repository.

# Justification
These changes should have very little affect on the current operation of OpenPnP but are rather intended to be stepping stones to future enhancements.

# Instructions for Use
The only operator observable change would be if they were to enter a length with units of microns.  For instance, before this change, entering "10.1um" would be incorrectly parsed to mean 10.1 millimeters.  It now gets parsed to mean 10.1 microns (micrometers) as intended.  Although very unlikely, if some observable output was intended to be in units of Feet, it would now have the correct value.

# Implementation Details
**1. How did you test the change?** The changes were tested by running nozzle tip calibration and bottom camera position and rotation calibration.  Several tests were also run with pipelines that utilized the above mentioned pipeline stages and the results were as expected.  Additional tests as discussed above were added to the unit test suite.
**3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.**There were changes to the model package but they were mainly additions of new functionality (added the new Area and Volume types and added two new methods to Point).  The Length type did have a few minor corrections to it as mentioned above but were in areas that were probably never previously used.
**4. Be sure to run `mvn test` before submitting the Pull Request.** All Maven tests were run and passed.


markmaker:
Hi @tonyluken 

This looks very nice, as far I can tell. I have only these questions:

1. Was it necessary to create the `AreaUnit` and `VolumeUnit` and their 1:1 mapping to `LengthUnit`? 
    IMHO, it could just as well have been handled as an implicit additional **exponent**, in addition to a base `LengthUnit`. All that I can see that would have to handle the **exponent**, are String parsing and String output methods.

2. The override property handling in the pipeline stage classes now has quite extensive code duplication per property. Couldn't this be refactored into common methods (Point and Length variants)? I see you already made it copy/paste friendly by assigning the variable `propertyName` or `property`, so this should be straight-forward (in fact, I suspect you prepared for refactoring, then forgot to pull it through).  

3. Maybe a chance for improvement, before it is multiplied: I see you copied the behavior of `Length` when it comes to `Length.parse(String, boolean)`, i.e. it returns `null` when it fails, but an explicit (undeclared) `NullPointerException` is then thrown in the `LengthCellValue(String)` (which is unused), `AreaCellValue(String)`, `VolumeCellValue(String)` constructors, and (implicitly) by `LocationButtonsPanel.getParsedLocation()`. I might overlook something here, but I find this strange. Maybe that could be reviewed, including for `Length`? @vonnieda, any special reason for not throwing tell-tale Exception right in `Length.parse(String, boolean)`? Would this not be more useful to the user than a nondescript `NullPointerException`? Note, that the most important application in `LengthConverter.convertReverse(String)` does itself throw a tell-tale Exception, but that should not be _its_ concern, should it?

Position Button throws Null Pointer Exception

_Mark

tonyluken:
Hi Mark,

1. AreaUnit and LengthUnit do have a 1:1 mapping but that's not true for VolumeUnit as it includes milliliters, microliters, and femtoliters as well as their equivalent cubic centimeters, cubic millimeters, and cubic microns respectively.  I guess I could eliminate those first three although I think they are the more commonly used volume units. It's not a big deal for me either way since Volume and VolumeUnit may never even get used.
2. Yeah I noticed there was a lot of duplication there.  Let me take a look to see if I can find a way to clean that up.
3. I was trying to not change the behavior of Length but I'm ok with making that change if you and @vonnieda both agree. On a related note, I noticed that LengthCellValue.compareTo(LengthCellValue) only compares the value and ignores the units.  Shouldn't it convert to the same units before doing the comparison? That's what I did in AreaCellValue and VolumeCellValue.

u0159:
Hi @tonyluken, your pr look very promising. 

To make pipelines more robust and upp independent it would be helpful when in the pipeline-editor some stages could handle system units instead of pixels. 

Here an example of my two bottom vision cameras where i need two different pipelines because of checking the size of MinAeraRect:

upp1
upp2

tonyluken:
Hi @u0159,
I totally agree that all pipeline stage fields that represent a length or area should be stored (and displayed) with the appropriate system units rather than pixels.  Then the stages would be independent of camera units per pixel and be much more transportable across different machines.  However, that is a bit out of the scope for this PR but will hopefully be in the works sometime soon.  The biggest issue I see with making such a change is making it backward compatible with existing pipeline definitions - there's not a trivial way to convert the current pipeline definitions with pixels to system units without knowing the specific units per pixel that were in effect when the pipeline was originally generated.

tonyluken:
@markmaker,
I pulled all the duplicate code out of the stages and put it into a static method of CvStage.

src/test/java/CvStageTest.java
```diff
@@ -0,0 +1,718 @@
...
+            assertEquals(originalParameter, overriddenParameter);
+            pipeline.setProperty(propName, overridingParameter);
+            overriddenParameter = CvStage.getPossiblePipelinePropertyOverride(originalParameter, 
+                    pipeline, propName);
+            assertEquals(overridingParameter, overriddenParameter);
+        }
+
+    }
+
+    static class TestCamera extends AbstractHeadMountable implements Camera {
```
markmaker, line 503: Hi @tonyluken,

could this CvStageTest.TestCamera be replaced by or inherit from VisionUtilsTest.TestCamera? It looks identical.

I'm just asking, because these mock-ups are a nuisance to maintain. 

_Mark

tonyluken: Hi @markmaker,

Yes! and done.

markmaker:
Hi @tonyluken,

Is this ready? Is it OK if I merge it to the `test` branch first, so it will be tested along with the other stuff?

_Mark

markmaker:
Hi @tonyluken,

It seems to show `DetectCircularSymmetry` in full conflict, even though I made the same CRLF → LF change in test. 

I guess you'll have to carefully resolve these only partially artificial conflicts, there are likely **true** conflicts as well, as I changed `DetectCircularSymmetry` in `test` too. See #1257.

_Mark

tonyluken:
Hi @markmaker,
Not sure what you meant by "in full conflict" - both yours and mine have LF endings and the only conflicts were where we had both made changes.  Anyway, I went ahead and resolved them.  The only remaining issue would be your point #3 above - I haven't done anything about it because I haven't heard anything back from @vonnieda.

markmaker:
It was just showing full conflict here on Github. 

When we check this out locally, the auto-lf crap will remove any difference (and potential conflict), that's why we don't find these wrong line-endings before pushing. 

_Mark

tonyluken:
I understand what you are saying but I did look at it on Github and it did **not** show full conflict.  I also downloaded a zipped copy of both test and feature/AllowPipelineParamsInPixels branches (did not check them out locally) and both files had LF line endings.  So I'm not sure why you were seeing them in full conflict.  Anyhow, it seems to have resolved itself.

markmaker:
I don't pretend to understand what Github does. You can still see it in the orginal commit:

https://github.com/openpnp/openpnp/pull/1251/commits/584fe18f4ff11804e6311770f81027d94b91e699#diff-b1dff1af25d96f30a05a031859c6ee1f9d1414245fb2cc8e12866c2bc8149320

Just to document how it looks from here:
grafik

_Mark

tonyluken:
I think that compare is against the develop branch and not the test branch where you fixed the line endings.

markmaker:
The compares are redone when you alter the merge branch (it actually warns you, it is about to do that, as you lose all prior comments on diffs). If you look at the timeline, you see how I switched it to `develop` and then back to `test`, _after_ I corrected the line-endings in `test`. To no avail. 

I think github does not actually compare the lines/files, it just stupidly marks up all lines it sees _recorded_ as changed in any of the commits leading up to the merge, regardless of whether the changes cancel each other out or result in the same end result on both branches. It is only when you manually resolve conflicts that it stops doing that. It's not the first time I observed such an effect.

_Mark

markmaker:
Just checking: Have you received my private email?

tonyluken:
I received this email.

markmaker:
Hi @tonyluken,

What is the status?

_Mark

tonyluken:
Hi @markmaker,

What status are you asking about?

markmaker:
Re my private email about point 3 of the list [here](https://github.com/openpnp/openpnp/pull/1251#issuecomment-901047070).

tonyluken:
Oh! No, I never received any private email about that - the only emails I received were the GitHub generated ones (including the one to which I responded).

markmaker:
Ahh, I misunderstood what you meant by "this email". I have now re-sent it. Please report if nothing arrives at your end.

tonyluken:
No private emails have arrived.

markmaker:
You got it this time? I used the new email address.

tonyluken:
No – even checked my spam folders – nothing.

markmaker:
tonyluken62@gmail.com is OK?

markmaker:
I'm merging this. We can tackle the point 3 issue later.

Big thanks, @tonyluken!

_Mark

PR #1252, opened 2021-08-17
Add enhancements to SimpleGraph

tonyluken:
# Description
Adds the following enhancements to SimpleGraph and SimpleGraphView:

- Adds a setting to force the X-Y aspect ratio to be 1:1 which is useful for graphs whose x axis may have the same units as the y axis
- Adds a setting to enable/disable showing of the line which can be used to de-clutter graphs where only the data points are important
- Adds a setting to enable/disable showing of a data marker at each data point which can be used to highlight the data points rather than the connections between the data points

# Justification
This change do not affect how OpenPnP currently operates but rather serves as a stepping stone to future enhancements.

# Instructions for Use
No special instructions for the operator is required as this change in not currently visible to the operator.

# Implementation Details
**1. How did you test the change?** Ran Vision tests with diagnostics enable to verify the existing graphs still work as usual.  Ran additional tests using the new settings to verify they work as intended.
**3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.** No changes were made to these packages.
**4. Be sure to run `mvn test` before submitting the Pull Request.** The Maven tests were run and all passed.

### DIFF
--- a/src/main/java/org/openpnp/gui/components/SimpleGraphView.java
+++ b/src/main/java/org/openpnp/gui/components/SimpleGraphView.java
@@ -141,6 +141,23 @@ public class SimpleGraphView extends JComponent implements MouseMotionListener, 
                         min.y = Math.min(-max.y, min.y);
                     }
                 }
+                if (dataScale.isSquareAspectRatio()) {
+                    if (min != null && max != null) {
+                        if (dataScale.isSymmetricIfSigned()) {
+                            if (min.x < 0.0 && max.x > 0) {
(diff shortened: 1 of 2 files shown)


markmaker:
Hi @tonyluken,

Brilliant! 

Wanted to do a "data marker" (or "scatter") option many times before. This would be the right choice for some of the new backlash calibration graphs. 

I'll handle the merge to `test`, where I've made some changes too. 

_Mark

Issue #1258, opened 2021-08-22
Movements are merged, that should be separated


doppelgrau:
# Problem Report
## Describe the Issue

Downloaded the latest version something in the motion planer has changed. The waypoints in the movement from a heap feeder to the drop box are ignored and merged in one motion causing a crash.

## Steps to Reproduce
1. Set up a heap feeder
2. click on "get samples"
3. watch how the head moves to Safe Z after finding a part, then straight to the drop location, ignoring the waypoints.

## Expected Result
The patch follows the waypoints

## Actual Result
straight to the drop location => crash

## Notes, Log Files, Screen Captures, Videos, etc. to Show the Issue

Thought it was a missing synchronization, but adding a wait for completion didn't help.
No open issues (expect 3d camera calibration) in issues.

    public void moveFromHeap(Nozzle nozzle) throws Exception {
        nozzle.moveToSafeZ();
        nozzle.waitForCompletion(CompletionType.WaitForStillstand);
        nozzle.moveTo(way1.derive(null, null, Double.NaN, Double.NaN), Motion.MotionOption.SpeedOverPrecision);
        nozzle.waitForCompletion(CompletionType.WaitForStillstand);
        nozzle.moveTo(way2.derive(null, null, Double.NaN, Double.NaN), Motion.MotionOption.SpeedOverPrecision);
        nozzle.waitForCompletion(CompletionType.WaitForStillstand);
        nozzle.moveTo(way3.derive(null, null, Double.NaN, Double.NaN), Motion.MotionOption.SpeedOverPrecision);
        nozzle.waitForCompletion(CompletionType.WaitForStillstand);
}

doppelgrau:
Even happens with the null motion planer, something is really strange there.

markmaker:
>  latest version

Do you mean the testing version?

If yes, there were some changes in the backlash compensation, where the `Motion.MotionOption.SpeedOverPrecision` is actually involved. Could you post a log?

_Mark

markmaker:
Or it could be something different, that just occurred to me... investigating.

doppelgrau:
Was the "develop"-branch.
Logfile will have to wait until tomorrow, maybe this evening (try to avoid crashing again after I cleaned up yesterday=> remove stuff from the table)
If it helps: the changes must have occurred in the last three month since my last PR.

markmaker:
[...]

PR #1259, opened 2021-08-23
Catches the situation if no nozzle tip is loaded

doppelgrau:
# Description
Fies a nullpointer exception

# Justification
Pushing covers of BlindsFeeder fails if no nozzle tip is loaded with an exception.

# Instructions for Use
No change

# Implementation Details
1. How did you test the change? Not yet on machine, bug #1258 is currently a show stopper. But quite simple.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java
@@ -1222,7 +1222,7 @@ public class BlindsFeeder extends ReferenceFeeder {
             if (preferredNozzle.getPart() == null) { 
                 // Nozzle is free
                 NozzleTip nozzleTip = preferredNozzle.getNozzleTip();
-                if (nozzleTip.isPushAndDragAllowed()) {
+                if (nozzleTip != null && nozzleTip.isPushAndDragAllowed()) {
                     // Return the nozzle and nozzle tip.
                     return new NozzleAndTipForPushing(preferredNozzle, nozzleTip, false);
                 }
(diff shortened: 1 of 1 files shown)


markmaker:
Good catch, @doppelgrau !

When I reviewed the code now, I found it is much better, of course, but still not water-tight. So please change the following too (just push another commit).

When it tries to load a nozzle tip, it should first check, if that nozzle tip is **not** currently loaded on **another** nozzle, especially since that other nozzle then surely has a part on, otherwise it would have been already chosen in the "Second" section. With the current code, it would then run into a "deadlock" situation, because the nozzle tip exchange would then fail, as it in turn checks for a part on the nozzle. Now, that I think about this, I sort of remember there was a user report to that effect, that I might have misunderstood after all.

https://github.com/openpnp/openpnp/blob/ac580c3c7493b37221b6890e24494aff548c6c44/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java#L1255-L1261

To something like this:

```java
                            if (pushNozzleTip.isPushAndDragAllowed()) {
                                boolean goodToGo = true;
                                for (Head otherHead : machine.getHeads()) {
                                    for (Nozzle otherNozzle :  otherHead.getNozzles()) {
                                        if (otherNozzle.getNozzleTip() == pushNozzleTip) {
                                            // Loaded on another nozzle, this means, it has a part on, otherwise 
                                            // it would already have been taken above.
                                            goodToGo = false;
                                            break;
                                        }
                                    }
                                }
                                if (goodToGo) {
                                    NozzleTip nozzleTip = nozzle.getNozzleTip();
                                    // Alas, found one for pushing, load it
                                    nozzle.loadNozzleTip(pushNozzleTip);
                                    // And return the nozzle and nozzle tip.
                                    return new NozzleAndTipForPushing(nozzle, nozzleTip, true);
                                }
                            }
```


Thanks a lot.

_Mark

markmaker:
Thanks, @doppelgrau!

Issue #1268, opened 2021-08-31
HeapFeeder: "Stiring" in the Heaps can damage parts


doppelgrau:
# Problem Report
## Describe the Issue

With larger parts (mainly 1206 capacitors, but also observed with some 0603 capacitors) the tactic to go down continuous can lead to problems if the tip gets stuck between parts. they can form quite robust structures and not move away like the smaller resistors.
As a result some force is applied to the parts that in some cases can even damage the part.

## Steps to Reproduce
1. Set up a HeapFeeder
2. fill the Heap with large capacitors (e.g. 1206 10uF 50V)
3. feed parts
4. repeat until behavior starts to get strange

## Expected Result
part gets picked without damage

## Actual Result
some parts get damaged

## Notes, Log Files, Screen Captures, Videos, etc. to Show the Issue
Will take care of the issue myself, just wanted to document it here.
Probably I'll add an option not to "stir" the parts, every time try to pick after lowering the tip (loweing, try, move up, try other location, ..., next round try it a bit lower...)

doppelgrau:
Can be closed on the merge of PR #1272

PR #1272, opened 2021-09-01
HeapFeeder Maintenance

doppelgrau:
# Description
Some maintenance on the HeapFeeder.

- Adresses bug #1268 
- Removes an exception if the template match pipeline throws an exception (template larger than are marked in the camera picture)
- Failing to get parts from the Heap now counts as a feed failure. Avoids that in some circumstances (e.g. camera failures) dozens of fetch operations are executed.
- Minor maintenance

# Justification
Improves the quality and usability of the HeapFeeder

# Instructions for Use
If a parts are stored in a Heap that tend to clog/build rigid structures where the nozzle can not move easily (e.g. 1206 capacitors), you can now enable in the settings the "Poke for Parts" option. That means that during each move Z is lifted 2 1/4 times part size (but at least one mm) before moving back down. So the nozzle tip should never apply large forces to the parts in the Heap.

# Implementation Details
1. How did you test the change? Simmulator and my machine


markmaker:
Looks good, thanks, @doppelgrau!

PR #1275, opened 2021-09-01
Feature/add maven checks for eol

tonyluken:
# Description
Adds Maven checks for file line endings that will cause the test to fail if the source code line endings are not LF (Unix) line endings.  It also adds a .gitattributes file to the repository that inhibits line ending conversions upon check-out/check-in.  And, finally it fixes the line endings in 76 *.java files.

# Justification
The repository keeps getting contaminated with various different line endings making file comparisons difficult.  This change should force everyone to use the same line endings.

# Instructions for Use
No special instructions for users.  Developers may need to update to modern tools that can use non-native line endings. 

# Implementation Details
**1. How did you test the change?** Ran the Maven test and it failed on the same set of 76 files with CRLF line endings as were found with an offline analysis.  Also deleted the local branch with this change and re-fetched it from the remote repository and verified no EOL conversion took place (prior to this change it used to convert every file to CRLF on my machine but now everything has LF line endings).
**2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?** N/A as no source code was changed other than line endings.
**3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.**  No changes were made to these packages other than correcting line endings.
**4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.**  After correcting the line endings in 76 java source files, the Maven test passes.


markmaker:
Wow! Big thanks, @tonyluken 

I gotta go to bed now, and the next days is full "real" work, so this will have to wait a moment (just so you know I'm not ignoring it).

_Mark

markmaker:
Hi @tonyluken,

I checked it out locally. 

```
git checkout -b tonyluken-feature/AddMavenChecksForEOL test
git pull https://github.com/tonyluken/openpnp.git feature/AddMavenChecksForEOL

```
And Checkstyle works! 

But git still does not force LF line-ending on checkout. Instead, I now have _all_ my files back in CRLF and lamented by Checkstyle. 

I found this solution:

https://stackoverflow.com/questions/29435156/what-will-text-auto-eol-lf-in-gitattributes-do/40977008#40977008
(No matter, it did not work then: a git bug was since fixed)

The combination between `text=auto` (which handles various extensions automatically as text files) and `eol=lf` seems to be the right thing. Now git knows when a file is a text file and will actively force it to LF, even if it already resides in the workspace. So I replaced the first directive in your `.gitattributes ` and it now reads:

```
# .gitatributes file for the OpenPnP repository

# Perform no line ending conversions
* text=auto eol=lf

# Binary files should never be converted nor diff'ed
*.icns          binary
*.png           binary
*.pxm           binary
*.wav           binary

```

Then I checked it in. And now it seems to work. When I check out `test`, I get CRLF, when I check out the new branch, I get LF:

CR_LF

Checkstyle is happy.

And that is with deliberately leaving my `core.autocrlf` on `true` (which I prefer, because I use git for other projects too).

Btw. I looked at Checkstyle config, and also blamed it back to 2017, and it seems **it never actually disallowed snake case**. The underscore is included:

```xml

<module name="MemberName">
    <property name="format" value="^[a-z][a-zA-Z0-9_]*$" />
</module>
<module name="TypeName">
    <property name="format" value="^[A-Z][a-zA-Z0-9_]*$" />
</module>
<module name="MethodName">
</module>
<module name="LocalVariableName">
    <property name="format" value="^[a-z][a-zA-Z0-9_]*$" />
</module>
```

When I remove it, there are tons of violations. Window Builder seems to generate such identifiers (numbering them _1 _2 etc,), so this is probably a no-go. 

### Next steps

Could you please test the new` .gitattributes` (above) on your side and the check it in/push it? 

After this, I think this is ready to merge, right?

_Mark

tonyluken:
Hmmm, @vonnieda was pretty adamant about not letting git make any line ending changes and I believe that is what "* -text" does.  I also think .gitattributes is supposed to override any local setting of core.autocrlf.  I have core.autocrlf set to true:
```
C:\Users\Tony\git\openpnp>git config --get-all core.autocrlf
true
true
```
 and with the .gitattributes setting of "* -text", all the local files show up with LF on my machine.  Could the problem be that you used a start point of test that already had local files with CRLF line endings?

> git checkout -b tonyluken-feature/AddMavenChecksForEOL **test**

Re: Snake case - interesting, once I included the test source files, the test was failing on the method names with multiple underscores in CalculateBoardLocationTests.java.

tonyluken:
Ah, I just noticed that in the pom, MethodName doesn't specify a format so it defaults to "^[a-z][a-zA-Z0-9]*$" - that explains why underscores are not allowed in method names.  I think I'll add that format explicitly to the pom so it's obvious to the next person that looks at it.

tonyluken:
BTW - you can see (and change) the line ending in Eclipse with [this](https://marketplace.eclipse.org/category/free-tagging/line-ending)

markmaker:
Truth be said, IMHO, git eol handling is broken. For instance, what they say here... 

https://www.git-scm.com/docs/gitattributes#_end_of_line_conversion

...is simply not true, otherwise we would not see the problems we see. We could all work happily with whatever platform specific eol there was. But somehow, somewhere, git fails and accepts CRLF files into the repo.

> not letting git make any line ending changes and I believe that is what "* -text" does. 

`-text` is the same as `binary`, according to the spec. The problem is that it seems _not_ to be effective when files have a history of being `text` and/or are already checked out and have different `eol`. If such a file was truly treated as binary, it would be replaced byte by byte, and we would see CRLF being replaced by LF as soon as you check out a branch with the `.gitattribute`. But that's not the case, **I tried**! 

> Could the problem be that you used a start point of test that already had local files with CRLF line endings?

I do not consider that "the problem" at all. It is simply unacceptable, IMHO, to require users to clean the working tree, using dangerous `rm -rf` commands, before they can check out a new branch of OpenPnP. The problem lies with git, not the user.

The `text=auto eol=lf` combination seems to work: git still knows it is a text file, and it will force working tree files to LF if needed. The end result is that no conversion will happen, _unless a legacy working tree state requires it_. I'm quite sure this exception is compatible with @vonnieda's goals. 

_Mark

tonyluken:
Ok, you’re the boss – I’ll change it to text=auto eol=lf

 

BTW – there’s a pretty good explanation of what may have been happening here Git not converting line endings to LF on commit - Stack Overflow <https://stackoverflow.com/questions/66577391/git-not-converting-line-endings-to-lf-on-commit> 

 

Tony

markmaker:
> Ok, you’re the boss

No, I'm not 🤭, and I would have liked to come across as convincing, not bossy 🤔. 

The question is: which option _does work_ better? Can you checkout `develop` for instance (which is not yet eol-corrected, and won't be for a while) and then checkout _this_ branch and everything works, back and forth?

Thanks for the [link](https://stackoverflow.com/a/66591083/15218951), it explains some interesting things.

> Unfortunately, Git checks neither core.autocrlf nor any settings in .gitattributes when deciding whether or not it needs to replace the index copy of some file. So changing either of these does not count as a "change" to any file.

On one hand, I guess that's exactly what happens, for the direction _into_ the repo. On the other hand, for the direction _out of_ the repo, it does not say, really, and I don't know why `text=auto eol=lf` then seems to work. 

I asked the author.

_Mark

tonyluken:
Oh, sorry, I didn't intend to imply you were "bossy" just that you're the one making the final decision as to what gets merged into the repository - somebody has to do it and you're that lucky one😉.

I'm still struggling to understand the git documentation - especially the last sentence of this in the description of the text attribute:

> Set to string value "auto"
> When text is set to "auto", the path is marked for automatic end-of-line conversion. If Git decides that the content is text, its line endings are converted to LF on checkin. **_When the file has been committed with CRLF, no conversion is done._**

It makes all sense to me up to that last sentence - what do you think that means?

markmaker:
> what do you think that means?

I guess that is some sort of "if it ain't broken, don't fix it" conservatism, i.e. taking the committed repo state as guidance. You had to actually convert and commit these 76+ files explicitly, to break that circle.

But if I understand correctly, once all the files are properly converted in the repo, adding `eol=lf` to the `text=auto` directive, will remove that ambiguity, not for _checkin_, but for _checkout_.  

I hope 😬

_Mark

markmaker:
Let's see how this goes.
_Mark

markmaker:
Thanks @tonyluken!

PR #1279, opened 2021-09-05
Add rectangle detection stage.

nuvotion:
Loosly based on the paper "Rectangle Detection based on a Windowed Hough
Transform" by Jung et al.

----------------------------------------------------------------------

# Description
This adds a new vision stage called DetectRectangle. It uses the HoughLines function of OpenCV and then does some further processing on the lines found. The first step is to organise lines into parallel pairs that are approximately symmetrical about the centre of the image. Pairs are then collected into quads that form rectangles. A score is applied to rectangles, based on the deviation from the expected size, along with a weighted score based on the strength of a line. Finally the corners are used to create a rotated rectangle, that can be used to get the part rotation and centre.

# Justification
I had quite a bit of difficulty getting the default bottom vision pipeline to recognise parts. This probably has something to do with the lighting/lens situation of my machine, however I've been getting really good results using this new model. It works very well for chip capacitors/resistors, as well as chips. Based on the idea that most electronic parts are rectangular.

# Instructions for Use
The module works well if it is preceded by the canny edge detector. Here is an example pipeline which I've been using:
```<cv-pipeline>                                                                   
   <stages>                                                                     
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRead" name="read" enabled="true" file="/home/kmtaylor/nuvotion/vision/radmic_images_0/cap502953625777304466.png" color-space="Bgr"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ConvertColor" name="bgr2gray" enabled="true" conversion="Bgr2Gray"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DetectEdgesCanny" name="canny" enabled="true" threshold-1="100.0" threshold-2="200.0"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DetectRectangle" name="hough" enabled="true" index="0" threshold="20" acc-weight="0.1" delta-theta="0.03" delta-alpha="0.03" delta-rho="20.0" min-rho="10.0" rho-a="15.0" rho-b="15.0"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.SetColor" name="black" enabled="true">
         <color r="0" g="0" b="0" a="255"/>                                     
      </cv-stage>                                                               
      <cv-stage class="org.openpnp.vision.pipeline.stages.DrawContours" name="draw" enabled="true" contours-stage-name="hough" thickness="1" index="-1">
         <color r="254" g="254" b="254" a="255"/>                               
      </cv-stage>                                                               
      <cv-stage class="org.openpnp.vision.pipeline.stages.MinAreaRect" name="min_area" enabled="true" threshold-min="100" threshold-max="255"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRecall" name="orig" enabled="true" image-stage-name="read"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DrawRotatedRects" name="result" enabled="true" rotated-rects-stage-name="min_area" thickness="3" draw-rect-center="false" rect-center-radius="20" show-orientation="false">
         <color r="255" g="0" b="0" a="255"/>                                   
      </cv-stage>                                                               
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageWriteDebug" name="debug" enabled="true" prefix="debug" suffix=".png"/>
   </stages>                                                                    
</cv-pipeline>
```

# Implementation Details
I have tested this code on my machine through a production run. It resulted in a substantial improvement of 0603 resistor placement. It was also effective in detecting SOT-23 parts which had low contrast on my installation.
I've tried to follow the coding style used in the other vision stages as closely as possible.
The maven test stage passes (mainly since it doesn't actually call this code)

### DIFF
--- a/src/main/java/org/openpnp/vision/pipeline/stages/DetectRectangle.java
+++ b/src/main/java/org/openpnp/vision/pipeline/stages/DetectRectangle.java
@@ -1,0 +1,311 @@
+package org.openpnp.vision.pipeline.stages;
+
+import java.awt.Color;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.ArrayList;
+
+import org.opencv.core.MatOfPoint;
+import org.opencv.core.Point;
+import org.opencv.core.Mat;
+import org.opencv.core.Scalar;
+import org.opencv.imgproc.Imgproc;
+import org.openpnp.vision.FluentCv;
+import org.openpnp.vision.pipeline.CvPipeline;
+import org.openpnp.vision.pipeline.CvStage;
+import org.openpnp.vision.pipeline.Property;
+import org.openpnp.vision.pipeline.Stage;
+import org.simpleframework.xml.Attribute;
+import org.pmw.tinylog.Logger;
+
+@Stage(description="Find expected rectangle centered on nozzle.")
+public class DetectRectangle extends CvStage {
+
+    @Attribute
+    @Property(description = "Index of rectangle to return.")
+    private int index = 0;
+
+    @Attribute
+    @Property(description = "Minimum accumulator count to be considered a line.")
+    private int threshold = 20;
+
+    @Attribute
+    @Property(description = "Weight of accumulator count that contributes to score")
+    private double accWeight = 0.1;
+
+    @Attribute
+    @Property(description = "Maximum angle deviation of lines to be considered a pair.")
+    private double deltaTheta = 0.03;
+
+    @Attribute
+    @Property(description = "Maximum angle deviation of pairs to be considered a rectangle")
+    private double deltaAlpha = 0.03;
+
+    @Attribute
+    @Property(description = "Maximum symmetry deviation of lines to be considered a pair")
+    private double deltaRho = 20;
+
+    @Attribute
+    @Property(description = "Minimum line separation to be considered a pair")
+    private double minRho = 10;
+
+    @Attribute
+    @Property(description = "Estimated pair distance A")
+    private double rhoA = 15;
+
+    @Attribute
+    @Property(description = "Estimated pair distance B")
+    private double rhoB = 15;
+
+    public int getIndex() {
+        return this.index;
+    }
+
+    public void setIndex(int v) {
+        this.index = v;
+    }
+    
+    public int getThreshold() {
+        return this.threshold;
+    }
+
+    public void setThreshold(int v) {
+        this.threshold = v;
+    }
+    
+    public double getAccWeight() {
+        return this.accWeight;
+    }
+
+    public void setAccWeight(double v) {
+        this.accWeight = v;
+    }
+    
+    public double getDeltaTheta() {
+        return this.deltaTheta;
+    }
+
+    public void setDeltaTheta(double v) {
+        this.deltaTheta = v;
+    }
+    
+    public double getDeltaAlpha() {
+        return this.deltaAlpha;
+    }
+
+    public void setDeltaAlpha(double v) {
+        this.deltaAlpha = v;
+    }
+    
+    public double getDeltaRho() {
+        return this.deltaRho;
+    }
+
+    public void setDeltaRho(double v) {
+        this.deltaRho = v;
+    }
+    
+    public double getMinRho() {
+        return this.minRho;
+    }
+
+    public void setMinRho(double v) {
+        this.minRho = v;
+    }
+    
+    public double getRhoA() {
+        return this.rhoA;
+    }
+
+    public void setRhoA(double v) {
+        this.rhoA = v;
+    }
+    
+    public double getRhoB() {
+        return this.rhoB;
+    }
+
+    public void setRhoB(double v) {
+        this.rhoB = v;
+    }
+    
+    private class HoughLine {
+        final private double rho;
+        final private double theta;
+
+        private HoughLine(double rho, double theta) {
+            this.rho = rho;
+            this.theta = theta;
+        }
+    }
+
+    private class HoughPair {
+        final private int[] lines;
+        final private double alpha;
+        final private double xi;
+
+        private HoughPair(int a, int b, double alpha, double xi) {
+            this.lines = new int[2];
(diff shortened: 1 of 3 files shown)


markmaker:
Hi @nuvotion,

Looks great! 

Some questions:

1. Why not directly return a `RotatedRect` as the result? 
    It would simplify the pipeline a lot, and seeing it only returns _one_ match, I don't see any use case for contours, or am I missing something? 
2. Could you rename it `DetectHoughRectangle`?
3. "Detect" stages should not unconditionally draw onto the image, could you add a boolean?

_Mark

nuvotion:
Hi @markmaker 

Answers to your questions:
1. No reason - The only reason it returns contours is because I based it on the DetectLinesHough stage which returns contours. I can easily change it to just output the rotated rect.
2. No problem - I'll rename it.
3. Yep - I'll add a boolean to disable drawing onto the working image. I suppose that would help slightly with performance as well.

I'll try and make these changes on the weekend and update the pull request.

nuvotion:
Ok - I've updated the pull request in response to your comments. Regarding the name change - I changed it to DetectRectangleHough - as this seemed more consistent with the other modules (like DetectLinesHough, DetectEdgesCanny etc...) Please let me know if you really wanted DetectHoughRectangle.

As you said it really simplifies the pipeline. Here is an updated example:
```
<cv-pipeline>
   <stages>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRead" name="read" enabled="true" file="/home/kmtaylor/nuvotion/vision/radmic_panel_photos/cap/cap14003197188032420850.png" color-space="Bgr"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ConvertColor" name="bgr2gray" enabled="true" conversion="Bgr2Gray"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DetectEdgesCanny" name="canny" enabled="true" threshold-1="100.0" threshold-2="200.0"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DetectRectangleHough" name="hough" enabled="true" index="0" threshold="20" acc-weight="0.1" delta-theta="0.03" delta-alpha="0.03" delta-rho="20.0" min-rho="10.0" rho-a="15.0" rho-b="15.0" draw-lines="true"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.ImageRecall" name="orig" enabled="true" image-stage-name="read"/>
      <cv-stage class="org.openpnp.vision.pipeline.stages.DrawRotatedRects" name="result" enabled="true" rotated-rects-stage-name="hough" thickness="3" draw-rect-center="false" rect-center-radius="20" show-orientation="false">
         <color r="255" g="0" b="0" a="255"/>
      </cv-stage>
   </stages>
</cv-pipeline>
```

markmaker:
Hey @nuvotion 

Great work! 

Yes, you're right, `DetectRectangleHough` is definitely better. 

The place to add a documentation link in the Wiki would be here 😁:

https://github.com/openpnp/openpnp/wiki/Computer-Vision

I, for one, would be very grateful, if somebody knowledgeable explained the Canny threshold 1 and 2 parameters in plain words. I do understand them in abstract/theoretical terms [from the doc](https://docs.opencv.org/4.5.2/d7/de1/tutorial_js_canny.html), but how to use them strategically in the context of bottom vision (for instance) is a leap that eludes me. As a consequence, I end up using trial and error, and the result then often turns out to be unstable with even slightly changing vision subjects. 

Thanks a lot,
_Mark

markmaker:
Hi @nuvotion 

Seeing you're knowledgeable in this stuff: Something different but related.

When I looked at your pipeline, I noted that [HoughLines](https://docs.opencv.org/4.5.2/dd/d1a/group__imgproc__feature.html#ga46b4e588934f6c8dfd509cc6e0e4545a) definitely needs a Canny up front. But if I understand the doc correctly, [HoughCircles](https://docs.opencv.org/4.5.2/dd/d1a/group__imgproc__feature.html#ga47849c3be0d0406ad3ca45db65a25d2d) has a Canny already built-in. This seems to be an inconsistency in OpenCV, do you agree?

Therefore, adding a second Canny in front is harmful, right? 

When I put two Cannies in sequence (in the context of a simulated strip feeder), I get this:

double edges

It makes sense: the edges as drawn have two edges, one from black to white and one from white to black. So doubling Canny will effectively double up the edges. A subsequent Hough now has two close-together edges to follow, and it will potentially "derail" between them. That's exactly what I see, when using the current OpenPnP standard pipeline:

https://github.com/openpnp/openpnp/blob/693671c6293819076b3ff8d4d14dfd29e3a9ca6e/src/main/resources/org/openpnp/machine/reference/feeder/ReferenceStripFeeder-DefaultPipeline.xml#L1-L15

hough circles derailed

Do you agree that the extra Canny stage should be removed?

When I remove it, I get this:

hough circles bang on

I have reported this before here, but it seems to have been lost in that discussion, or my point was not made clearly enough:

https://groups.google.com/g/openpnp/c/TRK5Q4vZUbc/m/M0XSA-A6AgAJ

I just want to be sure, before removing it.

_Mark

nuvotion:
Hi Markmaker - I've only just started getting my head around the opencv library - so I'm probably not able to offer too much wisdom - however I can comment on your question to some degree.

Yes, HoughCircles does do Canny edge detection internally. You can see it on line 1633 here:
https://github.com/opencv/opencv/blob/master/modules/imgproc/src/hough.cpp

As you can see the API doesn't allow you to select both thresholds, but forces a ratio of 1:2.

I agree that placing an extra Canny edge detection upstream will result in double edges like you have found.

Regarding an intuition for what the Canny thresholds actually do, like you I tend to just play with them until the resulting image looks to be picking up the edges that I want.

Thanks for merging the pull request - when I get a chance I'll try and update the wiki as you suggested, with an example of how to use the new module.

I also want to experiment with using the Generalised Hough algorithm to detect rectangles. If that's successful I'll try and do another pull request.

PR #1297, opened 2021-09-29
Add Advanced Camera Calibration as an experimental option

tonyluken:
# Description
This change allows cameras to be automatically calibrated using nothing more than a couple of fiducials (for top cameras) or just a nozzle tip (for bottom cameras).  The calibration includes location/head offsets, rotation (about the Z axis), tilt (about the X and Y axes), flips, units per pixels (at all Z levels), and lens distortion correction.

# Justification
This change unifies and simplifies (from the operator's perspective) the camera calibration process.

# Instructions for Use
At the current time, the operator can only access Advanced Camera Calibration by selecting the camera in the Machine Setup tree and selecting the Experimental Calibration tab.  A Wiki page has been setup [here](https://github.com/openpnp/openpnp/wiki/Advanced-Camera-Calibration) with further instructions. Eventually the hope is that this will be fully integrated into the Issues & Solutions system.

# Implementation Details
**1. How did you test the change?** This change was exhaustively tested on my machine by calibrating both top and bottom cameras with them mounted in varying different orientations. However, it is highly recommended that additional testing with other machines and cameras be performed to verify there are no major defects with this. 
**2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?** Yes
**3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.** No changes were made to these packages.
**4. Be sure to run `mvn test` before submitting the Pull Request.** Maven tests were run and all passed.


markmaker:
Hi @tonyluken,

This looks great, the diagnostics described on the Wiki look fantastic! 

For my understanding: I see you are pulling camera head offsets out of the calibration solution. Am I right to assume that this solves any concerns we discussed earlier? Any shift in the coordinate system due to camera tilt is now automatically compensated through the head offsets, right? Brilliant!

The only small improvement I see is moving that head offsets' calculation to the actual calibration computation and simply storing the result in the `AdvancedCalibration` (it is constant, right?). This would complete the already good separation of concerns you established, and it would remove that exception handling issue you have in there, too. 

I'm currently away for two weeks, and can't test this on the machine. But I'm really, really looking forward to it. 

Once you're ready, should I merge this anyway?

_Mark

tonyluken:
> For my understanding: I see you are pulling camera head offsets out of the calibration solution. Am I right to assume that this solves any concerns we discussed earlier? Any shift in the coordinate system due to camera tilt is now automatically compensated through the head offsets, right? Brilliant!

Well, it still doesn't solve the issue of capturing coordinates that are at different Z levels with a tilted camera prior to the camera being calibrated.  The single head offset can't compensate for the potentially many different offsets due to the tilted optical axis (assuming coordinates are captured at many different Z levels).

> The only small improvement I see is moving that head offsets' calculation to the actual calibration computation and simply storing the result in the AdvancedCalibration (it is constant, right?). This would complete the already good separation of concerns you established, and it would remove that exception handling issue you have in there, too.

It is not necessarily constant for non-default cameras if the default camera is physically touched.  It could be made a constant but that would mean forcing a recalibration of all head mounted cameras if the default camera is touched.

> Once you're ready, should I merge this anyway?

Yes, I think it is ready to be merged into the Test branch right now (it's all up-to-date with Test AFIK). I'd really like to see how it works on different machines with different cameras.

Oh, one thing I forgot to mention in the PR is that this does include my change to limit displacements to the soft limits during the backlash measurements.  If that's not ok I could pull that out and make it a separate PR.

markmaker:
> still doesn't solve the issue of capturing coordinates that are at different Z levels

Yes, but there is simply no way around this, as far as I can see. 

> It is not necessarily constant for non-default cameras if the default camera is physically touched. 

I don't understand what you mean (in fact, I also don't understand why the computation is different between cameras).

Plus, I thought some more about it, and I think this is not yet OK. The `ReferenceCamera.headOffsets`  getter and setter are asymmetric. If one presses **Accept** on the `ReferenceCameraPositionConfigurationWizard`, I expect the offsets to increment by the calibration shift, i.e. it is then doubly applied (and more times, if you repeat). The offsets exposed on the wizard should always be the _uncalibrated_ offsets. 

Now you could change the binding to the new `getUncalibratedHeadOffsets()` getter and create a setter too. But this starts to look ugly. It would be better to do it the other was around: create a new `getCalibratedHeadOffsets()` method and call it from wherever it is needed. What do you think?

> this does include my change to limit displacements to the soft limits during the backlash measurements

Please remove those. I will add them to my current round-up branch. Thanks!

_Mark

markmaker:
>>   this does include my change to limit displacements to the soft limits during the backlash measurements

> Please remove those. I will add them to my current round-up branch. Thanks!

Done in #1298 

_Mark

markmaker:
Hi @tonyluken,

I've tested your super calibration using the simulated camera imperfection of #1299 and .

I entered a simulated 5° yaw (rotation around machine Y axis) together with some pincushion distortion:

grafik

It nailed it pretty good:

grafik

My distortion simulation does not provide sub-pixel accuracy, therefore the remaining inaccuracies are expected, I guess: 

Top_2021-09-30_16 22 09 827

DISCLAIMER: This is all simulation, so the following might be **an artifact of the simulation**, i.e. it **might be my mistake**:

What I'm missing is the head offsets. They always come out as zero. With a yaw present, it should be non-zero, right?

Sure enough, after calibration the stored fiducial locations are no longer at the right coordinates:

advanced-calibration-on-off

I double-checked everything and also added #1300 to make sure the simulated fiducial 2 rendering is accurate. Note, my own distortion simulation preserves both the center and the size of the "captured" frame, i.e. I cannot think of a way how this spoils the calibration. 

Did the offsets work on physical cameras?

_Mark

tonyluken:
> I don't understand what you mean (in fact, I also don't understand why the computation is different between cameras).

I had to pick something to be the reference for the head offsets so I picked the default camera (actually the default camera minus the offset entered on its Position tab).  All other cameras are offset relative to that.  Remember when I originally conceived of this I was not thinking in terms of a fiducial mounted at a known location (I was thinking we would calibrate the camera before capturing any locations) so the head offset for the default camera would just normally be zero and any other cameras would just be offset from there.  But since you're insisting we need to capture the fiducial's location prior to calibrating the camera, I guess we could adjust the head offset to always bring the fiducial back to its original captured position.  I'll see what I can do.

It's good to see that it worked with the simulated errors.  As explained above, I'd expect the head offset of a single camera to always be zero.  Can you simulate multiple top cameras?

BTW - coming from the aerospace world, "yaw" is generally reserved for rotations about the Z axis, rotations about the Y axis are normally referred to as "pitch", and rotations about the X axis are referred to as "roll".  The normal aerospace coordinate system is X axis forward, Y axis to the right (starboard) and Z axis down.  Since OpenPnP's coordinate system is flipped and rotated WRT that, I'd be tempted to call rotation about the (OpenPnP) Y axis "roll" and rotation about the X axis "pitch" and keep "yaw" for rotation about the Z axis.

markmaker:
Hi @tonyluken 

> Remember when I originally conceived of this I was not thinking in terms of a fiducial mounted at a known location (I was thinking we would calibrate the camera before capturing any locations) so the head offset for the default camera would just normally be zero and any other cameras would just be offset from there. 

Yes, but I think we are not actually talking about the head offsets per se, we are talking about the head offsets **adjustment added on top**, i.e. the shift that happens when you switch from the "dumb" captured image half-width, half-height center to the "smart" truly perpendicular ray center your calibration can detect. I believe this is an entirely local/relative property of each camera to itself. It can be added on top of whatever "raw" head offsets the camera has according to its "dumb" center, i.e. that was calibrated using I&S earlier (or that will be calibrated later in other circumstances). 

> I guess we could adjust the head offset to always bring the fiducial back to its original captured position. I'll see what I can do.

Cool. 

> It's good to see that it worked with the simulated errors. As explained above, I'd expect the head offset of a single camera to always be zero. Can you simulate multiple top cameras?

Will do.

> BTW - coming from the aerospace world, "yaw" is generally reserved for rotations about the Z axis, rotations about the Y axis are normally referred to as "pitch", and rotations about the X axis are referred to as "roll". The normal aerospace coordinate system is X axis forward, Y axis to the right (starboard) and Z axis down. Since OpenPnP's coordinate system is flipped and rotated WRT that, I'd be tempted to call rotation about the (OpenPnP) Y axis "roll" and rotation about the X axis "pitch" and keep "yaw" for rotation about the Z axis.

Yeah, I was thinking the cockpit view is the camera view, i.e. the flight direction of the plane a.k.a. camera is pointing down Z. But of course, one can equally assume a bomb bay view. I will change it to only refer to rotation around machine Y.

_Mark

markmaker:
Note, I have since improved the simulated distortion to use anti-aliasing and Z rotation/scaling support on fiducial rendering. See `test` (merged soon).

Cam2_2021-10-01_10 28 14 485

Testing protocol using a secondary camera with these settings:

grafik

- Calibration will crash when `defaultZ` is not yet set (`null`). Calibration should check it before starting. 
    grafik

- After setting defaultZ and using the above settings, it won't converge (I have to go now, will investigate later).
grafik

- Suggestion: ask for both fiducials and diameters up front, then do the walking in one go.
- Suggestion: in the fiducial capturing step, move to the fiducial if already set on the head.
- Suggestion: initialize diameters with those stored on the head divided by UPP, if UPP is (preliminarily) set. 
- Motivation for the above: if users do the calibration multiple times, they can just cycle through, confirming each setting by pressing **Next**. 

I have to go now. 

_Mark

markmaker:
Merged #1301, you can reproduce the tests if you want.

markmaker:
Hi @tonyluken,

It seems the failure to converge can be attributed to the camera being mirrored. If I don't mirror, I get a solution:

grafik

The projection and the **Crop All Invalid Pixels** slider will always keep the original aspect ratio of a camera. This is not practical for 90° mounted cameras and for large errors. For instance, for my example, from this:

Cam2_2021-10-02_10 30 57 944

I get this at 100%=**Show all Valid Pixels**:

Cam2_2021-10-02_10 31 01 040

and all that is left at 0%=**Crop All Invalid Pixels**:

Cam2_2021-10-02_10 31 07 128

I also found and fixed a bug in my distortion simulation. Rotation around Z is not properly applied to Rotation around Y. Will create another PR.

_Mark

tonyluken:
Yeah, I hadn't thought about cameras that actually produce mirrored images - I guess that is a possibility if someone is using a mirror in their optical path like for coaxial illumination.  That's a bit tricky since it effectively introduces a left-handed coordinate system - I think I know how to fix it.

I think the slider is working exactly as I intended but I'll take a look at what it would take to flop the aspect ratio in those cases where it would increase the area of the image.

Note - rotation order is important - I've adopted the aerospace convention of rotate about Z then Y then X.  If your simulation is doing it in a different order, everything will still work ok except the rotation errors reported may be different than what was entered for the simulated rotations.

tonyluken:
> Theoretical: Now that the calibration tilt shift is expressed as head offsets, I agree. All things remain the same before/after calibration that way (at least on Zdefault). Back when you wanted to set the default camera head offsets to zero, and hide the tilt offsets, I did disagree, because it would change the frame of reference before/after calibration, and break things when revisited, and on existing machines with coordinates already captured.

Maybe I gave the wrong impression but I never (at least not for quite some time) intended to set the default camera head offsets to zero.

> Practical: I still see the issues of wanting to do backlash calibration, before doing precision camera and lens calibration. I can do backlash calibration with an imperfectly calibrated camera, but I need preliminary UPP, flip, 90° step rotation, hence the two-stage calibration approach. But you should not do a precision camera calibration without backlash calibration.

Note that I didn't say that we should calibrate the camera before backlash compensation is setup (but as you point out we could just use a worst case one-side positioning until the camera is calibrated).  I was just trying to point out that if you have a calibrated camera, it can do everything (and more) that an uncalibrated camera can do (including being used to re-visit backlash compensation).  Once you have the camera calibrated, I see no reason to ever use an uncalibrated camera for anything.

My preference would be to get rid of the two stage camera calibration - just set backlash compensation to handle a worst case (if it hasn't already been setup), calibrate the camera, and then move on to more sophisticated backlash compensation.  By eliminating the first stage of camera calibration, ReferenceCamera could be simplified considerably - no need to keep primary and secondary UPP, flips, rotation, etc. that you currently need to set in that first stage calibration. I'm not advocating we do it right now but just that's we we should be headed in the future.

markmaker:
> Maybe I gave the wrong impression but I never (at least not for quite some time) intended to set the default camera head offsets to zero.

Yes, it's been a while, I'm referring to [your pivotal statement, here](https://github.com/openpnp/openpnp/pull/1297#issuecomment-931488776):

> Remember when I originally conceived of this I was not thinking in terms of a fiducial mounted at a known location (I was thinking we would calibrate the camera before capturing any locations) so the head offset for the default camera would just normally be zero and any other cameras would just be offset from there. But since you're insisting we need to capture the fiducial's location prior to calibrating the camera, I guess we could adjust the head offset to always bring the fiducial back to its original captured position. I'll see what I can do.

Ever since then I was happy! 😁

> Once you have the camera calibrated, I see no reason to ever use an uncalibrated camera for anything.

Absolutely, I fully agree (except for this calibration itself, of course).

> My preference would be to get rid of the two stage camera calibration ... ReferenceCamera could be simplified considerably...

Yes that would simplify things, even more importantly, for the user! There are now several things you can do in two ways (or even three if you count the old manual way). UPP, nozzle head offsets, lens calibration, bottom camera position etc. Certainly doesn't make OpenPnP easier to understand or use. A cleanup will be in order.

Do you think you could automatically migrate the important existing settings (flip switches, rotation, UPP) into the new form? Existing users will probably accept to lose lens calibration on an upgrade, but it would be a bit hard to explain why they lose the basic orientation/mirroring of their camera, or the UPP needed for drag jogging etc. 

_Mark

tonyluken:
> Do you think you could automatically migrate the important existing settings (flip switches, rotation, UPP) into the new form?

I can't say for certain but think it would be possible.  The flips and rotation basically become the rectification matrix and the UPP settings drive the camera matrix and height of the camera.  The lens distortion parameters might just directly transfer.

markmaker:
Ho @tonyluken,

Testing this again, all the issues I had last time are resolved. Very nice.

### General

1. If for some reason I leave the Calibration Wizard during the calibration process, and then go back, its process state is lost. The Start Calibration button is enabled, so I could start it twice. What is worse, I cannot set the detection diameter and have to cancel the calibration when it asks for it for the secondary detection diameter after half the calibration: 

    Wizard gray

    It should somehow restore the process state when the Wizard is recreated. 

2. Why does it need the Approximate Camera Lens Z? If I'm not mistaken, that field is new(?): 

    Approx Camera Lens Z

   Is this is a starting point for the solver? 
   
   Couldn't there be a reasonable automatic guess/calculation? It is cumbersome if not outright impossible to measure (delicate diffusers are typically in the way, which may already be permanently mounted). The focal point distance can easily be calculated from the raw primary/secondary UPP. Why do you need the _lens_ Z? 

### Bottom Camera

Started to do bottom camera calibration. I first had to improve the `SimulatedUpCamera` to draw the nozzles/parts in a perspective projection. I'll make a PR.

3. Using simulation, this bangs into the edge of the camera view, because the simulated green nozzle "blob" is large (1mm). 

    Runs into the edge

   I was able to dance around this by setting the `AdvancedCalibration.testPatternFillFraction` in the `machine.xml`:

   `test-pattern-fill-fraction="0.80"`

   But I guess this should still be addressed, both in case of  large nozzle detection features and in case of narrow camera views. The `testPatternFillFraction` should be _minus_ half the detection diameter. 

4. Like for the down-looking camera, it should get/set the nozzle tip detection diameter from the configuration, if it already exists from preliminary calibration or from a previous run of the advanced calibration. You can take it from the *Vision Diameter* and convert it, if (preliminary) UPP are available:

   Nozzle Tip Vision Diameter

### Conclusion

Really looking good! I'm also amazed how, even in a purely simulated machine, you can get such "natural-looking" patterns:

grafik

_Mark

markmaker:
The SimulatedUpCamera "3D" simulation (#1321) is merged, if you want to use it for testing. 

perspective-and-blur-and-shading

_Mark

tonyluken:
> It should somehow restore the process state when the Wizard is recreated.

Yeah, I was aware of that issue but haven't come up with a good solution - I guess I need to add a state variable to AdvancedCalibration that tracks what is going on and is used to setup the Wizard correctly when it is recreated.

> Why does it need the Approximate Camera Lens Z?

Well, it's not _that_ new (it's in the Wiki pages I wrote a while back).  You are correct that it just is a starting point for the solver.  I found in some cases the solver would fail to converge or would converge to wildly incorrect solutions without this.  To me this seems like a relatively easy thing to measure and only needs to be very roughly correct - I think +/- 20% is plenty good enough.  No need to move diffusers or anything out of the way to make a more accurate measurement.  I didn't want to use the primary/secondary UPP settings because we had previously agreed that we were doing away with them and they might not even be set.

> The testPatternFillFraction should be minus half the detection diameter.

I'll work on that.

> Like for the down-looking camera, it should get/set the nozzle tip detection diameter from the configuration

I'll look into that.

tonyluken:
> Why does it need the Approximate Camera Lens Z?

After thinking about this some more, I think I have a different way to initialize the model that doesn't need the camera lens Z so I'll eliminate that input from the GUI.

markmaker:
> >   It should somehow restore the process state when the Wizard is recreated.

> Yeah, I was aware of that issue but haven't come up with a good solution - I guess I need to add a state variable to AdvancedCalibration that tracks what is going on and is used to setup the Wizard correctly when it is recreated.

That's probably hard to get right (races). Maybe it is sufficient to simply not disable the spinner. 🙈🙉🙊

_Mark

tonyluken:
> That's probably hard to get right (races). Maybe it is sufficient to simply not disable the spinner.

Yeah, that's what I've been worried about.  I suspect there are other places in the GUI where a similar sort of thing could occur.  I guess another alternative is to

markmaker:
Somehow, the message ends in the middle of the sentence?

>  I suspect there are other places in the GUI where a similar sort of thing could occur. 

Yes, it happens with other processes.

To really solve this, I guess a running process would have to register itself with the MainFrame (getter/setter). The setter could throw if an already registered process is still pending. 

_Mark

tonyluken:
Sorry about ending in the middle of a sentence.  I was intending to say another alternative would be to pop-up a dialog box if someone attempts to navigate away from the wizard while it still has some process running.  The dialog could give the operator the choice to either cancel the running process before continuing on to where they were trying go or to keep the process running and stay with the current wizard.  I know it's not as good as being able to navigate away and return but its better than the way it currently works.

tonyluken:
Hi @markmaker,

Give this latest version a try.  I think I worked-off most of the issues you had.  I also better randomized the data collection so that errors are more uniformly distributed.

markmaker:
Cool, I'll look into it tomorrow.
_Mark

markmaker:
Hi @tonyluken,

tested again and everything worked nicely. My test tilt was quite accurately detected (it found 1.408° vs. simulated 1.3° around Y, perfectly found 1.8° vs. simulated 1.8° around Z, another test found 5.447° vs 5° around Y, 1.787° vs. 1.8° around Z). 

The crop now works really nice even with portrait/landscape rotated cameras, the diagnostics are really, really great. 

I think this is ready, I'll merge it.

There is some potential for user confusion, now that this is seemingly in parallel with Issues & Solutions Calibration. Also, if your calibration is already active, mine will probably row in the empty and fail, but without telling the user. This could be very confusing and lead to frustration. 

So I propose trying to integrate it into I&S sufficiently, before announcing it big time in the group. What do you think?

Btw. do you really think it needs the full 32 radial lines? 

 **Great work!**

_Mark

markmaker:
Hi @tonyluken,

**Just for you to know:** Simulation has a bug, in that it uses the calibrated head offsets to simulate the camera view after calibration. Therefore the head offsets are doubly compensated, coordinates are wrong.

I will push a fix soon.

_Mark

markmaker:
Hi @tonyluken,

trying to wrap my head around integration your calibration into Issues & Solutions. 

My plans are like so:

1. The idea is to leave everything as it is with the `ReferenceCameraCalibrationWizard`. The users can always calibrate interactively. 
2. But as discussed earlier, a more guided option to calibrate, should also be proposed by Issues & Solutions, in the proper order of a machine setup.
3. To facilitate that, I would add an `auto` option to the `CalibrateCameraProcess`, to skip the interactive jog/spin/diameter process steps and just take the values already set through Issues & Solutions. The Process would therefore run through automatically, i.e. I could call it like a function. 
4. I intend to do this pragmatically, i.e. still use the `CalibrateCameraProcess` business logic, but just proceed to the next step automatically (while still supporting the **Cancel** button).
5. The assumption is that the  `CalibrateCameraProcess` can be run without the `ReferenceCameraCalibrationWizard`. If there are entanglements (I guess such as enabling/disabling the spinner), I will suppress them in `auto` mode (I hope there are very few).
6. Once I can call it like a function, ideally Issues & Solutions can schedule it once, quite early in the setup process, temporarily enabling worst-case backlash compensation. _Risks:_ some chicken and egg caveats will surely present themselves.

What do you think? Reasonable? Can I rely on (5) to be true?

If (6) holds, we could replace old style UPP, Transforms completely, i.e. we could remove the confusing double UI and functionality as soon as you can provide a migration function for Transforms, UPP, and lens calibration. You said earlier:

> The flips and rotation basically become the rectification matrix and the UPP settings drive the camera matrix and height of the camera. The lens distortion parameters might just directly transfer.

In fact it would be super-duper, if you could provide an API for setting up such a pseudo-calibration with the former Transform/UPP parameters. Users will no longer be able to orient/mirror cameras manually, once the Transforms are gone. They would have to be patient until the machine can move etc. i.e. until calibration becomes possible. So maybe we still want to provide minimal 90° rotation and mirroring functions through this API. 

In the Issues & Solutions fiducial location and diameter steps, I could still leave a minimal preliminary camera calibration in place, using the API. There are some practical considerations: Get the UPP to convert fiducial diameters to Length, provide the user with drag-jogging, correctly mirror/orient the cameras etc. To avoid confusion, I would no longer call the step "camera calibration" though, just "camera alignment" or so.

_Mark

tonyluken:
I don't think CalibrateCameraProcess really needs the wizard to run but you will need to override the abstract methods processRawCalibrationData() and processCanceled() similar to the way the wizard does.

I'll start thinking about what it will take to convert from the former Transform/UPP parameters.

BTW - you might want to look at the CameraWalker class in the org.openpnp.util package - it might be useful it for some of the preliminary stuff.  It can handle cameras at any arbitrary rotation.

markmaker:
Hi Tony,

I just noticed you assume a machine to have Safe Z at `0` at various places in the code. This is **not** a valid assumption, Safe Z could well be non-zero, and conversely, (although unlikely) zero could be a dangerous height, i.e. it could not be safe to move the nozzle to this height.

One example:
https://github.com/tonyluken/openpnp/blob/e52875365f40901be30ad349df9f9a04e85c4936/src/main/java/org/openpnp/gui/processes/CalibrateCameraProcess.java#L719

I'll fix it. I assume it is irrelevant for the calibration per se.

_Mark

markmaker:
Hi @tonyluken 

Another thing, I bumped into this:

https://github.com/openpnp/openpnp/blob/d5e649d5b76708a4c6ebab74c417843d4820eac6/src/main/java/org/openpnp/machine/reference/wizards/ReferenceCameraCalibrationWizard.java#L185-L187

If a bottom camera was manually 3D calibrated before (my case), or if the user changes focal plane etc. and recalibrates, the default Z will not be updated. Which results in completely inaccurate Units per Pixel.

Do you agree, that this should always be updated to the primary location Z? More specifically, it should be updated, when a new calibration was done?

_Mark

tonyluken:
Both changes look reasonable to me.

tonyluken:
Hi @markmaker,

> In fact it would be super-duper, if you could provide an API for setting up such a pseudo-calibration with the former Transform/UPP parameters. Users will no longer be able to orient/mirror cameras manually, once the Transforms are gone. They would have to be patient until the machine can move etc. i.e. until calibration becomes possible. So maybe we still want to provide minimal 90° rotation and mirroring functions through this API.

I've been successful at migrating the old upp and transform settings to the new settings (it took awhile to get it working because, confusingly, flipX means to flip vertically and flipY means to flip horizontally).  I can't really test the distortion correction migration very well because I don't have a lens with much distortion so it's difficult to tell if they'll just directly transfer but I still suspect they will.

As far as the preliminary "camera alignment" goes, I was thinking you could re-use the CameraWalker class to make the necessary measurements.  You'd just need to call its estimateScaling method (which collects four points) once over the primary calibration fiducial and once over the secondary (or for bottom cameras with the nozzle tip centered at two different heights) and pass the resulting scaling matrices along with the Z coordinates at which they were measured to an AdvancedCalibration method that does all the calculations - something like:

    /**
     * Sets-up a minimal set of advanced camera calibration parameters so that units-per-pixel 
     * scaling, flips, and rotation are approximately correct.
     * @param referenceCamera - the camera to which this calibration applies
     * @param scalingMats - an array of two or more scaling matrices as measured by the 
     * CameraWalker.estimateScaling method
     * @param calHeights - an array of the same length as scalingMats containing the Z coordinate 
     * at which the scaling matrices where measured
     * @throws Exception if less then two scaling matrices are provided or if the length of 
     * calHeights doesn't match the length of scalingMats
     */
    public void preliminaryAlignment(ReferenceCamera referenceCamera, RealMatrix[] scalingMats, 
            Length[] calHeights) throws Exception {
           .
           .
           .
}

What do you think?  Of course this requires two measurements (primary and secondary) before the camera gets "aligned."  I think I could modify this to allow passing each scaling matrix and Z coordinate one at a time so that the flips and rotation get set approximately correct on the first go (and scaling at the first height would be approximately correct as well).

markmaker:
Hi @tonyluken,

That's great news. Am I understanding this right? You can reimplement these flip and rotation setter/getters by manipulating the new camera projection on the fly? Cool!

I mentioned this already: after having put everything together, and doing real machine testing, I realized that in a realistic build a user would still want flip switches and at least 90° step rotation to make the cameras appear upright, possibly even _before_ the machine can move, i.e. before something like the walker could work. So if these flip and rotation setter/getters can be reimplemented, that's great news, but it then also follows that these setter/getters can still be used by my code. 

So I'm quite happy with the small _conceptual_ duplication that is now in the code (~50 lines is all [here](https://github.com/openpnp/openpnp/blob/0afc86dd84d7ccf0a3b0135e6d7b25daec0f2a15/src/main/java/org/openpnp/machine/reference/solutions/VisionSolutions.java#L985-L1035) and [here](https://github.com/openpnp/openpnp/blob/0afc86dd84d7ccf0a3b0135e6d7b25daec0f2a15/src/main/java/org/openpnp/machine/reference/solutions/VisionSolutions.java#L1135-L1137)). 

Frankly, I dread interfacing the  CameraWalker and Advanced Calibration just for this small task, they are mighty complex. Remember the data for the top camera calibration at two Z levels is collected in four separate I&S steps (2 x camera +  2 x nozzle Z). Today this is stateless, i.e. you can exit OpenPnP and continue later, revisit steps etc. I would still have to somehow (how?) intertwine the Auto Focus to calibrate the focal Z on the bottom camera. I currently do that _after_ the first pass, but _before_ the remaining passes for best accuracy of UPP, focal Z and vision diameter obtained for the nozzle tip. 3D calibration is also unnecessary for the bottom camera. I would also have to implement worst-case backlash compensation in the Walker. 

This is further complicated by the state machine style of the CameraWalker, and the Advanced Calibration. The underlying problem is that things are being done with blocking machine task Futures. It is neither callable from a machine task (would deadlock), nor directly from the GUI thread (would make the GUI unresponsive). I would have to wrap this into a SwingWorker or something (see P.S.). 

Summary: To make this all work without races is certainly more code than what I have now, and it would not be readable. 

_Mark

P.S.: Hint for future such programming: there is the `org.openpnp.spi.Machine.execute(Callable<T>, boolean, long)` to execute stuff on the machine no matter what. It will automatically decide whether a machine task Future is allowed/needed or not, i.e. it does the right thing regardless of the calling thread.

PR #1309, opened 2021-10-10
Feature/rotation limits

markmaker:
# Description

## Nozzle RotationMode

Introduces the `RotationMode` for nozzles, where the angular articulation between the pick and the place is subject to a modal angular offset. This serves the following purposes:

1. Support machines that have nozzles with restricted angular articulation, i.e. that cannot articulate the full 360°. Machines with a bit more than 180° (add the allowance for bottom vision alignment adjustments) are now supported. Even for machines with >= 360° articulation, but still limited range (e.g. tubing that is not free-running), the mode will reduce the angular movements to a minimum. 
2. Optimize the performance of shared rotation (`C`) axes machines: Parts are picked at angular offsets that make them pre-rotated for placement, i.e. all the nozzles are subsequently pre-rotated for alignment and placement at their shared 0° position. No large rotations are therefore needed for alignment and placement (only the small _and fast_ adjustments from bottom vision alignment offsets). The more precise pre-rotate bottom vision is therefore available without performance penalty. 
3. If none of the above applies, there is still an optimization available that simply picks the part at whatever rotation the nozzle currently has. Alignment and placement are then rotated _relative_ to that. This might speed up operation if multiple nozzles pick from the same feeder, or from feeders that are close together, i.e. where the move time between picks is not enough to rotate nozzles fast enough (e.g. in LED panel assembly). 

## Custom Angular Range

Rotation axes can now also optionally define **Soft Limits** to define a custom angular range. Formerly, the standard -180° ... +180° range was the only option. This has the following benefits:

1. Allows defining the angular range for machines with limited articulation (< 360°). See point (1) in the previous section.
2. For machines that have restricted rotation (e.g. tubing that is not free-running), a more relaxed custom range can be defined, e.g. -200° ... 200°, resulting in less "pirouetting", if the placement angle happens to be near the ±180° wrap-around limit. Alternatively, a shifted custom range, like -160° ... 200° can at least make it less likely, that a part placement angle is at the wrap-around limit (e.g. the common 180° placement angle). 
3. Allow defining the range for axes that are transformed, i.e. where 360 axis units ≠ 1 axis revolution. See next section.

## Rotation Axis Transforms

Rotation axes now support axis coordinate transformations. All the [Transformed Axes](https://github.com/openpnp/openpnp/wiki/Transformed-Axes) are supported, as long as the transformation in the articulated range is [strictly monotonic](https://en.wikipedia.org/wiki/Monotonic_function). Negation is also possible (e.g. for gear-coupled shared C axes).

All the relevant angular calculations (wrap-around at modulo 360° limits) are now properly performed in the transformed coordinate space. 

## Detailed Feature List

* Custom range rotational axis limits.
* Completely revised the way axis limits and wrap-arounds are applied in the `AbstractMotionPlanner`. It now supports transformed axes, i.e. the angular math is properly done in the `HeadMountable` coordinate space rather than in the raw axis coordinate space.
* Introduce the `RotationMode` for machines with shared C and/or limited angular articulation. An optimization mode is also available.
* Using `Nozzle.prepareForPickAndPlaceArticulation()`, the upcoming articulation must be announced before a pick-and-place-cycle is executed, so the proper `RotationMode` angular base offset can be applied to the nozzle.
* The Job Processor properly prepares the angular articulation.
* Issues & Solutions precision nozzle-to-head offset calibration properly prepares the angular articulation.
* The Discard cycle does not rotate a limited articulation nozzle.
* The Feeders Panel pick action properly prepares the angular articulation, using the **Test Alignment Angle** as the assumed placement rotation.
* Persist the **Test Alignment Angle** in `ReferenceBottomVision`, and make it publicly accessible.
* The rotation `P` button on the jog controls will now move to the 90° step angle nearest to the mid-position of a limited articulation nozzle (Note: this angle can vary, when a part is on the nozzle, i.e. when a `RotationMode` offset is applied).
* Nozzle tip calibration will respect rotation axis limits and make sure no `RotationMode` offset is currently applied.
* Introduced a `LocationOption.Quiet` option to squelch excessive chatter in the log (because angle math needs to be performed in transformed coordinate space, there are now many more transforms back and forth). 
* Added configuration error diagnostics, if the **Wrap Around** option is set on a limited articulation axis.
* Always apply limited articulation rotation mode offset, even if the articulation range is actually >= 360°. In order to center the rotation mostly around the mid-point, i.e. lessen the stress on tubes etc.

Extras:

* Added a Feed Rate per minute auto-conversion text edit to `ReferenceControllerAxisConfigurationWizard` for those that "think" in per minute units.
* Moved duplicate `getMachine()` code to base class `AbstractHeadMountable`.
* `GcodeServer`: Fixed bug in logging the parsed G-code.

# Justification
See the description.

# Instructions for Use

## Rotational Axis Limits

In the ReferenceControllerAxis configuration, as soon as **Limit to Range** is enabled, the **Soft Limit Low** and **Soft Limit High** are displayed:

ReferenceControllerAxis

The limits can be captured etc. as with linear axes. 

## Rotational Axis Transformations

You can now use axis transformations, the example shown here establishes a 1:10 scaling between the controller axis and the transformed axis: 

Mapped

The transformed axis can then be assigned to the Nozzle:

Nozzle axis mapping

## Set the Rotational Mode on the Nozzle

Also note the **Rotation Mode** that is configured on the Nozzle:

Rotation Mode

In the example above, we have a controller axis that rotates from -1 ... 19, which is 1:10 transformed to a -10° ... 190° angular range. This is more than 180° to allow for a ±20° alignment offset, but it is less than 360°. Hence, the **LimitedArticulation** mode must be configured on the nozzle. 

## Rotation Modes

The **Rotation Mode** entries work as follows:

- **AbsolutePartAngle**: the part is picked at its absolute angle, i.e. if the nozzle is at 0°, the part is rotated as drawn in the E-CAD library. This is the default and legacy mode, no angular offset is applied. The nozzle must be able to articulate 360° and more. 
- **PlacementAngle**: the part is picked with the placement angle compensated, i.e. if the nozzle is at 0°, the part is rotated as it will be aligned and placed. You must enable the **Pre-Rotate** option in bottom vision, for this to make sense. This is optimal for shared C axes. 
- **MinimalRotation**: the part is picked at whatever angle the nozzle happens to be at, i.e. no rotation is needed prior to the pick. Alignment and placement are done at _relative_ angles. This might speed up operation if multiple nozzles pick from the same feeder, or from feeders that are close together, i.e. where the move time between picks is not enough to rotate nozzles fast enough (e.g. in LED panel assembly). 
- **LimitedArticulation**: the articulation of the pick-to-placement rotation is centered around the midpoint of whatever angular range is available (taking into account a margin for alignment offsets). For the pick, the nozzle is automatically rotated towards either the minimum or maximum range limit, to allow for positive or negative relative rotation, as needed. Once the part is picked, only the placement angle is guaranteed to be reachable. This means you _must_ enable the **Pre-Rotate** option in bottom vision. 

## Working with Limited Articulation

This section describes working with a nozzle that has a limited articulation, i.e. less than 360°, and the **LimitedArticulation** rotation mode set.

Obviously you can no longer rotate the nozzle all the way. Without a part picked, the nozzle will articulate in its native range. If you try to jog beyond the range, an error is displayed. This does not apply to the camera virtual axis, so you should still be able to capture any rotation, e.g. from feeders. However, you might get errors when switching from camera to nozzle. 

Limited Rotation Error

Once a part is picked, the angular offset needed for placement articulation is taken into account (see the **LimitedArticulation** mode description). The effective _part_ angle is displayed as expected in the DRO and through the cross-hairs of the bottom camera view, i.e. the rotation offset is hidden from the user (use the log to see at what absolute angles the controller axis is driven). 

### Use Cases
 

- Inside a Job, the pick-to-place articulation is automatically controlled. 

- Pick Button The **Pick** action in the Parts and Feeders panels takes the **Test Alignment Angle** on the part alignment settings into account. 

- Pick Action
 
- The **Discard** cycle will not try to rotate the part for discard.
 
- [Issues & Solutions precision nozzle-to-head offset calibration](https://github.com/openpnp/openpnp/wiki/Calibration-Solutions#calibrating-precision-camera-to-nozzle-offsets) properly prepares the angular articulation.
 
- The rotation `P` button on the jog controls will now move to the 90° step angle nearest to the mid-position of a limited articulation nozzle (Note: this angle can vary, when a part is on the nozzle, i.e. when a `RotationMode` offset is applied).

**NOTE:** There might be other functions where limited articulation is not yet respected, especially when a rotation is applied to the nozzle were it does not actually matter. Please help with the improvement of OpenPnP and report such functions to the [discussion group](http://groups.google.com/group/openpnp). 

____
# !! HELP WANTED WITH TESTING !! 👍 💯 
____

# Implementation Details
1. Tested in simulation.
3. Changes in the `org.openpnp.spi` packages to introduce the `RotationMode`, the  `Nozzle.prepareForPickAndPlaceArticulation()` and some passed-through HeadMountables in existing methods in order to support axis coordinate transformations.

### DIFF
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -67,6 +67,7 @@ import org.openpnp.gui.support.MessageBoxes;
 import org.openpnp.gui.support.Wizard;
 import org.openpnp.gui.support.WizardContainer;
 import org.openpnp.gui.tablemodel.FeedersTableModel;
+import org.openpnp.machine.reference.vision.ReferenceBottomVision;
 import org.openpnp.model.BoardLocation;
 import org.openpnp.model.Configuration;
 import org.openpnp.model.Job;
@@ -79,7 +80,9 @@ import org.openpnp.spi.Feeder;
 import org.openpnp.spi.JobProcessor.JobProcessorException;
 import org.openpnp.spi.Nozzle;
 import org.openpnp.spi.NozzleTip;
+import org.openpnp.spi.PartAlignment;
 import org.openpnp.spi.PropertySheetHolder.PropertySheet;
+import org.openpnp.spi.base.AbstractPnpJobProcessor;
 import org.openpnp.util.MovableUtils;
 import org.openpnp.util.UiUtils;
 import org.pmw.tinylog.Logger;
@@ -644,6 +647,11 @@ public class FeedersPanel extends JPanel implements WizardContainer {
         // Do the feed an get the nozzle for the pick.
         Nozzle nozzle = feedFeeder(feeder);
 
+        // Make sure the nozzle can articulate from pick to placement. 
+        Location placementLocation = getTestPlacementLocation(feeder.getPart());
+        nozzle.prepareForPickAndPlaceArticulation(feeder.getPickLocation(), 
+                placementLocation);
+
         // Go to the pick location and pick.
         nozzle.moveToPickLocation(feeder);
         nozzle.pick(feeder.getPart());
@@ -660,6 +668,25 @@ public class FeedersPanel extends JPanel implements WizardContainer {
         }
         // The part is now on the nozzle.
         MovableUtils.fireTargetedUserAction(nozzle);
+    }
+
+    /**
+     * Create a test placement location from the discard location and the test alignment angle. 
+     * 
+     * @param part
+     * @return
+     */
+    public static Location getTestPlacementLocation(Part part) {
+        PartAlignment aligner = AbstractPnpJobProcessor.findPartAligner(Configuration.get().getMachine(), part);
+        Location placementLocation = Configuration.get().getMachine().getDiscardLocation();
+        placementLocation = new Location(placementLocation.getUnits(),
+                placementLocation.getX(), 
+                placementLocation.getY(), 
+                placementLocation.getZ(), 
+                (aligner instanceof ReferenceBottomVision ? 
+                        ((ReferenceBottomVision)aligner).getTestAlignmentAngle() 
+                        : 0.0));
+        return placementLocation;
     }
 
     protected static Nozzle getCompatibleNozzleAndTip(Feeder feeder, boolean allowNozzleTipChange) throws Exception {
--- a/src/main/java/org/openpnp/gui/JogControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/JogControlsPanel.java
@@ -62,6 +62,7 @@ import org.openpnp.spi.Head;
 import org.openpnp.spi.HeadMountable;
 import org.openpnp.spi.Machine;
 import org.openpnp.spi.Nozzle;
+import org.openpnp.spi.Nozzle.RotationMode;
 import org.openpnp.spi.base.AbstractNozzle;
 import org.openpnp.util.BeanUtils;
 import org.openpnp.util.Cycles;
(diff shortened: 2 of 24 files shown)


Shaische:
This is awesome @markmaker ! Will try to test this very soon on our [Rapid Star machine](https://github.com/deltaprintr/rapid-star). Is it possible to offset the homing with this so that C axis doesn't home where the tubes/wires are stretched/twisted and also sets that to 0 degrees? 

Example: Home and then reverse 100 degrees, set it to 0 degrees. Also allows easier accessibility for our design to switch the nozzles manually.

I realize you can do this with the HOME_TO_COMMAND but you can't set it to 0 degrees as far as I know? As a result, the crosshair in OpenPNP is off rotation and it becomes a problem if you want to use that C position to then also set up feeders, trays, etc.

markmaker:
/> //Example: Home and then reverse 100 degrees, set it to 0 degrees. /

Not sure I understand correctly, but as a guess, *yes*, this is 
possible. You can define any range and any home coordinate.

/> it becomes a problem if you want to use that C position to then also 
set up feeders, trays, etc./

You usually do such angular setup with the camera, not the nozzle. The 
camera virtual C axis is not restricted, i.e. you can correctly capture 
any angle.

It is only when you want to position the nozzle at an arbitrary camera 
location or /at any /stored location using the Positioning button (), 
that you then might hit the limit. In some cases that Error is good, as 
you then see that the location is invalid (such as in the tool changer 
locations).

The system needs the *prediction *of the articulation to correctly 
offset the nozzle. And that *prediction *is (currently) only supported 
for a *pick-to-place articulation*.

There might be more and more buttons/functions that we will be able to 
"smarten-up" to *predict (*or just ignore) the angular articulation. But 
there will always be /other /functions that simply don't support this. 
This is a physical restriction that you can't (and *sometimes don't want 
to*) make disappear completely.

_Mark


Am 10.10.2021 um 19:46 schrieb Shai:

PR #1318, opened 2021-10-26
Introduce CVPipeline management

Karlinik:
# Description

The pull request contains implementation of a new VisionSettings object and in overall tries to improves the Bottom Vision management. It aims on not just CVPipelines but the whole BottomVision to be reusable.

Originaly every change in `Parts-ReferenceBottomVision` tab created new `PartSettings` mapped to a specific `Part`. In this patch, I move all the elements of `PartSettings` to a new class `BottomVisionSettings` (BVS, implementing `VisionSettings` interface). Every instance of this class is mapped to a xml element having it unique Id and Name. For linking with a specific `Part` or `Package`, it uses "foreign key" type of reference so every `Part` or `Package` is given a new parameter `bottomVisionId`.

There is a new tab called `Vision` that serves for VisionSettings management. It gathers all the VisionSettings defined in the system and provides user with quick access and overview.

Currently the patch solves only BottomVision but in the future it should not be problem to move other Vision settings there. The Vision tab operates with general VisionSettings interface and is ready for future enhancements.

There remains the default CVPipeline encapsulated in the default built-in BVS defined on the Machine level, however, it is not editable.

# Justification

This patch aims to ease the BottomVision configuration and management. Users need to adjust the BottomVision elements to their needs quite often and in a lot of cases, especially with CVPipeline, the setup can be painful and unrealiable. I believe this patch could ease the process in the sense that users would be able to define multiple versions of BottomVision, test them easily and switch from one to another based on needs. The xml configuration could also ease the sharing between users. It is easy to share custom configuration and introduce default pipelines for each package or part.

# Instructions for Use

New BVS can be created or removed directly in the new `Vision` tab. Individual BVS can be assigned to an individual part or package directly in the `Parts/Packages` tab. The assigning is done by new column in the table with drop down, so user can easily apply a BVS by its id.
When selecting a settings it is possible to edit all the elements as it was previously done on the part level.

Alternative way is to create the settings directly in the xml file vision-settings.xml. The structure is briefly described here:
```
<openpnp-vision-settings>
    <vision-settings class="org.openpnp.model.BottomVisionSettings" id="BVS_TEST" name="Test Bottom Vision Settings" enabled="true">
        <cv_pipeline> ... </cv_pipeline>
    </vision-settings>
</openpnp-vision-settings>
```
To assing the settings to a package/part link it using a `bottomvision-id` element.

NewBottomVision

BottomVisionEdit

BottomVisionAssign

## Reset functionality and inheritance
With the new `Vision` tab and packages having the option to assign the BVS to them, the reset functionality extended and there is a new part-package-default inheritance model.

In every drop-down for the BVS selection there comes an extra empty option. When this option is selected the corresponding Part or Package BVS is set to `null`. When Part has null settings it automaticaly inherits BSV from its Package. If Package has null settings it automatically inherits the BottomVision default buil-it settings.

In the `Package-Vision` tab, there is new `BottomVision` section containing `Reset Parts` button. It resets all the parts (under that package) meaning it sets their BVS to `null` so they all inherits it from the package.
In the Vision tab, each BVS can be reset to the default one (Machine-ReferenceBottomVision). The option to reset just the CVPipeline still remains.

In the `Machine Setup-Vision-BottomVision-ReferenceBottomVision` tab there is an option to reset all the packages and parts so all the BVS elements withing parts or packages will be set to null.

# Implementation Details
1. Run the existing Unit tests and also created new ones to test the inheritance and reset model.
2. Every commit was followed by a manual tests.
2. Every change in the code was formatted using automatic formatting in IntelliJ. I tried to follow the existing naming conventions. Every change was also inspected by a basic SonarLint rules.
3. I made changes in the `org.openpnp.spi`  as I needed to create a VisionSetting class. 
4. As the configuration contains new xml file the related saving logic was extend to include these changes. The migration solution is also done.
4. Sucessfully run the existing Unit tests `mvn test` and also created new ones to test the inheritance and reset model.


markmaker:
Hi @Karlinik 

I read through everything again, but I still don't understand everything, so I guess I'll just try it out. But I can only do that over the week-end.

One big concern is migration: any solution needs to automatically migrate existing pipelines to the new system, ideally by eliminating duplicates and assigning tell-tale names. 

_Mark

Karlinik:
Hi @markmaker 

I am currently working on the migration solution. 
I was thinking about transforming current pipelines saved in the form of PartSettings to the new structures. During the loading of the configuration I will check the existence of PartSettings-Pipeline. From each, a new Pipeline will be created with ID containing the part_id it belongs to. My solution removes this PartSettings pipeline so it is a save check whether user is using old or new version of the configuration.
I am just not sure yet what to do with the 'inactive' PartSettings Pipelines, wheater to keep it there, remove it, or after transformation create a whole new configuration replacing the old one.

Different approach can be implemented using migration script. This way it could be even easier. Before loading the configuration, check for the PartSettings-Pipeline element. If found, run script that tranforms the xml elements and replace the configuration with the new one. Then proceed to loading.

Does it sound reasonable? Maybe the pipeline id should contain even more information? And is it reasonable to assign the pipeline directly to the package instead of the part?

Nikola

Karlinik:
More Im looking into the PartAlignment and PartSettings more it makes sense to create the whole BottomVision reusable. But I think my solution is like the first step and it wouldn't be hard to add other BottomVision settings.
Pipeline class could be just renamed to BottomVision and each <part-settings> entry in <part-settings-by-part-id> would be just rewritten to attributes and elements in this class. The manipulation would be also easier as each setting(BottomVision) would be part of the Part/Package directly so there would be no need for the map.

Once I'll finish with the migration, I can try to make a quick PoC of this solution if you want

Nikola

markmaker:
Hi @Karlinik,

> More Im looking into the PartAlignment and PartSettings more it makes sense to create the whole BottomVision reusable as you already suggested.

I'm glad this also makes sense to you. 

>  But I think my solution is like the first step and it wouldn't be hard to add other BottomVision settings.
Pipeline class could be just renamed to BottomVision and each <part-settings> entry in <part-settings-by-part-id> would be just rewritten to attributes and elements in this class. 

Exactly.

The most important thing that I would like to see addressed, is that there is some abstraction interface / base class that will allow for _other_ vision settings to be integrated. 

So rename your `Pipeline` into abstract `VisionSettings` or similar. Then derive a class `BottomVisionSettings` or similar. Use the [`WizardConfigurable`](https://github.com/openpnp/openpnp/blob/7f79b52ee31f8c99c7487e93eb24cc8908c22faf/src/main/java/org/openpnp/spi/WizardConfigurable.java#L5) interface to open a class specific Wizard. As a first step I'm happy with just bottom vision being adapted to the new system, but later we should add support for _any_ type of vision using a pipeline, like feeders etc. The new Vision tab should get a filter drop-down, so you can only see settings of one class.

> From each, a new Pipeline will be created with ID containing the part_id it belongs to. 

You have to be aware that users have huge part settings databases, it is the _union_ of all the parts that ever touched their machine. I was hoping that you could go through all the `PartSettings`, and group them by equal settings, i.e. all the `PartSettings` that have the same settings, including the same pipeline, should be transformed into a unique `BottomVisionSettings`. 

I suggest you use the `XmlSerialize` on the `PartSettings`, analog to this...

https://github.com/openpnp/openpnp/blob/7f79b52ee31f8c99c7487e93eb24cc8908c22faf/src/main/java/org/openpnp/machine/reference/solutions/GcodeDriverSolutions.java#L982-L985

... then use a fingerprint `DigestUtils.shaHex()` on the resulting String to map them into unique settings. You can use a `Map<String, BottomVisionSettings>`  to collect them. If an entry already exists, just add the part id to the name, so the users can later identify them. Some (passives) will probably get huge part lists, so perhaps you need to shorten them to a reasonable maximum. Users can then always beautify the names and clean up. 

This grouped migration should be much more manageable, than _one settings per part!!!_

Migration should always be active. The sheer presence of the `partSettingsByPartId` collection here...

https://github.com/openpnp/openpnp/blob/4777eac55463da9844dcaa054f20b8fe662f5c1b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java#L68

... should trigger it. 

Change that line to initialize to `null`, but for existing machines it will load the collection. Add `@Deprecated`.

Use a `@Commit` method to trigger migration. Once you're done, just set  `partSettingsByPartId` to `null`. The deprecated field needs to remain in the code, so users can migrate, but by being `null` after migration, it will no longer appear in the `machine.xml`. 

_Mark

Karlinik:
Hi @markmaker 

> So rename your `Pipeline` into abstract `VisionSettings` or similar. Then derive a class `BottomVisionSettings` or similar. Use the `WizardConfigurable` interface to open a class specific Wizard. As a first step I'm happy with just bottom vision being adapted to the new system, but later we should add support for _any_ type of vision using a pipeline, like feeders etc. The new Vision tab should get a filter drop-down, so you can only see settings of one class.
> 
I rewrote the implementation as you suggested. I've created a diagram, better than describing the structure here. `AbstractVisionSettings` and its use gives a space to create more _Vision settings_ and add it to the _Vision_ tab easily. So we are leaving the Vision tab for future extensions? Maybe, after this is done (hopefully), we can create another discussion so that the idea won't get lost.

OpenPNP-Vision

The migration solution will come in the next batch. 
As the functionality changed, I want to introduce also new unit tests so they are comming soon too.

Nikola

markmaker:
@Karlinik, brilliant!

Karlinik:
Hi @markmaker 

So the first migration solution is done. As the new `BottomVisionSettings` contains so far just the `CVPipeline` and some settings still remains in the `PartSettings` I did not deleted it during migration, I just set the pipeline to null. In the future as the rest of the attributes will move, the migration solution will have to take a slightly different form, but it should not be complicated.

I would not move the rest to BottomVisionSettings in this PR, it is now complicated enough and there are some question that need to answered first. For example the UI solution for reseting the whole `BottomVisionSettings` to default or just the pipeline or just the attributes.

For now I just want to ask a small question. What role does the `Machine->BottomVision->Pipeline` currently has? I wanted to remove it completely, but first I wanted to check here. Originally it was a place to refer for the default pipeline but currently it has no use. It can be possibly a problem, depends on the use case. If user will edit the default pipeline there is currently no way how to restore it to the original 'factory' value. It is simply treated as normal pipeline with default id. I think it is ok, but `Restore to Default` now can mean different things and it is important to make an agreement upon the final definition. There are three possible scenarios: 
1. Restore to original 'factory' default pipeline
2. Restore to pipeline with default ID
3. Restore to package pipeline (in case of `Part`)

My implementation defines it as follows:
1. In case of `Part`, restore to package pipeline (if user want to assign default pipeline, he/she can select it from the dropdown).
2. In case of Package, replace it with the default pipeline (with default id)

I can add an option to restore the pipeline to original default in the `Vision` tab, I would just keep the same id

Many thanks for the answers and suggetions.

Nikola

markmaker:
Hi @Karlinik, 

great progress there!

## Migration

> I would not move the rest to BottomVisionSettings in this PR, it is now complicated enough and there are some question that need to answered first. For example the UI solution for reseting the whole BottomVisionSettings to default or just the pipeline or just the attributes.

Hang in there @Karlinik , I know its a long haul, but its worth it! 

We really need to get rid of the `PartSettings` and the `partSettingsByPartId` collection _in one fell swoop_, otherwise it will become very messy semantically, to migrate the rest later. 

The way I understood your solution so far, I thought `BottomVisionSettings` are now a drop-downs in the Part/Package list. This means Parts will no longer have BottomVision Wizards of their own and the Packages never had. The system should really become symmetrical for Parts/Packages, so all the settings need to move to the `BottomVisionSettings` in the new Vision tab.

## Dynamic inheritance

Maybe that was wishful thinking on my part 🤪, but the way you described it, I just assumed you shared my vision of a system with dynamic inheritance: 

Both drop-downs in the list would have an empty/null entry, and if empty is selected, the Part inherits its  `BottomVisionSettings` from the Package. If on the package it is empty again, it inherits from the Default. The Default would be the `BottomVisionSettings` that is assigned in Machine -> BottomVision (which answers one of your next questions, see further below). 

Having inheritance makes this a _dynamic_ proposition _which is important!_ If you import an E-CAD project and a new resistor value of an already existing package is added, it dynamically inherits its settings from that package, simply by not having a `BottomVisionSettings` assigned (field is null). Furthermore, if you select a different `BottomVisionSettings` on the Package level, all Parts of this Package will automatically get the new setting, unless they have an exceptional `BottomVisionSettings` assignment, which is OK, because you likely don't want to lose these! (like you would through some copy-overwriting "Reset" operation). 

## Reset Functions

A **Reset** function on the Package would simply _unassign_ (set to null) the `BottomVisionSettings` in all the Parts of that Package. Another (global) **Reset** function on Machine -> BottomVision would _unassign_ the  `BottomVisionSettings` in all Packages _and_ all Parts. 

Each would of course need a stark warning and safety confirmation. 

## Default  `BottomVisionSettings`

>  What role does the Machine->BottomVision->Pipeline currently has? I wanted to remove it completely, but first I wanted to check here.

Yes, it is removed, i.e. replaced by a new default `BottomVisionSettings` to be selected from the list, i.e. a _third_ drop-down. 

**Important:** Pipelines are now exclusively a _property_ of (owned by) the `BottomVisionSettings`, i.e. as pipelines per se they  _never ever_ get "reset to all parts" or assigned, or whatever. All assignment goes _indirectly(!)_ through `BottomVisionSettings` drop-downs and/or inheritance.

In other words: we are finally moving on from "copy semantics" to proper "reference semantics", and in case of inheritance, this actually becomes "reference reference semantics". 

On the new Wizard for the `BottomVisionSettings` (in the new Vision tab) you can add a **Reset** button to reset the Pipeline to the factory default, same as today in Machine->BottomVision.

## Rest

I'm not sure I understand the rest of your text right, but my guess (and hope) is it clears itself up with the above explanations. 

If I missed something, or you have a different opinion about these things, please speak up (anybody!).

## Job Level Isolation (Future)

In the older discussion, @vonnieda strongly advocated a design where one can freeze all settings on a Job level and/or override setting, so other Jobs will not be affected. I would like to stress that this design does not at all preclude this, it is fully  [orthogonal](https://en.wikipedia.org/wiki/Orthogonality_(programming)) to it. 

Such a Job level isolation could be implemented by introducing a shadow copy of each `BottomVisionSettings` in the Job (job.xml) and an another shadow copy (Map) with the Package, Part, Default assignments of the `BottomVisionSettings`. You still get the full benefits of reference semantics and inheritance _within the Job_, while being fully isolated from other Jobs. Function could be added to recreate (reset) the shadow copy to the global default or vice versa on the Part/Packge/Default level. 

_Mark

Karlinik:
Hi @markmaker,

Thanks for the summary. I think I designed the inheritance and reset logic in the really similar manner, I just did it more complicated way.

I moved the PartSettings completely to BottomVisionSettings and I am almost done with the tests, just wanted to check one last thing:

* `VisionOffset` element in the `PartSettings` is really closely tight to logic related to `Part` so from this point of view it would make sense to move it there, but in the original code it was not part of the `Part` either. To be honest, I am not fully conviced about either (Part or BottomVisionSettings). What does make more sense to you (or others)?

Thanks!

markmaker:
Hi @Karlinik 

> VisionOffset element in the PartSettings ... What does make more sense to you (or others)?

I think there are multiple "techniques" in using OpenPnP.

I for one would put all of that on the package. IMHO, all vision related settings, including the vision offsets are a properties of the package, not an individual part. If you suddenly encounter a fluke part with different properties in vision, you'd have to create a _new_ package and assign it to that part, to handle the exception. 

But others, including @vonnieda, put all emphasis on the Part, they would even like to get rid of the whole Packages tab. For smaller projects it is certainly a big simplification in usage.

So to cover both these "techniques", the `BottomVisionSettings` becomes the neutral ground. It gets all the settings, regardless of whether it is attached to the Package or the Part. So yes, the vision offset must be moved there too. 

Realize this: The new "inheritance" design is the first step towards a system, where you can hide the whole Packages tab if you don't like it. You would then just inherit Part <- Default and not bother with Packages. Of course, we'd have to do a similar thing with nozzle tip compatibilities too, to do it. 

> Also, I got stucked on one UI thing, maybe someone can help. When Reset to Default is clicked, the BottomVisionSetting is indeed reset (values reset to default values) ...

First of all: are you sure we're on the same page with inheritance? If you want a Package or Part to inherit the default, you just unassign the `BottomVisionSettings`, right? `BottomVisionSettings` by themselves are independent. Having said that, it could make sense to overwrite one with the default settings from Machine->BottomVision. But be sure to make a clear distinction between resetting just the Pipeline to factory defaults and resetting a whole `BottomVisionSettings`.

Back to the UI problem:

All setters in OpenPnP should have their `firePropertyChange()`. I know this is not universally the case, but it really should be. 
```java
    public void setUsingLetterVariables(boolean usingLetterVariables) {
        Object oldValue = this.usingLetterVariables;
        this.usingLetterVariables = usingLetterVariables;
        firePropertyChange("usingLetterVariables", oldValue, usingLetterVariables);
    }
```

As soon as you have them, it should update the UI.

_Mark

Karlinik:
Hi @markmaker 

`BottomVisionSettings` (BVS) as a neutral ground for all the settings including vision offset sounds good, however, the offset detection should still stay in the Part BottomVision tab as it is tight to current machine selection. I am not convinced that it would be useful to put it (also) in the BVS tab.

Regarding the inheritance, yes, I think we are on the same page. If `Part` `BottomVisionSettings` is null, then it inherits the Package's one. If `Package` `BottomVisionSettings` is null, the built-in default settings (Machine->BottomVision) is used. I created new unit tests to cover this.
The the Vision tab, there are two reset option, _Reset to Default_ resets the whole BVS, _Reset Pipeline to Default_ resets just the pipeline.

Thanks for the hint for the UI, I was kind of fighting with that one.

And I think I am done, but too afraid to hit the _Ready for review_ button 😄 

Nikola

Karlinik:
The describtion was updated to match the discussed implementation and the patch was sent to review.

markmaker:
Hi @Karlinik 

Just from looking at the code online this looks quite nice! 

So I wanted to try this out and browse the code a bit using dev tools locally, but there are too many conflicts for a quick look. It seems you haven't merged the `test` branch for a while, or maybe you started with `test` being out of date? 

I'm just a reluctant user of `git` / Github myself, so I'm not really in a position to give "expert" advice. But just as a general rule, when working on a feature for a while, you need to merge the upstream intended target branch (e.g. `test`) into your local working feature branch from time to time, or it will become very frustrating to resolve the accumulated conflicts in the end. 

When you do the conflict resolving now, please be very careful. For instance, make sure that #1289 (which by nature of its intent  goes across the whole  GUI code) is not undone. 

Thanks!

_Mark

src/main/java/org/openpnp/gui/support/IdentifiableObjectToStringConverter.java
```diff
@@ -25,10 +25,7 @@
 public class IdentifiableObjectToStringConverter<T extends Identifiable>
         extends ObjectToStringConverter {
     public String getPreferredStringForItem(Object o) {
-        if (o == null) {
-            return null;
-        }
         T t = (T) o;
-        return t.getId() == null ? "" : t.getId();
+        return t == null || t.getId() == null ? " " : t.getId();
```
markmaker, line 29: I thing the `null` object should continue to be converted into the empty string. If I'm not mistaken, there is code testing against `.isEmpty()`.

Karlinik: If converted to an empty string then in the combobox it is almost invisible. The space is there intentionaly to add an empty option to the combobox with the same height as other fields. Afaik, this is just the cell renderer that sets the text as a placeholder, it should not have the impact on the value that it is representing but correct me if I am wrong.

Karlinik:
Hi @markmaker 

thanks for the check, I indeed did not merge the test branch for some time, now it should be fine. I also resolved few conflicts the Github suggested.

Now I see there are some checks failing, I will have a look.

I hope you'll like the code the same when checking locally :)

Nikola

src/main/java/org/openpnp/gui/support/IdentifiableObjectToStringConverter.java
markmaker: If it is only used in that context, then OK. But is it? Please check references.

Karlinik: I checked and tested during debugging. It is used only in Part/Package/JobPlacement table models and the text is not checked. When the `" "` value is set, the actual value is null as the null object.

markmaker: That's ok then. Thanks for checking.

src/main/java/org/openpnp/gui/VisionSettingsPanel.java
```diff
@@ -0,0 +1,203 @@
...
+                    }
+                }
+
+                for (Part part : Configuration.get().getParts()) {
+                    if (part.getVisionSettings() == settings) {
+                        usedIn.add(part.getId());
+                    }
+                }
+            }
+
```
markmaker, line 167: Should this also check against the BottomVision default?

src/main/java/org/openpnp/gui/VisionSettingsPanel.java
```diff
@@ -0,0 +1,203 @@
...
+import javax.swing.table.TableRowSorter;
+import java.awt.*;
+import java.awt.event.ActionEvent;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.prefs.Preferences;
+import java.util.stream.Collectors;
+
+public class VisionSettingsPanel extends JPanel implements WizardContainer {
+
```
markmaker, line 19: Should this extend AbstractConfigurationWizard, like all the other Wizards? 
(Note, I don't actually know this stuff very well, I just like things consistent).

src/main/java/org/openpnp/model/AbstractVisionSettings.java
```diff
@@ -0,0 +1,65 @@
+package org.openpnp.model;
+
+import org.openpnp.spi.VisionSettings;
+import org.openpnp.vision.pipeline.CvPipeline;
+import org.simpleframework.xml.Attribute;
+import org.simpleframework.xml.Element;
+
+public abstract class AbstractVisionSettings extends AbstractModelObject implements VisionSettings {
+    @Attribute()
```
markmaker, line 9: Should this implement the interface `Named` and make` getName()` an `@Override`?

src/main/java/org/openpnp/gui/VisionSettingsPanel.java
Karlinik: What do you mean?

markmaker: The Settings could be referenced here:
https://github.com/openpnp/openpnp/blob/a065d93a5987a3c7bf1963a62e51dd2dd30fd341/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java#L34
If yes, it should not allow to delete them.

markmaker:
Hi @Karlinik 

I'm now trying this locally.

First hiccup: 

`2021-12-04 19:43:36.729 SystemLogger ERROR: java.lang.Exception: Error while reading machine.xml (Element 'pipeline' does not have a match in class org.openpnp.machine.reference.vision.ReferenceBottomVision at line 430)`

You must keep old `@Elements` and `@Attributes` etc. (XML persistence), such as here:

https://github.com/openpnp/openpnp/blob/a065d93a5987a3c7bf1963a62e51dd2dd30fd341/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java#L34

Instead of _replacing_ the old field, you need to create a _new_ one. You can mark the old one as `@Deprecated`.

After migration, set it to `null`. It won't appear in the XML after that.

_Mark

markmaker:
...and btw. you should migrate the old pipeline to be the new default vision settings pipeline.

src/main/java/org/openpnp/gui/VisionSettingsPanel.java
Karlinik: `VisionSettingsPanel` follows the same pattern as other panels (Parts, Packages, ... ) in the `gui `folder. `BottomVisionSettingsConfigurationWizard `extends `AbstractConfigurationWizard` as it is the settings panel that uses the `Apply `and `Reset `buttons.

src/main/java/org/openpnp/model/AbstractVisionSettings.java
Karlinik: Good point, I didn't noticed the Named interface. Will add it.

Karlinik:
Oh, you are right, sorry for that, it completely skipped my mind. When testing the migration I created my machine.xml because my previous settings were already changed as I was testing the drafts and I must have forgot about the default pipeline as I removed it. 
Thanks! It should be fixed hopefully.

src/main/java/org/openpnp/gui/VisionSettingsPanel.java
markmaker: Ah ok!

markmaker:
Hi @Karlinik,

Good work, I like it!

Small stuff:

It should not be called "TEST" but "Default" or "Standard" or similar.

image

And if possible, make the ID column smaller.

___

The default settings (Drop-down) need to appear on the GUI somewhere here:

image

And it should be checked, when you try to delete that setting (I commented it earlier).

This is really a very nice step for OpenPnP... 😁

_Mark

Karlinik:
HI @markmaker 

> The default settings (Drop-down) need to appear on the GUI somewhere here:
> And it should be checked, when you try to delete that setting (I commented it earlier).

I thought this should reset everything, set all the `BottomVissionSetting` in Parts and Packages to `null` (almost) like it was in the original implementation.
Or maybe I don't fully follow. You mean to have the values there?
I was thinking that the default settings should be immutable. If any part was using custom default pipeline it should be migrated to a new VissionSettings so the data should not be lost. But now when I set the original default pipeline to BVS.pipeline it is quite the contrary. 
So shouldn't the default settings be just the immutable build-in default settings?

Nikola

src/main/java/org/openpnp/gui/VisionSettingsPanel.java
Karlinik: I'm still not sure I follow, sorry 😄 
So you mean to have the reference of used `BottomVissionSettings` in the `ReferenceBottomVision` instead of looping through all the parts and packages to check the usage right?

markmaker:
Hi @Karlinik,

Yes both solutions are possible. 

The way I was envisioning it, you can choose the default vision settings (drop-down where I indicated with the red arrow). The primary reason is so we can ship OpenPnP with stock vision settings. Users can then change from their (migrated) standard settings to the stock settings, and back, without losing anything. In the future I want to ship OpenPnP with multiple stock settings, for instance the rectlinear symmetry vision that I'm working on, that is tuning free. 

Now that we discussed this, I also think you should not migrate the default pipeline _into_ the stock settings, but make another, new vision setting and set it to be the default. Again so users can easily switch back and forth without losing anything. 

Another scenario is that users can optimize the settings/pipeline on a part/package and when they are particularly happy with it, they can simply choose _this_ proven vision setting as the new default, again without losing the former default setting, i.e. they can go back if it does not work out after all. 

I hope this makes sense?

In any case you need to prevent the assigned default vision setting from being deleted.

_Mark

src/main/java/org/openpnp/gui/VisionSettingsPanel.java
markmaker: Maybe I'm missing something. But from what I understand in the code, a user could delete the default  `BottomVisionSettings`, the one that is assigned here (if it is not the same time assigned to any part or package):

https://github.com/openpnp/openpnp/blob/a065d93a5987a3c7bf1963a62e51dd2dd30fd341/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java#L34

Note, when inheritance is active, the setting is `null`, i.e. the inheritance is _not_ evaluated in the in-use check loop above (and that's OK).

Karlinik:
Oh so not fill the drop down with the custom settings, but with the predefined settings (currently just the one default).
Honestly, I don't think it is in the scope of this PR. When new defaults will be created the UI should change correspondingly but now it would require quite another changes that we did not discuss. Now, when user resets everything it just sets null values and take the only one default value. If a stock of defaults is created it would require another logic how to deal with what is the default.

Regarding the migration I will undo the pipeline migration. But is it necessary to create the new settings from the default pipeline? Right now, when user opens `BottomVision` tab in `PartsPanel` or even when he start the alignment process, there is a `PartsSettings` created for every `Part` and the changes in the default pipeline won't affect it until user will restore it manually. So every change of pipeline or bottom vision is already recorded in the `machine.xml` and migrated to new `BottomVisionSettings`. I can migrate the default pipeline to a new settings, mark it as the migrated default but I don't think we need to assign it to any package/part.

Hope I described it well, there's a lot of visions, settings and pipelines already 😄 

Nikola

Karlinik:
Actually, the BVS_Test is not necessary at all. There is default settings created in the ReferenceBottomVision and original default pipeline migrated to BVS_migration_default:

image

so no need for another default pipeline, it would be useless.

src/main/java/org/openpnp/gui/VisionSettingsPanel.java
Karlinik: User cannot delete this default settings. It is the build-in one that is immutable. User can delete only the custom ones. The delete is performed upon the selection in the table of VisionSettings and the default settings in not present in that table.

markmaker:
I'm not sure we're understanding each other (?) 🤔

> Honestly, I don't think it is in the scope of this PR. 

Huh? You already have 99% in place. The only things I think are missing, are:

1. Add one stock `BottomVisionSettings` with the default pipeline from the resources, your `ReferenceBottomVision.createDefaultBottomVisionSettings()` already does that. Add it to the migration hash table.
2. Add another `BottomVisionSettings` with the pipeline that was formerly saved in `ReferenceBottomVision.pipeline`, assign it to `ReferenceBottomVision.bottomVisionSettings`. Add it to the migration hash table.
3. During `PartSettings` migration, if (2) is found to be equal, you can assign `null` to the Part, to let "default" inheritance work. So all users that frequently used "Reset to all Parts" would get a very clean migration, with only the exceptions assigned on the Part level. 
4. Expose `ReferenceBottomVision.bottomVisionSettings` on the GUI, using exactly the same drop-down that you used on the Part and the Package. The user can just choose any setting as the default.
5. In `VisionSettingsPanel.deleteSettingsAction` just check if `ReferenceBottomVision.bottomVisionSettings` is in the selection, if yes, refuse the delete. See below.
6. **Perfection Bonus:** After the `PartSettings` migration, for each Package find the most common/numerous vision setting assigned to Parts (use a `<BottomVisionSettings, Integer>` map to count them). Store the winner on the Package. Note, this could be `null`, if the default is the most common, i.e. inheritance goes _through_ to the default. After that, if not already `null`, each Part that has the same setting as the Package is set to `null`. This should result in "natural" package inheritance. I'm OK if you leave that "Bonus" point out. Maybe I'll add it later.

Point (5) could be implemented like so (note the new inner `partAlignment` loop):

```java
        @Override
        public void actionPerformed(ActionEvent e) {
            List<AbstractVisionSettings> selections = getSelections();

            List<String> usedIn = new ArrayList<>();

            for (AbstractVisionSettings settings : selections) {
                for (PartAlignment partAlignment : Configuration.get().getMachine().getPartAlignments()) {
                    if (partAlignment instanceof ReferenceBottomVision) {
                        if (((ReferenceBottomVision) partAlignment).getBottomVisionSettings() == settings) {
                            usedIn.add(partAlignment.getClass().getSimpleName());
                        }
                    }
                }

                for (Package pkg : Configuration.get().getPackages()) {
                    if (pkg.getVisionSettings() == settings) {
                        usedIn.add(pkg.getId());
                    }
                }

                for (Part part : Configuration.get().getParts()) {
                    if (part.getVisionSettings() == settings) {
                        usedIn.add(part.getId());
                    }
                }
            }

            if (!usedIn.isEmpty()) {
                String errorIds = String.join(", ", usedIn);
                MessageBoxes.errorBox(getTopLevelAncestor(), "Error",
                        "The selection cannot be deleted. It is used by " + errorIds);
                return;
            }

```

I hope this clears it up. 🤓

_Mark

markmaker:
> Actually, the BVS_Test is not necessary at all. There is default settings created in the ReferenceBottomVision and original default pipeline migrated to BVS_migration_default:
> 
> so no need for another default pipeline, it would be useless.

I diagree, see my points (1) and (2) in the last comment. These are two different settings. Yes, users can also reset the pipeline on their default setting, but that means losing the migrated pipeline. By letting user choose the default setting from the drop-down instead (see my point (4) above), they can keep the migrated setting. They can try the stock pipeline without risk.

Perhaps we should even somehow mark the stock settings as such and prevent the user from changing them at all. This way it always stays the "factory" choice, and we can even upgrade it with the OpenPnP version in the future (including especially the pipeline). 

_Mark

Karlinik:
Hi @markmaker, 

I got you, but still I think there are some things unclear and it is not like the 99% is done, there will be changes in data structures and ui, but I agree it can make the migration cleaner.

First, one scenario that's in my mind. During the migration, there can be two different part settings, one matching the build-in default pipeline, another matching the migrated default pipeline. Both corresponding parts will get `null` assigned to their `BottomVisionSettings`? If only the second Part get null, the first needs to have the new custom BottomVisionSettings assigned that will appear in the VisionTab, but then when the defaults are switched then what happens? The first will have custom BVS assigned that is exactly the same as the default right?

Regarding the (5) the `ReferenceBottomVision.bottomVisionSettings` is not in the `VisionTab` therefore it cannot appear in the delete selection. Those defaults should be immutable right? So there is no reason why they would be in the `VisionTab` table at all.

And BVS_Test ... it is completely different default pipeline, we don't need it. There will be the stock default created in `ReferenceBottomVision` from `ReferenceBottomVision-DefaultBottomVisionSettings.xml` and another default migrated from the previous default pipeline. This was just a test one to see if it gets created correctly from the `vision-settings.xml`.

Nikola

Karlinik:
Just an additional note, the way I did it in the last commit. I created new BottomVisionSettings from the previous default pipeline and put it in the VisionTab because that is the pipeline/settings that user were used to edit as the default so it appearing in the VisionTab mean it is something user can further use, edit or even delete if he/she wishes.

markmaker:
> During the migration, there can be two different part settings, one matching the build-in default pipeline, another matching the migrated default pipeline. Both corresponding parts will get null assigned to their BottomVisionSettings?

No, only the one matching the ReferenceBottomVision.bottomVisionSettings. See [my point (3)](https://github.com/openpnp/openpnp/pull/1318#issuecomment-986259317). 

> If only the second Part get null, the first needs to have the new custom BottomVisionSettings assigned that will appear in the VisionTab, but then when the defaults are switched then what happens? The first will have custom BVS assigned that is exactly the same as the default right?

Yes, that can happen. But that's actually a valid configuration: if the current ReferenceBottomVision.bottomVisionSettings is assigned to a part (or package) you are saying _"whatever happens with the ReferenceBottomVision.bottomVisionSettings later, I want this to stay unchanged"_. You are effectively freezing the choice on the part, which may be for good reason. 

Of course, we could later add a button that performs [steps (3) and (6) from the migration](https://github.com/openpnp/openpnp/pull/1318#issuecomment-986259317) whenever the user wants to, i.e. which "unfreezes" these on the part and package levels.

> Regarding the (5) the ReferenceBottomVision.bottomVisionSettings is not in the VisionTab therefore it cannot appear in the delete selection. Those defaults should be immutable right? So there is no reason why they would be in the VisionTab table at all.

I think _everything_ should appear in the table. The only thing that we should protect (disable the controls) are the built-in stock settings, i.e[ (1) from my list](https://github.com/openpnp/openpnp/pull/1318#issuecomment-986259317). Probably need to protect these from being deleted too, regardless of whether they are in use. I propose giving those a predefined ID, that makes them special. 

_Mark

Karlinik:
Hi @markmaker ,

I think we both have a slightly different solution in mind so we did not understand each other :)

I thought that every default settings that will be selectable in Machine-BottomVision tab should be immutable. But you mean to make the migrated default (MD) editable. First, I was not sure if it is the right step but maybe you are right that it would make the migration more clear and the new version would be easier to get used to. Just one note, during the migration, we are not migrating the pipelines, but the whole setting so it can happen that multiple parts will share the same pipeline but will have  other fields customed so the MD would not be used even when it maybe should have been. 

There are still some scenarious that are not really clear.

> you are saying _"whatever happens with the ReferenceBottomVision.bottomVisionSettings later, I want this to stay unchanged"_.

How would you imagine the freeze would work? Lets say MD is selected as the default, thus all the parts/packages will null value will ihnerit it. When user edits MD the corresponding parts will still have the original value? 

Next, assume that after migration we have one stock default settings, one MD and there is a part that is not in the partSettings map (is it even possible?). Should this part also inherit the MD? Otherwise there would be multiple defaults set at the same time which would be confusing.

Thanks, Nikola

markmaker:
>  But you mean to make the migrated default (MD) editable. First, I was not sure if it is the right step but maybe you are right that it would make the migration more clear and the new version would be easier to get used to. 

Yes, exactly. It was editable before, so it should remain editable. 

> we are not migrating the pipelines, but the whole setting so it can happen that multiple parts will share the same pipeline but will have other fields customed so the MD would not be used even when it maybe should have been.

Yes, that's expected behavior. The whole packet is now the unit. You can still copy/paste pipelines around via the editor. Despite the combinatorics, you will still end up with _much fewer_ distinct settings (there aren't that many combinations that make sense), so copy/pasting pipelines across will still be manageable. **And this is notably unlike before, when each part had its own copy of the pipeline, so changing a certain selection of pipelines was practically impossible!** 

> How would you imagine the freeze would work? Lets say MD is selected as the default, thus all the parts/packages will null value will ihnerit it. When user edits MD the corresponding parts will still have the original value?

Remember, this is **reference semantics**: the **choice** of the settings is freezed, but **not** the **contents** of the settings. If the users edit these contents, they will become effective for all the parts, packages and possibly the default `ReferenceBottomVision.bottomVisionSettings` i.e. _wherever_ they are assigned. 

_Mark

markmaker:
Hi @Karlinik 

This seems to have stalled... If you don't mind I'll take your version so far and implement the few things that are missing. Maybe it will become clearer. The reason is I'm blocked by this on my other two things that I have in my pipeline.

_Mark

Karlinik:
Hi @markmaker 

I am working on it, I just got stuck on one thing there. But if you have the patch prepared, you can push it here.

Nikola

markmaker:
Hi @Karlinik 

I've finally pushed my changes. Most of it already described in the last comment. 

As a take-away from the inheritance I've added a new option in OpenPnP to link the selections in the tables of the various tabs.

link tables

So when you click on a placement on a board, the part, package, vision setting, feeder will automatically be selected in their respective tabs. Or starting from any of these tabs. Obviously it can only link from the _more_ specific tables to the _less_ specific tables, although feeders and parts go both ways (the first feeder with that part will be selected, preferably an enabled one). 

Note, when you load your `machine.xml` you will get errors. I've had to remove the embedded `bottomVisionSettings` in `ReferenceBottomVision` (now also linked by ID, into the global visionSettings table). You need to delete it from the XML.

I'm quite happy with this system, thank you for bringing it so far. I really hope you don't take it the wrong way, that I kinda "kidnapped" your work, it is still the solid core in there, many of the changes you'll see are actually unrelated and long overdo improvements of how the GUI interacts (re-selection of rows after table data updates, re-selection of Wizard tabs where it was missing, etc). 😀

**Hope you can test this soon, and I'm curious to hear if this is in any bad way different from what you envisioned. Be frank!**

_Mark

P.S. There is a bugfix for limited articulation in there. See #1309. I wasn't able to test without fixing it, so I think it's OK to let it ride in this PR. The board rotation was not respected when the nozzle was being prepared for limited articulation.

Karlinik:
Hi @markmaker 

sorry for a little delayed answer, I just got to it.

Great work! A lot of enhancements and details that I didn't think of. No wrong way, I am happy I've contributed even a bit.

I like the Bottom Vision Settings tab now, looks better. Maybe we could move the _Bottom Vision - Reset Parts_ button from _Package - Vision_ tab and move it to the new _Bottom Vision Settings_ tab?

image

About the stock settings, I was thinking about adding the is_stock flag directly in the xml configuration. When a new stock settings is created it has to have the flag set as an easy way, how to create a new one without thinking about the naming convention. But the naming convention should follow anyway i guess.

From the code, I don't see anything that I would change. Unfortunately, I cannot test it on the machine, just the simulated one. But from the clicking manual testing, everything is fine so far.

I will review everything again tomorrow. Are you working on anything else there?

markmaker:
> Maybe we could move the Bottom Vision - Reset Parts button from Package - Vision tab and move it to the new Bottom Vision Settings tab?

Thanks for pointing it out, I wasn't actually aware of it. Once everything is finished (see below), I think that button is completely obsolete/replaced by inheritance. So I'll remove it. 

>  Are you working on anything else there?

The migration still needs to handle the package level inheritance, as [described in (6) here](https://github.com/openpnp/openpnp/pull/1318#issuecomment-986259317). I think I can do that over the weekend. 

Other that that, I' think it is ready. 😎

We're in the `test` branch, so we can still add finishing touches later. Let the community have a go at it, and have their say too. 

_Mark

markmaker:
>  Reset Parts button from Package 

I misunderstood what this function actually does. I will move it as you suggested.

markmaker:
Hi @Karlinik 

Again some work, as discussed above. 

The Part <- Package optimizer proved to be more difficult than anticipated. One problem was the order in which the configuration is loaded. When I was doing the migration, the part -> package ID resolution was not yet done. Took me a while to find 🤬🤬🤬. 

The Optimizer is now quite capable and can be re-run anytime: 

Optimizer

I've made things a bit cleaner too, among other things introducing `AbstractPartAlignment` and `AbstractPartSettingsHolder` for common base class stuff. 

I think this is ready for the `test` branch. 

I'll wait for your feedback then merge it!

_Mark

Karlinik:
Hi @markmaker 

Great work. When I was browsing the implementation I was thinking about some cases that could lead to duplicates or redundant settings, so the optimization will be definitely well appreciated.

From my side, u have a thumbs up! :)

markmaker:
Short intro video:
https://youtu.be/W63GbSf5BHk

Discussion:
https://groups.google.com/g/openpnp/c/BehcYrn-qhs/m/5fwyPCxvDAAJ

markmaker:
Hi,

The latest push implements the same principle for the Fiducial Locator Vision settings as for Bottom Vision. 

There is now a type filter on the Vision tab:

vision type filter

This has also revealed where the abstraction and the match-up between Machine Vision type (`ReferenceBottomVision`, `ReferenceFiducialLocator`) and Vision Settings type (`BottomVisionSettings`, `FiducialVisionSettings`) was not adequate. It was good to have this sorted out before we launch. 

Analog to Alignment (Bottom Vision), the Fiducial Locator can now be tested on the Parts and Packages tab. Just jog the machine near any matching fiducial and press **Test Fiducial Locator**:

Fid Test

_Mark

markmaker:
Nice work @Karlinik !!

Karlinik:
> Nice work @Karlinik !!

Same goes for you! Nice enhancements :) And many thanks for the cooperation :)

PR #1323, opened 2021-11-07
Feature / Chronological Ressource File Naming

markmaker:
# Description
Changes `Configuration.createResourceFile(Class, String, String)` to create files numbered by the [epoch time](https://en.wikipedia.org/wiki/Unix_time) in nanosecond, instead of random numbers. The nanosecond timer used is guaranteed to generate unique numbers (per Java VM). This makes sure files can be ordered chronologically by name, regardless of file system timestamps and their limited resolution (and whether they are preserved). 

These file names are typically generated when debug images are written in the `ImageWriteDebug` pipeline stage and from other places.

# Justification
At least Windows has limited timestamp resolution (1s), so rapidly written files will not be ordered chronologically. 

Furthermore, some tools require files ordered by name, as an example I used an online animated gif maker. 

# Instructions for Use
No change in usage. But if you go look into `$HOME/.openpnp2/org.openpnp.vision.pipeline.stages.ImageWriteDebug/` you will see files appear with chronological order names.

Note, obviously this does not apply to already existing files. Order is only guaranteed for new files. 

# Implementation Details
1. Used to create animations from `ImageWriteDebug` images.
3. The change is in the `org.openpnp.model` package.

### DIFF
--- a/src/main/java/org/openpnp/model/Configuration.java
+++ b/src/main/java/org/openpnp/model/Configuration.java
@@ -19,15 +19,18 @@
 
 package org.openpnp.model;
 
-import java.io.*;
-import java.nio.file.CopyOption;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
-import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
-import java.time.temporal.TemporalUnit;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


ian-arkver:
Hi Mark. There looks to be more happening in that commit than is described in the commit message, i.e. the includes have been cleaned/changed too?

markmaker:
Hi @ian-arkver 

Thanks for keeping an eye on things. 

Yes, I let Eclipse automatically canonize the import section. As it is clearly separated from the code and nothing bad can ever happen (or it would not compile), I think that is an acceptable "house cleaning" automatism. 

Changes will settle down, after every file has been through this once. And starting from there, we will see exactly what is added/removed, which is not the case with the many superfluous, unordered and/or wild-card imports that were often used in older code. 

Do you agree?

_Mark

ian-arkver:
Yeah, sure, that's a good target.

I only mentioned 'cos I spend most of my time lurking on various Linux mailing lists and they insist on the commits being squeaky clean - one relevant set of changes per commit, nothing else slipped in. I think that's a useful habit, since it's handy if you ever need to do a bisect.

So for example vscode likes to munge whitespace on saves, so I'd generally let it do that first and commit that with a "whitespace only" style comment, then do the actual changes I needed and commit those separately.

But it's all just personal style really I guess on a project like this which is a bit more laid back than Linux. :-)

All the very best,
Ian

markmaker:
C `#include`s are completely different animals. With a global namespace, where prior `#include`s pollute all the subsequent `#include`s, massive side effects are not the exception, but the norm. And yeah, there is a bit more prudence called for with the Linux [monolithic kernel](https://en.wikipedia.org/wiki/Monolithic_kernel) and using memory-unsafe C, as opposed to OpenPnP and using memory-safe Java 🤓

I have the same view as you on whitespace, though. Eclipse could automatically canonize that too, but that would be too disruptive to tracking back changes using `blame` & Co.

So I guess this is a somewhat subjective but balanced choice.

_Mark

PR #1332, opened 2021-11-20
Add processing to calculate axes scaling and non-squareness correction

tonyluken:
# Description
This PR just adds the processing that is needed to compute axes scaling and non-squareness correction from a series of ruler placements.  It does not include any means to collect the necessary data.

# Justification
This change is intended to be a stepping stone for later changes that will add the ability to collect the data needed.

# Instructions for Use
No changes are apparent to the operator.

# Implementation Details
1. How did you test the change? **Initially simulated the processing in Excel and added a test to Utils2DTest to simulate a data collection and processing.** 
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.  **Maven tests were run and passed.**

### DIFF
--- a/src/main/java/org/openpnp/util/Utils2D.java
+++ b/src/main/java/org/openpnp/util/Utils2D.java
@@ -26,6 +26,7 @@ import java.awt.geom.Point2D;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.commons.math3.linear.CholeskyDecomposition;
 import org.apache.commons.math3.linear.LUDecomposition;
 import org.apache.commons.math3.linear.MatrixUtils;
 import org.apache.commons.math3.linear.RealMatrix;
@@ -663,5 +664,128 @@ public class Utils2D {
 
     public static double angleNorm(double val) {
         return angleNorm(val, 45.);
+    }
+    
+    /**
+     * Computes the X-axis and Y-axis scaling and non-squareness correction factor based on a series
+     * of ruler measurements
+     * @param compensation - a 3 element array containing the X-axis scaling, Y-axis scaling, and 
+     * non-squareness correction factor in that order. This array should be initialized with the 
+     * values that were in effect when the ruler locations were captured and will be updated with 
+     * the new values when the call to this method returns. 
+     * @param rulerLocations - a list of collected ruler locations in the order they were 
+     * collected.  This should always contain two locations from each ruler placement and must 
+     * contain locations from at least three different ruler placements.
+     * @param rulerMultiple - the distance between the locations captured for each ruler placement 
+     * is expected to be an exact integer multiple of this length.  Larger values here help reduce
+     * any ambiguity in the actual distance between the locations.
+     * @param residualDistanceError - optional output parameter that contains the residual distance
+     * error for each ruler placement (pass null if not needed)
+     * @return the overall distance RMS error
+     * @throws Exception if the inputs violate any of the above restrictions
+     */
+    public static Length computeAxesScalingAndNonSquarenessCorrection(double[] compensation, List<Location> rulerLocations, Length rulerMultiple, double[] residualDistanceError) throws Exception {
+        //Validate inputs
+        if (compensation.length != 3) {
+            throw new Exception("The array passed for compensation must have exactly 3 elements.");
+        }
+        if (rulerLocations.size() < 6) {
+            throw new Exception("Must have 2 locations from at least 3 ruler placements.");
+        }
+        if (rulerLocations.size() % 2 != 0) {
+            throw new Exception("Each ruler placement must have exactly 2 locations captured.");
+        }
+        int n = rulerLocations.size() / 2;
+        
+        if (rulerMultiple.getValue() <= 0) {
+            throw new Exception("The ruler length multiple must be greater than zero.");
+        }
+        
+        if (residualDistanceError != null && residualDistanceError.length != n) {
+            throw new Exception(String.format("The array passed for residualDistanceError must have exactly %d elements.", n));
+        }
+        
+        //Decide which data to use
+        int nUsed = 0;
+        LengthUnit units = rulerMultiple.getUnits();
+        double[][] m = new double[n][3];
+        double[] d2 = new double[n];
+        RealMatrix uncompensatedDeltas = MatrixUtils.createRealMatrix(2, n);
+        RealMatrix actualDistanceMat = MatrixUtils.createRealMatrix(1, n);
+        for (int i=0; i<n; i++) {
+            Location delta = rulerLocations.get(2*i+1).subtract(rulerLocations.get(2*i)).convertToUnits(units);
+            uncompensatedDeltas.setEntry(0, i, delta.getX());
+            uncompensatedDeltas.setEntry(1, i, delta.getY());
+
+            double dxHat2 = delta.getX() * delta.getX(); 
+            double dyHat2 = delta.getY() * delta.getY();
+            double approximateDist = Math.sqrt(dxHat2 + dyHat2);
+            double actualDistance = rulerMultiple.getValue() * Math.round(approximateDist / rulerMultiple.getValue());
+            actualDistanceMat.setEntry(0, i, actualDistance);
+            
+            //Only keep data from ruler placements that have distances that are close 
+            //to an integer multiple of the specified ruler multiple
+            if (Math.abs(actualDistance - approximateDist) < rulerMultiple.getValue()/4) {
+                m[nUsed][0] = dxHat2;
+                m[nUsed][1] = dyHat2;
+                m[nUsed][2] = 2 * delta.getX() * delta.getY();
+                d2[nUsed] = actualDistance * actualDistance;
+                nUsed++;
+            }
+            else {
+                Logger.info("Distance between locations captured for ruler placement {} ({} and {}} appears to not be a multiple of {}, discarding the data for that ruler placement.", i, rulerLocations.get(2*i), rulerLocations.get(2*i+1), rulerMultiple);
+            }
+        }
+        if (nUsed < 3) {
+            throw new Exception("Data for too many ruler placements has been discarded, increase the ruler multiple being used and re-collect the data.");
+        }
+        
+        //Solve for a, b, and c
+        RealMatrix mMat = MatrixUtils.createRealMatrix(nUsed, 3);
+        RealMatrix bMat = MatrixUtils.createRealMatrix(nUsed, 1);
+        for (int i=0; i<nUsed; i++) {
+            mMat.setRow(i, m[i]);
+            bMat.setEntry(i, 0, d2[i]);
+        }
+        RealMatrix invMb = new SingularValueDecomposition(mMat).getSolver().solve(bMat);
+        
+        //Construct matrix P
+        RealMatrix pMat = MatrixUtils.createRealMatrix(2, 2);
+        pMat.setEntry(0, 0, invMb.getEntry(0, 0));
+        pMat.setEntry(0, 1, invMb.getEntry(2, 0));
+        pMat.setEntry(1, 0, invMb.getEntry(2, 0));
+        pMat.setEntry(1, 1, invMb.getEntry(1, 0));
+        
+        //Solve for matrix A
+        RealMatrix aMat = new CholeskyDecomposition(pMat).getLT();
+        
+        //Compute the residual distance error for each ruler placement as well as the overall RMS error
+        RealMatrix compensatedDeltas = aMat.multiply(uncompensatedDeltas);
+        RealMatrix bigC = compensatedDeltas.transpose().multiply(compensatedDeltas);
+        double drms = 0;
+        for (int i=0; i<n; i++) {
+            double rde = Math.sqrt(bigC.getEntry(i, i)) - actualDistanceMat.getEntry(0, i);
+            drms += rde * rde;
+            if (residualDistanceError != null) {
+                residualDistanceError[i] = rde;
+            }
+        }
+        drms = Math.sqrt(drms/n);
+        
+        //Construct the current A matrix
+        RealMatrix aMatOld = MatrixUtils.createRealMatrix(2, 2);
+        aMatOld.setEntry(0, 0, compensation[0]);
+        aMatOld.setEntry(1, 1, compensation[1]);
+        aMatOld.setEntry(0, 1, compensation[2]);
+        
+        //Compute the new A matrix
+        aMat = aMat.multiply(aMatOld);
+        
+        //Assign the axes scaling and non-squareness correction factors to the output array
+        compensation[0] = aMat.getEntry(0, 0); //scx
+        compensation[1] = aMat.getEntry(1, 1); //scy
+        compensation[2] = aMat.getEntry(0, 1); //ns
+        
+        return new Length(drms, units);
     }
 }
--- a/src/test/java/Utils2DTest.java
+++ b/src/test/java/Utils2DTest.java
@@ -1,9 +1,15 @@
 import java.awt.geom.AffineTransform;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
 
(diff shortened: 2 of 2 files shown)


markmaker:
Hi @tonyluken,

Brilliant! I will look at this, as soon as the I&S + Advanced Camera Calibration stuff is solved. 

_Mark

tonyluken:
Just adding a link to derivation of the algorithm used:
[Algorithm Derivation](https://github.com/openpnp/openpnp/files/7577154/NonSquarenessCorrection.pdf)

markmaker:
Hi @tonyluken,

I haven't forgotten this. The reason I have not started to build a solution is that I'm stuck with the _how_ and _where_ and even _why_, yes I'm procrastinating...

There are these problems:

1. Is it a problem in the first place? Non-squareness is not actually such a big problem in practice, because it is already compensated in the only place where it really matters: on the PCB. On other places it can even become a nuisance, for instance, if you have your feeders all aligned along an edge of the machine, it is more practical for them to all have the same X or Y coordinate, even if the machine geometry is actually a bit skewed. The skew does not practically matter between a few sprocket holes and the pick location.
2. The I&S/setup sequence, i.e. a chicken and egg problem: Non-squareness is such a fundamental thing that it must come very early, but practical solutions require a camera that is at least preliminary-calibrated, which means fiducials will already be captured in uncalibrated coordinates. It is even a greater problem for existing machines and their many captured locations. We would have to warn or even block existing users properly. It is easy to make situations worse than before, see (1). 
3. How to sample a number of measurements: It would require a table of measurements, and that does not integrate well with the "one-click" nature of I&S solutions. Even as a normal GUI Wizard it is hard work to implement. I guess we also want users to be able to delete/redo single (botched) measurements, a long "Process" to capture them all in one go, uninterrupted, is not practical. 
4. @crashmatt has put forth the [idea of using multiple homing fiducials](https://groups.google.com/g/openpnp/c/4vEk9DeM2zM/m/tmRlVS9ZAwAJ) for a true full-machine Affine transform, which is calibrated on each homing. This would be a completely alternative method, **if** it is practical to measure the distances of these fiducials criss-cross over the machine, i.e. do users have access to affordable reasonable precision rulers that are long enough?
5. Could you make your algorithm work with that method too? Example:
   
   multi-fiducials

   OpenPnP could then help with the measurement, i.e. precision-determine the fiducial and take a perpendicular to a ruler that is held closely to both fiducials, detecting the ticks (millimeter or 1/8th inch or whatever) and their nearer edge (ruler edge), so it can compute a precision measurement. I think such ticks can be quite robustly detected with `DetectHoughLines`. The user would enter the integral measurement (the tick to the left) and OpenPnP would add distance and sub-tick measurement to compute the _uncompensated controller coordinate_ fiducial distance. Because the vision distances are very small, it does not matter to mix them with the  _uncompensated controller coordinates_, locally.
   multi-fiducials-measurements
   Because the multiple homing fiducials would be captured prior to that, and are well known (and we could limit them to a reasonable fixed number to make the GUI simple), we can identify which fiducial-pair-measurement is taken, and easily record the measurements as a table _internally_ without making the GUI or interaction complicated. 

I'm sorry this has become somewhat of a moving target. 

_Mark

tonyluken:
1. I agree this may not be much of a problem and I don't think it is worth putting much effort into developing an automated solution - especially one that requires other things, such as the camera to be calibrated before it can be run.
2. I do not believe there is a chicken and egg problem if you keep this simple and use the operator's eyes and fingers to perform the collection.  It works with an uncalibrated camera and the machine only needs to be mechanically homed (that's probably optional as well).  Backlash probably should be compensated before this is run but that could be done with a very conservative one-sided compensation.  Really all you need is a down-looking camera that can be used to see things on the machine table and the ability to manually jog the X and Y axes (the squareness and axis scaling does need to be somewhat correct).
3. I just don't see this as a long process.  I just threw a ruler on my machine and using only the jog buttons was able to accurately capture a pair of points on it in less than a minute and a half (see images below).  For six pairs that would be about 9 minutes total.
firstPoint
secondPoint
I envision this as being pretty simple for the operator.  He/she only needs to jog the crosshairs of the camera over the end of an even numbered tick mark and click a button (and they don't even need to record what tick mark it is over).  Jog to another even numbered tick mark near the opposite end of the ruler, click the button, manually move the ruler to a new position, and repeat.  Botched measurements (like they jogged over an odd numbered tick mark) are automatically detected and tossed - that's why I'd like to see 6 pairs collected - it will still work if even half of the measurements are botched (3 pairs is the minimum needed).
4. The fiducial idea is interesting and kind of dovetails into my work with panels - every job will have a "root panel" (currently hidden from the operator) where all the boards and/or panels of the job reside.  This "root panel" is just like any other panel in that it could have fiducials and could serve the purpose needed for this.  Besides the issue you point out of being able to measure the distances between the fiducials accurately, my other concern is that it takes up valuable space on the machine table where these extra fiducials will be mounted.  
5. Yes, I believe the math can be made to work.

markmaker:
OK. I would never have thought this would be accurate enough...

tonyluken:
@markmaker,

> OK. I would never have thought this would be accurate enough...

Not sure I understand.

markmaker:
I just thought that these measurements must super-duper accurate to give the true squareness, and with our method even more so, given the added degrees of freedom. I mean, the alternative would be buying these very expensive calibrated 90° rulers/gauges or whatever they are called in English. 

_Mark

markmaker:
Hi @tonyluken,

I was thinking about this. 

You said 6 measurements. If we make it 6 "memorable" ones, it could be done in 6 guided steps in I&S, like so:

Non-Squareness-Method

Having them "memorable", you could revisit each of the six, if you think you botched one (like when you think the ruler may have shifted between the two end measurements). I could draw six simple icons to make them clear, and still tell the user to only take it as a rough guide and that it would be OK to use another position, should it be obstructed or something.

Would using such a "not-really-random" pattern be detrimental to accuracy? 

I figured the diagonals are more important to actually get the shear right, so better have two each. Conversely, the parallels to the machine axes are the least telling for any imperfection. True scale will still be derivable from the diagonals.

_Mark

crashmatt:
My problem with taking some of these measurements is that my nice long
reference ruler is too long to fit in some of these positions.
That and my machine is under a table which acts as a dust cover and
lighting control.  I would be lucky to get a 0.5mm measurement between my
fiducials.
If I had fiducials closer together I could get my 150mm digital calipers at
them.  Then I have a bunch of fiducials fixed in the middle of some prime
real estate.  A set of 1000mm calipers would work but could cost more than
the machine.

Getting some kind of precision square reference is not so expensive.  This
flat metal square
<https://www.amazon.se/Stanley-Platt-vinkel-600-400/dp/B00109FZUA/ref=sr_1_29?crid=39ZX0EW1DB3WT&keywords=precision+90+degree+square&qid=1659719938&sprefix=precision+90+degree+square%2Caps%2C78&sr=8-29>
seems
a good choice.
Problems with using it.
1. Getting it square with an axis to true against the other axis
2. Knowing if the camera is aligned with it.  A shorter reference can be
used if the measurement is more accurate.

Having a wizard to collect data points and correct for squareness should be
easy(ish?)
Having machine vision to home in on the line of the square reference would
make it even better.  This way you average the position over an axis of the
camera. Don't try to capture the corner which is difficult.  Capture two
points on each edge.
There will be some residual error caused by the ruler not being square with
the machine.  Say it is at 5 deg to your fixed axis.  If I have my thinking
right then there is a ~10% residual non squareness error.  This is ok.  Now
the machine is squarer, repeat the process.  Keep repeating until you have
a reasonable degree(literally) of confidence.
For extra bonus points, do the measurement in multiple locations on the
machine or along the reference edge to check for the machine being straight.

Now say that you are using a 100mm*100mm PCB to true the machine.  No
guarantees that the manufacturer made it square or straight. Maybe we can
use fiducials instead of edge cuts.  Load the PCB fiducals and use those to
get the machine square?

What if the PCB mask is not square?  What happens if that reference PCB
gets rotated?  Think of doing this with a 45deg angle.  It measures 45
degrees in one orientation.  It should measure the same angle in the other
orientation.  If the machine is not square, does it measure the same angle?
Does the reference item need to be square at all?

These are pretty random thoughts and any or all of it could be wrong.

/Matt

tonyluken:
@markmaker,

I think it would be better to collect the points with the ruler at (approximately) 0, 30, 60, 90, 120, and 150 degrees.  More of a star than a square.

crashmatt:
*Now say that you are using a 100mm*100mm PCB to true the machine.  No
guarantees that the manufacturer made it square or straight. Maybe we can
use fiducials instead of edge cuts.  Load the PCB fiducals and use those to
get the machine square?*

Which is pretty much what tony wrote earlier.  Need to read through
everything again.

/Matt

crashmatt:
I think rotating angle method I just described is pretty much the same as
tony suggested.  One uses an angle reference and his uses a length
reference.  They both need the reference to be rotated.

/Matt

tonyluken:
My method doesn't require any specific angles (and the angles of the ruler placements with respect to each other or to the machine don't need to be known) although it helps for them to be all as different as possible, i.e., collecting two or more sets at about the same angle doesn't help much.

crashmatt:
I don't think an exact angle reference is required.

Take a reference angle and rotate it 180deg.
[image: image.png]

Now split the difference between the two angles to get the perpendicular.
[image: image.png]

In reality the measurement will look more like this.  That is ok.  We can
do small rotations to bring the measurements back to a common vector.
Now we have a perpendicular reference.
[image: image.png]

When the machine is not square there is an offset to the distance between
the reference angles.
[image: image.png]

This looks independent of steps/mm.  It has influence on the rotation
correction which is a 2nd order problem.

I think this is very similar to using reference lengths. Say you get three
equal lengths to make a equilateral triangle.  Now a skewed reference
splits one of the lengths with an offset.
[image: image.png]

That is unless I have fallen for a trick and these corrections are not
possible with a non square reference.

/Matt

markmaker:
@crashmatt,

the images did not work.

I'm note entirely sure what you are proposing here, but there is  [@tonyluken's document about the algorithm we are discussing](https://github.com/openpnp/openpnp/files/7577154/NonSquarenessCorrection.pdf), and my later proposal with the fiducials was only a _side-track_, that I guess is (literally) off the table again, because (as you confirm) rulers are not long enough and the implementation is too complex.

_Mark

tonyluken:
@markmaker,

I think the fiducial idea still has merit when someone already has a machine up and calibrated and then wants to make changes to it that may affect its squareness and/or axes scaling but also wants to preserve locations that were previously captured.  The machine itself can be used to measure the fiducial locations both before and after the changes.  

I think the ruler idea works better on a machine that may never have been calibrated, i.e., during initial setup.  This method could also be used after a machine is calibrated but you potentially lose all previously captured locations.  Of course, if you have the axes transformations that were in effect when the previously captured locations were actually captured, and you have the new axes transformations, it should be possible to update all those captured locations to their new coordinates.

Tony

markmaker:
@tonyluken, yes that's how I meant it. I tried to combined the two but that's not a good idea/off the table.

crashmatt:
I can see the images in my mail but not the github pull request.

markmaker:
> I can see the images in my mail but not the github pull request.

I can see them in neither.

_Mark

tonyluken:
@crashmatt,

I can't see your images either.

crashmatt:
A side topic:  I have a working squareness compensation using my suggested method.  It needs to collect multiple points of data, similar to that required by the Tony.  Where should the user interface be?

1. Machine Setup -> Axes -> ReferenceLinearTransformAxis
2. Machine Setup -> Heads -> ReferenceHead
3. Issues and Solutions

#1 Has difficulty accessing the head Camera. Might be solvable.  Not my preferred choice. 
#2 No feasibility tested yet.
#3 Doesn't seem to clearly support multi-step user input.  Better for fully automated processes?

Any preferences?

This item should have an entry in issues.  Maybe in the calibration milestone before advanced camera calibration.  No point trying to fix the camera if the machine axis might rotate.

markmaker:
Hi @crashmatt 

I don't understand. I guess it would help if you uploaded/fixed the missing images for illustration.

After @tonyluken's feed-back, I was planning something like this:

1. Six new Issues & Solutions calibration steps.
1. Each captures two points from the ruler held at 30° steps.
1. To make the six captures unique in their role, I would include instructions referring to clock hands at the 6 positions, e.g. 
   _"Hold the ruler so it is roughly at an angle from 7 o'clock to 1 o'clock. The angle does not need to be precise, and you can place the ruler anywhere on the machine. It must be placed on an even surface/support structure. It must be at the Z level of the PCB, and the ruler must be quite thin/same thickness as a PCB. If this is not possible, all measurements must be at the same Z. In this case you can't use the results to set the scale of the axes."_ 
1. After getting the results, the user would have the choice of _not_ applying the transform at all, of applying _only_ the shear, or of applying both the shear _and_ the scale. With two-sub-variants of taking X or Y as the lead axis (i.e. the shear would be applied to the other one). 
1. The resulting non-squareness would be automatically applied to a pair of `ReferenceLinearTransformAxis`, which would be created on demand, and all head-mountables rewired. And vice versa, if no transform is applied, the head-mountables would be rewired to their raw axes and the `ReferenceLinearTransformAxis` removed.
1. The solutions can be re-opened to re-capture single measurements, and/or to revise the decision of how to apply.

_Mark

crashmatt:
Does this put all of the 6 new issues in the common issues list?  That would be a little messy.

Will keep my method descriptions to a new PR.  I need to write a wiki anyway.

tonyluken:
@markmaker,

>"Hold the ruler so it is roughly at an angle from 7 o'clock to 1 o'clock. The angle does not need to be precise, and you can place the ruler anywhere on the machine. It must be placed on an even surface/support structure. It must be at the Z level of the PCB, and the ruler must be quite thin/same thickness as a PCB. If this is not possible, all measurements must be at the same Z. In this case you can't use the results to set the scale of the axes."

I don't think you need to be so restrictive here.  The only real requirements are that both ends of the ruler be at the same Z level, it needs to be in good focus, and it can't move while the two points are captured.  There is no requirement for the different ruler placements to be at the same Z level nor do any of them need to be at the PCB Z level.  The scaling will be correct regardless of the Z levels used.

I think I'd word it something like, "Place the ruler on the machine at roughly the 7 o'clock to 1 o'clock angle. The angle does not need to be precise, and you can place the ruler anywhere on the machine. It must be placed on an even surface/support structure so that both ends are at the same Z level and in good focus of the down-looking camera.  It is important that it does not move or slip while the machine is jogged."

Tony

markmaker:
@crashmatt 
> Will keep my method descriptions to a new PR. I need to write a wiki anyway.

Cool, but you'll finish the other PR, right 😉?

> Does this put all of the 6 new issues in the common issues list? That would be a little messy.

Yes. Note these steps must be done very early, as was discussed before. They must also be blocked for existing users, as changing the squareness invalidates all captured locations! It would therefore likely be a separate Milestone "Machine Squaring", where only these 6+1 issues would be presented. Unlike other solutions, they would _only_ appear when inside this Milestone, i.e. they would not be inherited into the following Milestones, when left open. So I don't think it would be particularly "messy".

For the _general_ user, I imagine these measurements are not as easy as they sound. The ruler must absolutely not rotate or move between the two shots. it must no bend over an obstacle etc. Even the jogging might shake the machine too much (I'm talking true DIY). Weighing or taping down the ruler has its own potential complications. Some machines merely have some kind of "frame" rather than a flat table. 

@tonyluken 
> I don't think you need to be so restrictive here. 

Thanks, and of course you are right, as all the Z effects (camera tilt and whatnot) are cancelled out.

_Mark

crashmatt:
@markmaker, Got it.  Makes sense.
The other method might be easier to use.  It is pretty much just three fiducials on a piece of paper.  The machine can do a consistency check after every initial measurement to check the paper has not moved.  It can't learn steps/mm like the "Tony" can.  Might be good to give users the choice of methods.

The other PR is taking a short break for me to do something creative.  I have to complete it to get that feature into upstream.

PR #1335, opened 2021-11-24
Feature/Issues & Solutions: Integrate Advanced Camera Calibration

markmaker:
# Description
This PR integrates the Advanced Camera Calibration by @tonyluken (#1297) into the Issues & Solutions machine setup process.

Issues & Solutions

## Feature List

- Advanced Camera Calibrations is available as regular Issues & Solutions calibration steps.
- The invoked calibration is automatic, i.e. it does not interactively ask the user for fiducial location/diameters etc. Instead it reuses the already setup locations/diameters etc. from the preliminary calibration rig. 
- Camera feedback does no longer lag behind, camera image and cross-hairs do match. Furthermore, markers are drawn with anti-aliasing and sub-pixel accuracy:

   Screenshot_20211124_223601

- The classical Transforms and Units per Pixel wizard settings are grayed out when the Advanced Camera Calibration is active.
   
     Grayed out

- The preliminary calibration is locked, when the Advanced Camera Calibration is active:

   I&S blocked

- Default Z is now always set to the primary location Z when a new calibration is applied. 

# Justification
The Advanced Camera Calibration should be available like all the other automatic calibration steps, so users need not learn a new interface.

# Instructions for Use

See above and the Wiki:

https://github.com/openpnp/openpnp/wiki/Calibration-Solutions#advanced-camera-calibration

# Implementation Details
1. Tested in simulation.

### DIFF
--- a/src/main/java/org/openpnp/gui/IssuesAndSolutionsPanel.java
+++ b/src/main/java/org/openpnp/gui/IssuesAndSolutionsPanel.java
@@ -565,7 +565,7 @@ public class IssuesAndSolutionsPanel extends JPanel {
 
 
     /**
-     * 
+     * Rebuild the UI as needed, when solutions have changed state, perhaps asynchronously.
      */
     public void solutionChanged() {
         SwingUtilities.invokeLater(() -> {
--- a/src/main/java/org/openpnp/gui/MainFrame.java
+++ b/src/main/java/org/openpnp/gui/MainFrame.java
@@ -877,8 +877,10 @@ public class MainFrame extends JFrame {
     }
 
     public void hideInstructions() {
-        scheduledExecutor.shutdown();
-        scheduledExecutor = null;
+        if (scheduledExecutor != null) {
+            scheduledExecutor.shutdown();
+            scheduledExecutor = null;
+        }
         panelInstructions.setVisible(false);
         doLayout();
     }
--- a/src/main/java/org/openpnp/gui/processes/CalibrateCameraProcess.java
+++ b/src/main/java/org/openpnp/gui/processes/CalibrateCameraProcess.java
@@ -20,8 +20,12 @@
 package org.openpnp.gui.processes;
 
 import java.awt.Color;
+import java.awt.Graphics2D;
+import java.awt.RenderingHints;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.awt.geom.Ellipse2D;
+import java.awt.geom.Line2D;
 import java.awt.geom.Point2D;
 import java.awt.image.BufferedImage;
 import java.util.ArrayList;
@@ -29,13 +33,13 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Random;
 import java.util.concurrent.Future;
+
 import javax.swing.SwingUtilities;
 import javax.swing.SwingWorker;
 
 import org.opencv.core.KeyPoint;
 import org.opencv.core.Mat;
 import org.opencv.core.Size;
-import org.opencv.imgproc.Imgproc;
 import org.openpnp.gui.MainFrame;
 import org.openpnp.gui.components.CameraView;
 import org.openpnp.gui.components.CameraViewFilter;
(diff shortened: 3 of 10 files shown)


markmaker:
Hi @tonyluken,

I hope you can give this a spin soon. 

And then it is time for you to earn a big trophy, by announcing this formidable solution 🏆!

😄😄😄

_Mark

markmaker:
Hi @tonyluken,

I was testing the calib on my machine. I had to fix a race, see #1338.

The results are sobering, I mounted the cameras with so much tilt, it is embarrassing!! It is cropping away a quarter of my view. For long focal lengths, this is certainly hard. 😨

During testing, I started to wonder if the shuffling really makes sense. From the perspective of the points, the approach vectors are still very one-sided, there is only limited angular variance, not a game changer compared to the old radial walking. I don't think this will cancel out any backlash or similar. 
 
Also, it takes quite a long time to initialize the lines, and although it seems to only prepare data(?), it moves in a tiny star shape, and in case of the bottom camera, rotates the nozzle 180° back and forth. Obviously the actual walking then also takes much more time, when shuffled. 

I wonder if calibrated backlash compensation would not be sufficient (in the I&S sequence this is guaranteed). If this is still insufficient, one could temporarily replace it with a worst-case, one-sided backlash compensation and slow speed. 

Or, if randomization is strictly needed, one would need two-segment approaches, with the knee-points equally distributed around the target-points. 

What do you think? 

I tried to restore the old radial walking, at least as an option. But the code is not trivial, and I rather not break anything. 

**If you agree that shuffling is not optimal, would you please give me a hand?** 🙃

_Mark

tonyluken:
Hi @markmaker,

Thanks for catching (and fixing) the race conditions and other bugs you've found.

Yeah, it was somewhat surprising as to how much camera tilt we may have been living with - I guess that is why it was so important to have everything mounted as close to the same Z level as possible - hopefully that restriction can now be relaxed.

The shuffling is mainly to deal with errors that accumulate with time and not backlash type errors.  For instance, my camera's scaling tends to drift with temperature.  It takes 10 to 15 minutes for it to stabilize when I first power-up my machine.  It's not a large drift but it is definitely noticeable if I calibrate the camera right after power-up.  The shuffling spreads that error out rather than having it concentrate in a few areas allowing for a better model fit to the collected data.

WRT initializing the lines:  Both the old and the new process initialize all the lines at the center of the image.  It's just that the new process initializes all of the lines at the start while the old process did it one line at a time so in reality there is no difference as to the number of points captured between the two processes (ok, I did add one extra machine movement per line in the new process but I just don't think that adds much to the overall timeline).  Since the initialization is done at the center of the image where distortion and scaling errors are lowest, collecting the data there for all the lines first before "walking" on any of the lines also gives more time for things to settle (like the aforementioned temperature drift).

I grant you that the walking time will be longer for the new shuffled process but I didn't find it to be much more.  I found that the collection time was mostly dominated by the camera capture and detection of the fiducial/nozzle tip rather than the machine movement time.  I guess that could be different depending on your camera and machine kinematics though.

Bottom line is I think the shuffling should be kept but I'm not totally opposed to adding a switch to be able to turn it off.

markmaker:
Hi @tonyluken 

Now that I understand the reason, I agree that shuffling needs to be kept.

Re initializing: it looks as if the fiducial is captured at the same camera center point again and again. At the same time the machine moves in a star shaped motion pattern. It almost seems as if the fiducial is captured at the wrong end of each ray. But maybe I'm just missing info (again). 

You can see this in the video of my machine:

https://makr.zone/Advanced_Camera_Calibration.mp4

Some of the secondary Z footage is missing, that camera switches off after 5min of 4k. The diagnostics is actually from another run.

_Mark

tonyluken:
Hi @markmaker,

Yeah, that's the "extra" move I mentioned.  Since each line starts in the middle, I didn't want to just position the camera once in the middle and then capture the same point for all the lines.  So I added an extra move for each line that moves the camera away from the center and then back to the center before capturing the beginning of each line - hence the star pattern.  Just in case you're wondering, the order of the collected points is completely irrelevant to the model fitting algorithm.

PR #1336, opened 2021-11-25
Fix/OpenPnpCaptureCamera Video Format Identity Instability

markmaker:
# Description
`OpenPnpCaptureCamera`: Due to an instability in the `CaptureFormat.equals()` method, we use the `toString()` representation (Name) for video format selection in the GUI. A `formatName` getter/setter does the translation.

As an aside, this will also log a warning if two cameras have the same unique ID.

# Justification
Video format selection in the GUI was sometimes lost. Debugging has revealed that the `CaptureDevices` returned by `getCaptureDevices()` sometimes change between calls (different Java Objects). When this happens, the `CaptureFormat` Objects are also different. The `CaptureFormat.equals()` method then makes the formats _falsely_ differ, due to a changed `auto-allocated@` property. The ComboBox will then just select the first entry. 

# Instructions for Use
No changes.

# Implementation Details
1. Tested on the machine.


vonnieda:
Nice find, and fix!

Issue #1344, opened 2021-12-04
[Feature Request][Neoden4][FeederActuator] Add support to change Feeder IDs


jrobgh:
# Feature Request
## Describe the Feature
Enable driver support and GUI element to allow changing the Neoden4 Feeder ID. Each feeder internally stores an ID for the machine to address it (typically 1-50, new feeders ship default to 50).

## How Will It Look?
1. Neoden4Driver would add changeFeederId(int oldId, int newId), and changeFeederIdInternal
2. ND4FeederActuatorWizard would add two GUI dropdowns to select oldId, new Id, and a button 'Set' to issue the changeFeederId command

## Why Do We Need It?
1. Adding new feeders requires assigning them a non-default ID, also occasionally people might swap feeders/IDs.

## Provide Examples
In ND4 software this exists in the system menu. The underlying UART command is something like:

Neoden4Message message = new Neoden4Message(new byte[] { newFeederAddressByte, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01 }, oldFeederAddressByte);

markmaker:
I guess this is solved by #1345

PR #1345, opened 2021-12-05
[Neoden4][FeederActuator] Add support to change Feeder IDs

jrobgh:
# Description
Enable driver support and GUI element to allow changing the Neoden4 Feeder ID. Each feeder internally stores an ID for the machine to address it (typically 1-50, new feeders ship default to 50). We will allow range from 0-99. Feature request #1344 

# Justification
Updates the Neoden4 machine specific class

# Instructions for Use
In a Neoden4 Feeder Actuator, select the old and new ID numbers from the dropdowns, and click 'Change'

image

# Implementation Details
2. I think this follows similar style
3. Changes local to only Neoden4 machine classes
4. Ran mvn test

### DIFF
--- a/src/main/java/org/openpnp/machine/neoden4/NeoDen4Driver.java
+++ b/src/main/java/org/openpnp/machine/neoden4/NeoDen4Driver.java
@@ -698,7 +698,43 @@ public class NeoDen4Driver extends AbstractReferenceDriver {
     		throw new IOException("Feed error.");
     	}
     }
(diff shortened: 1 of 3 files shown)


jrobgh:
@markmaker is there someone who could help review this PR (localized to ND4 specific code)? thanks :)

markmaker:
I wasn't looking at it because it is still marked "<draft>".
_Mark

markmaker:
@jrobgh 

Looks good as far as I can tell.

Thanks!

_Mark

PR #1350, opened 2022-01-03
Adding MinAreaCircleStage

sovcharenko:
# Description
I find the existing pipeline for nozzle tip calibration not very accurate especially for small nozzle tips. No matter how hard I try i always get detected circles slightly off and the detection is not very repeatable.

# Justification

# Instructions for Use

# Implementation Details


markmaker:
Hi @sovcharenko 

Thanks for contributing, this looks promising!

I have the following comments:

1. There are other use cases where your new stage could be a useful alternative, so please make it compatible with existing circle returning stages, by returning a `List<Result.Circle>` with just one element, instead of directly the one `Result.Circle`.
2. Conversely, please revert the change in the `DrawCircles` stage. 
3. Please rename your stage into `MinEnclosingCircle` to reflect the proper OpenCV name.
4. Add a check to make sure the working image has only one channel (I know this is also missing in `MinAreaRect` that served as your template, you _could_ add it there too). 


_Mark

PR #1351, opened 2022-01-03
Adding MinAreaCircleStage

sovcharenko:
# Description
I find the existing pipeline for nozzle tip calibration not very accurate especially for small nozzle tips. No matter how hard I try i always get detected circles slightly off and the detection is not very repeatable.

This PR adds support for MinAreaCircle stage that can be used instead of DetectCircles and works more accurate and more reliable. Inspired by MinAreaRectangle stage used in part alignment pipeline.

Also this PR changes DrawCircles stage to support drawing a single circle in addition to an array. in the same way as DrawRectagles does, so that MinAreaCircle can be piped into it.

# Justification
Thought that others might be interested in this stage in their custom pipelines as well.

# Instructions for Use

1. Add MinAreaCircle to your pipeline.
2. Configure threshold min and max

MinAreaCircle will detect a minimal area circle that encloses pixels between threshold min and threshold max.

Note MinAreaCircle will only work with greyscale images.

# Implementation Details
Tested locally on my machine.

I believe my code follows the coding style

There are no changes in these packages.

ran `mvn -q -B test`


markmaker:
Hi @sovcharenko

Thanks for contributing, this looks promising!

I have the following comments:

1. There are likely other use cases where your new stage could be a useful alternative, so please make it compatible with existing circle returning stages, by returning a List<Result.Circle> with just one element, instead of directly the one Result.Circle.
2. Conversely, please revert the change in the DrawCircles stage.
3. Please rename your stage into MinEnclosingCircle to reflect the proper OpenCV name.
4. Add a check to make sure the working image has only one channel (I know this is also missing in MinAreaRect that served as your template, you could add it there too).

_Mark

sovcharenko:
@markmaker, done. please have another look.

Thanks

markmaker:
Very nice! Thanks, @sovcharenko

markmaker:
@sovcharenko 

In the PR you wrote:

> I find the existing pipeline for nozzle tip calibration not very accurate especially for small nozzle tips. No matter how hard I try i always get detected circles slightly off and the detection is not very repeatable.

I agree, and I'm working on a replacement. Out of personal interest, could you try the `DetectCircularSymmetry` for Nozzle Tip calibration?

https://github.com/openpnp/openpnp/wiki/DetectCircularSymmetry

Nozzle tip Pipeline here:
https://github.com/openpnp/openpnp/wiki/DetectCircularSymmetry#nozzle-tip-calibration

You need to set the **Vision Diameter** on the individual Nozzle Tip, i.e. the circular feature you want it to detect: 

image

Does this work for your tips too?

The idea of this stage is to be "tuning free". Apart from entering the **Vision Diameter**, which is an understandable physical property that is easily measurable or found in a datasheet, users do not have to edit the pipelines, set any thresholds etc. manually which may be quite unstable with ambient light and other camera conditions. The stage will also work for subjects where there are no clear dark/bright changes or merely blurred ones: 

blur

Or even nested/stacked circular features of which you want _one specific diameter_: 

radii

So far this stage/pipeline has captured everything, no user has reported otherwise. It is also used internally in the Issues & Solutions camera calibration. 

So what do you think? Could this be the new default? 

Note, I do not want to demerit the usefulness your new stage, it is certainly still a good choice in cases where the diameter is _not_ known.

_Mark

PR #1362, opened 2022-01-20
Fix / Issues & Solutions more Watertight - Round 2

markmaker:
# Description
This is a round-up of bug-fixes and improvements out of testing.

* Error when no nozzle tip loaded in _preliminary_ bottom camera
calibration.
* Disable Advanced Camera Calibration after it fails in the middle.
* Advanced Camera Calibration tool selection in Machine Controls from I&S: bug-fix.
* Limited articulation Rotation Mode support for Advanced Camera
Calibration.
* Log new rotation mode offset being set.
* Communications type visibility initially not established in Driver Wizard: bug-fix.
* Better serial flow control vs. confirmation flow control solutions.
* Optimize Safe Z moving: always move HeadMountables to the _nearer_ limit of the Safe Z Zone:
   image
* Feeder no part assigned caught in GUI actions.

# Justification
See these discussions:
https://groups.google.com/g/openpnp/c/MnKQK_Keww8/m/k9q2R3n7AQAJ
See #1360 (ongoing).

# Instructions for Use
No substantial change in use. 

Screenshot_20220120_172007

# Implementation Details
1. Tested in simulation, with user config.

### DIFF
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -705,6 +705,9 @@ public class FeedersPanel extends JPanel implements WizardContainer {
     }
 
     protected static Nozzle getCompatibleNozzleAndTip(Feeder feeder, boolean allowNozzleTipChange) throws Exception {
+        if (feeder.getPart() == null) {
+            throw new Exception("Feeder has not part set.");
+        }
         // Check the nozzle tip package compatibility.
         Nozzle nozzle = MainFrame.get().getMachineControls().getSelectedNozzle();
         org.openpnp.model.Package packag = feeder.getPart().getPackage();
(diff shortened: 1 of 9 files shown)


markmaker:
To explain the Safe Z bug-fix, I simulated the user's reported case, i.e. a nozzle offset calibration pick&place step, where it rotates the confetti by 180°. Done with the user's `machine.xml`.

Most importantly, consider the user machine's Safe Z Zone:

image

**BEFORE**: Because of the bug, N1 would stubbornly move to its **own** Safe Z, which is Z-16.6, then N2 would rotate 180° and at the same time go to the N2 Safe Z which is Z+16.6, then down to place the confetti.

image

**AFTER**: N1 realizes it is above the Safe Z Zone, but not inside it, so it will move to the upper limit of the Safe Z Zone, which is Z+16.6, then N2 would rotate 180° and at the same time go to the N2 Safe Z which is Z+16.6, i.e. no change, then down to place the confetti.

image

The Planned time tells us how much time the bug-fix saves.

_Mark

PR #1364, opened 2022-01-21
Feature/add additional advanced calibration debug

tonyluken:
# Description
Updates to advanced camera calibration:
- Adds a missing try-catch clause so that if an error occurs (fiducial/nozzle tip is not detected) during one of the steps on a radial line, the remaining lines still get collected - prior to this, the calibration collection would abort and none of the collected points at the Z level where the error occurred would get saved or processed leading to "The function cannot solve under-determined linear systems..." dialog box popping-up.
- Added addition validation of parameters before and during calibration.
- Additional logging has also been added to aid in future debugging.
- Adds a semaphore to protect against simultaneous reading and writing of the undistortion mappings and virtual camera matrix.
- Changed to a dashed yellow circle to mark the area being searched for the calibration fiducial/nozzle tip.
- If running advanced calibration manually on the default head camera, a dialog box now asks if the operator wants to home the machine when complete rather than just automatically doing it.
- Removed all blocking machine task Futures

# Justification
Bug fixes for everyone.

# Instructions for Use
No special instructions for use.

# Implementation Details
1. How did you test the change? **Intermittently obstructed the camera's view during calibration and verified calibration collection continues even when points are missed.**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **No changes to these packages were made.**


markmaker:
Hi @tonyluken,

Overall, this looks very good. 

I have some reservations about your hand-built semaphore and the endless loops for obtaining it. These are almost begging for undiagnosable 100% CPU hangs and races. Any race would only be remediable by killing OpenPnP.

From a quick look, I do not see why this is necessary. You could just mark `getUnitsPerPixel()`, `advancedUndistort()` and `clearCalibrationCache()` as `synchronized` and [it would guard agains simultaneous read/writes](https://docs.oracle.com/javase/tutorial/essential/concurrency/syncmeth.html), but safely. I think these three methods are all so fast, it is not worthwhile to distinguish between sharing readers and exclusive writers.

If I'm missing something, and you _really_ need a semaphore, you should guard against exceptions, even if you think at the moment, that you can rule out any exceptions in these code sequences (which I doubt seeing they use OpenCV). Use [try-with-resource-statement](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html). 

I will merge this now, so all the good stuff becomes available to testers, but hope you'll follow it up, thanks!

_Mark

tonyluken:
Hi @markmaker,

I was afraid you'd have reservations about my home-grown semaphore and I'm open to suggestions on how to improve it.  I don't believe just marking those methods as synchronized will fix the problem (if I understand what synchronized really means - basically only one thread at a time is allowed to execute the method).  The problem is that marking them synchronized doesn't prevent two different threads from accessing the same data via two different methods.  BTW - without the semaphore, I do occasionally see null pointer exceptions in the logs related to the undistortion mappings and the virtual camera matrix - of course they generally don't cause any real problems other than being a nuisance and cluttering up the log files.  Your point is well taken about accounting for the possibility of an exception occurring while the semaphore is being held - I'll look into fixing that.

ian-arkver:
I've not looked at the code, but I don't understand why a mutex (or java equiv) for undistortion/matrices would be a problem. Surely these are only altered, updated, i.e. write locked very briefly in the final throes of the calibration, during which time anything which is using the old values will be wrong. I don't see a problem locking them out while the new values are being written. You should only get a deadlock if the thread that's doing the writing depends on something that a thread that needs a reader lock produces **while it has the write lock held**. So, so long as the write lock is only held for a short time, and won't throw an exception during that short time, then a regular lock should be ok.

It could be you need to flush/discard old data (i.e. processed frames) from before the values were updated, but that's a different question and locking the updates won't fix that.

markmaker:
>  if I understand what synchronized really means - basically only one thread at a time is allowed to execute the method. The problem is that marking them synchronized doesn't prevent two different threads from accessing the same data via two different methods

That's not correct. The object is locked, not the single method. To cite from the [link](https://docs.oracle.com/javase/tutorial/essential/concurrency/syncmeth.html) I added earlier:

> -  First, it is not possible for two invocations of synchronized methods on the same object to interleave. When one thread is executing a synchronized method for an object, all other threads that invoke synchronized methods for the same object block (suspend execution) until the first thread is done with the object.
>  - Second, when a synchronized method exits, it automatically establishes a happens-before relationship with any subsequent invocation of a synchronized method for the same object. This guarantees that changes to the state of the object are visible to all threads.

So I'm still sure this would work here. 

And btw. that second point tells you that hand-rolling your own synchronization is a very, very difficult thing to get airtight on today's CPUs. 

https://en.wikipedia.org/wiki/Memory_barrier

_Mark

tonyluken:
Well, I can't argue with that - I thought I had tried making them synchronized methods and was still seeing the problem - I'll try it again.

tonyluken:
Hi @markmaker,

Well, it certainly appears to be working now with synchronized methods - I must have had some other problem when I originally tried it that I fixed inadvertently when I put in the semaphore.  Anyway, I pushed the change to remove the semaphore.  Thanks for the help.

markmaker:
It seems you can't push to an already merged PR. I guess you need to make a new one. Not sure if the branch name must be different.
_Mark

PR #1379, opened 2022-02-04
Fix / Trying to get rid of the Issues & Solutions Tab Race

markmaker:
# Description
There was a race/bug drawing/layouting the Issues & Solutions tab in a corrupted way. This is related to the initial automatic execution of **Find Issues & Solutions** when OpenPnP starts up.   

image
(Screenshot by Duncan)

This PR further decouples the initial automatic execution of **Find Issues & Solutions**, and subsequent solutions change and selection change notifications. This is after reading that a `TimerTask`, contrary to my expectations, does not run on the GUI event dispatching thread (as `SwingUtilities.invokeLater()` would). 

Everything is now done asynchronously. Only one notification is followed up, even if several have been registered. Note, I already attempted to implement that idea before, but there was a bug, setting the `dirty` flag asynchronously, instead of synchronously, defying the whole purpose. 🙄

# Justification
See the user report, and own experience:
https://groups.google.com/g/openpnp/c/Qk1Yxswmy-E/m/SbkM7d-3AgAJ

# Instructions for Use
No change.

# Implementation Details
1. Tested on Linux/Kubuntu. Could no longer reproduce the issue, while this was relatively easy before (no guarantees, though). 
4. Does not apply in `mvn test`.

### DIFF
--- a/src/main/java/org/openpnp/gui/IssuesAndSolutionsPanel.java
+++ b/src/main/java/org/openpnp/gui/IssuesAndSolutionsPanel.java
@@ -232,7 +232,7 @@ public class IssuesAndSolutionsPanel extends JPanel {
                             return;
                         }
 
-                        selectionActions();
+                        solutionChanged();
                     }
                 });
                 table.getColumnModel().getColumn(0).setPreferredWidth(150);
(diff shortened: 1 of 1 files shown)


ian-arkver:
I pulled the change, rebuilt and tried opening openpnp about 10 times. I _think_ I'm doing the same steps as before but can no longer make it fail. Looks like this is fixed now, though it's hard to be 100% sure since I didn't have a reliable test for it.

Excellent work, thanks Mark!

PR #1383, opened 2022-02-19
Feature/Bottom Vision Nozzle Tip Background Knockout Calibration

markmaker:
# Description

## Problem 

The Bottom Vision pipeline almost never works out of the box, users need to tweak the properties of select stages to make them work with their camera and their machine. Firstly, it is difficult to understand why a part detection does not work, and where in the pipeline the problem happens. Root causes may often be obscured and happen many stages before the visibly bad result is exposed. Secondly, once the proper stage is identified, it is often difficult to adjust its properties. Stages and individual properties are completely generic, and (often very academically) abstracted from their actual tasks, which is a good thing from a math and software design perspective, but a bad thing for them to be understood by OpenPnP users, in the context of the actual pipeline use case. 

One of the more complex to understand and tune stages is the `MaskHSV` stage, to detect and "knock out" the green Juki nozzle from bottom vision image (this works like a film-maker's [green-screen](https://en.wikipedia.org/wiki/Chroma_key)). Users tend to just "try & error", which often leads to just haphazardly working pipelines, i.e. the pipeline may momentarily work with one part or package, but not with the next. This leads to a large and hard to maintain number of pipelines, and tedious job setup. Even for users with prior computer vision knowledge, the required effort is considerable.

Furthermore, even slight changes in conditions (such as ambient light, slight differences in nozzle tips) will often break these pipelines, leading to errors in jobs, and _more_ work tweaking the many pipelines.

## Solution

This Pull-Request aims to solve this problem, by automatically (and repeatedly) calibrating the background with empty nozzle tip during the [Nozzle Tip Calibration](https://github.com/openpnp/openpnp/wiki/Nozzle-Tip-Calibration-Setup). Note: this is just **one** puzzle piece towards a tuning-free bottom vision (other PRs are in work). 

## Calibration Operating Principle

The new Background Calibration reuses the images that are already captured by the Nozzle Tip Calibration, so no extra machine time and motion is needed to obtain them. It analyzes these background images for the **Key Color** (Chroma Key) of the Juki (or similar) nozzle tip. Any dominant, vivid color is automatically detected and a likely bounding box computed in the [HSV Color Model](https://en.wikipedia.org/wiki/HSL_and_HSV). 

Furthermore, a cutoff brightness is determined where the non-color-keyed parts of the background must all be darker. This indicates the minimum brightness applicable typically in a `Threshold` stage (the subject of an upcoming PR). 

The background calibration automatically blots out the center piece in the calibration image, where the nozzle tip was detected. This will eliminate any shiny elements that are typically present (needles, scraped off points etc.). OpenPnP assumes this center part to be always covered by the picked part, i.e. there is no need to include its color in the background calibration, or these shiny elements treated as a problem (see Trouble Shooting). The blotted-out disc has a size of **Vision Diameter** + 2 x **Minimum Detail Size** (see Configuration).

The resulting calibration is also validated, and if the detected quality is not good, diagnostic guidance and visual feed-back is provided for users to remedy the situation (see section Trouble Shooting below).

## Calibration Pipeline Control

If the Background Calibration is enabled, the following is controlled in the bottom vision pipeline:

1. The `BlurGaussian` `kernelSize` property is controlled by a new **Minimum Detail Size** field, where the user can enter a length (in system units) that indicates the smallest sensible size of a feature on a part (such as the dimension of a pad, pin, ball of a package). The GaussianBlur is configured accordingly, to suppress image noise, dirt, texture etc. that is not relevant for detection. This is also applied to the Background Calibration itself, to get the same `MaskHSV` input data as the pipeline.
2. The `MaskCircle` `diameter` property is controlled by the already existing **Max. Part Diameter** field on the Nozzle tip. Peripheral background content that might disrupt the vision operation is masked out. This is also applied to the Background Calibration itself, to get the same `MaskHSV` input data as the pipeline.
3. The `MaskHSV` `hueMin`, `hueMax`, `saturationMin`,  `saturationMax`, `valueMin` and `valueMax` properties are all controlled by the Background Calibration bounding box.  

# Justification
See the Description above, and countless user support cases in the discussion group. 

# Instructions for Use

## Configuration
There is a new Background Calibration section on the Calibration tab: 

Screenshot_20220219_154325

**Method** controls the Background Calibration:

Screenshot_20220219_160944

- **None** switches Background Calibration off. The pipeline is no longer controlled, i.e. the user can edit the properties in the Pipeline Editor. 
- **Brightness** only determines the cutoff brightness of the nozzle tip and background, no key color is assumed to be present. 
- **BrightnessAndKeyColor** determines both the nozzle tip key color, and the remaining background cutoff brightness. 

**Minimum Detail Size** (in system units) configures the smallest valid detail of a part to be detected in bottom vision, such as the smallest dimension of a pad, pin, ball of a package. All smaller image details like image noise, textures, dirt etc. are considered irrelevant for detection. The **Minimum Detail Size** controls a blurring filter to suppress these artifacts.

**Minimum** and **Maximum** columns in the **Hue** (base color), **Saturation**, **Value** (Brightness) [HSV color model](https://en.wikipedia.org/wiki/HSL_and_HSV) indicate the calibrated bounding box of the key color (if enabled). 

The HSV color model can be illustrated like this (Wikipedia):

image

**Tolerance** can be used for each channel, to expand the detected bounding box. This provides robustness against changing ambient lighting, shadows cast by the picked parts, etc. Note: practical tolerance values are yet to be determined in testing. The tolerance is applied to both sides of the channel bounds, except for Saturation Maximum, and Value Minimum which are both  left unlimited in the bottom vision `MaskHSV` stage. OpenPnP assumes both darker and more vivid (greener) pixels are always part of the background, even if they were not detected, or cut off in the calibration. 

## Application

The Background Calibration data is then used to automatically parametrize the bottom vision operation. If the Background Calibration **Method** is not **None**, the relevant stage properties in the pipeline are fully controlled. If the background was successfully calibrated, the knockout should now be perfect:

nozzle-tip-background-knockout

As the Nozzle Tip Calibration is typically renewed at least whenever the machine was homed, or even with each nozzle tip change, the calibration data should be adaptive to changing ambient light and other conditions. A manual recalibration can also always be triggered, if needed.

Furthermore, any (color) differences between nozzle tips are also implicitly handled. This is important, when packages are compatible with multiple such nozzle tips. The pipeline (with is part/package dependent, but not nozzle tip dependent), can therefore dynamically react to these differences, through the calibration control. 

## Trouble Shooting

The calibration provides diagnostic feed-back regarding the quality of the background. A text describes the problem(s) and provides rough guidance as to how the problem can be remedied. Furthermore, the button **Show Problems** can be pressed to display the problematic image portions in the camera preview:

nozzle-tip-background-diagnostics

Credits: Nozzle tip images by Jan, with thanks!

# Implementation Details
1. Tested in simulation with real user images. 

### DIFF
--- a/src/main/java/org/openpnp/gui/components/HsvIndicator.java
+++ b/src/main/java/org/openpnp/gui/components/HsvIndicator.java
@@ -1,0 +1,189 @@
+/*
+ * Copyright (C) 2022 <mark@makr.zone>
+ * 
+ * This file is part of OpenPnP.
+ * 
+ * OpenPnP is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ * 
+ * OpenPnP is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public License along with OpenPnP. If not, see
+ * <http://www.gnu.org/licenses/>.
+ * 
+ * For more information about OpenPnP visit http://openpnp.org
+ */
+
+package org.openpnp.gui.components;
+
+import java.awt.Color;
+import java.awt.Dimension;
+import java.awt.Graphics;
+import java.awt.Graphics2D;
+import java.awt.image.BufferedImage;
+
+import javax.swing.JComponent;
+
+public class HsvIndicator extends JComponent {
+
+    public HsvIndicator() {
+        super();
+    }
+
+    private int minHue;
+    private int maxHue;
+    private int minSaturation;
+    private int maxSaturation;
+    private int minValue;
+    private int maxValue;
+
+    @Override
+    public Dimension getPreferredSize() {
+        Dimension superDim = super.getPreferredSize();
+        int width = (int)Math.max(superDim.getWidth(), 192);
+        int height = (int)Math.max(superDim.getHeight(), 128); 
+        return new Dimension(width, height);
+    }
+
+    protected void paintComponent(Graphics g) {
+        Graphics2D g2d = (Graphics2D) g;
+        int width = getWidth();
+        int height = getHeight();
+        int diameter = Math.min(width, height);
+        int unit = diameter/5;
+        double radius = diameter/2.0; 
+        BufferedImage image = new BufferedImage(width, height, BufferedImage.TYPE_INT_ARGB);
+        int [] data = new int[width*height];
+        for (int y = 0; y < diameter; y++) {
+            for (int x = 0; x < diameter; x++) {
+                double dx = x - radius;
+                double dy = y - radius;
+                double r = Math.sqrt(dx*dx + dy*dy);
+                if (r < radius) {
+                    double saturation = 255.0*r/radius;
+                    double hue = 255.0*((1.0 + Math.atan2(-dy, dx)/Math.PI/2.0) % 1.0);
+                    double alpha = Math.min(1.0, radius - r);
+                    double dHue = r/64;
+                    double dSat = radius/255;
+                    double included = Math.max(0, Math.min(1.0, Math.min(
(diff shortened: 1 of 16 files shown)


markmaker:
Added the Wiki page:
https://github.com/openpnp/openpnp/wiki/Nozzle-Tip-Background-Calibration

PR #1384, opened 2022-02-23
Feature/Pipeline Parameters

markmaker:
# Description

This the next step after #1383, aiming to provide a better user experience with vision operations. The aim is not to have to go into the Pipeline Editor at all. After having knocked out the background (usually green Juki nozzle tip) automatically, there still remain some setting that need tweaking. This Pull request addressed that.

Among the many stages and properties in a Computer Vision Pipeline, there are often only _very few_ crucial properties that need adjusting. But those are hard to find and even harder to adjust. By Introducing configurable parameters to the computer vision Pipeline, these can now be singled out by the Pipeline author and they are exposed as GUI controls on the Vision Settings. Currently the `ParameterNumeric` stage is implemented to control numeric properties of the stages. But other parameter data types can later be added easily, there is a nice abstraction already implemented, which allows mapping arbitrary types to sliders.

The GUI is implemented by the new `PipelinePanel` hat can be integrated into any Wizard form that wants to carry a pipeline. It also helps handling the **Edit** and **Reset** functions typically present on such a Wizard. The `pipeline` and `pipelineParameterAssignments`  properties can be bound. The latter is a simple Map, storing the parameter values. 

Sliders

The system is now available for Bottom Vision and Fiducial Locator Vision. But it is open to be introduced for any feeder etc. using computer vision and pipelines.

# Justification
Instead of digging into pipelines, users can simply adjust a few sliders. 

# Instructions for Use

## Application
Without ever going into the Pipeline Editor, users get the crucial parameters presented as sliders, which they can then interactively adjust. 

Sliders

The effect of the tweaking is directly shown in the Camera Preview. As configured in the parameter, the slider previews the stage that best shows the effect of any adjustments. In addition, if users stop dragging the slider, the pipeline end result is displayed with a slight delay. Watch it in action in the animation:

Parametric-Pipeline

The new default Bottom Vision Pipeline already has the parameters shown here implemented. 

### Parametric Control of a Stage 

In our example we want to control a `MaskHSV` stage named "threshold", and in it the "valueMax" property: 

target stage

The following will show you how to achieve that.

### Create New ParameterNumeric Stage

Create a new `ParameterNumeric` stage by pressing the  button. Type "para" to jump to the right stage implementation:

New Stage

Then drag & drop the new stage to the top of the list. Parameter stages can only influence the stages that come _after_ them, and it is also a well known convention to have the parameters at the top.

Stage Name

Change the new stage's **Name** in the Name column of the list. This name also sets the parameter name, i.e. the key under which the setting is stored in the vision settings. It is important that you choose a well-considered, stable name. If you later change that name, any stored parameter assignments will be disconnected. The **Name** must not be numeric, i.e. the default number that is assigned by the Pipeline Editor when a stage is created, is not allowed. The Parameter will remain inactive until you assign a proper **Name**. An exception is reported in the stage, as long as this is not yet the case. 

### ParameterNumeric Properties

The `ParameterNumeric` stage has the following properties:

ParameterNumeric Properties

The **parameterLabel** determines the label of the parameter, when it is added to the user interface. Unlike the stage name, the label may be changed anytime, as you like.

The **parameterDescription** determines the tool-tip of the parameter, when it is added to the user interface. You can use HTML markup, if you like (due to limitations of the property editor, editing is restricted to one line).

The **stageName** identifies the target stage that is controlled by this parameter. It is recommended to properly name the target stage and then copy the name over here. **Note:** The identified target stage must come _after_ this parameter stage.

The **propertyName** identifies the property in the target stage. This property must be numeric in nature (hence it is called "ParameterNumeric"). Future development may cover additional property types.  

The **numericType** sets the type and unit of the target property. Even though these properties are "all just numbers", there are some nuances to be distinguished:

image

- **Integer** controls the property as a whole number. It is typically used when color channels values (0 ... 255) are addressed, like in our example.
- **Double** controls the property as a plain floating point number.
- **Squared** controls the property on a squared scale, i.e. the linear scale of the slider is squared before being applied to the property. Effectively you can control a 2D area by its 1D dimension, which gives the slider a more natural control (before using this type, read about the **SquareMillimeters** and **SquareMillimetersToPixels** types below).
- **Exponential** controls the property on an exponential scale, i.e. the linear scale of the slider controls the _magnitude_ (exponent) of the number to be applied to the property. Use this if the property goes over a very large range. You can then control the small numbers finely, and the large numbers coarsely. 
- **Millimeters** controls dimensional properties that are given as absolute lengths/millimeters. Note: the parameter is always defined in millimeters, but the user interface will still convert everything to inches, and back, if that is the user's system preference. But the pipeline remains portable across systems that way. 
- **MillimetersToPixels** controls dimensional properties given in pixels, but the parameter is handled and stored as a length with units. This makes the parameter independent of the resolution, distance and viewing angle of the camera. It will automatically be converted to pixels using the proper **Units per Pixel** scale of the camera, and it can even adjust for the Z distance of the subject, [if 3D calibration is enabled](https://github.com/openpnp/openpnp/wiki/3D-Units-per-Pixel). 
- **SquareMillimeters** controls area properties that are given as areas/square millimeters. It is the squared variant of the **Millimeters** type, i.e. what was said about **Squared** and **Millimeters** applies in combination. 
- **SquareMillimetersToPixels** controls area properties given in pixels, but the parameter is handled and stored as an area with units.  It is the squared variant of the **MillimetersToPixel** type, i.e. what was said about **Squared** and **MillimetersToPixel** applies in combination. 

The **minimumValue** and **maximumValue** control the limits of the parameter range. If  **minimumValue** is larger than **maximumValue**, the slider control is effectively reversed, which may be more intuitive for certain parameters. The values are given as millimeters, or square millimeters if one of the corresponding **numericType**s is selected. Any squared or exponential scaling does only affect how the slider interpolates _between_ minimum and maximum, i.e. you do not need to do any math yourself.

The **defaultValue** is the preset for the controlled property, when the pipeline is first applied.  Again, the value is given as millimeters, or square millimeters if one of the corresponding **numericType**s is selected.

The **effectStageName** identifies the stage in the pipeline, where the effect of changes to the controlled property is best seen by the user. When users move the slider, the preview will dynamically show the effect. This is often but not always the same stage as the target stage given by the **stageName** (see above). If left empty, no effect preview is provided. 

The **previewResult** checkbox determines, if the result image of the pipeline is previewed, similar to the effect stage. If both  **effectStageName** and **previewResult** are given, the result image preview will only be shown after a delay, when the user stopped dragging the slider.  

# Implementation Details
1. Tested in simulation with real user images.


markmaker:
Added the Wiki page:

https://github.com/openpnp/openpnp/wiki/Exposed-Pipeline-Parameters

Links and more general instructions here:
https://github.com/openpnp/openpnp/wiki/Computer-Vision

_Mark

PR #1386, opened 2022-03-06
Feature/DetectRectlinearSymmetry Pipeline Stage

markmaker:
# Description

## Intro

The most important OpenPnP computer vision task is Alignment, or "Bottom Vision" to find the precise angle and position of parts on the nozzle. 

By default this is done by isolating the brightest elements of the part (the shiny contacts) and then finding the minimum area rectangle that bounds these. This works well, but it requires setting thresholds etc. to isolate the brightest elements, which can vary between packages, machines, cameras etc. PR #1384 made this easier, but it did not eliminate the need to manage different technical settings for different packages. 

## New DetectRectlinearSymmetry Stage

This PR add a new way to detect the parts that does not rely on specific thresholds etc. It exploits the fact, that in typical projects, an overwhelming majority of the parts/packages have a large degree of mirror symmetry (or [reflection symmetry](https://en.wikipedia.org/wiki/Reflection_symmetry)), i.e. their left side is a mirror image of the right side, and/or their upper side is a mirror image of the lower side. 

Symmetry has the nice property that it is self-contained, i.e. we only needs to compare the image to itself, no outside information (thresholds, template images, etc.) is needed and fewer assumptions (that contacts always appear bright, that the overall outline is well defined by a bounding rectangle etc.) are required to be true.

The mirror symmetry is present left-right and upper-lower, i.e. in the two 90° dimensions of the part, hence we call it Rectlinear Symmetry. 

## Operation Principle

The new stage basically works in two phases:

1. Firstly, it determines the angle of the subject, by analyzing 90° cross-sections of the image at varied angles (in reality this happens iteratively), and then taking the one with the largest contrast. It is a bit like shooting x-rays across the subject, and looking for the crispiest result. The following animation shows how this works. Cross-sections are indicated in blue, the yellow graph indicates the recorded contrast at various angles:

    find-maximum-angular-contrast-small-ic

2. Secondly, the resulting cross-sections (at the best angle) are analyzed for symmetry. The stage will select a mid-point for each 90° cross-section that shows the best-matching left-right and upper-lower mirror image. 

For truly symmetric parts, the stage is completely tuning free. Assuming there are no disruptive patterns in the background (and we took care of that in #1383) it will be very robust in determining the alignment. 

The stage is robust against irregularities, because it works in a _probabilistic_ way. There are no hard edges and thresholds involved where the image is broken down into either black or white. Instead, the _most likely_ match is detected from the _integral_ image information. For this reason the bent pin in the following example will not disrupt the bounding box, as it would in the default OpenPnP pipeline. These irregularities, and their robust suppression can make a crucial difference in placement accuracy: 

Irregular pin

Rect-linear symmetry does not imply a rectangular contact arrangement, parts could well be cross-shaped, for instance. Furthermore, a certain tolerance is also implied, parts with fewer/smaller pins on one side (e.g. SOT23-5) are not a problem. The cross-section symmetry is not perfect, but still the best match: 

Semi-asymmetric

## Asymmetric Parts

Some parts are not symmetric. For those, the stage can be made to work with some extra tweaking. However, because there is no symmetry, we lose the self-tuning property. We must adjust the threshold and optionally the minimum detail size:

Asymmetric-rectlinear

1. Specialize the asymmetric package.
2. Switch the left-right and/or upper/lower symmetry off. 
3. Adjust the threshold to capture the span of the asymmetric dimension.
4. Optionally adjust the minimum detail size to suppress artifacts that are not relevant.

## Soft MaskHsv

Because of the probabilistic nature of the DetectRectlinearSymmetry stage (see above), we need to provide it with a "natural" image. However, background removal using MaskHsv (see #1383) is usually "hard", i.e. it sets all the pixels to deep black, which then makes these different from the rest of the not-so-black background. Furthermore, the Mask is also hard on the edges, i.e. it either masks a pixel, or not. Amplified by some properties of typical MJPEG color encoding, and compression, this may result in a jagged image and block artifacts: 

Hard Mask

To address this, the MaskHsv stage has therefore been extended with two new properties:

**softEdge** adds a soft edge around the boundaries of the HSV model mask. Pixels can now be somewhere in between being masked and not. This naturally maps to focal blur and other effects where the key color to be masked is mixed with others. 
   
hsv-soft-edge

**softFactor** determines, how strongly the mask should be applied. At 1.0 fully masked pixels will become all black (default). At 0.5, only half the brightness is taken out. 

Soft Mask

## Pipeline Stage Properties in Sequence

In previous versions, the pipeline stage properties were always sorted alphabetically, which often resulted in a strange order ("Max" came before "Min" etc.). This has been fixed, properties now appear as intended by the stage author. You can press the **A-Z** button to sort them alphabetically as before:

stage-properties

## ImageRead Stage Simulates Camera Capture

To facilitate testing with pre-captured images, potentially from other users, the ImageRead stage can now emulate a camera-captured image: 

image

**handleAsCaptured**, if enabled, handles the loaded image as if captured by the camera. The image resolution and aspect ratio will be adapted, and if information is present (in a upp.txt file side-by-side with the image), the image is scaled to camera Units per Pixel. Any pixel coordinates obtained from the image are therefore correctly interpreted, as they would from a camera captured image. The image is also registered as the pipeline captured image.

Images are scaled, cropped and/or extended with black borders, as needed. 

# Justification
Make bottom vision tuning free, at least for symmetric parts. 

# Instructions for Use

## Using the new Rectlinear Pipeline Pipeline

Go to the Vision tab, select the Rectlinear Symmetry entry and copy the pipeline:

Copy

Then go to the part where you want to try this, and press **Specialize for ...**.

Then paste the pipeline:

Paste

## Bottom Vision Pipeline Control

The Bottom Vision pipeline is parametrized by OpenPnP, so that some stage properties are controlled from the calling function:

- Background removal is controlled as described on the [Nozzle Tip Background Calibration page](https://github.com/openpnp/openpnp/wiki/Nozzle-Tip-Background-Calibration#calibration-pipeline-control)
- The location where the DetectRectlinearSymmetry stage searches for the part is controlled through the "alignment.center" pipeline property. Typically just the center of the camera. 
- The **expectedAngle** property of the DetectRectlinearSymmetry stage is controlled through the "alignment.expectedAngle" pipeline property. This is set to the nominal alignment angle.
- The **searchDistance** property of the DetectRectlinearSymmetry stage is controlled through the "alignment.searchDistance" pipeline property. This is taken from the Nozzle Tip configuration, **Max. Pick Tolerance**:

    Max. Pick Tolerance

-  The **maxWidth** and **maxHeight** properties of the DetectRectlinearSymmetry stage are controlled through the "alignment.maxWidth" and "alignment.maxHeight" pipeline properties, respectively. If the **Part size check** is enabled on the Vision Settings, the effective part size is taken (with the **Size tolerance** added). If no part size info is present, the **Max. Part Diameter** from the nozzle tip is taken. 

## Configuring the DetectRectlinearSymmetry Stage

DetectRectlinearSymmetry Stage

**expectedAngle** tells the stage the expected angle of subject to be detected. 

**searchDistance** sets the search distance around the center.

**searchAngle** sets the search angle (two-sided) around the expected angle.

**maxWidth** sets the maximum cross-section width of the subject to be detected.

**maxHeight** sets the maximum cross-section height of the subject to be detected.

**symmetricLeftRight** tells the stage whether the subject is left/right-symmetric (in the sense as seen at 0° subject rotation). According to this switch, the stage either takes the **symmetricFunction**, or the **asymmetricFunction**.

**symmetricUpperLower** tells the stage whether the subject is upper/lower-symmetric (in the sense as as seen at 0° subject rotation). According to this switch, the stage either takes the **symmetricFunction**, or the **asymmetricFunction**.

**symmetricFunction** determines how the cross-section is evaluated for **symmetric** subjects. 

- **FullSymmetry** looks for full inner and outline symmetry. Use for truly symmetric subjects and best precision.
- **EdgeSymmetry** looks for full inner and outline symmetry of edges. Use for partially symmetric subjects, where some, but not all features are present on both sides, or where shades differ.
- **OutlineSymmetry** looks for outline symmetry only. Used for subjects that are symmetric on their outline, but not on the inside.
- **OutlineEdgeSymmetry** looks for outline symmetry of edges only. Used for subjects that are symmetric on their outline, but not on the inside, and where some, but not all features are present on both sides, or where shades differ.
- **OutlineSymmetryMasked** looks for outline mask symmetry only. Use for quite asymmetric subjects. Requires setting a mask threshold.

**asymmetricFunction** determines how the cross-section is evaluated for **asymmetric** subjects. The choice is the same as for **symmetricFunction**.

**minSymmetry** minimum relative symmetry. Values larger than 1.0 indicate symmetry. The value is highly image-dependent and can only be used for very individual settings.

**subSampling** is used to speed things up: Only one pixel out of a square of **subSampling** × **subSampling** pixels is sampled. The best region is then locally searched using iteration with smaller and smaller **subSampling** size. The **subSampling** value will automatically be reduced when other search properties require it. Use **BlurGaussian** before this stage if **subSampling** suffers from moiré effects. 

**superSampling** can be used to achieve sub-pixel final precision: 1 means no super-sampling, 2 means half-pixel precision etc. Negative values can be used to stop refining **subSampling**, -2 means it will stop at a 2-pixel resolution.

**smoothing** is applied to the sampled cross-sections. Given as the Gaussian kernel size. This is needed to eliminate interferences, when angular sampling coincides with the pixel raster or its diagonals.

**gamma** determines the [Gamma correction](https://en.wikipedia.org/wiki/Gamma_correction) to be applied to the image. The input signal is raised to the power of **gamma**. With **gamma** > 1.0 the bright image parts are emphasized.

**threshold** is only used when the **OutlineSymmetryMasked** function is active. Only pixels with luminance greater than the **threshold** are considered when determining the outline of the part.

**minFeatureSize** is only used, when the **OutlineSymmetryMasked** function is active. Pixels are masked by the **threshold** property, and a cross-section pixel count over these masked pixels is then determined. A cross-section bin only counts as "detected" when the pixel count is larger than **minFeatureSize**. This is used to remove masking imperfections, i.e. image specks and impurities up to a certain size and frequency.

**propertyName** determines the pipeline property name under which this stage is controlled by the vision operation. If set, these will override some of the properties configured here. Defaults to "alignment" for Bottom Vision.

**diagnostics**, if enabled, indicates the detection with cross-hairs and bounds.

**diagnosticsMap** , if enabled, overlays a diagnostic map indicating the angular reclinear contrast and rectlinear cross-section.

# Implementation Details
1. Tested in simulation and with many user images collected in the [openpnp-test-images/Parts/Bottom/](https://github.com/openpnp/openpnp-test-images/tree/master/Parts/Bottom) repository.
6. Did follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style).
7. No changes in the `org.openpnp.spi` or `org.openpnp.model` packages.
8. Successful `mvn test` before submitting the Pull Request.


ian-arkver:
This looks really nice Mark. Looking forward to testing it once I get "Real Work (tm)" finished for a bit.

markmaker:
> This looks really nice Mark. Looking forward to testing it once I get "Real Work (tm)" finished for a bit.

Thanks, and looking forward to getting your feed-back!

markmaker:
Added the Wiki:
https://github.com/openpnp/openpnp/wiki/DetectRectlinearSymmetry
https://github.com/openpnp/openpnp/wiki/MaskHsv

PR #1387, opened 2022-03-09
Feature/More Transparent Pipeline Stage Property Control

markmaker:
# Description

When pipeline stage properties are controlled by the pipeline caller, this is not currently made transparent to the user. Manual changes to the stage properties remain without effect, which can be quite confusing. 

The pull request changes this.

## Controlled Stage Properties More Transparent

Controlled Properties

Controlled properties are now grayed-out, and their value hidden. When users select the property, the control value is displayed along with the property description. 

This applies both to properties controlled from computer vision code in OpenPnP (built-in) and from [Exposed Pipeline Parameters](https://github.com/openpnp/openpnp/wiki/Exposed-Pipeline-Parameters).

**Note**: I found no way to _directly_ display the control value in the property table without actually changing the property inside the stage. But changing the stage properties is a no-go, because this then wrongly marks the Pipeline as modified, and it also leaves behind the last control value, instead of the authored default value. The default value may still be needed as fall back, in case the control value is not set (some pipeline callers set override properties conditionally).

## Other Related Changes

- To unify the handling, the `SimpleBlobDetector` control is now also subject to a configurable `propertyName`. 
- The `subSampling` property of the `DetectRectlinearSymmetry` from #1386 was added to Bottom Vision caller control, from the **Minimum Detail Size** of the [nozzle tip background calibration](https://github.com/openpnp/openpnp/wiki/Nozzle-Tip-Background-Calibration#other-settings). 
- The Minimum Details Size is now  assumed to explicitly refer to the smallest contact size, rather than some unspecified "smaller size". Therefore, the `GaussianBlur` property `kernelSize` is now set to _half_ that (same as `subSampling` mentioned above).
- The Wizard tooltip and default value was changed accordingly. 
  Minimum Detail

# Justification
See the discussion here:
https://groups.google.com/g/openpnp/c/kZLS7PnSE5Y/m/aJA7XabWBAAJ

# Instructions for Use
The Wiki has been amended:
https://github.com/openpnp/openpnp/wiki/DetectRectlinearSymmetry#automatic-bottom-vision-pipeline-control
https://github.com/openpnp/openpnp/wiki/Nozzle-Tip-Background-Calibration#calibration-pipeline-control
https://github.com/openpnp/openpnp/wiki/Nozzle-Tip-Background-Calibration#other-settings

# Implementation Details
1. Tested in simulation with user images. 


markmaker:
@jrobgh 

Thanks for the feedback. I'm also thinking about providing a _comment_ argument for calling code, so developers can state the _origin_ and/or _intent_ of the property control. 

There are usually very good reasons why a property is controlled and you should not need to override it.

Have you seen the conversation about why you should **not** override the MaskCircle diameter? How and why it is controlled by the Nozzle Tip **Max. Part Diameter** property now?

See here (including what I linked there):
https://groups.google.com/g/openpnp/c/lGikHtY5lh8/m/gSYvnrw6DAAJ

_Mark

PR #1391, opened 2022-03-17
[NeoDen4Driver] Multi-axis Z/C support

jrobgh:
# Description
Neoden4 head has 4 nozzles, previously our ND4 driver relied on a single axis shared among all 4 nozzles. By enabling named axis support in the ND4 driver, we can create dedicated Z and rotation axis for each nozzle. These should be named  zN1/zN2/zN3/zN4/rotationN1/rotationN2/rotationN3/rotationN4 for the movement function to find them correctly.

# Justification
This is an improvement for the NeoDen4 driver.

# Instructions for Use
1) NeoDen 4 users should add ReferenceControllerAxis for zN1/zN2/zN3/zN4 and rotationN1/rotationN2/rotationN3/rotationN4
2) Map the heads nozzles to the corresponding zN# and rotationN# axis

image

image

# Implementation Details
1. Tested on ND4 machine, ran maven test
2. Yes passed code style checks in Maven test
3. Changes are contained to ND4 class
4. mvn test ran and passed


markmaker:
Hi @jrobgh,

Can I persuade you to do it "the right way"? 🥳

I can see from where this code originated, _historically_, but it would be _so much better_ if you adopted the proper OpenPnP way. 😃

You should no longer reference axes by some "magic" name, instead just take them from the `hm`, they are directly mapped there. Use `hm.getAxisX()` or `hm.getAxis(Axis.Type.X)` etc. 

https://github.com/openpnp/openpnp/wiki/Mapping-Axes

I suggest you try to make it as "standard" as possible. That way [Issues & Solutions](https://github.com/openpnp/openpnp/wiki/Issues-and-Solutions) will "understand" the configuration correctly. A growing number of other functions also rely on that mapping to provide correct functionality. 

**Or better yet, don't use the `hm` at all, that's the "old way"! Read on...**

I suggest to use G-code compliant axis letters, even if the Neoden is not a G-code driver. The axis lettering is a machine control convention that goes beyond G-code, IMHO. So letter them X Y Z U V W A B C D  and map them to the ordinal numbers you seem to need for the proprietary driver. 

If you convert your `moveToInternal()` to use a loop over the axes in the `AxesLocation`, you can directly take the axes from there and drive the proprietary driver according to the above-mentioned letters, and their mapping to ordinal numbers. I also suggest to compare with the previous coordinate and not move the axis if unchanged.

Question: is there really no way to move X, Y together with one or more C (or Z)? Won't this slow down the machine?

_Mark

jrobgh:
Hi Mark, thanks very much for the comment, sure I am open to that. I guess
considering there are multiple Z axes, need to loop through all of them and
update the changed ones? (Can getAxis return multiple axes of the same
type? If not maybe using A,B,C,D makes sense here?)

From what I've gathered the head has the 4 nozzles (each with an axis with
this update) and the head x/y movement only happens with the nozzle raised
up in Z. There are two motors to actuate in Z, one shared between nozzles
1-2, the other between 3-4.

John

markmaker:
Hi @jrobgh 

>  I guess considering there are multiple Z axes, need to loop through all of them and update the changed ones?

Consider the following code. It shows it for the C axes:

```java
        AxesLocation location1 = move.getLocation1();
        AxesLocation location0 = move.getLocation0();
        // Take only the changed axes and only those driven by this driver
        AxesLocation displacement = location0.motionSegmentTo(location1)
                .drivenBy(this);
        // Drive rotation axes.
        for (ControllerAxis axis : displacement.byType(Axis.Type.Rotation)
                .getControllerAxes()) {
            // map from axis letter to driver index.
            int index = 1+"ABCD".indexOf(axis.getLetter());
            if (index < 1 || index > 4) {
                throw new Exception("Invalid axis letter "+axis.getLetter()
                +" for rortation axis "+axis.getName());
            }
            moveC(index, location1.getCoordinate(axis, getUnits()));
        }
       ...
```

The Z axes can be done the same way. 

The X, Y axes need to be extracted as before.

Again the question:  _is there really no way to move X, Y together with one or more C?_ 

_Mark

markmaker:
is this still **`*wip*`** as the title implies?

jrobgh:
Hi Mark, thanks for the follow up- is this more in line with the desired approach (multiple named Z and C rotation axes following naming as Z1/Z2/Z3/Z4 and C1/C2/C3/C4)? 

This class was holding z and c location as doubles directly instead of using the AxesLocation objects (homingOffsets, etc.) so still working to get those updated properly.

markmaker:
Hi @jrobgh 

>  is this more in line with the desired approach (multiple named Z and C rotation axes following naming as Z1/Z2/Z3/Z4 and C1/C2/C3/C4)?

Well, it would be better for the **Letter** to really be a single _letter_ (literally), i.e. conformant to G-code (RS-274), as I suggested. If you'd rather not use letters (X Y Z U V W A B C D), then better leave them empty and take the index number from the axis **Name** rather than the **Letter** (in future version of OpenPnP such a "not-a-letter" state might be rejected or cause problems).

Also, for the X, Y axes, rather than checking for `if(distance(deltaX, deltaY) > 0.0001)` you should instead check whether the axis is in the `displacement`, as it is then already checked as _changed_ according to the **Resolution/Steps per unit** configured on the axis. If a user chooses to set a coarser resolution (to avoid run-out/non-squareness compensation etc. related micro-motion overhead), everything in the planner and driver should act consistently.  Something like this:

```java
if (displacement.getAxis(Axis.Type.X) != null || displacement.getAxis(Axis.Type.Y) != null) {
  ...
```

>  yes the head can support each nozzle having a different rotation value concurrently so having multiple C axis here is important.

I was talking about _concurrent motion_, i.e. rotating the axes _while_ the head is also moving in X, Y to pick a part or to align it to its placement angle. For a typical OpenPnP machine, this saves quite a lot of time.


_Mark

markmaker:
@jrobgh 

you still on this?

_Mark

jrobgh:
Hi Mark, thanks for the follow up, sure so if I change Z1/Z2/Z3/Z4, C1/C2/C3/C4 to A/B/C/D, E/F/G/H (types Z, Rotation) would that be compliant with future schemes? I'll get that rolled in and try to validate the other changes on my machine

markmaker:
No, the letters should G-code (RS-274) standards compliant.

So I would use `Z` `U` `V` `W` for the Z axes and `A` `B` `C` `D` for the C axes. 

These are the best match, i.e. they are officially axis letters (other letters have different meaning), plus linear axes are linear and rotary axes are rotary. 

https://en.wikipedia.org/wiki/G-code#Letter_addresses
https://www.nist.gov/publications/nist-rs274ngc-interpreter-version-3

_Mark

jrobgh:
Thanks, pulled in those changes and ran both maven checkstyle and a test assembly to validate the functionality

markmaker:
Hi @jrobgh 

You are now going for standard letters, very good, but I guess you forgot to change it back to `getLetter()` here:

https://github.com/openpnp/openpnp/blob/262a51ea8cb7da4da002014b79eace8a631beec3/src/main/java/org/openpnp/machine/neoden4/NeoDen4Driver.java#L775-L789

and here:

https://github.com/openpnp/openpnp/blob/262a51ea8cb7da4da002014b79eace8a631beec3/src/main/java/org/openpnp/machine/neoden4/NeoDen4Driver.java#L811-L825

For a more compact way to do this, look at the code [I originally proposed here](https://github.com/openpnp/openpnp/pull/1391#issuecomment-1073038168);

```java
            // map from axis letter to driver index.
            int index = 1+"ABCD".indexOf(axis.getLetter());
            if (index < 1 || index > 4) {
                throw new Exception("Invalid axis letter "+axis.getLetter()
                +" for rotation axis "+axis.getName());
            }
```

Also, please have a look at [my earlier comment](https://github.com/openpnp/openpnp/pull/1391#issuecomment-1081556494), about the X, Y displacement (resolution).

_Mark

jrobgh:
Thanks Mark for the reply, in this section:

https://github.com/openpnp/openpnp/blob/262a51ea8cb7da4da002014b79eace8a631beec3/src/main/java/org/openpnp/machine/neoden4/NeoDen4Driver.java#L854:L876

- per your comment modify to check for non-null displacement (instead of distance > 0.0001)

it seems like motion planner can be used instead of code in this section, so removed a few lines related to:
- low speed for short displacement
- restrictions on moves with nozzles lowered
- backlash compensation

markmaker:
Perfect. 
Is this ready to merge?
_Mark

jrobgh:
Yeah should be ready to merge, thanks for your comments

markmaker:
Thanks @jrobgh !

JeremiahGillis:
I can confirm this works with my NeoDen4. This fixed some of the rotational issues I encountered previously. Great Job!

PR #1395, opened 2022-04-06
Feature / Issues & Solutions Dynamic Safe Z and Max. Part Heights

markmaker:
# Description
With **Dynamic Safe Z** enabled, nozzles need a head-room, the so-called Safe Z Zone to lift the nozzles higher with taller parts. Furthermore, on nozzles with shared Z axes the part heights compete for the same head-room. The maximum expected part height is specified on the nozzle tip **Max. Part Height**.

These Issues & Solutions detect when the (combined) **Max. Part Height**(s) on single or shared Z axis nozzles violate the Safe Z Zone head-room.

# Justification
See discussion group:
https://groups.google.com/g/openpnp/c/kQZzXxshQog/m/1kGQF2XYBwAJ

# Instructions for Use
Are provided in the Solutions, and in the [Wiki](https://github.com/openpnp/openpnp/wiki/Kinematic-Solutions), that was already written to accompany the Solutions with links:

Safe Z Zone Isssue

New, more specific Wiki sections and links were also added for existing Dynamic Safe Z and Kinematics solutions.

https://github.com/openpnp/openpnp/wiki/Kinematic-Solutions

# Implementation Details
1. Tested in simulation.
3. There is a minimal change in `org.openpnp.spi.NozzleTip` to include the `Solutions.Subject` interface, so nozzle tips can become I&S subjects.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/solutions/HeadSolutions.java
+++ b/src/main/java/org/openpnp/machine/reference/solutions/HeadSolutions.java
@@ -216,53 +216,7 @@ public class HeadSolutions implements Solutions.Subject {
                         addInconsistentAxisIssue(solutions, camera, hm, Axis.Type.X);
                         addInconsistentAxisIssue(solutions, camera, hm, Axis.Type.Y);
                         if (hm instanceof Nozzle) {
-                            if (hm.getAxisZ() == null) {
-                                solutions.add(new Solutions.PlainIssue(
-                                        hm, 
-                                        "Nozzle "+hm.getName()+" does not have a Z axis assigned.", 
-                                        "Please assign a proper Z axis. You might need to create one first.", 
-                                        Severity.Error,
-                                        "https://github.com/openpnp/openpnp/wiki/Mapping-Axes"));
-                            }
-                            if (hm.getAxisRotation() == null) {
-                                solutions.add(new Solutions.PlainIssue(
-                                        hm, 
-                                        "Nozzle "+hm.getName()+" does not have a Rotation axis assigned.", 
-                                        "Please assign a proper Rotation axis. You might need to create one first.", 
-                                        Severity.Error,
-                                        "https://github.com/openpnp/openpnp/wiki/Mapping-Axes"));
-                            }
-                            if (hm.getAxisZ() != null && hm.getAxisRotation() != null) {
-                                for (Camera camera2 : head.getCameras()) {
-                                    if (camera2.getAxisZ() == hm.getAxisZ()) {
-                                        sameAxisAssignedIssue(solutions, camera2, hm, camera2.getAxisZ(), Axis.Type.Z);
-                                    }
-                                    if (camera2.getAxisRotation() == hm.getAxisRotation()) {
-                                        sameAxisAssignedIssue(solutions, camera2, hm, camera2.getAxisRotation(), Axis.Type.Rotation);
-                                    }
-                                }
-                                for (Nozzle nozzle2 : head.getNozzles()) {
-                                    if (nozzle2 == hm) {
-                                        break;
-                                    }
-                                    if (nozzle2.getAxisZ() == hm.getAxisZ()) {
-                                        solutions.add(new Solutions.PlainIssue(
-                                                head, 
-                                                "Nozzles "+nozzle2.getName()+" and "+hm.getName()+" have the same Z axis assigned.", 
-                                                "Please assign a different Z axis.", 
-                                                Severity.Error,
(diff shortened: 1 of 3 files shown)


ian-arkver:
Hi Mark,

I don't think this part height addition should apply to cam style shared axes where one nozzle being raised doesn't push the other one down. I couldn't easily tell from the commit whether you only applied this to geared shared axes.

Regards,
Ian

markmaker:
The code does all the necessary transformations, I hope 😅

https://github.com/markmaker/openpnp/blob/c7f9014571e394e9f9e7da743d188fda85255707/src/main/java/org/openpnp/machine/reference/solutions/KinematicSolutions.java#L450-L460

The text  says _"after transformation"_ to make it clear that this is not _necessarily_ a mere addition. However, in most practical applications it actually _is_, even with the regular cam transformation, because it is symmetric. Only if the cam is asymmetric (lopsided nozzles), or has V shaped arms at an angle other than 180° would this not be true. And I guess that is rare.

_Mark

markmaker:
...thinkink a bit more, I guess it is even true with asymmetric/lopsided cam, as long as it has arms at 180°.

For each distance one nozzle goes up, the other goes down, regardless of the angle of the cam and the actual (sinus) transform. In other words, each Millimeter of head-room that is consumed by one nozzle is not available for the other. 

It is only for the V-shaped cam arms, that these ratios are not **1 : -1**.

https://github.com/openpnp/openpnp/wiki/Transformed-Axes#referencecamcounterclockwiseaxis

_Mark

markmaker:
... sorry for the spam ... just to make it clear: the code _does_ handle the V-shaped cam or any other strange and future transformation (as long as it is [strictly monotonic](https://en.wikipedia.org/wiki/Monotonic_function), which it must be anyway). 

_Mark

ian-arkver:
Yeah, maybe, but I don't quite follow it for the cam case. Maybe I don't understand what rawAxisZ's safeZoneHigh and Low are.

You start at the nozzle position, then move to there plus max height, then move from there to one of the safe heights and throw the warning if the two moves go in different directions (I think).

But, for a cam nozzle you can't lift the part higher than that nozzle's zero anyway (it'll hit the endstop), so maybe for those the max part height always has to be within the safe zone, so this error is covered.

Anyway, I'll revisit my safe zones etc. when I get back to the machine (whenever that might be) and will post to the group if something seems astray.

[ps - thanks for the additional thoughts re odd-shaped cams. Mine's just a regular straight one though.]

Thanks!
Ian

ian-arkver:
ps.
"For each distance one nozzle goes up, the other goes down, regardless of the angle of the cam and the actual (sinus) transform."

This is the bit that's not right. Only one nozzle moves at a time. The other sits at its zero position.

markmaker:
Ahh now I understand.

The OpenPnP cam transformation does not model any blocking. It assumes that the nozzle keeps going up with the cam. So if your  **Max Part Height** is more than half the Safe Z Zone, it will not be reported. So yeah, this is a loophole. 

However, the two nozzles compete for the headroom, so as long as all your nozzle tips have the same **Max Part Height** it will be reported correctly. 

> You start at the nozzle position, then move to there plus max height, then move from there to one of the safe heights and throw the warning if the two moves go in different directions (I think).

Correct, and you see the second nozzle's nozzle tip **Max Part Height** transformation code here:

https://github.com/markmaker/openpnp/blob/c7f9014571e394e9f9e7da743d188fda85255707/src/main/java/org/openpnp/machine/reference/solutions/KinematicSolutions.java#L497-L501

This time it uses the transformation of `nozzle2`.

_Mark

ian-arkver:
Yes, having the same MaxPartHeight on both nozzles/tips should cover it, but I could imagine people might have different setups (eg. a nozzle only used for small parts, only compatible with small nozzles). Maybe that's a weird corner case.

I was wondering if for the cam case only an additional check that any compatible MaxPartHeight isn't > safeZone/2 might be worth it?

ps. Minor typo in the text:
https://github.com/markmaker/openpnp/blob/c7f9014571e394e9f9e7da743d188fda85255707/src/main/java/org/openpnp/machine/reference/solutions/KinematicSolutions.java#L529
s/Save/Safe/

All the best,
Ian

markmaker:
> I was wondering if for the cam case only an additional check that any compatible MaxPartHeight isn't > safeZone/2 might be worth it?

We would have to define the blocking height in the Cam transform. It is not necessarily at the balance point, it can be below or above. It would be quite hard to measure. 

Don't know how important that is, just for I&S.

_Mark

PR #1398, opened 2022-04-08
pom: Target Java17 and add add-opens flags to tests and launchers

ian-arkver:
# Description
Update target compiler to Java 17. Add --add-opens flags to surefire test launcher and openpnp batch scripts to allow internal access to java.lang and java.awt.

# Justification
This should allow developer use of Java 17. Further work may be needed on prebuilt packages.

# Instructions for Use
N/A for the user, other than allowing Java 17 to be used.

# Implementation Details
1. How did you test the change? Maven tests pass. On my machine started, ran homing, closed. Machine.xml survived.
2. Did you follow the coding style? N/A for pom/scripts
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages... N/A
mvn tests completed using default java...
openjdk 17.0.3 2022-04-19
OpenJDK Runtime Environment (build 17.0.3+3)
OpenJDK 64-Bit Server VM (build 17.0.3+3, mixed mode)

## Breaks GH CI pipelines?
GitHub emailed saying all the CI runs failed with an invalid target version. Seems they don't like "17" though it is the correct version afaik for Java 17. Should this be passed in from the build environment somehow? None of the CI pipelines use 17.

### DIFF
--- a/openpnp.bat
+++ b/openpnp.bat
@@ -2,7 +2,7 @@
 
 set archi=%PROCESSOR_ARCHITECTURE%
 
-if not x%archi:86=%==x%archi% java -jar target\openpnp-gui-0.0.1-alpha-SNAPSHOT.jar
-if not x%archi:64=%==x%archi% java -jar target\openpnp-gui-0.0.1-alpha-SNAPSHOT.jar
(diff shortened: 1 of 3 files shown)


ian-arkver:
I'm not sure what the correct approach is here. Setting the maven target java version breaks the CI pipelines @vonnieda has set up. Should maven continue to target version 8? I don't know enough about maven/java to know exactly what the pros & cons of that are.

Since the --add-opens flags are only needed for versions >= 17 and the flag isn't present on version 8 we'd need to add a JRE version dependency to the launch script.

I'll close this PR since it clearly doesn't work and needs more thought.

PR #1399, opened 2022-04-14
Fix BlindsFeeder fiducials with increased wall_width

qwertymodo:
# Description
The fiducials on the BlindsFeeder OpenSCAD model are not correctly positioned if the wall_width parameter is changed.

# Justification
This is a simple fix to allow for a different wall_width so you can space out the strip lanes to account for things like passive spool hangers.  By default, the model is completely unchanged, this only corrects the calculation of the fiducial position to account for the wall_width if it does change.  Currently, any wall_width value other than 1 results in an incorrect position of the fiducials, such as this:

blindsfeeders_fiducials

# Instructions for Use
Create an instance of a BlindsFeeder object and pass in a wall_width value of anything other than the default of 1.  The walls between lanes will be increased and the fiducials will remain in their proper positions.

# Implementation Details
This is a simple 2-line change to include the wall_width parameter when calculating the position of fiducials.  It should be fairly self-explanatory.


markmaker:
Great, thanks!

PR #1401, opened 2022-04-18
Add capability to disable distortion and tilt corrections

tonyluken:
# Description
Add the capability to disable lens distortion correction and camera tilt correction.

# Justification
May help in debugging an issue where the camera appears to have a tilt to it after calibration has been run. It is suspected that fitting of a camera model that includes both tangential lens distortion and camera tilt may not be robust in its ability to decern the difference between the two effects when they are both relatively small.  This change allows one or the other or both to be independently disabled to see what if any affect they have on the calibration.

# Instructions for Use
Everything works just as it did previously unless the machine.xml file is manually edited outside of OpenPnP.  The advanced-calibration section for each camera now includes entries for the following three attributes:
disable-tilt-correction="false"
disable-distortion-correction="false"
disable-tangential-distortion-correction="false"

Setting any of these to true, disables the respective correction.  The first disables any correction for camera tilt, i.e., rotation of the camera about the X and Y-axes is assumed to be exactly zero (correction for rotation about the Z-axis is still performed).  The second disables both radial and tangential lens distortion correction.  And the third disables only tangential lens distortion (assuming the second is set to false).

# Implementation Details
1. How did you test the change? **Ran the same calibration data through the algorithm with the new attributes at each of their true/false settings and verified the results were as expected.**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.   **No changes were made to those packages.**
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.  **Maven tests were run and passed**.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/camera/calibration/AdvancedCalibration.java
+++ b/src/main/java/org/openpnp/machine/reference/camera/calibration/AdvancedCalibration.java
@@ -145,6 +145,15 @@ public class AdvancedCalibration extends LensCalibrationParams {
     
     @Attribute(required = false)
     private double virtualHeightFov;
+    
+    @Attribute(required = false)
+    private boolean disableTiltCorrection = false;
+    
+    @Attribute(required = false)
+    private boolean disableDistortionCorrection = false;
+    
+    @Attribute(required = false)
+    private boolean disableTangentialDistortionCorrection = false;
 
     
     private Mat virtualCameraMatrix = Mat.eye(3, 3, CvType.CV_64FC1);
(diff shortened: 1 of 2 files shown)


markmaker:
Hi @tonyluken 

Thanks!

Regarding the **Instructions for Use**: 

1. Am I right to assume, that after changing any of the switches, the calibration **must** be redone before the switches become effective? 

1. If yes, is it sufficient to enable **Skip New Collection An Reprocess Prior Collection?** and then just press **Start Calibration**?

_Mark

markmaker:
@tonyluken, please disregard my last question, I now saw that you left detailed instructions in the group:

> You will need to startup OpenPnP and then just shut it down (this will write some new fields into your machine.xml file).  Then open your machine.xml file in a text editor, find the section with your down camera, and below that, find the line that start like:
> 
>  `<advanced-calibration enabled="true" overriding-old-transforms-and-distortion-correction-settings="true" valid="true" ... `
> 
> On that line, find:
> 
> `disable-tangential-distortion-correction="false"`
> 
> and change false to true.  Save the file and restart OpenPnP.  Then on the Machine setup tab, select the down camera, go to its Advanced Calibration tab, set the "Skip New Collection and Reprocess Prior Collection" check box and then click the Start Calibration button.  That should recalibrate the camera without having to collect a new set of data.  Once that is done, try the selfie test again.
> 
> 

https://groups.google.com/g/openpnp/c/SgrK9A_69lQ/m/KucCOSZGAAAJ

_Mark

PR #1402, opened 2022-04-21
Disable tangential distortion correction by default

tonyluken:
# Description
This is a follow on to PR #1401 - this change disables tangential distortion correction by default

# Justification
Testing has shown that camera calibration performs better with tangential distortion correction disabled. 

# Instructions for Use
For new camera calibrations, no special action is required as this will be applied by default.

For cameras that had previously been calibrated, they can either be fully recalibrated, or the collection step can be skipped and the change applied by navigating to each camera in the Machine setup tree, selecting the Advanced Calibration tab, enabling the "Skip New Collection and Reprocess Prior Collection" check box, and then clicking the "Start Calibration" button.

Note - in the off chance that someone wants to re-enable tangential distortion correction, they can manually do so by editing their machine.xml file and changing the attribute disable-tangential-distortion-correction to the value of "false".

# Implementation Details
1. How did you test the change? **Started OpenPnP, saved the configuration, and then examined the machine.xml file to verify that all cameras had disable-tangential-distortion-correction set to "true".  Also verified that if it is manually changed back to "false", the value is retained.**
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.  **All maven tests were run and passed.**

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/camera/calibration/AdvancedCalibration.java
+++ b/src/main/java/org/openpnp/machine/reference/camera/calibration/AdvancedCalibration.java
@@ -47,6 +47,12 @@ import org.simpleframework.xml.core.Commit;
 import org.simpleframework.xml.core.Persist;
 
 public class AdvancedCalibration extends LensCalibrationParams {
+    // Prior to version 1.2, no explicit version attribute existed. Version 1.0 (implicit) was the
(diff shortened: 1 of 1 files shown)


markmaker:
Cool!

PR #1409, opened 2022-05-06
Feature/Parts Table Placement and Feeder Counts

markmaker:
# Description
This adds **Placements** and **Feeders** counts to the **Parts** table.

# Justification
Users can quickly see how many parts are going to be required for one Job. The table can also be cleverly sorted by these columns to help provisioning feeders.

# Instructions for Use
The **Placements** count is automatically computed and updated whenever placements are added/changed. Note, the full count is used, regardless of enabled/disabled boards, and regardless of enabled/disabled or already placed placements.

The **Feeders** count is automatically computed and updated whenever parts are assigned/unassigned on feeders. Note, the full count is used, regardless of enabled/disabled feeders. 

For feeder provisioning, first click twice on the **Placements** column header to sort descending and then click on the  **Feeders** column header to sort ascending. This will show all parts requiring additional feeders at the top, starting with those that require the most parts:

Screenshot_20220506_114720

# Implementation Details
1. Tested in simulation.
3. Changes `org.openpnp.model.Part` to add the counts as (pseudo) properties.

### DIFF
--- a/src/main/java/org/openpnp/gui/tablemodel/PartsTableModel.java
+++ b/src/main/java/org/openpnp/gui/tablemodel/PartsTableModel.java
@@ -36,9 +36,9 @@ import org.openpnp.model.Part;
 @SuppressWarnings("serial")
 public class PartsTableModel extends AbstractObjectTableModel implements PropertyChangeListener {
     private String[] columnNames =
-            new String[] {"ID", "Description", "Height", "Package", "Speed %", "BottomVision", "FiducialVision"};
+            new String[] {"ID", "Description", "Height", "Package", "Speed %", "BottomVision", "FiducialVision", "Placements", "Feeders"};
     private Class[] columnTypes = new Class[] {String.class, String.class, LengthCellValue.class,
-            Package.class, String.class, BottomVisionSettings.class, FiducialVisionSettings.class};
+            Package.class, String.class, BottomVisionSettings.class, FiducialVisionSettings.class, Integer.class, Integer.class};
     private List<Part> parts;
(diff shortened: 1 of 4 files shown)


maholli:
@markmaker - FYI: not sure if this commit or #1408, but I now receive the error below each time I trigger the ReferencePushPullFeeder auto-setup (didn't use to occur). 

image

markmaker:
Hi @maholli,

Judging from the stack trace this is related to OCR and I cannot see a relation to this PR. 

Please check this:

1. In your template feeder, are you using the default pipeline? 
1. If not: the pipeline is supposed to contain a stage named "OCR", and it must be enabled. Is this the case?
1. In your template feeder: Is OCR set up properly? 
1. Assuming you just updated to the newest testing version, what was your previous version? 
   (maybe another, earlier change triggered this).

I will add better diagnostic feed-back to the feeder for when this happens.

_Mark

maholli:
Hi Mark. No action needed on your part, I was just adding this error info since you're generous enough to distribute your test branch. 

FWIW, this occurs on feeders with and without OCR "enabled," but feeder positions still update properly despite the error message. Pipeline is your circular symmetry from [here](https://github.com/openpnp/openpnp/wiki/DetectCircularSymmetry#referencepushpullfeeder). I'm playing with lots of different physical configurations so I don't have any feeders marked as template at the moment. As for test version, I catch each update as it's released. I haven't been able to find a test version commit hash anywhere in openpnp, just figured it wasn't something you spent the time implementing.

markmaker:
Solved in #1410

PR #1242, opened 2021-08-08
Feature / Better DetectCircularSymmetry Search Ranges

markmaker:
# Description
Improves the DetectCircularSymmetry stage to support cropped search ranges. 

It automatically crops the search range to the image edges. Note, the area analyzed for circular symmetry must still fully be inside the image, that's why the search area indicated by the reddish heat map still has a margin to the edge (half of **maxDiameter**): 

runoutCalibration_result825748987901124084

If the nominal target location is close to the edge of the image, the search range can become asymmetric:

grafik

A search width and height can now also (optionally) be specified, in addition to the circular maximum search range (this is not yet employed/implemented in the feeder classes):

search dims

The PR also improves diagnostic heat-map and cross-hairs. The indicated circle hugs the circumference very precisely. Observe how asymmetries in the [Bresenham approximation](https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm#Method) show that sub-pixel accuracy is in-deed present (jogging around with 0.01mm steps):

sub-pixel -accuracy

The nozzle tip calibration Wizard pipeline edit button has been changed to take the nozzle location as the nominal target location. This can be used to test nozzle tip detection at an off-center position. 

This PR also fixes a bug, where multi-target detection did not actually use the **superSampling** i.e. sub-pixel best result (it _did_ work in single-target). In this illustration with the fixed version, sub-pixel accuracy is again made obvious by the asymmetric [Bresenham approximations](https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm#Method) that vary from sprocket hole to sprocket hole:

sub-pixel -accuracy-r

# Justification
See here:
https://discourse.openpnp.org/t/camera-calibration/79/39?u=markmaker

# Instructions for Use
The [Wiki page](https://github.com/openpnp/openpnp/wiki/DetectCircularSymmetry#properties-used-when-controlled-by-the-stage) is already updated.

# Implementation Details
1. Tested in simulation and on machine debug images saved earlier.
2. Did follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style).
3. No changes in the `org.openpnp.spi` or `org.openpnp.model` packages.
4. Successful `mvn test` before submitting the Pull Request.

PR #1420, opened 2022-05-20
Fixes to the Advanced Camera Calibration.

markmaker:
# Description
Fixes to the Advanced Camera Calibration.

1. Automatic progression to the second Z level restored. 
2. Coverage of calibration increased from 90% to 98%.
3. Restriction of search radius removed. The circular symmetry detection can work with cropped search areas, since #1242.
4. For some reason, the **Auto Camera View** mode was switched off by Advanced Camera Calibration. This was removed. 

# Justification
## Automatic Calibration from issues & Solutions
In automatic mode, as employed by Issues & Solutions, the calibration would still require the user to press "Next" to go to the second Z level. The change introducing this was probably an unintended left-over from a debugging session. This PR restores the original code.

## Coverage of Search
Changing the coverage from 90% to 98% increases the precision. See the results.

Before:
Before

After:
After

The precision was increased by approx. √2.

## Search Diameter
The Advanced Camera Calibration would no longer work with strong distortion (such as in simulation), the search diameter would be so constrained at the edge that the fiducial was no longer found. Restricting the search radius is no longer necessary due to #1242.

# Instructions for Use
No change in usage.

# Implementation Details
1. Tested in simulation.
5. No changes in the `org.openpnp.spi` or `org.openpnp.model` packages.
6. Successful `mvn test` before submitting the Pull Request.

### DIFF
--- a/src/main/java/org/openpnp/gui/processes/CalibrateCameraProcess.java
+++ b/src/main/java/org/openpnp/gui/processes/CalibrateCameraProcess.java
@@ -440,7 +440,6 @@ public abstract class CalibrateCameraProcess {
      */
     private boolean estimateUnitsPerPixelAction() {
         Logger.trace("Entering estimateUnitsPerPixelAction");
-        ((ReferenceCamera) camera).setAutoVisible(savedAutoVisible);
         
         //Capture the movable's coordinate as the approximate center of the camera's field-of-view
         centralLocation = movable.getLocation().convertToUnits(LengthUnit.Millimeters).
@@ -763,10 +762,7 @@ public abstract class CalibrateCameraProcess {
                 }
                 // Note, automatic only available when location set. 
                 if (automatic) {
-                    setInstructionsAndProceedAction("If necessary, use the jog controls on the Machine Controls panel to jog the camera so that the calibration fiducial at Z = %s is approximately centered in the green circle.  Click Next when ready to proceed.", 
-                            ()->requestOperatorToAdjustDiameterAction(),
-                            calibrationLocations.get(calibrationHeightIndex).getLengthZ().toString()); 
-//                    requestOperatorToAdjustDiameterAction();
+                    requestOperatorToAdjustDiameterAction();
                     return true;
                 }
             }
@@ -973,13 +969,6 @@ public abstract class CalibrateCameraProcess {
         //Mask off everything that is not near the expected location
         pipeline.setProperty("MaskCircle.center", expectedPoint);
         pipeline.setProperty("DetectCircularSymmetry.center", expectedPoint);
-        
-        //Restrict the mask diameter so as to not extend beyond the edge of  the image
-        maskDiameter = 2*(int)Math.min(expectedPoint.getX(), 
-                Math.min(pixelsX-expectedPoint.getX(),
-                Math.min(expectedPoint.getY(),
-                Math.min(pixelsY-expectedPoint.getY(), maskDiameter/2))));
-        
         pipeline.setProperty("MaskCircle.diameter", maskDiameter);
         pipeline.setProperty("DetectCircularSymmetry.maxDistance", maskDiameter/2.0);
 
--- a/src/main/java/org/openpnp/machine/reference/camera/calibration/AdvancedCalibration.java
+++ b/src/main/java/org/openpnp/machine/reference/camera/calibration/AdvancedCalibration.java
@@ -53,7 +53,8 @@ public class AdvancedCalibration extends LensCalibrationParams {
     // tangential distortion correction by default.
     // Moving to version 1.3 - version 1.2 accidentally disabled all distortion correction by 
     // default rather than only tangential distortion correction.
-    private static final Double LATEST_VERSION = 1.3;
+    // Moving to version 1.4 - increase the testPatternFillFraction. 
+    private static final Double LATEST_VERSION = 1.4;
     
     @Attribute(required = false)
     private boolean overridingOldTransformsAndDistortionCorrectionSettings = false;
@@ -122,7 +123,7 @@ public class AdvancedCalibration extends LensCalibrationParams {
     private int desiredRadialLinesPerTestPattern = 32;
     
     @Attribute(required = false)
-    private double testPatternFillFraction = 0.90;
+    private double testPatternFillFraction = 0.98;
     
     @Attribute(required = false)
     private double walkingLoopGain = 0.50;
@@ -277,6 +278,11 @@ public class AdvancedCalibration extends LensCalibrationParams {
             //Fix version 1.2 where the initial values of these two were accidentally swapped
             disableDistortionCorrection = false;
             disableTangentialDistortionCorrection = true;
+        }
+        if (version <= 1.3) {
+            // Allow for a larger fill fraction, thanks to better circular symmetry detection,
+            // including even cropped circles. 
+            testPatternFillFraction = 0.98;
         }
     }
     


markmaker:
Hi @tonyluken,

I guess we need to discuss these little changes. This is just a proposition.

_Mark

tonyluken:
Hi Mark,

1. I purposefully put that back because my calibration fiducials are "stacked" - the secondary fiducial only gets placed on the machine when necessary and it goes right over the top of the primary.  That pause is needed to give me a chance to place the secondary calibration rig on top of the primary.  I don't think it is too onerous for those that have both fiducials on the machine at the same time to have to make one additional mouse click.  If you think otherwise, please make it optional.
2. & 3. I'm ok with both of those as long as they get tested with real cameras.  I have two main concerns: Can the model of lens distortion accommodate the actual lens distortion out near the edge of the field-of-view (FOV) while still fitting well with that near the center of the FOV? I'd hate to drive the model fit by large distortions near the edge of the FOV (where it doesn't matter much) at the expense of the fit near the center (where it is most important).  Secondly, especially for bottom cameras, does the lighting etc. hold up well enough to get reliable detections very near to the edge of the FOV?
4.  That was probably a mistake on my part - I'm fine with your fix.

markmaker:
Hi @tonyluken,

Thanks for the fast answers.

(1.) 

Oh, I understand. Assuming such a stacked setup will be very rare, I've added the `automationLevel` property, visible in the `machine.xml` as `automation-level`. It defaults to 2, meaning full automation. If you set it to 1, it will pause at the Z level change. Users can now also set it to 0 to get full process step interaction.

(2. & 3.)
  > _"Can the model of lens distortion accommodate the actual lens distortion out near the edge of the field-of-view (FOV) while still fitting well with that near the center of the FOV?"_ 

I might be mistaken, but the Residual Error Map seems to suggest that the match is even better at the center, see the "before" and "after" screen shots in the PR description. AFAIK, polynomial fitting is behaving well for interpolation, but badly for extrapolation. These are the differences I observed:

Raw simulated distortion (radial displacement by `f*r²`):
Top_2022-05-21_20 30 59 851

Calibrated 90% coverage:
Top_2022-05-21_20 31 02 475

Calibrated 98% coverage:
Top_2022-05-21_20 47 16 344



   > _"Secondly, especially for bottom cameras, does the lighting etc. hold up well enough to get reliable detections very near to the edge of the FOV?"_ 

The circular symmetry detection is quite sturdy. And I guess this concern is about equally valid at 90%, especially since the probing radii are scaled to match the camera view rectangle, i.e. it happens well beyond the actual "pre-rotated bottom vision ROI" circle. 

If it turns out to be a problem, I can quickly fix it in both `develop` and `test`. Users can change the `machine.xml` in the mean-time.

What do you think?

_Mark

tonyluken:
> I might be mistaken, but the Residual Error Map seems to suggest that the match is even better at the center, see the "before" and "after" screen shots in the PR description. AFAIK, polynomial fitting is behaving well for interpolation, but badly for extrapolation.

I don't know if this is a real problem or not, but my concern is that you are testing with a simulated camera that has a perfect (and simple) mathematical description of the lens distortion.  A real lens may have distortion artifacts that don't fit the polynomial distortion model very well.  An example to illustrate the potential problem - imagine fitting a polynomial curve to a set of data that is actually exponential in nature.  Over any segment of the exponential data you can always find a polynomial curve that will fit the data to any desired level of residual error.  But if you increase the domain of the data segment, you would need to increase the order of the polynomial being fit to the data in order to achieve the same level of residual error.  In the plot below, the blue curve is fit to all the data (blue and orange) while the orange curve is only fit to the orange data.  Clearly the residual error near the center is worse for the blue curve than for the orange curve.  Conversely, if the orange curve were extrapolated out to edges of the domain of the blue curve, its residual error would be much higher (but only out at those far edges).  BTW - the below data is modelled as exp(-(x^2 / 1e5)).
 
PolynomialFitToNonPolynomialData

markmaker:
@tonyluken,

I agree, my simulated example is no "proof" at all.

However, your showcase is clearly an example of _under-fitting_. What I'm trying to say is exactly the opposite,  namely that I believe the advanced camera calibration is showing signs of _over-fitting_. 

The undistorted image is perfect _inside_ the coverage area, then goes off very "nervously" _outside_ of it. As far as I understand polynomial fittings, and admittedly partly based on gut feeling, this could mean there is an excess of freedoms of degrees in the approximation function. For my simulation, one can argue, this is due to the distortion being too simple (I see that now). But IMHO the same behavior was [confirmed on a real camera](https://groups.google.com/g/openpnp/c/62B-K-IPm-A/m/KYZSwkuRAQAJ), admittedly with less dramatic edge distortion, but still clearly visible (Note, the image seems cropped to valid pixels, so it is likely even stronger without crop):

Down camera graph paper and grid

Due to the nature of lenses, I believe distortions are comparatively very low in the center so I think it is valid to "clamp down" the polynom as far on the edge as possible. If as a result some probes are being smoothed in the center area, I believe we are suppressing measurement errors, rather than missing out on actual distortion. 

Unless I misunderstand, this is also what the "Before" and "After" Residual Error plots reflect (in my screenshots in the Description), i.e. errors become more evenly distributed with smaller extremes. As would be expected, knowing the samples are obtained using the same mechanical and optical process. But this would have to be validated on a _real_ camera.

_Mark

markmaker:
Latest additions: 

- Makes sure the base angle line walker (and its 180° opposite) aims at the corner of the camera view, so the very extreme of the lens is probed. Switches corners at secondary Z.
- Furthermore, the starting point radius is scattered randomly, so the probing is a bit less concentric, effectively increasing radial probing density. 
- Finally, this solves an issue with limited articulation nozzles. 

I now get good results and reliable convergence with as little as 4 radial lines. 

Will now try to test on real machine.

_Mark

markmaker:
Machine Testing:

- No problems with 98% coverage. 
- Happy to see that your #1416 did solve my **imaginary** camera mounting problem. I can finally use the Advanced Calibration, yippee!
- Same with bottom camera/nozzle detection. Reliable right to the very corner:
    grafik

Top camera with standard 32 radial lines:

grafik

Top camera with 12 radial lines:

grafik

Top camera with merely 4 radial lines:

grafik

Bottom camera with standard 32 lines (do not mind the scale, I was not able to exactly match the focal plane):

grafik

_Mark

markmaker:
OK, this is embarrassing. 😏

Turns out the update-handler was faulty. When the `machine.xml` is from _after_ Advanced Camera Calibration was introduced, but _before_ the `version` property was added, then the coverage of calibration (`testPatternFillFraction`) was **not** updated to 98%. This is [now fixed](https://github.com/openpnp/openpnp/pull/1420/commits/6a8fadd8b371b1ab562ec2dd923fbedca0c936d1).

As it happens, my `machine.xml` for the physical machine falls exactly into that category. So what I [tested and documented above](https://github.com/openpnp/openpnp/pull/1420#issuecomment-1137698527) is the _original_ 90% coverage. 

Now I tested again, this time _really_ 98%:

- The calibration woks flawlessly at 98%. Detection is robust to the very edge:
   Edge probe
- Same for the bottom camera/nozzle detection:
  bottom edge probe
- The resulting _undistortion_ is significantly better than [with 90% coverage:](https://github.com/openpnp/openpnp/pull/1420#issuecomment-1137698527) 
  Undistortion
- However, one can argue that the grid match (Units per Pixel) is slightly less precise. _Or maybe not?_ The paper _does_ add a a slight paper thickness Z delta, i.e. it _is_ slightly nearer to the camera, so the millimeter paper appearing slightly larger might actually be correct. Not sure. 
- The bottom camera is fine too (again, I was not able to match the focal plane): 
  Undistortion bottom

@tonyluken, what do you think? 

_Mark

tonyluken:
Hi Mark,

Re: Polynomial overfit versus underfit - in either case a polynomial can have large excursions outside the domain of the data points (all polynomials of degree greater than zero eventually go off to +/- infinity).  An overfit (too high of a degree) will pass exactly through the data points but will probably have large excursions _between_ the data points as well.  An underfit (to low of a degree) will probably miss all of the data points but will not have any large excursions between the data points.

In our case, we're fitting a 6-degree polynomial but it really only has three degrees of freedom since only the even non-constant terms are computed (radial distortion is modelled as  1+k1*r^2+k2*r^4+k3*r^6 where r is the radial distance from the image center).  So as long as we collect points at least 5 different radial distances from the center, we will be underfitting (4 would be an exact fit and 3 or fewer would be an overfit).

> The resulting undistortion is significantly better than https://github.com/openpnp/openpnp/pull/1420#issuecomment-1137698527

I just don't see that.  If anything, I'd say that the 98% coverage one looks worse.  Look at the vertical graph paper lines on the left side of the image - they deviate significantly from being parallel with the reticle lines.  Now, granted the 90% coverage one does show some deviation on the right side of the image but it doesn't appear as bad to me.  Maybe you're seeing something I'm not?

markmaker:
> > The resulting undistortion is significantly better than https://github.com/openpnp/openpnp/pull/1420#issuecomment-1137698527

> I just don't see that. If anything, I'd say that the 98% coverage one looks worse. Look at the vertical graph paper lines on the left side of the image - they deviate significantly from being parallel with the reticle lines. Now, granted the 90% coverage one does show some deviation on the right side of the image but it doesn't appear as bad to me. Maybe you're seeing something I'm not?

Well, I was probably a bit too much preoccupied with the classical barrel/pincushion distortion. If you crop the relevant parts and rotate 90°, it becomes clearer:

90%:
image

image

98%:
image

image

The 98% does seem less curved edge-to-edge (i.e. less barrel), and less "worst" grid deviation than the 90% (in the opposite corner).

But I have to agree that the grid deviation of the 98% image does seem to extend further _into_ the image, i.e. into the area that really matters for our vision stuff. 

90% (rotated 180° to match corner):
image

98%:
image

The fact, that the 98% coverage also changed the detected tilt angle (evident in the fact that the millimeter paper view crop is shifted ~0.5mm horizontally between 98% and 90%), makes it likely that a non-uniform corner lens distortion affects the overall fit too much. 

So I will go back to 90%. 

Thanks for the patience in talking this through.😁

_Mark

tonyluken:
@markmaker,

Ok, I'm good with it.  BTW - I made a few edits to my previous comment.

Tony

markmaker:
Wiki image (somewhat related):
___
perpendicular-constant-xy-diff-z

PR #1422, opened 2022-05-23
Eagle import pads

qwertymodo:
# Description
Add SMD pad definitions to packages on EAGLE board import

# Justification
Currently, the Eagle Board Importer does not import pad outline definitions for packages, even though the information exists inside of the board file, and OpenPnP has the data structures in place for storing those definitions.  There is room for improvement here, such as possibly filtering out only pads that are located on the top layer of the board, or handling of mirrored parts to bring them onto the top layer.  Also, this makes no attempt to import through-hole pads, as it seems like OpenPnp's package model doesn't really have a way to define those.

# Instructions for Use
Click File>Import Board>CadSoft EAGLE Board
Click Browse and select an EAGLE .brd file
Check the box labeled "Create Missing Parts"
Click Import
Navigate to the Packages tab in the main window
Select a newly created package
Footprint outlines should show up outlined in the top vision window
Select the Vision tab in the lower panel
There should be a list of pads, showing their location, size, rotation, and roundness

openpnp_pad_import

# Implementation Details
1. This change was tested by importing several EAGLE board files.
2. I followed the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style) and ran the provided code formatter.
3. I did not make any changes in the `org.openpnp.spi` or `org.openpnp.model` packages.
4. Be sure to run `mvn test` before submitting the Pull Request.

Note: the very large diff in the first commit is because I pulled several outermost layers of the large nested loop out into a separate loop, then reduced the indentation of the remaining loop, causing the entire thing to show as changed.  I did not make any actual changes to the code inside that large loop starting around line 300.

### DIFF
--- a/src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java
+++ b/src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java
@@ -99,8 +99,8 @@ public class EagleBoardImporter implements BoardImporter {
         return board;
     }
 
-    private static List<Placement> parseFile(File file, Side side, boolean createMissingParts, boolean updateExistingParts, boolean addLibraryPrefix)
-            throws Exception {
+    private static List<Placement> parseFile(File file, Side side, boolean createMissingParts,
+            boolean updateExistingParts, boolean addLibraryPrefix) throws Exception {
 
         String dimensionLayer = "";
         String topLayer = "";
@@ -118,12 +118,12 @@ public class EagleBoardImporter implements BoardImporter {
         List<BoardPad> pads = new ArrayList<>();
 
         // Keep track of unique parts as we see them, so we don't update them for every placement
-        HashMap<String,Part> parts = new HashMap<String,Part>();
+        HashMap<String, Part> parts = new HashMap<String, Part>();
 
         ArrayList<Placement> placements = new ArrayList<>();
-        // we don't use the 'side' parameter as we can read this from the .brd file
-        // in the future we could use the side parameter to restrict this from only parsing one side
-        // or the other or both
+        // we don't use the 'side' parameter as we can read this from the .brd file in the future we
+        // could use the side parameter to restrict this from only parsing one side or the other or
+        // both
 
         EagleLoader boardToProcess = new EagleLoader(file);
         if (boardToProcess.board != null) {
@@ -131,19 +131,24 @@ public class EagleBoardImporter implements BoardImporter {
             // first establish which is the Dimension, Top, Bottom, tCream and bCream layers in case
             // the board has non-standard layer numbering
             for (Layer layer : boardToProcess.layers.getLayer()) {
-                if (layer.getName().equalsIgnoreCase("Dimension")) {
+                if (layer.getName()
+                         .equalsIgnoreCase("Dimension")) {
                     dimensionLayer = layer.getNumber();
                 }
-                else if (layer.getName().equalsIgnoreCase("Top")) {
+                else if (layer.getName()
+                              .equalsIgnoreCase("Top")) {
                     topLayer = layer.getNumber();
                 }
-                else if (layer.getName().equalsIgnoreCase("Bottom")) {
+                else if (layer.getName()
+                              .equalsIgnoreCase("Bottom")) {
                     bottomLayer = layer.getNumber();
                 }
-                else if (layer.getName().equalsIgnoreCase("tCream")) {
+                else if (layer.getName()
+                              .equalsIgnoreCase("tCream")) {
                     tCreamLayer = layer.getNumber();
                 }
-                else if (layer.getName().equalsIgnoreCase("bCream")) {
+                else if (layer.getName()
+                              .equalsIgnoreCase("bCream")) {
                     bCreamLayer = layer.getNumber();
                 }
             }
@@ -151,10 +156,10 @@ public class EagleBoardImporter implements BoardImporter {
             // Now we want to establish the width of the board which we need to record
             Double x_boundary = 0.0;
             for (Object e : boardToProcess.board.getPlain()
-                    .getPolygonOrWireOrTextOrDimensionOrCircleOrRectangleOrFrameOrHole()) {
+                                                .getPolygonOrWireOrTextOrDimensionOrCircleOrRectangleOrFrameOrHole()) {
                 if (e instanceof org.openpnp.model.eagle.xml.Wire) {
                     if (((org.openpnp.model.eagle.xml.Wire) e).getLayer()
-                            .equalsIgnoreCase(dimensionLayer)) {
+                                                              .equalsIgnoreCase(dimensionLayer)) {
                         x_boundary = Math.max(x_boundary,
                                 Double.parseDouble(((org.openpnp.model.eagle.xml.Wire) e).getX1()));
                         x_boundary = Math.max(x_boundary,
@@ -166,23 +171,25 @@ public class EagleBoardImporter implements BoardImporter {
                                                          // the Y=0;
 
             // determine the parameters for the pads based on DesignRules
-            for (Param params : boardToProcess.board.getDesignrules().getParam()) {
+            for (Param params : boardToProcess.board.getDesignrules()
+                                                    .getParam()) {
 
-                if (params.getName().compareToIgnoreCase("mlMinCreamFrame") == 0) { // found exact
-                                                                                    // match when 0
-                                                                                    // returned
-                    mmMinCreamFrame_string = params.getValue().replaceAll("[A-Za-z ]", ""); // remove
-                                                                                            // all
-                                                                                            // letters,
-                                                                                            // i.e.
-                                                                                            // 0mil
-                                                                                            // becomes
-                                                                                            // 0
-                    if (params.getValue().toUpperCase().endsWith("MIL")) {
+                if (params.getName()
+                          .compareToIgnoreCase("mlMinCreamFrame") == 0) { // found exact match when
+                                                                          // 0 returned
+                    mmMinCreamFrame_string = params.getValue()
+                                                   .replaceAll("[A-Za-z ]", ""); // remove all
+                                                                                 // letters, i.e.
+                                                                                 // 0mil becomes 0
+                    if (params.getValue()
+                              .toUpperCase()
+                              .endsWith("MIL")) {
                         mmMinCreamFrame_number =
                                 Double.parseDouble(mmMinCreamFrame_string) * mil_to_mm;
                     }
-                    else if (params.getValue().toUpperCase().endsWith("MM")) {
+                    else if (params.getValue()
+                                   .toUpperCase()
+                                   .endsWith("MM")) {
                         mmMinCreamFrame_number =
                                 Double.parseDouble(mmMinCreamFrame_string) * mil_to_mm;
                     }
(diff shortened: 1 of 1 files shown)


markmaker:
Hi @qwertymodo,

This is a very useful addition. I would use it myself 🤪!

I have several questions (numbers for reference):

1. In the future, please do not apply code formatting outside the region that you changed. I had to look at the first two commits and even those contain code reformatting, i.e. your actual changes are hard to see. See [here](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style) (I just updated that section for more clarity).
2. The footprint will only be added if the package is new. Could it be made to always update? 
   https://github.com/openpnp/openpnp/blob/608f77cd58e48bfd14fe04800af280972929e9a1/src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java#L347-L349
4. Regarding polygons: the (already existing) code converts them to their bounding rectangles. 
    https://github.com/openpnp/openpnp/blob/608f77cd58e48bfd14fe04800af280972929e9a1/src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java#L571-L580
   Looking at my own packages in Eagle, I believe this will not come out well. Polygons are rarely used, but if they are used, it is specifically for non-rectangular, often even concave shapes (otherwise you would use one or more SMDs). I wonder if you could extend the `Footprint` to include a list of polygons, aside from the pads. Footprints are always handled as Java `Shape`, so they could easily contain polygons. It is clear that these would not be visible in the OpenPnP GUI, but IMHO that is not necessary, if we can directly import them. Ideally, the GUI would have a field "Number of extra polygons" (bonus 🤗).
7. I wonder if you could add an option (checkbox) on the Import dialog, to **not** prefix parts and package IDs with the library ID.
   https://github.com/openpnp/openpnp/blob/608f77cd58e48bfd14fe04800af280972929e9a1/src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java#L334-L335
    IMHO these long IDs are impractical, prevent easy sorting, and are too large for [feeders with OCR](https://youtu.be/5QcJ2ziIJ14?t=69). Does Eagle itself support having the _same_ ID in multiple libraries, and then use these libraries in the same schematic? Even if it does, I don't think that's a good idea, especially in a pragmatic OpenPnP/DIY environment
_Mark

qwertymodo:
1. I didn't run code formatting until the end in order to avoid that, but the fact that I removed the outer layers of a huge and messy loop caused the entire loop to get marked as changed. If you'd like, I could redo that to pull some of that into a separate commit, but there really wasn't any avoiding it, unless you wanted a huge chunk in the middle left indented several extra levels after those outer loops were removed.
2. In principle, I don't like the idea of overwriting existing data when you're only expecting to import new data. I could perhaps run an extra check and add it footprint to existing packages where the package exists with no footprint. Would that work for you?
3. I left the existing polygon code in there, but I only imported Smds as package pads. I don't understand the intent of the original code well enough to remove what was already there. If you want that gone, feel free to remove it, but I feel that's outside of the scope of this PR.
4. I can look into adding that optional behavior change, but I can say that yes, definitely it can happen that you have an identical package name in multiple libraries, and they can be different, so in that case you'd end up with two different parts getting assigned whichever package happens to come first. This might be ok for PnP purposes though.

markmaker:
About 2: I really think that the OpenPnP Job should reflect the latest import, and whatever changed in the ECAD. Don't you think the ECAD **must** be the "boss"? 

About 3: OK. I think there are actually two kinds of "pads" in there: a) for the footprint, and b) for solder paste lands (back when OpenPnP supported a solder paste dispenser). If we wanted to use the footprint for bottom vision, then adding the polygons would still be required. Are you sure you can't spend the extra effort? 👍👍👍

About 4: Like I said, an option (checkbox) would be best.  😇 Btw. there is a more or less "official" ULP export from Eagle that I think most users actually use for OpenPnP. It just takes the plain IDs. So I guess this is a valid approach.

_Mark

qwertymodo:
On number 2, if the intention was to always update/overwrite existing data, there shouldn't be an optional checkbox to create missing parts, it should always just do it. The existing design pattern is to fill in missing data, not overwrite existing data, so I chose to retain that existing intent.

On 3, the limitation is that I don't see anywhere in the openpnp data model to create pads as arbitrary polygons. If I missed that, can you point me at an example, and I'll be happy to reconsider.

4 is a fair point, I'll look into it.

markmaker:
Hi @qwertymodo,

>  I don't see anywhere in the openpnp data model to create pads as arbitrary polygons.

True, that's why I wrote the following earlier: 

>  I wonder if you could extend the `Footprint` to include a list of polygons, aside from the pads. Footprints are always handled as Java `Shape`, so they could easily contain polygons. It is clear that these would not be visible in the OpenPnP GUI, but IMHO that is not necessary, if we can directly import them. Ideally, the GUI would have a field "Number of extra polygons" (bonus hugs).

True, that would be an extension of your PR. I was just hoping for _extra_ effort towards _extra_ usefulness. 😇

>  if the intention was to always update/overwrite existing data, there shouldn't be an optional checkbox to create missing parts, it should always just do it

I don't understand the argument. If the import was always just adding new parts/packages, then why is there need for that checkbox? If nothing was ever overwritten then it would mean that with the checkbox **off** the import would do **nothing**. Doesn't make sense. 🙃

_Mark

qwertymodo:
> If nothing was ever overwritten then it would mean that with the checkbox **off** the import would do **nothing**. Doesn't make sense. 🙃
> 
> _Mark

That IS what happens. With the checkbox unchecked it doesn't import any parts or packages into the openpnp library, it only imports the board placements. It never overwrites existing parts or packages.

markmaker:
OK, I see. Turns out the footprint is the only actual data being imported (everything else is already **in** the IDs, including the Part-to-Package-assignment), so before you added the footprint, there was no precedence _either_ way, LOL. 

We better go on and both admit the best solution would be to add _another_ checkbox and let users decide. :-)

_Mark

qwertymodo:
Here are my thoughts. First of all, where I'm at IRL, I just can't dive down the rabbit hole of arbitrary polygons, but if somebody adds it into the openpnp core in the future, I may be willing to come back and add it into the eagle importer. As for the rest, I propose:

Remove the existing checkbox to import missing parts, just make this always happen. There's no reason not to.

Add 2 new checkboxes:
-Update existing parts/footprints
-Use globally unique part names

Maybe "globally unique" isn't the best term for the second one, but basically if the first box is unchecked, then only parts and footprints that don't already exist get added, and if it is checked then everything gets imported, overwriting existing parts. The second one toggles whether to include the library prefix on part names.

How does that sound?

markmaker:
Sounds nice, thanks for making the extra effort. Some feed-back in detail:

- Do not remove the "missing parts" checkbox and functionality. There is the scenario that you already imported the project, then deleted all the unwanted placements (you usually import many "things" that are not really parts to be placed: through-hole components, copper-only artifacts such as PCB edge connectors, copper lead jumpers, copper lead antennas, etc), then have to re-import, to correct a detail error or to update a minimal board revision.
- I would name the unique ID checkbox for what it does: "Add library prefix".

Thanks,
_Mark

qwertymodo:
I should have a bit of time to work on this over the next few days.  I've started by cleaning up the original commits to remove all of the extraneous code reformatting.  There will still probably be a big ugly reformat commit at the end due to the deeply nested loop that I removed several outer loops from, but I'll hold off on that until I finish adding in the rest of the actual changes.

markmaker:
Hi @qwertymodo,

thanks for the changes, this is much appreciated! 🌟🌟🌟🌟🌟

Looks good, just these issues:

A new packages should only be created if `pkg` is `null`, otherwise you are creating a duplicate package object _per imported placement_ that is then referenced just by this new part, all other parts will still reference an older one. Later, when config is serialized, these older packages are lost, _including the original one from before the import, and its settings (nozzle compatibility etc.)_:
https://github.com/openpnp/openpnp/blob/7d2d065df4e248bf607bd93b215629a3ba18071d/src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java#L324-L325

In the same new `if {}` you should also just do the `cfg.addPackage(pkg)` (in case you're wondering, this is just adding the Java object reference, you can still modify the object after adding it). So move this line up into the new `if {}`:
https://github.com/openpnp/openpnp/blob/7d2d065df4e248bf607bd93b215629a3ba18071d/src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java#L350

Furthermore, also remove the following section for pretty much the same reason. The part must **not** be removed and recreated, or original settings will be lost. Note: as the part ID is prefixed by the package ID, this code has never made sense anyways:

https://github.com/openpnp/openpnp/blob/7d2d065df4e248bf607bd93b215629a3ba18071d/src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java#L351-L356

While you're at it, you could also remove the second `addPart()` here (if `part` is not `null`, it was already found in `cgf`):

https://github.com/openpnp/openpnp/blob/7d2d065df4e248bf607bd93b215629a3ba18071d/src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java#L363-L365

Thanks again! 👍

_Mark

qwertymodo:
Ok, the new package per placement thing is definitely a problem, good catch.  Unfortunately, it's not as simple as "only create a new package if it's null" because if you're updating an existing package, it will be non-null.  So I guess if it's non-null, instead of creating a new package I should just take the existing package, delete all of its internal shapes and re-add them from the imported footprint.  Then there's the problem of doing it for every placement, which is inefficient, but perhaps it's not too bad unless the part count gets really large.

The line where it removes the part and re-adds it was already there, so I just left it that way.  I'll see what I can do about changing that.

markmaker:
>  So I guess if it's non-null, instead of creating a new package I should just take the existing package, delete all of its internal shapes and re-add them from the imported footprint.

That's a cautious approach if there are other settings on a footprint, that might get lost (there are currently none, AFAIK, but it might be good to allow that for the future). Currently, you are creating a new footprint and the assigning it to the package, so the old one will be collected by the GC.

>  Then there's the problem of doing it for every placement, which is inefficient. 

True, of course an import-only-once-map would help. But as you said, I don't think the performance is a problem.

> The line where it removes the part and re-adds it was already there...

Yes, but the actual problem was created by introducing the `updateExistingParts` switch. Just delete the section, it was non-sensical in the first place. 

Thanks.

_Mark

qwertymodo:
> That's a cautious approach if there are other settings on a footprint, that might get lost (there are currently none, AFAIK, but it might be good to allow that for the future). Currently, you are creating a new footprint and the assigning it to the package, so the old one will be collected by the GC.
You're correct. I dug a bit deeper, and the behavior of the existing setPackage and addPart functions are to simply replace the existing one using HashMap.put(), so there's no need to delete the existing one first.
> True, of course an import-only-once-map would help. But as you said, I don't think the performance is a problem.
It was easy enough to just go ahead and do it right, so this is done.
> Yes, but the actual problem was created by introducing the `updateExistingParts` switch. Just delete the section, it was non-sensical in the first place.
I've restructured the logic in such a way that this should no longer be an issue.  Now, when updateExistingParts is true, the footprint inside the package is replaced, but the original package object retrieved from the config is retained, so when that package object is then "replaced" inside of the part object, anything else attached to it should remain intact.

markmaker:
@qwertymodo,

this is even better! Thanks.

Just one detail:

https://github.com/openpnp/openpnp/blob/e852cdb8e3570a705c5ddb106aab4c3083aa897b/src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java#L325

Should be:

```java
if ((pkg == null && createMissingParts) || (pkg != null && updateExistingParts)) {
```
Otherwise it will create missing parts when just `updateExistingParts` is set, but `createMissingParts` is not. Which, as I explained earlier, is a relevant use case, to import last minute changes from the ECAD.

The same here:

https://github.com/openpnp/openpnp/blob/e852cdb8e3570a705c5ddb106aab4c3083aa897b/src/main/java/org/openpnp/gui/importer/EagleBoardImporter.java#L357

_Mark

qwertymodo:
Good catch. I had a feeling there was bound to be something amiss in there, unwrapping and restructuring control logic is tricky that way...

qwertymodo:
Ok, there we go.  I went ahead and amended the last commit instead of creating a new one just for that little change.  Let me know if there's anything else you'd like to see before I run the code formatter and blow up the combined diff.

markmaker:
# 👍 !

Go ahead and do the planned loop extraction refactoring, as a separate commit.

Thanks again!

_Mark

qwertymodo:
I ran the formatter, but I just realized that the one thing it didn't automatically fix is the comments at the end of deeply nested lines

// that
// look
// like
// this

so I'm going to go back through tomorrow and try to collapse those down a bit manually.

markmaker:
Are you using and IDE that can refactor?

It would be ideal to extract some inner blocks of that loop into separate methods. 

_Mark

qwertymodo:
Ok, that looks better.  The commit diff is awful, but it did shorten up the file by about 10% just by collapsing all of those single-word-per-line comments at the end of long/heavily indented code lines.

markmaker:
@qwertymodo, again thanks a lot for this addition!

# 🚀 🏆

markmaker:
Please announce it in the discussion group. 

And if you could add a screenshot and short explanation in the Wiki, that would be much appreciated:

https://github.com/openpnp/openpnp/wiki/Importing-Centroid-Data#eagle

qwertymodo:
Which discussion group would you like me to post to?  The Google group, or the Discord server?  Or both?

markmaker:
I'm only on the google group. But "both" is also OK. 

Personally, I don't like these chats for non-trivial technical stuff. Chats have no no beginning and no end, and the form seems to make people write without thinking first. Speed and quantity over quality.

_Mark

markmaker:
For the record: I made a mistake here, not merging this into `test` branch first. But luckily it doesn't matter, as I just merged `test` into `develop` anyways, as planned.

See here:
https://groups.google.com/g/openpnp/c/2GHhEK8UpMQ/m/5ex1-pP4BQAJ

_Mark

qwertymodo:
Ok, good to know. I'm working on another unrelated PR right now, so in the future should I be submitting against test instead of develop?

markmaker:
Yep.

PR #1433, opened 2022-06-29
Feature / GcodeDriver, GcodeAsyncDriver: Add $-Command Wait Time

markmaker:
# Description

The TinyG controller is traditionally configured via the `CONNECT_COMMAND` using a series of `$var=<value>` statements. These might cause TinyG to write some settings back to the EEPROM. During that [EEPROM write, interrupts on the MCU have to be disabled](https://github.com/synthetos/TinyG/wiki/Tinyg-Communications-Programming#configuration-commands) therefore no further serial communication is allowed. 

There is now a **$-Command Wait Time [ms]** that pauses sending for that many milliseconds after each command starting with a `$` sign:

Driver Settings

The 50ms default is taken from the [Liteplacer Software source code](https://github.com/jkuusama/LitePlacer-DEV/blob/0607c29c5cca4654fa8e55aa3854325d2108b8df/LitePlacer/TinyG.cs#L364).

## Limitations

With **Confirmation Flow Control** enabled, this is always going to work. This is what **Issues & Solutions** automatically suggest for the TinyG, so it should be OK.

With **Confirmation Flow Control** disabled (i.e. truly asynchronous operation), this is only guaranteed to work, if the $-commands come as the first thing inside the `CONNECT_COMMAND`, or if the controller is otherwise guaranteed to be idle before sending $-commands.

The wait time will now be added to any $-commands, regardless of whether it actually writes to the EEPROM or not. But I guess that's no problem in practice, you can still set 20 settings per second. 

With the default set to 50ms on all machines, including for controllers _other than TinyG_, it might also affect other controller that use $-commands. However, only the `$H` homing command comes to mind, and this usually takes a long time anyway. 

# Justification

See user discussions:
https://groups.google.com/g/openpnp/c/M0K_TirXPcs/m/QGzm2y1BAgAJ

# Instructions for Use

Normally, no action is needed, the default of 50ms is automatically set on the driver. 

You can then see how the commands are staggered by 50ms in the log (go to the **Log** tab):

Log

# Implementation Details
1. Only tested in simulation. I must leave true TinyG testing to its users.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/GcodeAsyncDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/GcodeAsyncDriver.java
@@ -316,8 +316,44 @@ public class GcodeAsyncDriver extends GcodeDriver {
             Logger.debug("{} empty command after pre process", getCommunications().getConnectionName());
             return;
         }
+        if (command.startsWith("$")) {
+            waitForEmptyCommandQueue();
+        }
         CommandLine commandLine = new CommandLine(command, timeout);
         commandQueue.offer(commandLine, writerQueueTimeout, TimeUnit.MILLISECONDS);
+        if (command.startsWith("$")) {
+            waitForEmptyCommandQueue();
+            Thread.sleep(dollarWaitTimeMilliseconds);
+        }
+    }
+
+    /**
+     * A crude way to at least wait for all prior commands to have been sent. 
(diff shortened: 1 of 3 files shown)


markmaker:
Added to Wiki:
https://github.com/openpnp/openpnp/wiki/GcodeDriver#dollar-command-wait-time

PR #1440, opened 2022-07-12
Feature / Vision Compositing - Multi-Shot Bottom Vision

markmaker:
# Description
Adds Vision Compositing, i.e. multi-shot bottom vision, where multiple corners of a part can be aligned individually and then combined into the whole-part alignment. 

multi-shot

The following features are included:

- Align packages that are larger than the camera view.
- Align packages that are not rectangular in their overall shape. 
- Improve accuracy by detecting in the camera center, avoiding errors from parallax, viewing & lighting side angles, and residual lens distortion. 
- Align packages based on pad corners facing inwards, i.e. parts that are asymmetric on the outside, or so large you can't capture their outside corners.
- Enable the Pipeline Editor to cycle through multiple shots.
- Use the nozzle Rotation Mode offset (see #1309) to align the nozzle rotation with the aligned part. 

# Justification
Between tiny 0201/0402 passives and enormous LQFP-256 packages, it is quite hard to get the bottom camera view right. One has to trade resolution against view size. Attempting to escape this dilemma by using a higher resolution camera comes at a significant additional cost in processing power, more lighting required, more compression artifacts, or (worse) reduced frame rate (fps) and additional camera lag. 

Wanting a large camera view also adds constraints to the machine build. The ideal camera has a long focal length ("telephoto") lens to minimize parallax errors, which means it has to be far away from the subject. However, this usually means building a taller machine, especially for table-top design. Consequently, designers often use (very) wide-angle lenses, which are detrimental to alignment accuracy (see the parallax error illustration further below). 

Ironically, the very large parts are often quite rare in projects, just the _one_ MCU, for example. The overwhelming majority of parts are rather small. It hurts to make poor tradeoffs for the few exotic parts.  

Vision Compositing helps with all these problems. 

Cameras can now be optimized to the brunt of the work with medium and small sized parts. You can use medium resolution cameras with high fps, longish focal length lenses and still get a reasonable compact table-top design. A narrower view means more accuracy, and faster convergence on multi-pass alignment. 

The few large parts can then still be aligned with multiple shots. Obviously, there is an extra cost in alignment time, which may or may not be compensated by the avoided tradeoffs overall (more fps, less lag, faster convergence on multi-pass alignment). 

As an independent benefit, multi-shot alignment increases accuracy, particularly for very large parts. In some cases it might actually be the key to successful placement, regardless of the package actually being too large for the camera view. 

Finally, the multi-shot feature also enables alignment of non-rectangular parts (best see the video for examples). 

# Instructions for Use

For an easy overview, [please watch the video](https://youtu.be/P-ZudS7QQeE).

## Camera Roaming Radius

A **Camera Roaming Radius** of zero (default) effectively switches off Vision Compositing.

A **Camera Roaming Radius** of more than zero allows Vision Compositing. The nozzle is allowed to hold the part anywhere within this radius of the bottom camera location. 

Camera Roaming Radius

Within the **Camera Roaming Radius**, the nozzle is allowed to move freely in X, Y, i.e. without going to Safe Z first. This is also used to optimize in-camera drag-jogging. However, this is **not** a soft-limit, you still have to be careful not to bump your nozzle and/or part into any obstacles near your camera. 

The radius must accommodate both the nozzle distance from the camera center, and the part protruding from the nozzle. This is automatically observed during Vision Compositing, taking the the hull of the package footprint into consideration. Because the part is typically held in its center, and because the corners are the package's obvious multi-shot targets, this means that part sizes are more or less constrained to _those with diameters smaller than the **Camera Roaming Radius**_. 

The following illustrates an LQFP-144 package within a 30mm camera roaming radius: 

Roaming as a limit

**CAUTION**: Your camera "pit" must physically provide this freedom of movement, when the nozzle tip and part is at Camera Z (focal plane), and with the worst pick tolerances and other deviations to spare. 

## Other Machine Preparation

It is recommended to enable the **Align with Part?** option on the nozzles:

Align with part switch

It offsets the nozzle rotation coordinate to match that of the part, once the part is aligned. This means that the cross-hairs and the DRO are nicely aligned with the part. This is especially useful in the multi-shot scenario, where you often only see a fraction of the whole part, and the reticle (cross-hairs, grid etc.) provides valuable visual feedback, 

## Package Footprint

In order for Multi-Shot Bottom Vision to work, the footprint must be defined. Most common footprints can easily be generated from very few datasheet parameters, right inside OpenPnP:

Footprint generator

Note: the Eagle E-CAD importer can also import the pads along with the packages. 

## Package Vision Compositing

Basic operation should be automatic. Given the **Camera Roaming Radius**, the multi-shot is enabled. It will automatically be employed with parts nearly as large as the camera view. 

The package **Vision Compositing** tab can give a quick preview of the footprint and its treatment by bottom vision:

Vision-Compositing-Tab

You can hover your mouse over the shots, indicated by corner marker and masking radius. It then indicates the camera view rectangle according to your camera size. The thick red/dashed circle around the package indicates the **Camera Roaming Radius**, i.e. where the nozzle and part can be.

You can press the mouse to see how the pads are fused together, where they are too close to isolate. Only the outside corners of these blocks can be considered as shot locations. But even these are subject to many more constraints.

### Compositing Settings

**Method** determines how the footprint is treated:

Compositing Method Drop-Down

- **None**: the part is aligned with one shot, even if its size or other properties of the footprint would suggest otherwise. This can be used to basically enforce the old behavior. It might be a way to tell OpenPnP to still align select packages with one shot, even if the footprint (plus tolerances) are actually overlapping the camera central circle. Or when pre-rotate is switched off and the part just happens to fit inside the rectangular camera view.
- **Restricted**: the part is always aligned with one shot if it fits inside the camera view (central circle), regardless of other properties of the footprint, like convex pad patterns. This is the default method, intended to provide continuity for the majority of packages. 
- **Automatic**: the part is aligned with one shot if it fits inside the camera view (central circle), and if its corners are understood to be symmetric and convex, to be capture as one bounding rectangle. If this is not the case, a multi-shot is automatically performed. It might combine pairs of corners into "brackets" or "diagonals" shots.
- **SingleCorners**: every isolated corner is shot individually. This can be used for the best accuracy, for the following reasons: 
  - Only the camera center , i.e. a perpendicular "ray" up Z will provide positional information. 
  - Parallax as well as view and light side angle errors are minimized. 
  - Multiple shots reduce random errors, staticstically.

**Extra Shots** determines how many extra shots are taken for this package. The minimum is automatically computed from the geometry of the pads. For example, a square package shape can be aligned with just two corners shots, but the opposing two corners are made available as extra shots, to improve accuracy.

**Max. Pick Tolerance**: determines by how much a package, or rather its corners, can deviate after the pick. When set to zero, the analog setting on the nozzle tip is taken. If non-zero, this value overrides the nozzle tip's. The larger this tolerance, the more the corners must be isolated from other, obstructing corners, or from the limits of the camera view. Finding a solution can therefore fail, if tolerance are too large. 

**Min. Angle Leverage**: determines how far away two corners have to be, in order to define the detected angle of the package, _relative_ to the dimension of the package (the lesser of width and height). A setting of 0.5 means that the corners must be half-way apart across the package. 

**Allow inside corner?**: determines if inside corners are allowed. For very large parts, where even the **Camera Roaming Radius** is not enough to position the part to its corners, the alignment might still be gleaned from the inside corners of large pads.

Inside corners

## Improving Accuracy

The following illustration (exaggerated) shows how a slightly tilted nozzle might result in large placments errors (red) due to large parallax errors in (very) wide angle lenses:

illu

Detecting the same corners using two shots reduces these errors to nothing, as the parallax is negligible when looking straight up from the camera center, and even what little remains, is symmetric left and right, and cancels itself out.

Similar effects might come from viewing pins from the side, and/or having asymmetric lighting. Or from residual lens distortions.

# Implementation Details
1. Tested extensively in simulation and some machine tests.
3. Changes in the `org.openpnp.spi` or `org.openpnp.model` packages are: elevating nozzle tip part dimension  properties to the model. Introducing the VisionCompositing model. Adding the Camera Roaming Radius. Moving the Footprint generator into the model. Adding missing `add(),multiply(), divide()` to the Point model.

Known TODOs for a later PR; the `VisionCompositingTest` unit test is unfinished. I would like to base it on a custom `packages.xml` file, ideally with my own but also problematic footprints from the testing round.

### DIFF
--- a/src/main/java/org/openpnp/gui/PackageCompositingPanel.java
+++ b/src/main/java/org/openpnp/gui/PackageCompositingPanel.java
@@ -1,0 +1,313 @@
+/*
+ * Copyright (C) 2022 <mark@makr.zone>
+ * inspired and based on work
+ * Copyright (C) 2011 Jason von Nieda <jason@vonnieda.org>
+ * 
+ * This file is part of OpenPnP.
+ * 
+ * OpenPnP is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ * 
+ * OpenPnP is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public License along with OpenPnP. If not, see
+ * <http://www.gnu.org/licenses/>.
+ * 
+ * For more information about OpenPnP visit http://openpnp.org
+ */
+
+
+package org.openpnp.gui;
+
+import java.awt.BorderLayout;
+import java.awt.event.ActionEvent;
+
+import javax.swing.AbstractAction;
+import javax.swing.Action;
+import javax.swing.JButton;
+import javax.swing.JCheckBox;
+import javax.swing.JComboBox;
+import javax.swing.JLabel;
+import javax.swing.JPanel;
+import javax.swing.JTextField;
+import javax.swing.border.BevelBorder;
+import javax.swing.border.TitledBorder;
+
+import org.jdesktop.beansbinding.AutoBinding;
+import org.jdesktop.beansbinding.AutoBinding.UpdateStrategy;
+import org.jdesktop.beansbinding.BeanProperty;
+import org.jdesktop.beansbinding.Bindings;
+import org.jdesktop.beansbinding.Converter;
+import org.openpnp.gui.components.ComponentDecorators;
+import org.openpnp.gui.components.VisionCompositingPreview;
+import org.openpnp.gui.support.DoubleConverter;
+import org.openpnp.gui.support.Icons;
+import org.openpnp.gui.support.LengthConverter;
+import org.openpnp.machine.reference.vision.ReferenceBottomVision;
+import org.openpnp.model.BottomVisionSettings;
+import org.openpnp.model.Configuration;
+import org.openpnp.model.Length;
+import org.openpnp.model.Location;
+import org.openpnp.model.VisionCompositing;
+import org.openpnp.model.VisionCompositing.Composite;
+import org.openpnp.model.VisionCompositing.CompositingMethod;
+import org.openpnp.model.VisionCompositing.Shot;
+import org.openpnp.spi.Camera;
+import org.openpnp.spi.Head;
+import org.openpnp.spi.Machine;
+import org.openpnp.spi.Nozzle;
+import org.openpnp.spi.NozzleTip;
+import org.openpnp.util.UiUtils;
+import org.openpnp.util.VisionUtils;
+
+import com.jgoodies.forms.layout.ColumnSpec;
+import com.jgoodies.forms.layout.FormLayout;
+import com.jgoodies.forms.layout.FormSpecs;
+import com.jgoodies.forms.layout.RowSpec;
+
+@SuppressWarnings("serial")
+public class PackageCompositingPanel extends JPanel {
+    private final org.openpnp.model.Package pkg;
+    private JPanel compositingPanel;
+    private JLabel lblMethod;
+    private JLabel lblMaxPickTolerance;
+    private JTextField maxPickTolerance;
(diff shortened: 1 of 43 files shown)


markmaker:
Added the Wiki page.
https://github.com/openpnp/openpnp/wiki/Vision-Compositing

Issue #1109, opened 2021-01-20
ReferenceTrayFeeder: Selecting a move to the current pick location ignores feed count


BaxEDM:
# Problem Report
## Describe the Issue
For a ReferenceTrayFeeder, selecting the "move the camera to the selected feeders pick location" button always moves the camera to the first position of the feeder, regardless of the feedcount. Other feeders take the feedcount into consideration

## Steps to Reproduce
How can a developer reproduce the issue you are experiencing? Use a step by step format like this:
1) Add a ReferenceTrayFeeder
2) Pick an initial pick location in the ReferenceTrayFeeder configuration, click apply
3) add some X offset in the ReferenceTrayFeeder configuration, e.g. X 10, click apply
4) Set tray count X =10, click apply
5) press the "move the camera to the selected feeders pick location" button, the camera moves to the pick location
6) Change the feedcount to 1, click apply
7) press the "move the camera to the selected feeders pick location" button, the camera moves to the pick location

## Expected Result
for step #7 of the steps to reproduce, I expected the camera to move to the picklocation + 1*X offset, since the feedcount was set to 1

## Actual Result
The feedcount is ignored and the camera always moves to the initial pick location

doppelgrau:
The pick-location is only updated after a feed-operation, so without a feed operation, the pick-position is not changed.

diorcety:
> 
> 
> The pick-location is only updated after a feed-operation, so without a feed operation, the pick-position is not changed.

This is not the case for ReferenceStripFeeder for exemple.

ozzysv:
> The pick-location is only updated after a feed-operation, so without a feed operation, the pick-position is not changed.

but the ReferenceStripFeeder for some reason simultaneously updates the position and this is convenient. And with ReferenceTrayFeeder  the question has already been raised several times, but the authors pretend that they don't care...)))

doppelgrau:
It's OpenSource.
Just create a good Pull-Request after thinking if in some corner-cases the change might cause problems.

markmaker:
As @doppelgrau correctly said, that's expected behavior. 
_Mark

Issue #1016, opened 2020-05-28
 change in feeder pick location, not taken into account.


JulienDevillers:
Last version of develop [branch](url)

# Problem Report
## Describe the Issue
A change by typing  coordinates of pick location is not taken into account when "move tool to
the selected feeder's current pick location".

## Steps to Reproduce
- open the joined machine.xml.
- connect and zero the machine
- goto to feeders panel
- click on feeder2
- see that Z for pick location is -26
- click on the button "move tool to the selected feeder's current pick location."
- check on tool coordinates, in status bar that  Z is  -26
- set  Z for pick location to -28.
- click apply
- select another feeder and go back to feeder2
- see that Z for pick location is -28 as it should
- click on the button "move tool to the selected feeder's current pick location."
- check on tool coordinates, in status bar, Z is -26 while it should be -28.

If I close the app and restart, new coordinates are taken into account.

## Expected Result
tool should go to the new coordinates.

[machine.zip](https://github.com/openpnp/openpnp/files/4696263/machine.zip)
[OpenPnP.0.log.zip](https://github.com/openpnp/openpnp/files/4696267/OpenPnP.0.log.zip)

Whenever possible, please use the Menu -> Help -> Submit Diagnostics feature and include the URL here. Submit Diagnostics makes it easy to include log files, screen capture, debug images, configuration files, etc.

-> bad api key.

doppelgrau:
Duplicate of #1109 ?!
Pick location is only updated after a feed operation (can argue if that's good or not, but since a feed operation can be "expensive" it makes somehow sense.)

markmaker:
`pickLocation` should be made a local variable of `getPickLocation()` initialized to `location`. It is reset in each call to `feed()` I see no sensible use of this copy.

_Mark

markmaker:
But for the `ReferenceStripFeeder` that the OP referred to, I see no redundancy. Maybe it was solved in #1031. 

I also cannot reproduce it.

_Mark

markmaker:
See #1449.

markmaker:
Closing this as no longer reproducible for `ReferenceStripFeeder` and solved by #1449 for `ReferenceDragFeeder`. 
_Mark

surabibio:
Wow and Thanks

PR #1446, opened 2022-07-21
Move Camera to Pick Location

surabibio:
# Description
"Move Camera to Pick Location" button in the Feeders tab moves the camera to an empty position except when Feed Count is 0, making this behavior not very useful to the me - an ordinary user.

# Justification
This is a long standing BUG. Also Feed Feeder was moving the nozzle. Moving the camera would be a more useful activity.

# Implementation Details
1. Moved increment of FeedCount to a Post Pick routine. Removed special handling when FeedCount was zero.

### DIFF
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -578,39 +578,14 @@ public class FeedersPanel extends JPanel implements WizardContainer {
                 Feeder feeder = getSelection();
                 // Do the feed and get the nozzle that would be used for the subsequent pick. 
                 Nozzle nozzle = feedFeeder(feeder);
+		feeder.postPick(nozzle);	// moveCameraToPickLocation BUG. Until distinct Increment FeedCount is implemented.
 
 		// Tracking with camera rather than nozzle
-                /*
-		// Note, we do not use nozzle.moveToPickLocation(feeder) as this might involve 
-                // probing, which we don't want to happen here. Instead we need to simulate a preliminary 
-                // pick location. 
-                Location pickLocation = preliminaryPickLocation(feeder, nozzle);
-                MovableUtils.moveToLocationAtSafeZ(nozzle, pickLocation);
-                MovableUtils.fireTargetedUserAction(nozzle);
-		*/
-		    
              	Camera camera = MainFrame.get().getMachineControls().getSelectedTool().getHead()
                         .getDefaultCamera();
                 Location pickLocation = preliminaryPickLocation(feeder, nozzle);
                 MovableUtils.moveToLocationAtSafeZ(camera, pickLocation);
                 MovableUtils.fireTargetedUserAction(camera);
-            });
-        }
-    };
-
-    public Action pickFeederAction = new AbstractAction() {
-        {
-            putValue(SMALL_ICON, Icons.pick);
-            putValue(NAME, "Pick");
-            putValue(SHORT_DESCRIPTION, "Perform a feed and pick on the selected feeder.");
-        }
-
-        @Override
-        public void actionPerformed(ActionEvent arg0) {
-            UiUtils.submitUiMachineTask(() -> {
-                Feeder feeder = getSelection();
-
-                pickFeeder(feeder);
             });
         }
     };
@@ -642,9 +617,25 @@ public class FeedersPanel extends JPanel implements WizardContainer {
         // Perform the feed.
         nozzle.moveToSafeZ();
         feeder.feed(nozzle);
-	feeder.postPick(nozzle);	// moveCameraToPickLocation BUG. Execute postPick if implemented
         return nozzle;
     }
+	
+    public Action pickFeederAction = new AbstractAction() {
+        {
+            putValue(SMALL_ICON, Icons.pick);
+            putValue(NAME, "Pick");
+            putValue(SHORT_DESCRIPTION, "Perform a feed and pick on the selected feeder.");
+        }
+
+        @Override
+        public void actionPerformed(ActionEvent arg0) {
+            UiUtils.submitUiMachineTask(() -> {
+                Feeder feeder = getSelection();
+
+                pickFeeder(feeder);
+            });
+        }
+    };
 
     /**
      * Perform a job-like feed and pick operations sequence. 
@@ -720,7 +711,7 @@ public class FeedersPanel extends JPanel implements WizardContainer {
         }
         // Check the nozzle tip package compatibility.
         Nozzle nozzle = MainFrame.get().getMachineControls().getSelectedNozzle();
-        org.openpnp.model.Package packag = feeder.getPart().getPackage();
+        org.openpnp.model.Package packag = feeder.getPart().getPackage();d
         if (nozzle.getNozzleTip() == null || 
                 !packag.getCompatibleNozzleTips().contains(nozzle.getNozzleTip())) {
             // Wrong nozzle tip, try find one that works.
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -579,12 +579,21 @@ public class FeedersPanel extends JPanel implements WizardContainer {
                 // Do the feed and get the nozzle that would be used for the subsequent pick. 
                 Nozzle nozzle = feedFeeder(feeder);
 
-                // Note, we do not use nozzle.moveToPickLocation(feeder) as this might involve 
+		// Tracking with camera rather than nozzle
+                /*
+		// Note, we do not use nozzle.moveToPickLocation(feeder) as this might involve 
                 // probing, which we don't want to happen here. Instead we need to simulate a preliminary 
                 // pick location. 
                 Location pickLocation = preliminaryPickLocation(feeder, nozzle);
                 MovableUtils.moveToLocationAtSafeZ(nozzle, pickLocation);
                 MovableUtils.fireTargetedUserAction(nozzle);
+		*/
+		    
+             	Camera camera = MainFrame.get().getMachineControls().getSelectedTool().getHead()
+                        .getDefaultCamera();
+                Location pickLocation = preliminaryPickLocation(feeder, nozzle);
+                MovableUtils.moveToLocationAtSafeZ(camera, pickLocation);
+                MovableUtils.fireTargetedUserAction(camera);
             });
         }
     };
@@ -633,6 +642,7 @@ public class FeedersPanel extends JPanel implements WizardContainer {
         // Perform the feed.
         nozzle.moveToSafeZ();
         feeder.feed(nozzle);
+	feeder.postPick(nozzle);	// moveCameraToPickLocation BUG. Execute postPick if implemented
         return nozzle;
     }
 
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -711,7 +711,7 @@ public class FeedersPanel extends JPanel implements WizardContainer {
         }
         // Check the nozzle tip package compatibility.
         Nozzle nozzle = MainFrame.get().getMachineControls().getSelectedNozzle();
-        org.openpnp.model.Package packag = feeder.getPart().getPackage();d
+        org.openpnp.model.Package packag = feeder.getPart().getPackage();
         if (nozzle.getNozzleTip() == null || 
                 !packag.getCompatibleNozzleTips().contains(nozzle.getNozzleTip())) {
             // Wrong nozzle tip, try find one that works.
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -685,8 +685,8 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
             for (int i = 0; i < 1 + feeder.getPickRetryCount(); i++) {
                 try {
                     pick(nozzle, feeder, jobPlacement, part);
-                    postPick(feeder, nozzle);
                     checkPartOn(nozzle);
+                    postPick(feeder, nozzle);
                     return;
                 }
                 catch (Exception e) {
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceRotatedTrayFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceRotatedTrayFeeder.java
@@ -55,13 +55,14 @@ public class ReferenceRotatedTrayFeeder extends ReferenceFeeder {
 	@Element
 	protected Location firstRowLastComponentLocation = new Location(LengthUnit.Millimeters);
 
-	private Location pickLocation;
+	//Private Location pickLocation;  Fix for: change in feeder pick location, not taken into account. #1016
 
 	@Override
 	public Location getPickLocation() throws Exception {
-		if (pickLocation == null) {
-			pickLocation = location;
-		}
+		//if (pickLocation == null) {
+		//	pickLocation = location;
+		//}
+		Location pickLocation;
 		int partX, partY;
 
 		if (feedCount >= (trayCountCols * trayCountRows)) {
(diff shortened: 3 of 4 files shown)


markmaker:
Hi @surabibio,

You're not the first to be bothered by this. See #1016. But it actually behaves correctly as it is.

1. First, you must understand there are **two** operations: **Feed** and **Pick**. 
2. **Feed** makes the feeder produce a new part. In some feeders, this **transports the tape**. In the strip feeder, or grid feeder, it just **advances the pick location to the next location**. 
3. **Pick** then actually picks the part.
5. So _before_ the **feed**, the pick location in a strip feeder is still the _last_ pick location, it will go to an **empty** pocket. 
6. This is actually the same for most other feeders, like for instance an automatic feeder: the pick location will go to an empty pocket, as the tape is not yet transported by the **feed** operation. This is correct, as we don't want to expose the part too early, it might fall out of the tape where the cover tape is removed. In the strip feeder this is just a bit unexpected, but it is **consistent**. You just need to get used to it. 
7. We cannot change how this works, it would break normal operation. One of these is the **Recycle** special operation, where a part is put back into its feeder, so it must still be pointing at the last pick location. 
4. The two separate functions are also somewhat reflected in the fact that there are two buttons on the feeder panel: two button. One does a **feed** _and_ **pick**. The other just a **feed**. 

I hope that you will now see how your changes are not correct. You also added the **postPick** after the **feed**. But is must only come after the **pick**. 

_Mark

surabibio:
Hi Mark, I could understand the funda of Feed and Pick, but as a user this reasoning is not convenient. Now with the new fix everyone in our team is happy and the Move Camera button is doing useful work. As for issue #1016 I felt this BUG in Drag Feeder. This also should be corrected. I will give it a try.
Ravi

markmaker:
> Hi Mark, I could understand the funda of Feed and Pick, but as a user this reasoning is not convenient. Now with the new fix everyone in our team is happy and the Move Camera button is doing useful work. 

I have to say that unless you find a way for it to work seamlessly for _all_ the feeder types, I'm afraid, this will never be acceptable. 

For many feeders, feeding _after_ picking is not practical or acceptable. Often this will expose a (potentially expensive) part (peel off the protective cover tape), so this must only be done shortly before a part is used. 

You also completely changed the meaning of the **feed** button, moving the camera, in addition to doing the **feed**, which makes no sense, because there is already a button for moving the camera to the pick location. Furthermore, this will fail for feeders where the camera can't reach the pick location at some edges of the machine, which is a valid scenario. 

Imagine users with automatic feeders. They expect the **feed** button to just do a tape transport, instead it moves the camera to the pick location. This might even be dangerous, they might be holding their heads or expensive camera close, in order to observe/record the feeder mechanism. Ouch!

I agree, it takes a bit of _education_ to understand the two-step behavior of a feeder that **changes its pick location** with the **feed** operation (strip feeder, tray feeder, blinds feeder, feeders that produce multiple parts per feed actuation, like the push-pull feeder with 2mm pitch 0201/0402 parts or using the multiplier). But this is really something that users _must_ understand sooner or later, at the latest when your team advances to more elaborate feeder types. So better educate them now.

> As for issue #1016 I felt this BUG in Drag Feeder. This also should be corrected. I will give it a try.

I'm sorry, I actually meant #1109 (which is referenced from #1016 for explanation).

surabibio:
Hi Mark, This fix is only for the StripFeeder.  I have not altered jobPlanner or any other feeder or played with the base files. 
_public void postPick(Nozzle nozzle)_ is an already existing function that is not used currently by any other feeder and I just leveraged it to fix this issue and I think you should give 5 starts for this approach. 

>  _Imagine users with automatic feeders. They expect the feed button to just do a tape transport, instead it moves the camera to the pick location. This might even be dangerous, they might be holding their heads or expensive camera close, in order to observe/record the feeder mechanism. Ouch!_

Wait a second. Right now as per the old code
`                MovableUtils.moveToLocationAtSafeZ(nozzle, pickLocation);
                MovableUtils.fireTargetedUserAction(nozzle);
` the behavior is to move the nozzle to the pick location. Getting banged by the camera is better than getting banged by the nozzle. Just kidding. To the best of my knowledge I have not made any dangerous alterations.

I will post a video of how many of the small snippets including Feeder Groups script help us in setting up and finishing the job much faster.
Ravi

markmaker:
> This fix is only for the StripFeeder. I have not altered jobPlanner or any other feeder or played with the base files... 

You added `postPick()` to the `feedFeeder()` method in the `FeederPanel`. That will now be called for all feeders and is completely wrong after **feed** (or it would be called `postFeed()`).

> `postPick(Nozzle nozzle)` is an already existing function that is not used currently by any other feeder

It is:

References of postPick


> Wait a second. Right now as per the old code ... is to move the nozzle to the pick location. Getting banged by the camera is better than getting banged by the nozzle. Just kidding. 

You're right, this is wrong and _as_ dangerous! 

Having said that, your change is still a bit "less correct", as the _camera_ might not be able to reach the pick location or make something collide with the feeder if on the far side, whereas the _nozzle_ obviously always must be able to go to the feeder pick location.

Whatever, this _must_ be removed. 

**Back to the actual intent.** I think I now understand what you _actually_ want to achieve: 

You want the `ReferenceStripFeeder` to increment its feed count _after_ the pick in `postPick()`, instead of in the `feed()`, _before_ the pick. 

If I understand it correctly, the camera moving is a mere side show and could be removed, i.e. the user can always press the proper camera positioning button, which then positions to the right tape pocket (and you might want to consider changing the PR title).

There is a problem with that: 

The `ReferenceStripFeeder` uses computer vision to realign the pick location with the tape. It also serves the purpose of detecting when it runs out of tape (no more sprocket holes = no more parts). Running out of parts will make the `JobProcessor` retry a few times, then switch over to the next feeder with the same part, or if there is none, it will interrupt the Job for a reload, or _defer_ the placement.

Vision checking happens in the call to `feed()`, which is the correct place, _semantically_, as you'll see below. If you put it in `postPick()`, it is the wrong place, for these reasons:

1. `JobProcessor` retry loops will not work: it will just _stick_ at the same feed count and pick location. Therefore, after a part was missing in the tape for instance, or badly picked, the feeder vision pick location realignment, pick, vacuum sensing, bottom vision, size check operations will not work, the part is obviously still missing at the _same_ pick location over again. 
2. You might then consider moving the computer vision to `postPick()` and maybe even adding your own retry loop there, etc. ... but that's even worse:
3. There will be no vision, no retry loop etc. for the first part of a job. A moved strip or missing part is not detected/handled.
4. The same is true after a failure happened before `postPick()` was even called. 
5. If it actually fails in `postPick()`, it will _fail_ the **already successful** last placement, instead of the **next** placement of that part.
6. If there _is no next placement of that part_ in the job, the `postPick()` failure is artificial, e.g. if a user prepared a strip of _exactly_ the needed number of (expensive) parts, they will get a _bogus_ hard to understand error for its last part.

Conclusion:

The **feed** and **pick** operations as **two steps**, and **in that order** is not a random design choice Jason made because he did not know better. It is designed that way for many very good reasons.

As I said: in the end it is a matter of some investment in _team education_ rather than of "fixing what ain't broken".

_Mark

markmaker:
Just to show how I solved the education/usability issue in the `BlindsFeeder`, I made the vision operation number the part locations with their **feed count** (**Show Features** button):

Numbered Parts

With a similar thing for the `ReferenceStripFeeder`, users would perhaps better understand how the relationship between **feed count** and pick location works, but without breaking the system. 

_Mark

markmaker:
See #1448.

surabibio:
> postPick(Nozzle nozzle) is an already existing function that is not used currently by any other feeder

I agree. I have missed on this. This will mess with those feeders unless - the user is using postPick for something like "increment feeder" in which case they would welcome this....
I am just curious to know what the feed and postfeed does for those autoFeeders.

Other than this I don't think it will have any other side effects and vision will continue to function as usual, but you know better, so I leave it to you.

Yes feed and pick are two operations. Sometimes feed should be inhibited - example the fist part after setup in Drag Feeder. For some feeders Feed should happen after Pick - e.g Strip Feeder, Rules like Feed should precede Pick is Draconian.

markmaker:
Hi @surabibio,

> Other than this I don't think it will have any other side effects and vision will continue to function as usual, but you know better, so I leave it to you.

Well, it is all explained in detail [here](https://github.com/openpnp/openpnp/pull/1446#issuecomment-1192665293). If my explanation is not understandable in some points, or you think I'm mistaken, please tell me in what way.

> For some feeders Feed should happen after Pick - e.g Strip Feeder, Rules like Feed should precede Pick is Draconian.

How is this supposed to work, if the availability of a part (running out of tape, switching to next feeder) is determined in the feed operation? Do you have a crystal ball or a time machine?

_Mark

surabibio:
Hi Mark, 
I have only moved `setFeedCount(getFeedCount() + 1);` to postPick. Other than this there is no change in the feed.
Vision checking, end of tape checking are as usual and they are operating on the part before the pick as old code.

OK 
`if ((maxFeedCount > 0) && (feedCount > maxFeedCount)) {`
needs to be changedf to 
`if ((maxFeedCount > 0) && (feedCount >= maxFeedCount)) {`

I hope I have made my case clear. Do you still feed it will produce the side effects you mentioned?
(If so will accept with no further ado)

Ravi

markmaker:
Hi Ravi,

>... I have only moved `setFeedCount(getFeedCount() + 1);` to postPick. Other than this there is no change in the feed. Vision checking, end of tape checking are as usual and they are operating on the part before the pick as old code.

> I hope I have made my case clear. Do you still feel it will produce the side effects you mentioned? (If so will accept with no further ado)

OK let's try the source code approach.

Inside the `JobProcessor` there are three _nested_ retry loops, each retrying over an exception handler: 

The _outer_ loop is a two-step **feed** _then_ **pick** retry loop:
https://github.com/openpnp/openpnp/blob/af43024e03f6d9e98c738cff8fb1424be52c487b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L574

The first _inner_ loop is nested _inside_ the outer loop's **feed** step:

https://github.com/openpnp/openpnp/blob/af43024e03f6d9e98c738cff8fb1424be52c487b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L648

The second _inner_ loop is nested _inside_ the outer loop's **pick** step:

https://github.com/openpnp/openpnp/blob/af43024e03f6d9e98c738cff8fb1424be52c487b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L685

You are moving the pick location increment from the front of the _first_ inner loop to (almost) the back of the second _inner_ loop. This changes the behavior severely. The feeder will be stuck at the same feed count if it fails anywhere _before_ the `postPick()`. 

If it fails inside `feed()` (failure to detect a sprocket hole), it will retry the `feed()` a few times, but due to your change, without advancing the pick location. If a sprocket hole cannot be recognized by vision, because e.g. it is a bit torn, this will happen again and again, and the job will be stuck. 

If it fails after a successful `feed()`, but before `postPick()`, the outer loop will try to feed and pick the _same_ part again and again, _despite_ calling `feed()` repeatedly. It will also uselessly repeat the vision operation again and again, at the same location. 

Even after the `JobProcessor` gives up, and the user restarts the Job later, it will still be stuck at that same _already failed_ feed count. Users can also press the **feed** button as many times as they like, it won't help. Users would have to _manually_ set the feed count to the next usable part. This is a complete break with how other feeders behave.

On the other hand, if it fails _after_ `postPick()`, the inner loop, that should try to pick the _same_ pick location a few times (perhaps the vacuum didn't hold), will now unexpectedly _advance_ the pick location and pick new parts each time. Butn now _without performing the vision check_, as the vision is in `feed()` and _not_ in `postPick()`. So it will miss detecting the end of the tape, which will cause the nozzle to pick beyond the tape (and in the worst case might even cause damage).

Another point is the `takeBackPart()` function that you would have to fix too. 

_Mark

surabibio:
Hi Mark,
Thanks for the very detailed explanation. Now we are on the same page.

> Vision checking happens in the call to feed(),...  If you **put it in postPick(),** it is the wrong place, for these reasons:

I admit was confused by that bit of your message.

Made the changes to source code and submitting for your kind evaluation.

markmaker:
Hi @surabibio,

I don't see how the latest changes address the issues listed by me. In fact, some open new cans of worms.

_Mark

surabibio:
### 1
> If it fails inside feed() (failure to detect a sprocket hole), it will retry the feed() a few times, but due to your change, without advancing the pick location. If a sprocket hole cannot be recognized by vision, because e.g. it is a bit torn, this will happen again and again, and the job will be stuck.

The below change in feed(Nozzle nozzle) takes care ...
```
        try {
		updateVisionOffsets(nozzle);
	} catch (Exception e) {
		setFeedCount(getFeedCount() + 1);
		throw e;
	}
```
### 2
> If it fails after a successful feed(), but before postPick(),

ie, if it fails in this code block
`                /**
                 * Currently this will throw and abort the placement if it fails. Probably it should
                 * discard and retry, and really it should probably be done before we attempt to
                 * feed. I *think* this has been debated as to whether or not it's useful
                 * and should maybe be done at the end of the cycle, rather than here. Maybe it just
                 * gets removed completely.
                 */
                checkPartOff(nozzle, part);
`
As the comments say it all, we can just ignore it as of now. 

### 3

> On the other hand, if it fails after postPick(), the inner loop, that should try to pick the same pick location a few times (perhaps the vacuum didn't hold), 

this piece of code will handle it
```
                    pick(nozzle, feeder, jobPlacement, part);
                    checkPartOn(nozzle);
                    postPick(feeder, nozzle);
                    return;
```

ie. we check the vacuum first before calling postPick. 
PostPick() as we discussed is not used in any feeder except the auto feeders. Is this change welcome or not by AutoFeeders is the question? My strong guess is it is good for autofeeders.
My reasoning is as below:
For Machines operating at high CPH feed time is comparable to pick and place time and therefore feed should start right after the pick else the machine will stall. So autofeeder users do the feed (or tape advancement) in the postPick() function. If so this change will be beneficial to them because previously when pick fails (due to vacuum) it was advancing the feeder. Now it will not and is consistent with the behavior in other feeders.

markmaker:
You cannot just address selected points and choose to ignore the others. Each of your adjustments changes _more_ of OpenPnP's current behavior. 

The exception handler in `feed()` now actually _feeds_ when it fails and does _not feed_ when not. You kind of inverted the semantics! 

And you did not address all the _other_ failure paths I listed. Your strategy seems to ignore these and even swap `postPick()` with `checkPartOn(nozzle)` at your convenience for _all_ the feeders, which opens another can of worms. 

And all that, just because you refuse to learn the simple and natural logic that the pick location of a _positional_ feeder is not updated until you actually _feed_ it, simply because _feeding_ means incrementing the position. I genuinely don't understand, why this is so much of a problem for you, or your team. 

It is normal to be surprised first, ask the question, but after the explanation, it should be a non-issue. See it three times and don't even think about it. 

https://youtu.be/6n9ESFJTnHs?t=14

_Mark

surabibio:
Can you remove the terminology Feed and change it to 
PrePick -> Pick -> PostPick -> PrePlace -> Place -> PostPlace
Feeders can choose to do whatever they want when each of these functions are called by the Job Processor.

I intentionally do not want to be specific to questions, but if this suggestion is not acceptable, you may close the PR.

markmaker:
Hi @surabibio,

I think the processor cannot be agnostic to what really happens when, because it must know it for the retry loops. The term "feed" is semantically loaded as the "advancing" element. To effectively move the `feed()` to after the pick (which is what you actually want), it would have to be known as such by the `JobProcessor`, so it can span up the retry loop properly. And it would actually have to break this in two halves, as both the first halve (the feed, in this case the pick location increment) and second halve (the vision operation checking if there is still a sprocket hole) can fail. The retry loop must span both. But now the first halve is at the end of the `JobProcessor` cycle, and the second halve is at the front. So the retry loop would have to be some sort of **in-and-out-of-cycle-hybrid** depending on where it fails.

Again, I don't think the **initial justification** is valid for making everything so much more complex. Users just need to learn and understand how it works, and why the pocket is one off _before_ the feed.

The technical principle that the _next_ slot will be used is quite common. I wish I had found a less martial example, but maybe being martial actually helps remember it 🤠

https://youtu.be/GLD3IX4eXKk?t=58
 
Unlike me, you are likely all too young to know what a slide projector is, and that its lever is also always one off the next slide. 😆

_Mark

markmaker:
As you suggested, I will close this now.

surabibio:
What about this:
https://www.youtube.com/watch?v=SG-FXLxLOSM

PR #1448, opened 2022-07-22
Fix / Remove Nozzle Move from `feedFeederAction`

markmaker:
# Description
Remove the nozzle move to pick location from the `FeedersPanel`'s `feedFeederAction`.

# Justification
User broke nozzle tip, because he rightly did not expect the move to pick location when he pressed the **feed** button on a drag feeder:

https://groups.google.com/g/openpnp/c/qOlRCwZeIWw/m/HlDPTYjqAAAJ

# Instructions for Use
Use the **feed** button just for the feed:

Feed button

If you want to position the nozzle as well, use the **position nozzle** button _explicitly_:

Position button

# Implementation Details
1. Tested in simulation.

### DIFF
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -578,13 +578,6 @@ public class FeedersPanel extends JPanel implements WizardContainer {
                 Feeder feeder = getSelection();
                 // Do the feed and get the nozzle that would be used for the subsequent pick. 
                 Nozzle nozzle = feedFeeder(feeder);
-
-                // Note, we do not use nozzle.moveToPickLocation(feeder) as this might involve 
-                // probing, which we don't want to happen here. Instead we need to simulate a preliminary 
-                // pick location. 
-                Location pickLocation = preliminaryPickLocation(feeder, nozzle);
-                MovableUtils.moveToLocationAtSafeZ(nozzle, pickLocation);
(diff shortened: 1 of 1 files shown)


surabibio:
After feeding what does the user do?
He moves the camera to the pick location to see if all is well.
Previously the nozzle was moved to the pick location - which also made sense.
I have not complaints for removing the nozzle move. But I am disappointed that "Move Camera to pick location" is not implemented.
_**Feed has been paralyzed**_

markmaker:
Hi @surabibio,

I don't understand your point. There is a button to move the camera (or the nozzle) to the pick location, if you like that. We are talking about manual GUI actions here, for setup or troubleshooting. There is no call for production level automation. 

As for the `ReferenceStripFeeder` the feed will do the feed vision operation and the camera will be moved over the sprocket hole. So a user already sees the feeder state. Like I said earlier, the vision operation, that already displays a pipeline result image, could paint the numbers of the pick locations into that result image, for clarity. Adding this would be a very welcome contribution. 💯 

I agree that removing the nozzle move might be a bit unintuitive for feeders like the tray feeder, where nothing physical happens. But even for those feeders, it was strange that the nozzle actually moved to the part, pressing down on it. Not something I would appreciate for an expensive IC, for instance, when I just wanted to test **feed** to see how it advances the pick location. 

_Mark

surabibio:
Ok! Got the point. But for users (like me) who don't use computer vision the feed button will not move the head now. So please consider moving the Camera To Pick Location after Feed.

PR #1454, opened 2022-07-26
Fix ReferenceLeverFeeder Feed Start Location Z Height Ignored

qwertymodo:
# Description
The ReferenceLeverFeeder currently throws away the Z height of the Feed Start Location, instead using the safe Z height configured in the Z axis.  This causes two main problems.  First of all, you can't set the start height to anything other than the Z axis safe height.  Not a huge issue, but still a bug.  Second though, because a lever feeder will often be actuated with a pure up-down motion of the Z axis, this means that the "return to start" move becomes a no-op, causing the actuator to be turned off before the lever arm has properly returned to its starting height.  This results in a very sudden snap-back motion, which can lead to parts jumping, or just additional wear and tear on the lever arm.  You can see the result very clearly by slowing down the feed speed and noting that the actuator turns off before the lever arm has a chance to be returned to its starting position, as shown here:

https://youtu.be/Ma-H7PO320E

# Justification
This fixes a bug.

# Instructions for Use
Add a ReferenceLeverFeeder to your machine, and configure the Feed Start Location to a Z height other than your Z Safe Height, and/or with a functional feed actuator configured.  Perform a feed operation, and note that prior to this fix, the Feed Start Location did not respect the configured Z height, and the actuator will turn off before the lever is returned to its start height.

# Implementation Details
I chose to break the initial move to the Feed Start Location into two moves, starting with an XY-only move at the Safe Z Height, as the current implementation already does, because it seems prudent to make that move at a safe height, not knowing where the head might be prior to initiating the feed command.  The code already issues a head.moveToSafeZ() before moving to the pick location, so it's not necessary to add that after the return to start position.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceLeverFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceLeverFeeder.java
@@ -148,6 +148,7 @@ public class ReferenceLeverFeeder extends ReferenceFeeder {
 		    
 	            // Move the actuator to the feed start location.
 	            actuator.moveTo(feedStartLocation.derive(null, null, Double.NaN, Double.NaN));
+	            actuator.moveTo(feedStartLocation);
 		    
 	            // enable actuator (may do nothing)
 		    	actuator.actuate(true);
@@ -160,7 +161,7 @@ public class ReferenceLeverFeeder extends ReferenceFeeder {
 		    		peelOffActuator.actuate(true);
 		    	}
 	            // Now move back to the start location to move the tape.
-	            actuator.moveTo(feedStartLocation.derive(null, null, Double.NaN, Double.NaN));
+	            actuator.moveTo(feedStartLocation);
 		    
 			    // Stop the take up actuator
 		    	if (peelOffActuator != null) {
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceLeverFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceLeverFeeder.java
@@ -144,12 +144,13 @@ public class ReferenceLeverFeeder extends ReferenceFeeder {
             Location feedStartLocation = this.feedStartLocation;
             Location feedEndLocation = this.feedEndLocation;
 
+            // Move the actuator to the feed start location, at a safe height.
+            actuator.moveTo(feedStartLocation.derive(null, null, Double.NaN, Double.NaN));
+
+            // Move the actuator to the actual feed start location height.
+            actuator.moveTo(feedStartLocation);
+
 		    for (double i = partPitch.convertToUnits(LengthUnit.Millimeters).getValue(); i > 0; i=i-4) {  // perform multiple feeds if required
-		    
-	            // Move the actuator to the feed start location.
-	            actuator.moveTo(feedStartLocation.derive(null, null, Double.NaN, Double.NaN));
-	            actuator.moveTo(feedStartLocation);
-		    
 	            // enable actuator (may do nothing)
 		    	actuator.actuate(true);
 		    


qwertymodo:
I took a look through the other reference feeder implementations to see if a similar fix might be relevant to any of the others, and as far as I can tell, it doesn't seem to be.

markmaker:
Hi @qwertymodo,

Thanks for having a go at that.

When I made the [`ReferencePushPullFeeder`](https://github.com/openpnp/openpnp/wiki/ReferencePushPullFeeder), it started out as an extension to the `ReferenceLeverFeeder`. After a while it became too complex, and I made a new class instead. But while doing that, I too noticed the missing Z. So why did I not fix it?

Because it is a change of behavior. Users now have an arbitrary Z value stored in the start location. Maybe they tried something and when they saw that it does not work, they just left some odd Z value there. Then one day they upgrade OpenPnP and suddenly the machine now _uses_ that Z coordinate! If the Z is higher than Safe Z, the actuator might not properly insert with the lever, go half above it, then make a diagonal move down, and badly chew things up, if the Z is low, it might crash.

I guess you'd have to make an update handler that stores the current `actuator.getSafeZ()` into the Z location on all existing `ReferenceLeverFeeder`s. And you would have to remember the update handler _state_, so even if users swap out multiple `feeder.xml` files, it will still work correctly, on the _individual_ feeder. A typical pattern is to add an `@Attribute Double version` and a `@Commit` handler to fix it, based on the stored or null `version`, then set the new `version`.

Furthermore, unless I'm missing something, you could just remove the `head.moveToSafeZ() `here:

https://github.com/openpnp/openpnp/blob/f19179baeb24e175de91824f37e3a9aa893c2a07/src/main/java/org/openpnp/machine/reference/feeder/ReferenceLeverFeeder.java#L139

And instead use a `MovableUtils.moveToLocationAtSafeZ()` here, and get rid of the deprecated ~~`Double.NaN`~~ hack and the second move:

https://github.com/openpnp/openpnp/blob/f19179baeb24e175de91824f37e3a9aa893c2a07/src/main/java/org/openpnp/machine/reference/feeder/ReferenceLeverFeeder.java#L148-L151

The `MovableUtils` method is preferred because it is a (future) location for motion optimization.

Also move the second  `head.moveToSafeZ() ` into the `if(...) {  ... }` body. So if the feeder does not need to actuate, it will not do anything. 

https://github.com/openpnp/openpnp/blob/f19179baeb24e175de91824f37e3a9aa893c2a07/src/main/java/org/openpnp/machine/reference/feeder/ReferenceLeverFeeder.java#L185

While at it, please fix the mixed tab/space indenting (use only space) in that method (but not the whole file). 

Thanks!

_Mark

markmaker:
Ah, and I forgot to mention: 

The  [`ReferencePushPullFeeder`](https://github.com/openpnp/openpnp/wiki/ReferencePushPullFeeder), can do (almost) anything the `ReferenceLeverFeeder` can do and _much more_. 

Best watch the video, but be ware that the featured 3D printed feeder is just one design example, the [`ReferencePushPullFeeder`](https://github.com/openpnp/openpnp/wiki/ReferencePushPullFeeder) can be used with virtually _every_ mechanical and electro-mechanical feeder design.

https://youtu.be/5QcJ2ziIJ14

_Mark

qwertymodo:
But while doing that, I too noticed the missing Z. So why did I not fix it?
> 
> Because it is a change of behavior. 

Ok, I can understand that, makes perfect sense.

> I guess you'd have to make an update handler that stores the current `actuator.getSafeZ()` into the Z location on all existing `ReferenceLeverFeeder`s. And you would have to remember the update handler _state_, so even if users swap out multiple `feeder.xml` files, it will still work correctly, on the _individual_ feeder. A typical pattern is to add an `@Attribute Double version` and a `@Commit` handler to fix it, based on the stored or null `version`, then set the new `version`.

Ok, do you have any examples I could look into that would show how to implement that?  It's definitely more work than I'd hoped for, but better to do it right.

> Furthermore, unless I'm missing something, you could just remove the `head.moveToSafeZ() `here:
> 
> https://github.com/openpnp/openpnp/blob/f19179baeb24e175de91824f37e3a9aa893c2a07/src/main/java/org/openpnp/machine/reference/feeder/ReferenceLeverFeeder.java#L139
> 
> And instead use a `MovableUtils.moveToLocationAtSafeZ()` here, and get rid of the deprecated ~`Double.NaN`~ hack and the second move:

Does MovableUtils combine the multiple moves in the single call (move Z-only to safe height, then move XY-only to destination, then move Z to destination)?  If so, absolutely that's a better option, I simply wasn't aware of it.

> While at it, please fix the mixed tab/space indenting (use only space) in that method (but not the whole file).

Will do.  The mixed indentation annoyed me too, but that was how it already was, so I opted to make the new code match the old.  I'll leave it for now, and run a whitespace-only commit at the end to keep the commits clean in the meantime.

markmaker:
> Ok, do you have any examples I could look into that would show how to implement that?

Perhaps this:

https://github.com/openpnp/openpnp/blob/af43024e03f6d9e98c738cff8fb1424be52c487b/src/main/java/org/openpnp/machine/reference/camera/calibration/AdvancedCalibration.java#L270-L280

> Does MovableUtils combine the multiple moves in the single call (move Z-only to safe height, then move XY-only to destination, then move Z to destination)? 

Yes. 

> Will do. 

Cool! Thanks.

_Mark

markmaker:
> Will do.

@qwertymodo  you still on this?

qwertymodo:
Yes, I'm still looking to finish this up.  I somehow managed to break builds on my Windows box that's actually attached to my machine, so I haven't been able to test anything.  Not really sure what's going on, I can't even build clean checkouts right now, and the machine that I *can* build on isn't anywhere close to my PnP to actually test functionality.  Hopefully I can figure it out soon.

markmaker:
@qwertymodo ping.

qwertymodo:
Sorry I've been AWOL on this. I recently moved, and the process of refinishing my basement which will become my primary workspace, which was supposed to take 6 weeks, instead took 6 months. I literally just got concrete poured yesterday. So I'm definitely still going to be finishing this, things have just been on hold over here due to life circumstances. But things are finally moving forward now, so hopefully I'll have time soon.

markmaker:
@qwertymodo  ping again.

markmaker:
I'll close this for now. 

I guess ReferencePushPullFeeder can do everything the ReferenceLeverFeeder can do and more, and things are fixed there.

If need be, it can be reopened.

_Mark

Issue #1459, opened 2022-08-01
BlindsFeeder - OCR not working on Test branch


crashmatt:
# Problem Report
Doing OCR on BlindsFeeder always throws an error reported as **missing an "OCR" stage in the pipeline** 

## Steps to Reproduce
How can a developer reproduce the issue you are experiencing? Use a step by step format like this:
1. Configure a BlindsFeeder to use OCR
2. Select the "OCR Detect" button
3. Machine moves to OCR location to take image

## Expected Result
Should collect OCR as normal

## Actual Result

![BlindsFeeder OCR Error](https://user-images.githubusercontent.com/351261/182206342-f8a5478d-b1e0-4194-8417-d28a57097f6e.png)

## Notes, Log Files, Screen Captures, Videos, etc. to Show the Issue

The pipeline including SimpleOCR stage
![BlindsFeeder OCR error pipeline](https://user-images.githubusercontent.com/351261/182206431-efae4574-1284-41f2-8eb2-de7321e2d07e.png)

[The check for the property is here](https://github.com/openpnp/openpnp/blob/af43024e03f6d9e98c738cff8fb1424be52c487b/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java#L711
)

![Debug showing alternate property naming](https://user-images.githubusercontent.com/351261/182206632-70f10454-329a-44b1-a3d3-976089d269b9.png)

-----------------------------------------------------------------------

# Feature Request
## Describe the Feature
Change the checked property name from "alphabet" to "SimpleOcr.alphabet" to match the pipeline.

markmaker:
Please test with newest `test` version. 

@crashmatt, thanks again for reporting.

I'm closing this as you already tested. Should there be a problem left, we can reopen.

_Mark

PR #1461, opened 2022-08-01
Fix / BlindsFeeder: Qualified Pipeline Property "SimpleOcr.alphabet" (Cherry-Pick)

markmaker:
This is simply the cherry-pick from #1460.


markmaker:
This also fixes #1459 for the `develop` branch, where it is also present.

PR #1463, opened 2022-08-02
PushPullFeeder ocr read offset

crashmatt:
# Description
Adds an option to move the OCR identification position along the axis of the feeder.
Moves the OCR part check to before the new position of alignment holes is submitted to the feeder.  This avoids new hole alignment being applied to the wrong feeder.

# Justification
A camera that has limited scope on the machine y-axis can't see both the tape holes and a position to read OCR or QRCode.  The OCR read position needs to be offset.

# Instructions for Use
On PushPullFeeder configuration tab use the "OCR offset" value.  Positive offset is away from the feeder.  Negative offset is towards the feeder.
A zero offset results in default behavior where the OCR is checked at the mid tape hole location together with the first tape hole position check.  A non zero offset results in the OCR being checked just after the first tape hole position check.
Use "Setup OCR region" to pick an new region of interest at the new offset location.  The region moves relative to the OCR offset.  If the offset changes then the region may need to be changed.
Selecting "Part by OCR" results in the camera moving to the new position before OCR check.
Feeding results in the additional OCR check only as required.

# Implementation Details
Zero offset shows no additional camera movement or image capture to check OCR.
Non zero offset moves the camera along the feeder axis.  This has only been tested with a feeder mounted at one side.
Feeder direction is derived from the direction between tape hole 1 and 2.
Using a feeder with offset as a template results in the offset being copied.

  With luck yes.
3. No changes to `org.openpnp.spi` or `org.openpnp.model` packages 
4. `mvn test` passed

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java
@@ -211,6 +211,8 @@ public class ReferencePushPullFeeder extends ReferenceFeeder {
     protected double ocrFontSizePt = 7.0;
     @Element(required = false)
     protected RegionOfInterest ocrRegion = null; 
+    @Element(required = false)
+    protected double ocrOffset = 0.0; 
 
     public enum OcrWrongPartAction {
         None,
@@ -997,6 +999,16 @@ public class ReferencePushPullFeeder extends ReferenceFeeder {
         return ocrRegion;
     }
 
+    public void setOcrOffset(double ocrOffset) {
+        Object oldValue = this.ocrOffset;
+        this.ocrOffset = ocrOffset;
+        firePropertyChange("ocrOffset", oldValue, ocrOffset);
+    }
+    
+    public double getOcrOffset() {
+        return ocrOffset;
+    }
+    
     public void setOcrRegion(RegionOfInterest ocrRegion) {
         Object oldValue = this.ocrRegion;
         this.ocrRegion = ocrRegion;
@@ -1229,6 +1241,24 @@ public class ReferencePushPullFeeder extends ReferenceFeeder {
         }
     }
 
+    public Location getOcrVisionLocation() throws Exception {
+        if (!(hole1Location.isInitialized() && hole2Location.isInitialized())) {
+            // not yet initialized, just return the current camera location
+            return getCamera().getLocation();
+        }
+        else {
+//            if (getCamera().isUnitsPerPixelAtZCalibrated()) {
+//                if (!getLocation().getLengthZ().isInitialized()) {
+//                    throw new Exception("Feeder "+getName()+": Please set the Pick Location Z coordinate first.");
+//                }
+//            }
+            Location ocrVectorOffset = hole1Location.unitVectorTo(hole2Location).multiply(ocrOffset, ocrOffset, 0.0, 0.0);
+            return getHole1Location().add(getHole2Location()).multiply(0.5).add(ocrVectorOffset)
+                    .deriveLengths(null, null, getLocation().getLengthZ(), 
+                            getLocation().getRotation()+getRotationInFeeder());
+        }
+    }
+
     protected void setupOcr(Camera camera, CvPipeline pipeline, Location hole1, Location hole2, Location pickLocation) {
         pipeline.setProperty("regionOfInterest", getOcrRegion());
         pipeline.setProperty("SimpleOcr.fontName", getOcrFontName());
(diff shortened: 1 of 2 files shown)


markmaker:
Hi @crashmatt,

Thanks for the initiative.

Some comments (numbered for easy reference):

1. We want to support hand-mounted, continuous feeder arraying, so we want a huge tolerance of ±2mm
1. The Region of Interest (ROI) must be quite tight, as edges etc. might be misinterpreted as characters in OCR.
1. This is probably less of a concern with modern barcodes where you can add a generous margin with anything in it, but I'd like to keep it functional for all use cases, whenever possible (personally I still find human readable OCR more practical in a DIY environment).
1. Therefore, I still consider it necessary to calibrate **first**, then cut out the ROI to read the OCR/barcode **second**.
1. You seem to be concerned that the wrong feeder will be calibrated before OCR action is applied, i.e. before it is replaced with the _right_ feeder. But unless I'm missing something, that concern is **unfounded**: The vision operation will just "see" the sprocket holes, not the pockets, i.e. it will calibrate the sprocket hole locations, but the pick location is just translated _relatively_, according to the sprocket holes. In fact the calibration does nothing to any of the _stored_ locations, it just determines the _relative_ offsets and rotation, that are then _dynamically_ applied to the feeder, whenever the locations are evaluated. So it will not "overwrite" the pick location of a configured 12mm tape, 8mm pitch feeder with that of a 8mm tape, 2mm pitch feeder that happens to be physically present at its former "slot" position (assuming this is your concern).
1. This is of course under the assumption that you use more or less the same model archetype for all your feeders, i.e. where the sprocket holes and ROIs are all in the same _typical_ positions. But then again, in your approach you are assuming the same, implicitly, otherwise the feeder replacement (translation) and cloning won't work. 
1. So I'm fine with the second OCR/barcode shot, but I think you should keep the calibration order intact, suppress OCR in the calibration loop, and instead just launch a second camera capture when the calibration loop has finished. 
1.  Separating it like this, I think you can reduce or eliminate the amount of changes inside `FindFeaturesMode` and `FindFeatures`. I'm a bit concerned about these changes, because this is heavily optimized combinatorial multi-purpose code (Auto-Setup, Cloned Setup, Calibration, Calibration and OCR, just OCR). I can't put my finger on anything that won't work with your code, but I would be happier if we kept the changes small. It took me quite a while to get it right in the myriad of use cases.
1. Furthermore, I'm not happy with just a one-dimensional offset. A universal solution would allow for both horizontal and vertical displacement (in feeder relative terms).
1. Instead of making the GUI and the offset vs. ROI definition/validity more complex, I propose you extend the `RegionOfInterest` and `RegionOfInterestProcess` like so:
1. As an option of the `RegionOfInterestProcess`, make a new process step, where you tell the user to jog to the desired location where the camera sees the ROI.
1. Then record the jog offsets (camera location distance before/after) in the ROI.
1. Then use that for the second shot (if `offsets.isInitialized()`).
1. This will make sure there is never any question of invalid ROI vs. offsets, and it is much more intuitive for the user.

I know I'm demanding a lot, but it is really worth it, IMHO. Seeing how fast you were after the initial idea, I hope you'll take it as an easy challenge! 💪😁

_Mark

crashmatt:
@markmaker  Thanks for the detailed review.

I did a one dimension offset because I found a single space in the GUI for
a value.  Your suggested method is better for the user.

Having a completely separate OCR check after alignment makes sense.  In
most cases it only needs to be done once so it won't slow down the process
much.

I don't think there will be much code in common on the new implementation.
To make this clean I will start a new branch from test and a new PR.

I still have the concern about doing alignment before properly identifying
the correct feeder.  I will test this to convince myself that this is not a
real problem.  First I need to print some new 2040 feeders.

markmaker:
Hi @crashmatt 

it's a good thing discussing this with you, it made me thinking more about all this and I think I actually found a bug in my original concept. 

Currently the "origin" of a feeder is the pick location. If another feeder is swapped-in due to OCR, it will be swapped-in at their pick locations. This works with all the same types of tapes (unfortunately, I only tested this case really), but it is wrong if that other feeder has different tape width or different first uncovered part pick location. 

Currently it takes the two sprocket holes closest to the pick location to calibrate the position and then records the pick location as the feeder origin, and the mid-point of the sprocket holes as the reference for the ROI:
relative coordinates old

As it should be, it must always take the sprocket holes closest to the reel, and their mid-point as the feeder origin and the reference point for the ROI. So feeders of arbitrary tape width and part size can be swapped-in:
relative coordinates

I will need to fix that myself. But I thought you should know.

_Mark

PR #1468, opened 2022-08-07
Push pull feeder - OCR position offset

crashmatt:
# Description
Adds an option to offset the OCR identification camera position when performing the OCR region of interest.

[Improvement on this PR](https://github.com/openpnp/openpnp/pull/1463)

# Justification
A camera that has limited scope on the machine y-axis can't see both the tape holes and a position to read OCR or QRCode. The OCR read position needs to be offset.

# Instructions for Use
When setting the OCR region of interest jog the camera to the required position before setting the region.
The offset is relative to the position and rotation of the feeder.

# Implementation Details
How did you test the change?
Zero offset shows no additional camera movement or image capture to check OCR.
Non zero offset moves the camera to the preset position. This has only been tested with a feeder mounted at one side.
Feeder direction is derived from the direction between tape hole 1 and 2.
Using a feeder with offset as a template with copy vision enabled results in the offset being copied.
Auto setup works as expected.
Adds an additional camera capture and pipeline run which slows down the process.  This will only happen when the OCR is first verified.

Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?
With luck yes.

No changes to org.openpnp.spi or org.openpnp.model packages

mvn test passed


crashmatt:
There may be a more streamlined way to do the position rotation.  Currently using atan2 of vector between the tape holes.

markmaker:
Hi @crashmatt,

I see this is still "WIP", so I might be kicking at an open door, then please disregard the following 😇:

I just wanted to know why you put the OCR offsets in the feeder and not the `RegionOfInterest` (and the `RegionOfInterestProcess`), as discussed [in points 10 - 14 here](https://github.com/openpnp/openpnp/pull/1463#issuecomment-1203733055).

There would be no `atan2()` sorcery needed on the offsets, instead it would just have to be rotated in `RegionOfInterest.rotateXy(double)`, along with the rest of the locations. And the rotation is already applied properly here:

https://github.com/openpnp/openpnp/blob/baa3d0ae380f7a5c64be2e9055663b9a4509076e/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L2224-L2225

I'm also thinking about other _future_ uses of `RegionOfInterest` and how to make it as easy as possible.

_Mark

crashmatt:
> https://github.com/openpnp/openpnp/blob/baa3d0ae380f7a5c64be2e9055663b9a4509076e/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L2224-L2225
> 

I was struggling to read and understand enough of the existing design concepts to make an elegant solution.  I had too many open questions about what ROI is ,how the process works, what is responsible for moving cameras etc.  For my own sanity I just had to make something work.   Now I have a basic working system I can take some time to refine it.

/Matt

crashmatt:
Maybe the transform could be sent to the ROI Process to return a location transformed to the feeder.   Then the ROI could store the offset instead of the feeder storing the offset.  Not sure about this.

crashmatt:
Bug : When entering the wizard and adjusting the camera position, the camera view does not always live update.

markmaker:
@crashmatt, I'm not sure I understand. What I proposed is not duplicating  the `RegionOfInterestProcess` but to extend it with the jogging location delta, perhaps as an option (boolean) to be passed to the constructor. And there is no need to store the offsets as a location in the feeder. The relative before-after offsets should be stored as such in the ROI and should always be taken from there. The offsets should be handled in `rotateXy()`, like the other Locations.

> Bug : When entering the wizard and adjusting the camera position, the camera view does not always live update.

That's strange. But when you jog otherwise, does it always update?

_Mark

crashmatt:
> @crashmatt, I'm not sure I understand. What I proposed is not duplicating the `RegionOfInterestProcess` but to extend it with the jogging location delta, perhaps as an option (boolean) to be passed to the constructor. And there is no need to store the offsets as a location in the feeder. The relative before-after offsets should be stored as such in the ROI and should always be taken from there. The offsets should be handled in `rotateXy()`, like the other Locations.
> 
> > Bug : When entering the wizard and adjusting the camera position, the camera view does not always live update.
> 
> That's strange. But when you jog otherwise, does it always update?
> 
> _Mark

I only just worked out what rotateXy is doing just before you sent your message.  Had not seen setFeederLocation and relocateFeeder.  Now it makes sense.  Still learning.

Using a boolean to configure the process was one of the options I had considered.
RegionOfInterestProcess was easy to put an extra step at the end.  Putting an extra step at the start was not so easy.  There were very many constants spread around that were dependent on the sequence.  I am sure there might have been a hack but I decided to tidy-up instead.  I made a new process to keep the large change separated. 
I will add the boolean to RegionOfInterestLocationProcess.  If the new style passes inspection then we can use it as the original RegionOfInterestProcess.

/Matt

crashmatt:
I am testing with a duplicated RegionOfInterestProcess for now. Once I have
it under control I will change RegionOfInterestProcess.  Have just
submitted something that works closer to the way you say.

*That's strange. But when you jog otherwise, does it always update?*
I have not noticed any problems in normal operation.  Jsut got a
capture in RegionOfInterestProcess
that was blurred.
I don't have a clear pattern yet on when it happens.  Failure rate is about
10%.

markmaker:
> Jsut got a capture in RegionOfInterestProcess that was blurred.

Is this related to #1462 (points 11. and 12.)?

https://youtu.be/Pxg6g3KI5_E

_Mark

crashmatt:
Not sure which point 11/12 is.  It is not due to camera setting time.  The camera image should live update as normal.  For some reason it gets frozen just as the machine completes its move.  It may have been a problem before but not noticed since ROI does not normally use a dynamic image. Will keep an eye on this and report an issue if required.

markmaker:
>  since ROI does not normally use a dynamic image

I don't see why it should not. 

I just tested with **Suspend in Tasks** and even with **Preview FPS** = 0. Obviously, the camera view is not live with FPS=0, but it _does_ update after jogs. I also confirmed that it _does_ proper Camera Settling after each jog. 

Strange. You would have to log this at Trace level. 

_Mark

crashmatt:
Now fixed in RegionOfInterestWizard.

markmaker:
Hi @crashmatt 

I like the enum in the process, nicely done!

Earlier I made a mistake: I said that calibration is done before the OCR pass. This is not true as it turns out (see the `i == 0` condition). So I guess your PR will actually improve that as well! 👍👍👍

There are still a few details that I propose to make it even better:

1. No need to subclass the `RegionOfInterestOffset`. It is not really OOP if both the regular and the sub-class are then created in the end. Just add the offsets to the regular `RegionOfInterest` and leave it `null` or `Location.origin` if not set in the process.
1. Call it `offsets` / `getOffsets()` (with "s" for plural, i.e. multiple coordinates), as is the convention in OpenPnP.
1. Like I said earlier, I'd like to keep the performance roughly the same for those that do not use offsets, i.e. there should be no extra camera capture. Propose to implement as follows:
1. In the loop, keep the OCR prepared on this condition:
   ```java
   boolean ocrPass = (ocrAction != OcrWrongPartAction.None && getOcrRegion() != null && !getOcrRegion().getOffsets().isInitialized());
   ```
1. After the loop inside the `if` check for `getOcrRegion().getOffsets().isInitialized()`: 
   - if `true`, perform the extra camera offsets move, call `setupOcr()`, and call `pipeline.process()` which settles and captures a new camera frame.
   - else just use the pipeline as is, it should then just use the already captured camera frame and OCR results from the last loop iteration (it will still be slightly slower than before, as OCR vision is now performed in every loop iteration).
1. Make `FindFeaturesMode.ocrOnly` capital case, as per coding style.
1. Cosmetics: You can remove the empty `else {}` at the end. 
1. Cosmetics: simplify `getOcrLocation()` to use one `if (offset == null || !offset.isInitialized())`. Use the local variable in the `add()`.

Correct me if I'm wrong, it is easy to overlook something. 😬

_Mark

crashmatt:
Agree on everything but there might be a small problem with 5.

OCR gets switched off in the pipeline unless it is needed.  A processing speed saving.
The hole calibration position loop can exit at the end when it finds its position is within tolerance.
OCR is now done at the end to improve accuracy.
If OCR is normally not done in the pipeline then there is no OCR data ready to use.
The decision to enable OCR has to happen both before and after the pipeline has run.

The only alternative is to run the pipeline a second time with OCR enabled.  OCR should only happen once per machine r so the penalty is small.  The extra camera capture can be avoided.

No need to put ocrPass in the loop if the decision to run the OCR pipeline has to happen after the loop.

/Matt

crashmatt:
Unless processing the pipeline and taking a new image can't be separated.

markmaker:
> If OCR is normally not done in the pipeline then there is no OCR data ready to use.

See my point 4. We must keep the OCR prepared in the loop (unless there are ROI offsets). I'm aware that this means some additional computation cost, see my point 5, second sub-point, where I already mentioned that. 

But I think that extra cost is tolerable. Note, the loop will only iterate, when the feeder / tape has in-deed moved by more than `calibrateToleranceMm` from its nominal/old position, i.e. more than 0.3mm by default. This should not routinely happen. And the OCR stage also has a lesser cost than taking a second camera shot, with settling (~12ms on my feeder/camera setup). 

> The only alternative is to run the pipeline a second time with OCR enabled. OCR should only happen once per machine r so the penalty is small. The extra camera capture can be avoided.

This is not possible, at least not without major changes to how pipelines work. When you do `pipeline.process()` it also executes the `ImageCapture` stage, which settles and captures the camera. 

_Mark

markmaker:
Almost...

Read my point 4 carefully. We still need to check whether OCR is at all wanted, and that a region is defined. Bascially, you need to restore the `if {} else {}` from _before_ this PR, but with the slightly changed condition that I formulate in my point 4.

One additional point that I just saw here: 

https://github.com/openpnp/openpnp/blob/1cf8d9bb09daf46734405f9a5fe3d43c16300a6b/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L1482-L1483

Make it so:
```java
                if (autoSetupMode != FindFeaturesMode.OcrOnly || calibrationTrigger == CalibrationTrigger.None) {
                    // No vision calibration wanted - just copy the pre-set locations
                    calibratedHole1Location = getHole1Location();
                    calibratedHole2Location = getHole2Location();
                    calibratedPickLocation  = getLocation();
                }
                else {
```
and remove the matching curly brace here:

https://github.com/openpnp/openpnp/blob/1cf8d9bb09daf46734405f9a5fe3d43c16300a6b/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L1720-L1721

That's cleaner. Btw. make sure you have no tabs in the indents, it currently looks wrong on GitHub, that's usually a sign that there are tabs. 

Point 8 is still open.

Thanks for the endurance... :+1: 

_Mark

markmaker:
Mistake in my code proposal (I also fixed it above, to avoid confusion). 

It should read:

_Make it so:_
```java
                if (autoSetupMode == FindFeaturesMode.OcrOnly || calibrationTrigger == CalibrationTrigger.None) {
                    // No vision calibration wanted - just copy the pre-set locations
                    calibratedHole1Location = getHole1Location();
                    calibratedHole2Location = getHole2Location();
                    calibratedPickLocation  = getLocation();
                }
                else {
```

crashmatt:
_Btw. make sure you have no tabs in the indents, it currently looks wrong on GitHub, that's usually a sign that there are tabs._

Code style is my nemesis.  Never did get the code style check to work.

crashmatt:
Yes.  I managed to miss/delete this logic. 

`                if (autoSetupMode == FindFeaturesMode.OcrOnly || calibrationTrigger == CalibrationTrigger.None) {
                    // No vision calibration wanted - just copy the pre-set locations
                    calibratedHole1Location = getHole1Location();
                    calibratedHole2Location = getHole2Location();
                    calibratedPickLocation  = getLocation();
                }
                else {`

crashmatt:
May have solved the mystery of the tabs.  I neglected to indent the whole code block around the extra logic on point 4.  Now it is merged the indent is correct as it originally was.

markmaker:
Looks good, but (in case you missed them) points 4, 8 are still open.

crashmatt:
This logic doesn't fit any more:

`boolean ocrPass = (ocrAction != OcrWrongPartAction.None && getOcrRegion() != null && !getOcrRegion().getOffsets().isInitialized());`

Offsets can be uninitialized with zero offset.  We still want an ocr pass. The initialization state is checked during the ocr pass.

`boolean ocrPass = (ocrAction != OcrWrongPartAction.None && getOcrRegion() != null)`

markmaker:
> This logic doesn't fit any more:
> 
> `boolean ocrPass = (ocrAction != OcrWrongPartAction.None && getOcrRegion() != null && !getOcrRegion().getOffsets().isInitialized());`

Well, there are now two possible OCR operation modes (mutually exclusive):

1. The one in the loop. It is only used if there are **no offsets**, because in case of offsets, it will be done after the loop, see (2). 
2. The one after the loop. If **there are offsets** it needs to move the camera and take a new camera shot and process the pipeline.

So yes, I still think the logic is correct inside the loop. 

_Mark

crashmatt:
Mode 1 does not exist the way it did.  It has changed to running ocr after location is finalized rather than on the first pass.

An extra optimization is to set ocrPass before location finding so it does not have to do the ocr processing if it is not required.  I don't see how to do it better than this.  What am I missing?

markmaker:
Probably easier to explain in code, I've added comments inline:

```java
         final boolean ocrWanted = (ocrAction != OcrWrongPartAction.None && getOcrRegion() != null);

        // Calibrate the exact hole locations by obtaining a mid-point lock on them,
        // assuming that any camera lens and Z parallax distortion is symmetric.
        for (int i = 0; i < calibrateMaxPasses; i++) {
            // move the camera to the mid-point 
            Location midPoint = runningHole1Location.add(runningHole2Location).multiply(0.5, 0.5, 0, 0)
                    .derive(camera.getLocation(), false, false, true, false)
                    .derive(null, null, null, runningPickLocation.getRotation()+getRotationInFeeder());
            Logger.debug("calibrating sprocket holes pass "+ i+ " midPoint is "+midPoint);
            MovableUtils.moveToLocationAtSafeZ(camera, midPoint);
            if(ocrWanted 
                    && !(getOcrRegion().getOffsets() != null && getOcrRegion().getOffsets().isInitialized())) {
                // OCR wanted and we do not have ROI offsets, so we need to prepare the OCR now, in order to get calibration and OCR
                // in one fell swoop. Note, only the last loop iteration will actually be evaluated, after the loop.
                setupOcr(camera, pipeline, runningHole1Location, runningHole2Location, runningPickLocation);
            } else {
                // OCR either unwanted, or the ROI has offsets and we will need an extra camera move and capture, 
                // therefore disable OCR in this pass.
                disableOcr(camera, pipeline);
            }
            // take a new shot
            pipeline.process();
            feature = new FindFeatures(camera, pipeline, 2000, FindFeaturesMode.CalibrateHoles)
                    .invoke();
            runningHole1Location = feature.calibratedHole1Location;
            runningHole2Location = feature.calibratedHole2Location;
            runningPickLocation = feature.calibratedPickLocation;
            // calculate the worst pick location delta this gives, cycle part 1 is the worst as it is farthest away
            Location uncalibratedPick1Location = getPickLocation(1, runningVisionOffset);
            Location calibratedPick1Location = getPickLocation(1, feature.calibratedVisionOffset);
            Length error = calibratedPick1Location.getLinearLengthTo(uncalibratedPick1Location);
            Logger.trace("new vision offset "+feature.calibratedVisionOffset
                    +" vs. previous vision offset "+runningVisionOffset+" results in error "+error+" at the (farthest) pick location");
            // store data if requested
            if (storeHoles) {
                setHole1Location(runningHole1Location);
                setHole2Location(runningHole2Location);
            }
            if (storePickLocation) {
                setLocation(runningPickLocation);
            }
            if (storeVisionOffset) {
                // update the stats
                if (visionOffset != null) {
                    // Only when a previous vision offset has been stored, should we store the error
                    // because the feeder might have been moved physically. The user's actions are 
                    // not part of the calibration error. :-)
                    addCalibrationError(error);
                }
                setVisionOffset(feature.calibratedVisionOffset);
            }
            // is it good enough? Compare with running offset.
            if (error.convertToUnits(LengthUnit.Millimeters).getValue() < calibrateToleranceMm) {
                break;
            }
            runningVisionOffset = feature.calibratedVisionOffset;
        }
        
        // setup OCR if wanted
        if (ocrWanted) {
            Location ocrOffsets = getOcrRegion().getOffsets();
            if (ocrOffsets != null && ocrOffsets.isInitialized()) {
                // We have ROI offsets, need to move the camera, capture and process for OCR.
                Location ocrLocation = getOcrLocation();
                MovableUtils.moveToLocationAtSafeZ(camera, ocrLocation);
                setupOcr(camera, pipeline, runningHole1Location, runningHole2Location, runningPickLocation);
                pipeline.process();
                feature = new FindFeatures(camera, pipeline, 2000, FindFeaturesMode.OcrOnly)
                        .invoke();
            }
            // else : we already have OCR results from the last calibration loop·iteration (see in the loop above).
            if (feature.detectedOcrModel == null) {
                Logger.warn("Feeder "+getName()+" OCR operation expected, but no \"OCR\" stage result obtained from pipeline.");
            }
            else {
                Logger.trace("got OCR text "+feature.detectedOcrModel.getText());
                triggerOcrAction(feature.detectedOcrModel, ocrAction, ocrStop, report);
            }
            disableOcr(camera, pipeline);
        }
```

crashmatt:
Yes. The penny has finally dropped.
Have made slight modification which I think is more human readable.

Taking this out of WIP

markmaker:
Brilliant, many thanks, @crashmatt !

PR #1473, opened 2022-08-17
Machine squaring with triangles - Proof of concept

crashmatt:
# Description
Using an unknown reference angle to correct machine squareness.  With preliminary support in Issues and Solutions.

# Justification
A simple method for correcting machine squareness without the need for any absolute reference.

# Instructions for Use

- Print the [reference fiducials from this document](https://drive.google.com/file/d/1jhcAxBoRd40y0sn8uLP7Qn4ZjcLJz6V9/view?usp=sharing).  This is three homing fiducials with unknown angle between them.  The accuracy of your printer does not matter.
-  Run OPnP.
- Go to Issues and find "Calibrate machine squareness"
- Adjust the fiducial size.  Try 1.5mm.
- Place the paper at the fiducial height oriented so you can see the rotation 1 orientation text correct when you are at the front of the machine.  Exact orientation does not matter.
- Jog the camera to fiducial 1 and press the button the capture Rotation 1 Position 1.  If the fiducial can be captured then the machine will correct its location to it.  If not then the set camera location will be captured.
- Repeat for positions 2 and 3
- Rotate the paper so the rotation 2 text can be seen properly.
- Repeat the same process capturing positions 1,2,3 for rotation 2.
- Press the accept button.  Then the open button and then accept again.  Its a bug.
- The text should show the machine skew angle and the skew coefficient.
- Now go to your Machine Setup -> Axes -> ReferenceLinearTransformAxis y -> X input -> Factor.  Remove the skew coefficient from that factor.  Apply the new coefficient. 
- Repeat the positions/points capture again to confirm it is correct.

My initial corrections and measurements seem to reduce skew to below 1mm / m.  More 

# Implementation Details
Has a test module which can be used to verify measurement on different X and Y input skews.
Not yet verified with a good physical standard.

Code style broken.  Proof of concept only.
org.openpnp.model modifications
mvn test included

**Notes and todo list:**
- Using fiducials is highly recommeded for usability.
- Add automated modification of the ReferenceLinearTransformAxis factors. 
- Using the Solutions format makes the user interface difficult to use.  This can probably be fixed.
- Much more automation can be done to make this process easier.  For example, moving the camera to the estimated location of the next fiducial.
- The current implementation needs a specific orientation and pattern. I don't see a real requirement to need anything else.
- The diagonal angles of the triangle are not so sensitive to skew so they can't be used.  It might help to change to a square pattern.  Then there are 4 angles of interest instead of one.  More measurement redundancy is good.
- With a known square pattern the measurement might be automated after the first fiducial.
- Need the auto fiducial size feature from calibration.
- Some method to identify the fiducials would help to prevent user errors.


tonyluken:
Hey @crashmatt, I suspect you really intended to merge this into test rather than develop.  As it is now it is almost impossible to review since GitHub is showing 128 files have changed - I assume most of those are not due to your changes.

crashmatt:
I did notice the rather long change list.  Thanks for the correction.

crashmatt:
Closing due to PR into develop instead of test

markmaker:
No need to close. Just change the branch.

markmaker:
You have the `ReferencePushPullFeeder` changes in there. 

1. Finish your PR #1468 
2. I'll merge it.
3. Pull.
4. Then merge `test` into your `feature/SquaringByTriangles`.
5. Push.

This will sort it out.

_Mark

markmaker:
Just noticed: There are Advanced Calibration solutions still in there from copy&pasting.

tonyluken:
I worked out the least-squared-error math for using three points in a triangular arrangement to compute the non-squareness correction.  See the last section here [NonSquarenessCorrection.pdf](https://github.com/openpnp/openpnp/files/9377949/NonSquarenessCorrection.pdf)

Note that using this method you can still compute a scaling factor for one of the axes (you have to assume the other axis has a scaling factor of 1.000) in addition to the non-squareness correction.

crashmatt:
@markmaker This PR is a hack.  No code style intended yet.

@tonyluken I had noticed that a relative scaling factor would be possible.  If an absolute measurement of one triangle lengthis done then this can be used to calibrate the absolute machine scaling.

Thanks for including this in your analysis.  Good to see a proof of my random thoughts.  
It makes sense to solve this matrix style.  I just don't have any intuition about what characteristics this calculation has.

With machine skew some angles increase and some decrease.  Angles in some positions are not influenced by skew at all.  If you take a line through the half angle between the points. Skew sensitivity requires having half angle components on both axis.  Sensitivity to skew is  Sin(halfAngle)*Cos(halfAngle).  Diagonal half angles are 100% influenced.   Half angles at the axis are 0% influenced.

The angle magnitude has influence on sensitivity to skew.  The half angle can be on the diagonal but with a small angle difference.  This behavior might be linear with angle.

All said, using a 90 degree angle aligned with an axis is ok.  Anything else is questionable.

After closing the PR I started working on a Quadrilateral version of the same thing.  Once you know the position of three points then it is easy to automate the collection of the fourth.  More information makes for a more robust measurement.  Your matrix calculation might be applied four times, once per corner.

/Matt

markmaker:
Hi @crashmatt, @tonyluken,

you guys lost me. Can you explain to me _why exactly_ we are discussing this second variant? 

I really don't want to be a killjoy here, but unless I'm missing something, I don't see how a printed paper is better than the ruler. 

When I was squaring my machine physically ([see here](https://github.com/openpnp/openpnp/wiki/Machine-Squaring-Guide)), I had big troubles with the millimeter paper, very hard to really get down flat all over. I think a ruler is a much better proposition, especially for the more advanced "frame style" machine designs, that allow for sunken/docked feeders etc. rather than the simple table top designs. 

Just as an example:

Peter Betz Parametric pick and place (PnP) machine
Image Credit: [Betz Technik Parametric pick and place (PnP) machine](https://www.betztechnik.ca/store/p44/Parametric_pick_and_place_(PnP)_machine_design_with_BOM!.html)

Those would need laying the ruler across gaps. Either using a stiff ruler, or it being suspended on a bar, which is much easier to find than suspension for a full sheet of paper. 

If this alternative is about using fiducial vision for better precision, I believe the same precision can be achieved with a ruler, using a small circle-masked crop of the first measurement for [template matching](https://docs.opencv.org/4.x/d4/dc6/tutorial_py_template_matching.html) in the second. So any inaccuracies in the manual click capture simply do not matter, and the user is free to actually choose any good non-repetitive ruler pattern as the template: 

Template Matching

This also allows hands-off re-capturing of both points with slow-acceleration moves to make it irrelevant if the ruler has slipped a bit during the typically jerky hand jogging, where almost all DIY machines tend to shake a bit. Can even do it in multiple passes/reverse direction, to self-check that the reduced machine acceleration did in-deed prevent the ruler from slipping. 

_Mark

crashmatt:
@markmaker The method you find easiest will depend on your machine and your preference.  The triangles method could work by putting three pen dots on a piece of cardboard.  
The homing fiducial has already been used so it doesn't take extra effort.  With automation only the first fiducial will need to be found.  The machine will search for the others in a pattern or at least assist in moving to the right location.

Preferred method comes down to user choice.  Some users might like to compensate with one method and verify with another.  Some users might might wish to use a good absolute angle reference.  Lets not restrict users to one method.

Being honest about this I am biased. I get a kick out of the idea of getting the machine square by using three random dots on piece of paper.

/Matt

tonyluken:
@markmaker,
I'm not advocating for this method but merely providing a more solid mathematical development for it.  I still think the ruler method is the way to go.

@crashmatt,
I wouldn't bother with developing a Quadrilateral version.  I think you'd accomplish the same thing by just collecting the "triangle pattern" more than twice (each rotated differently than the others).  The matrix math I developed works with as many triangle collections as the operator cares to collect (minimum of two).  Also note that it does not use/care about any angles, it just needs the observed locations of the dots to work.  And, yes, if you knew the actual spacing between any two of the dots, it should be possible to compute the true scaling for both axes.

crashmatt:
@tonyluken
I was thinking of sensitivity to measurement error.  I had not thought about this in depth and collecting an extra data point is  easy.  I have not thought enough about what accuracy I need.

My home fiducial is in one corner and my tool changer (without vision) is in the other.  To keep accuracy with that I need about +- 0.25mm max over 500mm.  That would be a squareness better than +-0.0005.  This requirement is removed after implementing multiple fiducials for machine correction since the machine becomes more repeatable.

Next is PCB placement accuracy.  I have not thought about this but you probably have.  
Say you need no more than 0.1mm error over a 150mm*150mm PCB, assuming there is no skew correction from the PCB fiducials.  That is a maximum skew of 0.00067.  


My measurement repeatability on fiducials is in the order of +-0.02mm.  Double that for a line.  Double that for an angle.  Double that when comparing two angles.  Say the measurement is over 150mm.  Skew measurement error could be 0.001 peak.  This peak should be reduced over multiple measurements.

/Matt

tonyluken:
@crashmatt,

I think you are way too concerned about the accuracy you need from one corner of your machine to the other corner.  Generally, you use the machine to capture the locations of tool changers, feeders, PCBs, etc.  For those purposes, it only matters that your machine can repeatably return to the same location.  The absolute scaling and squareness of the machine doesn't matter much at all!

The only time it really matters is when a location needs to be computed based on information that was not directly captured by the machine.  For instance, placement locations on a PCB are computed based on their theoretical CAD location on the PCB.  But since the real PCB has unknown manufacturing errors (scaling and non-squareness) as well as the errors of your PnP machine, we correct for both by using fiducials to determine the transformation between the theoretical placement locations and their real locations in machine coordinates.  So, in the end, the machine scaling and non-squareness doesn't even matter here.

The one place where having axes square and scaled correctly is important is for some machine calibration steps.  For instance, camera calibration involves moving the machine in an expected pattern to determine the pixel scaling of the camera.  If the axes scaling is incorrect, that will directly contribute to camera scaling errors.

markmaker:
> For instance, camera calibration involves moving the machine in an expected pattern to determine the pixel scaling of the camera. If the axes scaling is incorrect, that will directly contribute to camera scaling errors.

But if the scaling is _so much off_ that it matters in the small area the camera view covers then the algorithm would probably also have problems guessing the right tick on the ruler, right? 

https://github.com/openpnp/openpnp/blob/c1760cc56c7775ff1b1d2e9224cd51f4ccbada42/src/main/java/org/openpnp/util/Utils2D.java#L723

_Mark

crashmatt:
@tonyluken I agree.  I had not thought about it properly. Just throwing numbers around to figure out what I actually need.

Closing PR.

tonyluken:
@markmaker,

>But if the scaling is so much off that it matters in the small area the camera view covers then the algorithm would probably also have problems guessing the right tick on the ruler, right?

That's the reason for having the rulerMultiple.  It is used to round the measured distances to figure out the true distance.  Just about any practical amount of scaling and non-squareness error can be accommodated by using a large enough rulerMultiple.  As long as measuredDistance - trueDistance is within the range +/-0.25rulerMultiple, it works correctly.  If it is outside that range but within +/-0.75rulerMultiple, the measurement is tossed as being invalid, otherwise it is accepted as a valid (but erroneous) measurement.  So, it is important to use a rulerMultiple that is large enough to prevent that from occurring.  Note that rulerMultiple can also be set using imperial units (something like 2 inches) if an imperial ruler is used for the calibration.

PR #1475, opened 2022-08-28
Performance improvement: Add support for caching of scripting engines

alexanderwiller:
# Description
With this PR, a `HashMap` is introduced into the `Scripting` class that serves as a cache, indexed by the file extensions, storing references to instances of `ScriptEngine` for future use after the first instantiation of any `ScriptEngine`.

In the general machine configuration dialog, a checkbox "Cache scripting engines?" with default `false` is added to allow opt-in into this feature. In the main window `Scripts` menu, a menu item `Clear Scripting Engines Cache` is added that can be used to clear the cache, e.g. for diagnosing erroneous behavior of any script in case global state is persisted in the scripting engine between invocations.

Timers are added for `TRACE` loglevel to allow script performance analysis.

# Justification
Depending on the specific scripting engine, requesting the engine from the factory and initialization of the engine can take quite a lot of time, e.g. for the Jython engine ~250 ms, even on a high-performance machine.

As scripts may be executed e.g. for every part picked, this time adds up and causes the machine to be slower than neccessary.

In my even more affected use case, I want to execute a script for every single move to check if the target position is inside the non-rectangular working envelope of the machine. 250 ms delay for every move when using a python script is a very large performance penalty and also affects the jogging experience significantly. (The script hook for this use case doesn't yet exist / is WIP and will be a follow-up PR)

Concrete numbers for the engine loading time can be found here, with caching completely eliminating those:
```
# Without cache
2022-08-28 15:44:30.681 AbstractHeadMountable DEBUG: Top.moveTo((7.964311, 6.616022, 0.000000, 0.000000 mm), 1.0)
2022-08-28 15:44:30.682 Scripting TRACE: Scripting.on Motion.Move.Before
2022-08-28 15:44:30.682 Scripting TRACE: Scripting.on found Motion.Move.Before.bsh
2022-08-28 15:44:30.683 Scripting TRACE: Caching disabled, bsh.engine.BshScriptEngine@49dc7ef8scripting engine loaded in 0.489 milliseconds
2022-08-28 15:44:30.687 Scripting TRACE: Script Motion.Move.Before.bsh executed in 4.0558 milliseconds
2022-08-28 15:44:30.688 Scripting TRACE: Scripting.on found Motion.Move.Before.js
2022-08-28 15:44:30.690 Scripting TRACE: Caching disabled, jdk.nashorn.api.scripting.NashornScriptEngine@73b5518dscripting engine loaded in 2.2571 milliseconds
2022-08-28 15:44:30.692 Scripting TRACE: Script Motion.Move.Before.js executed in 1.6064 milliseconds
2022-08-28 15:44:30.692 Scripting TRACE: Scripting.on found Motion.Move.Before.py
2022-08-28 15:44:30.945 Scripting TRACE: Caching disabled, org.python.jsr223.PyScriptEngine@1f8ec228scripting engine loaded in 252.9509 milliseconds
2022-08-28 15:44:30.950 Scripting TRACE: Script Motion.Move.Before.py executed in 4.7353 milliseconds

# With cache
2022-08-28 15:44:50.761 AbstractHeadMountable DEBUG: Top.moveTo((13.733733, 10.786311, 0.000000, 0.000000 mm), 1.0)
2022-08-28 15:44:50.761 Scripting TRACE: Scripting.on Motion.Move.Before
2022-08-28 15:44:50.762 Scripting TRACE: Scripting.on found Motion.Move.Before.bsh
2022-08-28 15:44:50.762 Scripting TRACE: bsh.engine.BshScriptEngine@49dc7ef8scripting engine found in cache
2022-08-28 15:44:50.765 Scripting TRACE: Script Motion.Move.Before.bsh executed in 2.4536 milliseconds
2022-08-28 15:44:50.765 Scripting TRACE: Scripting.on found Motion.Move.Before.js
2022-08-28 15:44:50.765 Scripting TRACE: jdk.nashorn.api.scripting.NashornScriptEngine@73b5518dscripting engine found in cache
2022-08-28 15:44:50.766 Scripting TRACE: Script Motion.Move.Before.js executed in 0.6984 milliseconds
2022-08-28 15:44:50.766 Scripting TRACE: Scripting.on found Motion.Move.Before.py
2022-08-28 15:44:50.767 Scripting TRACE: org.python.jsr223.PyScriptEngine@1f8ec228scripting engine found in cache
2022-08-28 15:44:50.770 Scripting TRACE: Script Motion.Move.Before.py executed in 3.2768 milliseconds
```

# Instructions for Use
## Scripting engines caching
By default, every time a script should be executed, a new instance of the appropriate scripting engine is requested. By enabling the scripting engines caching in the general machine configuration, a new instance of any scripting engine is only requested for the first script execution. Following executions re-use the already initialized scripting engines.

This removes the performance penalty that is associated with scripting engine initialization, which, especially for Jython, can be a few hundred milliseconds for every single script execution.

**Scripting engines caching can - depending on the scripting engine implementation - cause global state to be kept in following invocations. Pay attention to global variable declarations inside scripts and keep them idempotent or inside a smaller than global scope.**

In the main window `Scripts` menu, the menu item `Clear Scripting Engines Cache`  can be used to clear the cache. This function is useful for diagnosing erroneous behavior of any script in case global state is persisted in the scripting engine between invocations.

# Implementation Details
### Open questions
Im not sure if the checkbox for enabling / disabling this feature belongs in the machine configuration, but adding a separate scripting section to the config seemed excessive at this point as well. Maybe this would be indicated when the scripting class gets more configurable functionality.

Also, it would be nice to disable the menu item `Clear Scripting Engines Cache` when no engines are cached or when caching is disabled, but I don't know how to signal this state back to the `ScriptFileWatcher` class in an elegant way.  This is just for semantics, so that users don't think 'oh something isn't working maybe I'll reset that ominous cache??".

Both configurations (caching on / caching off) were tested for functionality with all three supported scripting languages (see output above). In case the cache is disabled, a new engine object is shown in `TRACE` output every script call. With caching enabled, `TRACE` output shows initialization of a new engine and subsequent reuse of the same object.
Clear cache is tested by observing the corresponding message in the `TRACE` log and also construction of a new object for the next script call.
Yes, but I left non-conforming lines outside of the PRs scope untouched.
3. If you made changes in the org.openpnp.spi or org.openpnp.model packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.
I just realized that I in fact added a field to `spi/Machine`, does this already qualify as a change to this package?
5. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.
Done.


src/main/java/org/openpnp/scripting/Scripting.java
```diff
@@ -126,4 +161,24 @@ public void on(String event, Map<String, Object> globals) throws Exception {
...
+        }
+        engineCache.clear();
+        Logger.info("Cleared " + clearedEngines.size() + " scripting engines from cache: "
+                + String.join(", ", clearedEngines));
+    }
+
+    // To be used to disable/enable the cache clear menu button
+    // public int getCachedScriptingEnginesCount() {
+    //     return engineCache.size();
+    // }
```
alexanderwiller, line 183: This can be removed or used when the open questions are addressed.

markmaker:
Hi @alexanderwiller,

very nice. This looks peachy to me. 

> I just realized that I in fact added a field to spi/Machine, does this already qualify as a change to this package?

Yes, but this seems legitimate. 

> This can be removed or used when the open questions are addressed.

Sorry, I don't understand what you mean by "open questions", i.e. how it is related to this piece of code?

> I want to execute a script for every single move to check if the target position is inside the non-rectangular working envelope of the machine

Given the coding skills demonstrated here, are you sure you can't implement this as a genuine OpenPnP feature, i.e. in Java, so everybody can benefit, rather than as your "personal" script solution? 😁 (I'm just goading you!)

A list of no-go-boxes (or similar definition) would be very nice. 🤠
 
_Mark

src/main/java/org/openpnp/scripting/Scripting.java
markmaker: Sorry, I don't understand what you mean by "open questions", i.e. how it is related to this piece of code?

alexanderwiller:
Hey Mark, thanks for the feedback.

These are the ominous open questions, maybe they got lost visually in the PR text:

- Im not sure if the checkbox for enabling / disabling this feature belongs in the machine configuration, but adding a separate scripting section to the config seemed excessive at this point as well. Maybe this would be indicated when the scripting class gets more configurable functionality.
=> As you didn't object on that, I assume it's fine?

- Also, it would be nice to disable the menu item `Clear Scripting Engines Cache` when no engines are cached or when caching is disabled, but I don't know how to signal this state back to the `ScriptFileWatcher` class in an elegant way. This is just for semantics, so that users don't think 'oh something isn't working maybe I'll reset that ominous cache??".
=> I implemented this in a new commit, bc18e6e78496c54f548b7e6dd55e4fe3d6f736af.

Regarding the no go box-feature, native support in OpenPnP would be nice for some cases, but I think many of those "special" geometries can't be covered by just boxes. Imo its a decision between user convenience and OpenPnP programming effort to generalize sufficiently flexible exclusion zones. I will elaborate more in the follow-up PR.

If you don't see any need for changes, from my side this PR should be complete.

Thanks,
Alex

src/main/java/org/openpnp/scripting/Scripting.java
alexanderwiller: resolved in https://github.com/openpnp/openpnp/pull/1475#issuecomment-1229542829

markmaker:
Hey @alexanderwiller, 

> Im not sure if the checkbox for enabling / disabling this feature belongs in the machine configuration

One option would have been just making it configurable in the **Scripts** menu, same way as e.g. table linking. 

menu option

But I think it is fine on the machine, as the machine also handles the somewhat related _machine task queue_. 

It is a good thing we are talking about this, because now that I think about it, I see there is a danger that we get threading and/or reentrance issues with the engine cache. Some scripts are called from the GUI thread, some from the machine thread. Some scripts call machine actions that in turn could trigger scripts, while the outer script engine is still running. As scripts often use motion, that would certainly be the case if you introduced scripts for motion. In theory, this could also happen with even more levels of nesting.

One option would be to remove the engine from the cache while in use. So if a script were to call machine action with scripts triggered, it would find the cache empty and create a new engine. If the script calls many such actions, the engine would at least be cached again during that whole script. Once the script ends, the outer engine would be reinstated, which unreferences the inner engine, to be garbage collected. A more elaborate/more performant solution could cache a _pool_ of engines and their busy states, so even nested scripting would always be fast. 

All that state handling would have to be thread safe _and_ exception safe. The state changes before/after the script must be properly `synchronized`, but it must _not_ be blocking during script evaluation. A `finally { }` block would have to do the cleanup (you could also move your time measurement there, so even scripts that throw exceptions would be properly measured). 

While you're at it, please merge the latest `test` into your branch, for the diff to show up properly.

_Mark

alexanderwiller:
Regarding the config, I would then leave it as it is. I feel like for "true" settings that are not just view-related, there are better places than the main window menu bar.

And the caching, well, that would have been too easy 😄 Totally forgot about thread safety and concurrent invocations.

I'll take some time to implement a pooled threadsafe version as you suggested and will also do the time measurement improvements.

markmaker:
Hi @alexanderwiller 

This looks really nice. 

As you have already added the cool unit test, could you add a more "brutal" threading test, with multiple threads running in parallel calling scripts many times in parallel and perhaps some way of checking if they work as expected, like checking for lost updates or something?

> For existing users, I guess them reading the release notes or stumbling across the feature in the Wiki would be the only practical way?

No, you can add an Issues & Solutions suggestion. Analog to the one here:

https://github.com/markmaker/openpnp/blob/abf6dad08d822619927aebdc20b41a051999230e/src/main/java/org/openpnp/machine/reference/solutions/VisionSolutions.java#L215-L234

This solution can link to the Wiki (like in the example above). You can add your nice explanation text there. 

_Mark

markmaker:
Just saw the test fails:

https://github.com/openpnp/openpnp/runs/8097302329?check_suite_focus=true

_Mark

alexanderwiller:
> Just saw the test fails:
> 
> https://github.com/openpnp/openpnp/runs/8097302329?check_suite_focus=true
> 
> _Mark

> ScriptingTest.testScripting:67  Script execution for js file extension didn't ...

The output of the test is not too informative, but I have an idea why the test may fail. In the pom.xml, there is no entry for the Nashorn scripting engine, which maybe isn't included by default in the test environment, but is included in the JDK I use on my machine.

If thats the issue, I would suggest to add Nashorn to the pom.xml and maybe remove the extra code from [ee4de22](https://github.com/openpnp/openpnp/pull/1475/commits/ee4de229d75f34e9f7f35d4983e76c52b6c4e968) again, as all officially supported scripting engines ideally should be tested.

---

I will add a better threading test as well, my first shot wasn't exactly marked by motivation implementing some advanced solution 😄

Same for Issues & Solutions, thats a very nice feature, I was already wondering how it was implemented to support some pretty complex tasks with reasonable effort.

alexanderwiller:
Thanks for starting the run, it looks like the Nashorn engine is in fact missing for the Java 15 environment: https://github.com/openpnp/openpnp/runs/8121123075?check_suite_focus=true#step:4:1239

And a quick search revealed that Nashorn isn't bundled with Java 15 and higher anymore: https://stackoverflow.com/questions/65265629/how-to-use-nashorn-in-java-15-and-later

The maintainer has published it separately after the deprecation: https://search.maven.org/artifact/org.openjdk.nashorn/nashorn-core/15.4/jar

I hope that adding that dependency to the pom.xml doesn't cause any conflicts with older Java versions.

Last, the test failed because I hardcoded the number of engines, gotta fix that as well: https://github.com/openpnp/openpnp/blob/ee4de229d75f34e9f7f35d4983e76c52b6c4e968/src/test/java/org/openpnp/scripting/ScriptingTest.java#L120

alexanderwiller:
Turns out that Nashorn 15.1 and higher in fact cause a build error:

```
[INFO] --- maven-surefire-plugin:2.22.2:test (default-test) @ openpnp-gui ---
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.openpnp.scripting.ScriptingTest
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.111 s <<< FAILURE! - in org.openpnp.scripting.ScriptingTest
[ERROR] testScripting  Time elapsed: 0.104 s  <<< ERROR!
java.lang.UnsupportedClassVersionError: org/openjdk/nashorn/api/scripting/NashornScriptEngineFactory has been compiled by a more recent version of the Java Runtime (class file version 59.0), this version of the Java Runtime only recognizes class file versions up to 52.0
        at org.openpnp.scripting.ScriptingTest.testScripting(ScriptingTest.java:38)
```

Nashorn 15.0 luckily works fine with Java 8, so if its also compatible with Java 15 that would allow just adding it to the pom.xml, without the need for any Java version-specific mechanism.

alexanderwiller:
Now everything that was open except the Issues & Solutions wizard should be implemented.

Regarding the added Nashorn 15.0 dependency, I would suggest to see if the tests pass now and then remove most of ee4de229d75f34e9f7f35d4983e76c52b6c4e968, as all three engines should be available and be tested.

Working on the wizard now.

alexanderwiller:
Sorry for those "works on my machine"-type test failures, should be fixed now.

markmaker:
Hi @vonnieda, 

I hope you could quickly look at [this one issue](https://github.com/openpnp/openpnp/pull/1475#issuecomment-1233392478):

> Turns out that Nashorn 15.1 and higher in fact cause a build error:... 
> Nashorn 15.0 luckily works fine with Java 8, so if its also compatible with Java 15 that would allow just adding it to the pom.xml, without the need for any Java version-specific mechanism.

Is the `pom.xml` dependency on Nashorn 15.0 OK? 

Is that what you meant here?

> I'm not worried about Nashorn going away. The javax scripting system allows for drop in scripting languages - we do this for BeanShell and Python already - so if it does get removed from Java we'll just drop the latest version back in to OpenPnP.

https://groups.google.com/g/openpnp/c/_25dKmJEqwE/m/lHIMRCwrBwAJ

According to this, it probably only works from Java 11: 

https://stackoverflow.com/questions/66061017/different-nashorn-engine-for-java-15-and-15/66190018#66190018

Could we drop Java 8, 9, 10 support from OpenPnP and make the default installer packed engine one of the newer ones? If only I knew how 🤔

_Mark

vonnieda:
> Hi @vonnieda,
> 
> I hope you could quickly look at [this one issue](https://github.com/openpnp/openpnp/pull/1475#issuecomment-1233392478):
> 
> > Turns out that Nashorn 15.1 and higher in fact cause a build error:...
> > Nashorn 15.0 luckily works fine with Java 8, so if its also compatible with Java 15 that would allow just adding it to the pom.xml, without the need for any Java version-specific mechanism.
> 
> Is the `pom.xml` dependency on Nashorn 15.0 OK?

Yep, I think it's a good choice.

> 
> Is that what you meant here?
> 

Not exactly - because I didn't realize someone would keep packaging it. I assumed we'd just have to find the oldest version and fork it and maintain it. But, it seems someone else is doing that so we can just include it.

> > I'm not worried about Nashorn going away. The javax scripting system allows for drop in scripting languages - we do this for BeanShell and Python already - so if it does get removed from Java we'll just drop the latest version back in to OpenPnP.
> 
> https://groups.google.com/g/openpnp/c/_25dKmJEqwE/m/lHIMRCwrBwAJ
> 
> According to this, newer versions probably only work from Java 11:
> 
> https://stackoverflow.com/questions/66061017/different-nashorn-engine-for-java-15-and-15/66190018#66190018
> 
> Could we drop Java 8, 9, 10 support from OpenPnP and make the default installer packed engine one of the newer ones? If only I knew how 🤔

Yes, I think we are far overdue to drop 8 - 10. 

Unfortunately there is no magic bullet for doing this. The short version is you (we) need download and install Install4J, learn how to use it (again), and figure out how to bundle the new binaries, or come up with something different.

I've been trying to catch up on maintenance tasks recently, so I'll see if I can fit this in soon, and if not I'll document where I get stuck in case someone wants to take a stab at it.

I'll put my notes here: https://github.com/openpnp/openpnp/issues/1478

Jason



> 
> _Mark

markmaker:
@alexanderwiller,

after @vonnieda's OK for Nashorn, I think we can merge this. Once the older Java versions are dropped, we can modify the `pom.xml` to use a newer version with minimal effort.

Do you agree?

_Mark

alexanderwiller:
Hey Mark and Jason,

thanks for the insights regarding the Java version situation.

I have two more things to do, first I wanna remove the code from https://github.com/openpnp/openpnp/pull/1475/commits/ee4de229d75f34e9f7f35d4983e76c52b6c4e968 again as all engines should be tested and a missing engine should cause the test to fail.

Also, the wizard for suggesting to turn on the cache is still missing.

Will do these two things soon, then I think its ready for merge.

alexanderwiller:
I just installed Java 15 locally and found out that Nashorn 15.0 doesn't work with Java 15 and Nashorn 15.1 and up don't work with Java 8.

As JS scripting support is already broken for Java 15, this allows the tests to pass till Java 8/9/10 support is dropped and Nashorn can be bumped to 15.4.

I'd say this PR is complete then, given that the tests pass now.

I'm still unappy with storing the script engine pooling in the machine class, but will open another issue for that.

markmaker:
Hi @alexanderwiller,

Seeing your attention to detail (which I appreciate!), I guess you want to fix these:

There is a typo here:

```java
solutions.add(new Solutions.Issue(machine,
                        "Script exeuction performance can be improved by enabling engine pooling.",

```

and this should be "Caught":

```java
System.out.println("Catched test exception from script");
```

Made that same mistake for years before I started noticing, _Fremdsprachen!_ 😅 

_Mark

alexanderwiller:
Thanks, hope now its done 😃

markmaker:
Thanks, @alexanderwiller, cool addition!

matmair:
@alexanderwiller thank you for your work on this, it helped me out a lot. Very valuable addition to user of the scripting environment!

PR #1476, opened 2022-09-01
Limit low value of speed slider to 1 %

alexanderwiller:
Moving a machine with 0 % speed doesn't seem to provide a lot of value.
Ticks are kept at 1, 25, 50, 75, 100 %.

# Description
Speed slider range is now limited to 1 % at the bottom, but labels are kept tidy as they are set manually.

# Justification
I don't think that normal usage of the UI should trigger an exception if there is nothing exceptional to report.

# Instructions for Use
No instructions required.
=> Loading OpenPnP demo machine, checking correct debug log speed info from 0.01 - 1 while jogging.
=> Yes.
3. If you made changes in the org.openpnp.spi or org.openpnp.model packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.
=> No changes
4. Be sure to run mvn test before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. 
=> Done, passed

### DIFF
--- a/src/main/java/org/openpnp/gui/JogControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/JogControlsPanel.java
@@ -364,15 +364,7 @@ public class JogControlsPanel extends JPanel {
         speedSlider.setMajorTickSpacing(25);
         speedSlider.setSnapToTicks(true);
         speedSlider.setPaintLabels(true);
-        speedSlider.setMinimum(1);
         speedSlider.setOrientation(SwingConstants.VERTICAL);
-        Hashtable<Integer, JLabel> speedSliderLabels = new Hashtable<Integer, JLabel>();
-        speedSliderLabels.put(1, new JLabel("1"));
-        speedSliderLabels.put(25, new JLabel("25"));
-        speedSliderLabels.put(50, new JLabel("50"));
-        speedSliderLabels.put(75, new JLabel("75"));
-        speedSliderLabels.put(100, new JLabel("100"));
-        speedSlider.setLabelTable(speedSliderLabels);
         panelControls.add(speedSlider, "20, 4, 1, 9"); //$NON-NLS-1$
         speedSlider.addChangeListener(new ChangeListener() {
             @Override
--- a/src/main/java/org/openpnp/gui/JogControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/JogControlsPanel.java
@@ -364,7 +364,15 @@ public class JogControlsPanel extends JPanel {
         speedSlider.setMajorTickSpacing(25);
         speedSlider.setSnapToTicks(true);
         speedSlider.setPaintLabels(true);
+        speedSlider.setMinimum(1);
         speedSlider.setOrientation(SwingConstants.VERTICAL);
+        Hashtable<Integer, JLabel> speedSliderLabels = new Hashtable<Integer, JLabel>();
+        speedSliderLabels.put(1, new JLabel("1"));
+        speedSliderLabels.put(25, new JLabel("25"));
+        speedSliderLabels.put(50, new JLabel("50"));
+        speedSliderLabels.put(75, new JLabel("75"));
+        speedSliderLabels.put(100, new JLabel("100"));
+        speedSlider.setLabelTable(speedSliderLabels);
         panelControls.add(speedSlider, "20, 4, 1, 9"); //$NON-NLS-1$
         speedSlider.addChangeListener(new ChangeListener() {
             @Override
--- a/src/main/java/org/openpnp/gui/JogControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/JogControlsPanel.java
@@ -369,9 +369,14 @@ public class JogControlsPanel extends JPanel {
         speedSlider.addChangeListener(new ChangeListener() {
             @Override
             public void stateChanged(ChangeEvent e) {
-                Configuration.get()
-                .getMachine()
-                .setSpeed(speedSlider.getValue() * 0.01);
+                try {
+                    Configuration.get()
+                    .getMachine()
+                    .setSpeed(speedSlider.getValue() * 0.01);
+                }
+                catch (Exception ex) {
+                    // TODO: I don't like this
+                };
             }
         });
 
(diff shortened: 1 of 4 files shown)


markmaker:
Hi @alexanderwiller. 

I'm not convinced, this will actually result in a better user experience. 

Currently, when they just mindlessly draw the slider all down, they get an exception which makes them think about what minimum speed they actually want. They then have to manually set it, which more likely results in them setting it to something like 5% (you can't actually set 1% with the mouse, the slider is too coarse).

With your change, when they just mindlessly draw the slider all down, they get 1% which means a move will take a 100 times longer than nominal. New machines (still in the process of being built) are often configured with very low feed-rate/acceleration limits, e.g. the OpenPnP GcodeDriver default is 1000mm/min. At 1% it would take a minute to travel 10mm. This can easily look as if OpenPnP and even the controller is hanged (motion LEDs on, and seemingly nothing happens). Albeit for different reasons, I've had that happen to me and it was very frustrating until diagnosed! I also got multiple such reports from users in the past (again for different reasons). 

Maybe you should limit it to 5% (20 times longer time). You could add a new "Minimum Speed"` @Attribute` in the `AbstractMachine` that is only accessible in the `machine.xml`, so if somebody is unhappy with 5% they can change it.

_Mark

alexanderwiller:
Okay, I get the issue of "1 % looks like nothing happens" and agree that it could cause a bad UX.

After thinking about why that wasn't an issue for me but the 0 % exception was, I realized the dimensions involved:
> the OpenPnP GcodeDriver default is 1000mm/min. At 1% it would take a minute to travel 10mm.

As my machine can move at least with ~1000 mm/s, even 1 % is 10 mm/s and not very slow, a fixed lower limit of 5 % would therefor be quite high. So for me a common UI action is to just drag the slider to the minimum for some set-up tasks, triggering the exception.

Is there any reason against having a default of 5 %, but making it available as a regular setting in the Machine?

I fear that nice tick spacing could get difficult, maybe we can provide a table for 5 % and 1 %, but otherwise just let the JSlider setup evenly spaced labels?

markmaker:
> Is there any reason against having a default of 5 %, but making it available as a regular setting in the Machine?

No, not at all.

> I fear that nice tick spacing could get difficult, maybe we can provide a table for 5 % and 1 %, but otherwise just let the JSlider setup evenly spaced labels?

Don't know what you mean. Frankly, I don't know much about these slider controls.

_Mark

alexanderwiller:
Hello @markmaker 

I added a basic implementation of this feature, but have a few questions:

- The speed slider limits aren't updated when the model is updated, even though they are updated when the config is loaded: https://github.com/openpnp/openpnp/pull/1476/files#diff-0c459df782aa4cd05b008197076e878d8abf7f719f3f8f307b3c0206226accbdR776-R777
How can I update them on a model property change? I already tried the following without luck:
```
BeanUtils.addPropertyChangeListener(machine, "speedLimitLow", (e) -> {
    if (e.getNewValue() != null) {
        speedSlider.setMinimum((int) ((double)e.getNewValue() * 100));
    }
});
```
- Setting a speed outside of the allowable limits should cause an exception, because it's not allowed. Still, just swallowing that exception silently feels wrong: https://github.com/openpnp/openpnp/pull/1476/files#diff-0c459df782aa4cd05b008197076e878d8abf7f719f3f8f307b3c0206226accbdR372
Do you think that this is justified or do you see another way? This is basically an error that shouldn't ever be triggered from this class.
- How is it possible to validate the data entered? In general, I looked around and it seems like I can enter any invalid number anywhere, i.e. that no validation is in place anywhere.

  I'm pretty sure that the permissible value range belongs in the model. One idea was to just throw an exception: https://github.com/openpnp/openpnp/pull/1476/files#diff-06a414473442f35df9becae03fbe4315e315977656ed58213fcf9d984e85d0a5R733
  This attempt wasn't too succesful tho, as `addWrappedBinding` seems to just silently swallow any exception, they appear in the log, but are not shown to the user.

  Also, a better option would probably be to use annotations for that purpose. What do you think of finishing this PR first without validation, then exploring the use of https://www.baeldung.com/javax-validation#validation?

Alex

markmaker:
Hi Alexander

The setters must use `firePropertyChange()`. Just a random sample:

https://github.com/openpnp/openpnp/blob/af43024e03f6d9e98c738cff8fb1424be52c487b/src/main/java/org/openpnp/model/Package.java#L92

> setting a speed outside of the set limits should cause an exception

Why would this ever happen?

> How is it possible to validate the data entered? 

Throw in the setter.

> In general, I looked around and it seems like I can enter any invalid number anywhere, i.e. that no validation is in place anywhere.

Of course, one could add a gazillion validations in OpenPnP. But given the fact that the same users usually have to edit controller configuration files in obscure syntax, where they can easily mess up things very badly, I think a certain responsibility can still be taken as granted. Sometimes, blocking validations are also a PITA! For instance, if they insist on setting values on previously uninitialized properties, that one does not yet know... and then instead of leaving it uninitialized for a later revisit, users set it to a random value that the validation is happy with... which causes hard to diagnose problems later. Or when one value is validated to be larger than the other, and you want to change them in the wrong order. I absolutely hate these. 

For critical values, that users tend to misunderstand and therefore misconfigure, I added Issues & Solutions warnings/errors. These also catch existing/old config errors as soon as a user updates (and on my side it flags errors if users send their `machine.xml`, regardless of their OpenPnP version). 

_Mark

alexanderwiller:
Okay, thanks, things are a lot more clear now.

> Why would this ever happen?

It probably won't 😅 Maybe in tests, but thats a far stretch. Just feels kinda wrong to not enforce these limits at all if they are part of the `Machine` class.

I see all the issues you mentioned regarding validation as well and also thought about the fact that a user must somewhat know what they are doing no matter what. Way simpler to just let things fail downstream in case something wrong is entered.

markmaker:
In case you overlooked it, you need to merge the new test branch with your #1475 into this one and resolve the **conflict**.
_Mark

markmaker:
The setters have some "Percent" variable names, when in fact they are factors `0.0 ... 1.0`.

markmaker:
@alexanderwiller,

Are you still on this?

Needs resolving conflicts.

_Mark

alexanderwiller:
@markmaker yes, following up on all my open issues/PRs is on my to-do list, will get back to this soon.

markmaker:
@alexanderwiller ping.

alexanderwiller:
@markmaker please don't act on the latest push, I will clean up the code first.

markmaker:
I think this is now addressed in #1515 where the Minimum Speed now also serves some functional purposes.

_Mark

alexanderwiller:
Just today/now I was pondering about if and where to enforce the upper / lower speed limits to not run into some issue with double conversion imprecision etc., to finally finish this PR, when I got the notifications 😁 

#1515 is obviously far more comprehensive and I feel like moving the speed limits to the motion planner is cleaner than any solution I would have come up with 👍

PR #1484, opened 2022-09-07
Feature / Better Blow-Off Control

markmaker:
# Description
Some machines can perform a blow-off air pulse when parts are placed. This PR improves the handling of this:

- So far one had to define a blow-off level on each package, which is too much work. This PR adds a default blow-off level on the nozzle tip for a more central configuration:
  Default blow-off level
- For some machines it is assumed that the blow-off command happens to implicitly close the vacuum valve too, i.e. the valve has a sort of _three-state operation_. For this behavior one can enable the **Closes Vacuum Actuator?** option on the nozzle (default is _enabled_, to get the same behavior as before):
   Blow-off closes valve
- Otherwise, the place operation now closes the vacuum valve, even if a blow-off value is set. This also allows for actuators on different drivers/controllers (previously impossible).  

- The Issues & Solutions nozzle offsets calibration a.k.a. "Confetti Test", now looks for a part with ID `TEST-OBJECT` to use for the calibration. Users can now create such a part in order to set a custom part height and package specific blow-off value, if needed. 
- This is [documented in the Wiki](https://github.com/openpnp/openpnp/wiki/Calibration-Solutions#calibrating-precision-camera-to-nozzle-offsets), linked from the Issues & Solutions, item (which explicitly tells the user to read the Wiki). 

# Justification
See discussion:
https://groups.google.com/g/openpnp/c/P2-vgP8-6qo/m/yagtfRQ7CQAJ

# Instructions for Use
See the Description.

# Implementation Details
1. No explicit testing was done (no such a blow-off valve on my machine, no simulation implemented). It is assumed this is done quickly by the reporting user in the testing version.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
@@ -983,6 +983,14 @@ public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMoun
         blowOffActuator = actuator;
     }
 
+    public boolean isBlowOffClosingValve() {
+        return blowOffClosingValve;
+    }
+
+    public void setBlowOffClosingValve(boolean blowOffClosingValve) {
+        this.blowOffClosingValve = blowOffClosingValve;
+    }
+
     protected void actuateVacuumValve(boolean on) throws Exception {
         if (on) {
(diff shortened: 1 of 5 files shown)


vonnieda:
Nice work @markmaker - this feature has needed some love for some time!

PR #1485, opened 2022-09-12
Fix/Issues and Solutions HOME_COMMAND

markmaker:
# Description
Issues & Solutions now proposes a `HOME_COMMAND`, even if one is already present. The proposed command is built and compared to the present one. If the present one does not _contain_ the proposed lines, the solution is presented. Users can still add custom lines before and after the command, without I&S wanting to undo them. 

As always, users can also Dismiss the proposed solution, if they know their present command is better.

As an aside, this PR also fixes a typo/text sequence in the `CameraSolutions`.

# Justification
Support cases related to `HOME_COMMAND`, the latest of which is the following:
https://groups.google.com/g/openpnp/c/2DKOxhBc6pg/m/2HgSqTmCCQAJ

# Instructions for Use
No change,

# Implementation Details
1. Tested with Issues & Solutions.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/solutions/CameraSolutions.java
+++ b/src/main/java/org/openpnp/machine/reference/solutions/CameraSolutions.java
@@ -333,14 +333,14 @@ public class CameraSolutions implements Solutions.Subject  {
                                             + "by Issues & Solutions. Press <strong>Find Issues & Solutions</strong> to check for the latest status. "
                                             + "Proceed only when you are sure the properties are OK, white balance must be computed as the last step.</p><br/>")
                                     + "<p>Color-keyed computer vision needs stable, and accurate colors. Neither the unstable Auto white balance, "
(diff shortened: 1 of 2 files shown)


lags:
Hi @markmaker,
after update, I'll report what happens to me (with tinyg).
I&S suggest as you can see in gif the right HOME_COMMAND but when confirmed and refresh, it still appears but under driver section, HOME_COMMAND is present and correct. So seems as if I&S not mark as applied.

Still appears also if close openpnp and reopen

markmaker:
@lags, thanks for the report!

Can you send the machine.xml?

_Mark

lags:
Of course...
[machine.xml.zip](https://github.com/openpnp/openpnp/files/9550586/machine.xml.zip)

markmaker:
Should be fixed in the testing version, see also #1486. 

@lags please test, thanks!

_Mark

lags:
Good job @markmaker.
It works!

Issue #1325, opened 2021-11-10
Panel options window not showing OK and Cancel buttons in Linux


neliooliveira:
# Problem Report
## Describe the Issue

Panel options window not showing OK and Cancel buttons in Linux.
The buttons are there because I can still select them using TAB.
It's probably just the size of the window that is not high enough to fit all the elements in the Linux theme.

## Steps to Reproduce

Open the panel options in a Linux machine.

## Expected Result
Buttons to be visible

## Notes, Log Files, Screen Captures, Videos, etc. to Show the Issue
![Captura de ecrã de 2021-11-10 11-01-35](https://user-images.githubusercontent.com/16232608/141104420-ac81cded-f789-460e-a377-8482ef212ae9.png)

kaklik:
The same happens to me: 

![image](https://user-images.githubusercontent.com/1194177/193769605-a06c6da5-7715-48dd-a27c-18277b127d55.png)

I should use 3x Tabulator and enter to hit the "hidden" OK button. 

I am running that version of OpenPnP
![image](https://user-images.githubusercontent.com/1194177/193770250-9f5528b5-537d-401a-b346-9ee5abccabf4.png)

on Ubuntu:

Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.1 LTS
Release:	22.04
Codename:	jammy

kaklik:
The recent version of OpenPnP has changed the Panelization Settings layout which resolved this issue. 

![image](https://user-images.githubusercontent.com/1194177/222408985-632a5dff-7bfe-4908-bdde-703b5ae5c635.png)

Therefore the issue should be closed. 

FIY @neliooliveira

markmaker:
Thanks, @kaklik!

PR #1492, opened 2022-10-29
Fix/Panelization Settings window height

lags:
# Description
Fixes Panelization Settings window height reported in #1325

### DIFF
--- a/src/main/java/org/openpnp/gui/panelization/DlgAutoPanelize.java
+++ b/src/main/java/org/openpnp/gui/panelization/DlgAutoPanelize.java
@@ -140,6 +140,7 @@ public class DlgAutoPanelize extends JDialog {
 
         setSize(400, 600);
         setResizable(false);
+        pack();
         setLocationRelativeTo(parent);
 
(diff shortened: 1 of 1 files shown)


markmaker:
I'm not a Java GUI expert, but can this not be made dynamic, to shrink-wrap the controls?

_Mark

lags:
Me too not expert but `pack();` should be better than raise height value.

image

As you can see, pack not shrink control for parent frame but resize window for the best view referred setSize function.

image

Have sense?

markmaker:
I'm not sure I understand. Do you think this is better or not?

For me `pack()` sounds exactly right. Can you  add it and push again?

_Mark

lags:
Yes Mark. I think that with `pack()` function is better than raise height size. I make new PR.

lags:
I think not need make new PR because when commit with `--amend` this PR will be updated

markmaker:
> I think not need make new PR because when commit with `--amend` this PR will be updated

No need. You could just change the code, commit and push again.

I'll merge this now, thanks, @lags 

_Mark

Issue #1363, opened 2022-01-21
Java17: Simple XML serializer ignores converter annotations under Java 17


ian-arkver:
I commented out the test serialization before the machine.xml file write so we get the partially written file before the exception is thrown.

Comparing [the Java 17 version](https://github.com/openpnp/openpnp/files/7909150/machine.xml.txt) with [the Java 8 version](https://github.com/openpnp/openpnp/files/7909196/machine.xml.txt) we can see that the next element to be written would have been:

```
                  <cv-stage class="org.openpnp.vision.pipeline.stages.DrawContours" name="7" enabled="true" contours-stage-name="9" thickness="2" index="-1">
                     <color r="255" g="255" b="255" a="255"/> 
                  </cv-stage>
```
which is an element with the custom serializer. So I guess this isn't working with 17, at least when run by the testbench.

_Originally posted by @ian-arkver in https://github.com/openpnp/openpnp/issues/1361#issuecomment-1018029545_

ian-arkver:
Running BasicJobTest under VSCode debugger with a breakpoint in the ColorConverter code. The breakpoint is hit under JDK 1.8, but not hit under JDK 17 resulting in an illegal reflective access to the Color object.

![image](https://user-images.githubusercontent.com/17830580/150503346-f4b52160-7bd4-47b0-bb14-3a815ac770f3.png)

The version of simpleframework.xml used is 2.7.1 which is the latest but dates from 2013. I guess this package is abandonware and perhaps a different xml serialiser should be used.

markmaker:
I understand you found a workaround, and this can be closed, right?

ian-arkver:
Well, I didn't, @lags did, but yeah I think this can be closed.

Issue #1361, opened 2022-01-19
RFC: Fixes for recent JDK support (eg JDK 17)


ian-arkver:
# Feature Request / Request for Comment
## Update dependencies and other fixes to allow OpenPnP to run with a recent Java VM.

I didn't find an online list, or any open Issues describing the current blockers for running with a more recent JDK,
so thought I'd create this as a place to collect the information and discussion.

Currently OpenPnP requires an old, insecure JDK version to work properly due to illegal reflective access failures in various places, for example when trying to write java.awt.Color to machine.xml. There are probably other places, possibly within package dependencies. These may be harder to work around than the java.awt.Color one.

Since [JEP403](https://openjdk.java.net/jeps/403) was implemented in Java 17, there's now no command line option to re-enable illegal reflective accesses.

## How Will It Look?
There should hopefully be no changes to the user experience, unless some UI element dependencies are badly broken by Java 17.

## Why Do We Need It?
1. What problem does it solve?

Fixing this would avoid a common problem where new users try to run OpenPnP on a too recent Java version and report that it doesn't work.

Presumably the driver behind JEP403 and its predecessors was to improve security of the Java VM.

2. Is it useful for everyone, or does it only solve a problem for your machine?

This could be useful to many people, especially new users installing it on a new machine.

## Provide Examples
N/A

ian-arkver:
Thanks Mark, understood. Me neither. However if we have a place where we can collate what's broken then those with some coding time and the necessary knowhow will have a place to start.

I don't know whether it's just a matter of a few missing accessors or an explicit serialiser/deserialiser for some data type or whether something more fundamental will break like maybe "beans bindings won't work and the whole subsystem needs replaced" [Not saying beans bindings are broken, just an example].

For java.awt.Color there's [some information here](https://stackoverflow.com/a/60683175) and some [maybe relevant code here](https://github.com/shamanDevel/SimpleVideoEditor/blob/master/src/org/shaman/sve/SerializationUtils.java), though we'd need to make sure backward compatability with existing machine.xml files was preserved, i.e. the color was serialised to/from the existing format.

ian-arkver:
Thanks Jason, I'll see what I can do.

ian-arkver:
Interestingly there's already a custom serialiser in the code for awt.Color, eg. in CvStage/DrawCircles:

```
    @Element(required = false)
    @Convert(ColorConverter.class)
    private Color color = null;
```

[...]

PR #1493, opened 2022-11-07
JDK17 support

lags:
# Description
Allow openpnp to compile and run in Java 17

# Justification
`help-wanted` as described in #1361 and #1363 

# Implementation Details
1. Tested by trying to edit a java file and trying to recompile.
2. Did follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)
4. Successful  `mvn test` before submitting the Pull Request.


ian-arkver:
Why does ReferenceCamera load OpenCV? Not really related to this PR but seems wrong.

I also discovered that changing the target in the pom breaks the GH CI actions. I suspect these need fixed by Jason.

markmaker:
@vonnieda, just so you know, this is above my "pay grade". I hope you can have a look.

lags:
Hi Ian,
> I also discovered that changing the target in the pom breaks the GH CI actions. I suspect these need fixed by Jason.

The build errors report that seems advice an unavailability of `onvif-java-lib`.



I am not expert for this but maybe if rerun and this artifact is available should be ok

ian-arkver:
OK, different error then. I tried this in PR #1398 but the error there was to do with "invalid version" so I gave up and closed the PR.

lags:
I also had this error message too and had to change a few things in my environment:

- JAVA_HOME global variable to point to `C:\Program Files\Java\jdk-17.0.5`
- change Java version inside pom from `1.8` to `17`
- change Java compiler inside IDE (I use IntelliJ IDEA) to use `jdk-17.0.5`

For this commit I was inspired by your great work.
I just had to add ` --add-opens=java.desktop/java.awt.color=ALL-UNNAMED` clausole because as you reported `java.awt.color` not work properly when used

ian-arkver:
Yeah, it can be fixed up locally, but the CI actions deliberately call up specific versions of JDK. Targetting "17" with Java 8 doesn't work (afaik)

btw, the add-opens hint came from another author:
https://groups.google.com/g/openpnp/c/2cUIHXxVpXQ

lags:
Yes you are right. Maybe in CI action should be added new Java 17 but shouln't cancelled if other action failed

ian-arkver:
I'm not sure if the installers create desktop shortcuts that then call openpnp.bat/sh or whether they call java directly and need the extra options added there too.

I'm also not sure that changing the maven compiler plugin target version in the pom is the right thing to do. I think that's what breaks the CI pipelines since they pull in older JDKs.

vonnieda:
Thanks @ian-arkver - that's helpful. I think removing JDK 8 build support entirely should be part of this. 17 is currently the LTS versions, so that seems like a fine version to target to me. I'll update the installers; they do not call the sh or bat, so they need a manual update. I'll also get the embedded JDK updated to 17.

pom.xml
```diff
@@ -50,9 +50,9 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
-			<groupId>org.simpleframework</groupId>
+			<groupId>com.subshell.simpleframework</groupId>
```
vonnieda, line 53: @lags What does this fork of simple-xml do? Fix https://github.com/openpnp/openpnp/issues/1363 perhaps? I looked through https://github.com/subshell/simplexml but didn't see anything.

vonnieda (commented):
I'm working on merging this, along with some other minor changes that are all related. Those are: fixing opencv and openpnp-capture on macos-arm64, updating Install4J, and fixing Nashorn (JS engine). I'll merge all that together soon to close this.

pom.xml
lags: Hi @vonnieda yes, this fork solve the errors raise when try to use Java 17. At the time I too hadn't found anything related [here](https://github.com/ngallagher/simplexml/compare/master...subshell:simplexml:master) but it works.
Probably the version upgrade has dependencies that fix the compatibility?

vonnieda:
Hi all,

I fixed up the Github Actions, Installers, and a bunch of MacOS issues as part of this. Those changes, and the ones in this PR are now in https://github.com/openpnp/openpnp/pull/1534. If everyone is good with the changes we can close this one and merge that one into test. 

Thank you @lags, @ian-arkver and @markmaker for all your work on this!

Thanks,
Jason

Issue #2, opened 2013-01-27
Google Code skipped issue 18


vonnieda:
_Skipping this issue number to maintain synchronization with Google Code issue IDs._

PR #1494, opened 2022-11-07
ReferenceLoosePartFeeder - Human Vision Fallback

surabibio:
# Description
In ReferenceLoosePartFeeder when computer vision fails, the user get stuck and has to tweak CV settings which is often hindering especially in the middle of a job. An option to manually jog to the parts pick location could be time saving.

# Instructions for Use
When CV fails, the "Instructions Panel" shows up. The current job is paused. The user has to manually jog to the parts pick location and click "Accept" The job resumes.

# Implementation Details
1. How did you test the change? Tested in simulation.

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -103,7 +103,7 @@ import com.google.common.eventbus.Subscribe;
 
 @SuppressWarnings("serial")
 public class JobPanel extends JPanel {
-    enum State {
+	public enum State {
         Stopped,
         Paused,
         Running,
@@ -747,6 +747,10 @@ public class JobPanel extends JPanel {
         return true;
     }
 
+    public State getJobState() {
+    	return state;
+    }
+
     public void importBoard(Class<? extends BoardImporter> boardImporterClass) {
         if (!checkJobStopped()) {
             return;
@@ -1042,13 +1046,13 @@ public class JobPanel extends JPanel {
         }
     };
 
-    public void PauseJob() {
+    public void pauseJob() {
     	if (state == State.Running) {
             setState(State.Pausing);
     	}
     }
 
-    public void ResumeJob() {
+    public void resumeJob() {
     	 if (state == State.Paused) {
              setState(State.Running);
              jobRun();
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -1042,6 +1042,19 @@ public class JobPanel extends JPanel {
         }
     };
 
+    public void PauseJob() {
+    	if (state == State.Running) {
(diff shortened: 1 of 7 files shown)


markmaker:
Hi @surabibio 

you need to properly handle the OpenPnP threading architecture. GUI stuff and machine task stuff must not be mixed.

1. You must leave the machine task (which may or may not be a job) by (re-) throwing the exception, there is no other proper way. This will automatically pause the job. In fact it is more complicated, we would need some sort of unconditional exception, so that the retry loops are broken.
2. To resume you must first check if there is a paused job, it might have been a simple GUI button feed. 

I was thinking about user intervention inside a task before (auto-template image learning in bottom vision) and it is not easy at all. 

_Mark

surabibio:
Hi,

Nowhere am I interfering with the threading like wait() or Sleep() or notify().

 

This approach is like the user clicking the pause button right after feed(). Will it not pause the job? So does this approach.

Only thing is, it eats the exception and the machine thinks feed() has succeeded for good, which is OK for us.

 

Regarding #2 I think that is already taken care of. i.e pressing GUI button feed will not inadvertently start the job!!

 

Any other critical review of this approach is welcome.

markmaker:
> Nowhere am I interfering with the threading like wait() or Sleep() or notify().

You are calling GUI stuff from `feed()`, which is always running in a machine task thread. You cannot do that. Use `SwingUtilities.invokeLater()` or similar. 

> Only thing is, it eats the exception and the machine thinks feed() has succeeded for good, which is OK for us.

Before claiming such things, take the time to see where the `feed()` call originates from in the `JobProcessor` and in the GUI action and what happens when _no_ exception is thrown. For the  `JobProcessor`, check what further actions take place, before the processor next checks for `State.Paused`. Frankly, you should have done that before even _thinking_ about intercepting `feed()` like that. 

> Regarding [#2](https://github.com/openpnp/openpnp/issues/2) I think that is already taken care of. i.e pressing GUI button feed will not inadvertently start the job!!

When I said "resume" I meant the `ResumeJob()` call in your `doProceed()`. The `feed()` could have been called from a job or from the GUI action. The  `ResumeJob()` must only be called in the former case.

I've spent a lot of time on your last PR, explaining things again and again. You should have learnt to second-guess your approach properly, before claiming things, notably _directly_ against my advice. Show some readiness to actually check out what I try to explain, or ask back if it is not clear what I mean ... oh and it doesn't hurt to do it nicely. :-)

_Mark

markmaker:
Please read more carefully, what I actually wrote. Investigate what I pointed out. 

_Mark

surabibio:
Here is the [last occured discussion](https://groups.google.com/g/openpnp/c/OyZd4SEXhmA/m/f0GQGBTvAwAJ) on this feature. Even thought the PR could not be consumed, wish this idea lives on and hope one day it may find takers....

PR #1496, opened 2022-11-20
UI translation improvements

VSSavin:
# Description
* Firstly, the 'Translations' class has been changed:
  + added and used 'static class UTF8Control extends ResourceBundle.Control' to use UTF8 encoding directly in editors (IDEs) for .properties files.
  + added method 'String getStringOrDefault(String key, String defaultValue)'. Returns a value from the 'RESOURCE_BUNDLE' for the specified key or return defaultValue if none is found.
* Secondly, everywhere if necessary was used the method 'Translations.getStringOrDefault' in the code for UI translations. For example, the code `new JLabel("Search");` has been changed to `new JLabel(Translations.getStringOrDefault("FeedersPanel.SearchLabel.text","Search"));`
* Also, after using the 'UTF8Control' class we can't use https://github.com/jcbvm/i18n-editor to read/write translation files properly because it  doesn't support UTF-8 directly. To solve this problem I forked sources to https://github.com/VSSavin/i18n-editor and made some fixes (you can now choose the encoding in the 'Settings' menu). Now we have to use it. 

# Justification
This change improves UI translations.

# Instructions for Use
No change.

# Implementation Details
1. Manual UI testing.
2. Did follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)
3. There are some changes in `org.openpnp.spi` and `org.openpnp.model` packages. All changes  affect only 'name' and 'short description' of Actions (like 'deleteAction', 'permutateUpAction' etc.). For example `putValue(SHORT_DESCRIPTION, Translations.getStringOrDefault("AbstractAxis.Action.Delete.Description",
   "Delete the currently selected axis."));`
4. Successful mvn test before submitting the Pull Request.


markmaker:
Hi @VSSavin, 

this seems like a _huge_ amount of work, all the more I _really, really_ wish you had discussed this with us beforehand, as we clearly demand both in the [developer guide](https://github.com/openpnp/openpnp/wiki/Developers-Guide#discussion-first) and in the [pull request template](https://github.com/openpnp/openpnp/blob/develop/PULL_REQUEST_TEMPLATE.md). 🙁

I hope @vonnieda can have a look, as I don't really know the canonical "Java way" of doing such translations. 

Is this really the purely manual coding process it looks like? If true, it seems very cumbersome to do it this way, with no (apparent?) help from any tools. I was under the impression, that some IDE tools (Window Builder?) provide translation support with automatic source code parsing and refactoring (that's where the strange `//$NON-NLS-1$` annotations come in, right?). Plus the UI builder seems to maintain the translation files automatically, at least it did add new strings, when I changed a part of the UI that already had translations enabled. 

Is there a way with your solution to incrementally (!) generate the missing entries in the language files, and populate default English?

There are [some strings with variables in them](https://github.com/openpnp/openpnp/blob/39a6e1aa625ff82193bfc3b1fd0e018f580e27d9/src/main/java/org/openpnp/gui/IssuesAndSolutionsPanel.java#L407-L411), I guess this does not work, you must provide (numbered) placeholders, and then substitute them after the translation is applied. 

Another problem is the fact that this PR touches countless files, while e.g. @tonyluken has announced extensive work, so conflicts are very likely. We should really try to _coordinate_ such "cross-section" work. 

I really hope my concerns are unfounded, there seems to be a lot of hard work in it. @vonnieda?

_Mark

ian-arkver:
Does this break other translations, eg. zh_CN which you have not converted to UTF-8?

vonnieda:
Hi all,

First, thanks for the PR @VSSavin! I have only had a brief look, as I am prepping for my biggest work day of the year, but it looks quite nice to me so far!

I can come back to this after Thanksgiving if needed, but here are my quick thoughts:

- Converting the language files to UTF-8 is a huge win. This lets us edit these files as plain text and will make it far easier for people to contribute. They'll be able to edit the files right on Github to submit a PR. I should have done it that way from the start. (There were reasons, but they turned out not to be good ones)

- Can someone address Mark's comment:
  > Plus the UI builder seems to maintain the translation files automatically, at least it did add new strings, when I changed a part of the UI that already had translations enabled.

  I actually don't think I knew that, but if that is the case it would be a shame to lose it.

- It doesn't look like it will break anything as @ian-arkver asked, but if someone could throw a script together to convert the rest of the files to UTF-8 that'd be nice.

Thanks,
Jason

vonnieda:
> Is there a way with your solution to incrementally (!) generate the missing entries in the language files, and populate default English?

@markmaker The resource bundle system is supposed to fall back to the default properties file if no mapping is found, so no need to add English to the non-English files. 

So, it would be very nice to have an always up to date defaults file that was generated automatically, but I don't think we need to push the defaults into the other files.

Jason

markmaker:
> So, it would be very nice to have an always up to date defaults file that was generated automatically, but I don't think we need to push the defaults into the other files.

Like I said I might be missing something, but I don't see how new entries find their way into any of the translation files, neither the English nor any other. So translators would have to scan the source code to find new translatable items and their default meaning in English, which I do not think is practical. If it is expected of the initial developer to manually add texts to the English language file, I don't see the point of putting the default text in the source code redundantly. It would also be a revision nightmare, when the English language file then either unexpectedly overrides the source code default, or the language file is out of sync. 

Translators usually need the English text to do their work, so populating missing entries in _all_ the language files with the default text would be very helpful for them. They can then just overwrite the English with their language. 

Well, that's how _we_ do it for 30 years in my regular job, using a completely different system (not Java), obviously. Frankly I'm also surprised that there is no native support in Swing. In our UI system, privileged users (translators, developers) can alternatively translate/label everything interactively, simply while running the application: right click on any label/caption/button/menu and then edit text, tooltip, assign icon, etc. And this is from 1994.

When I find time, I will do tests with Window Builder, to see what really happens.

_Mark

VSSavin:
@ian-arkver 
> Does this break other translations, eg. zh_CN which you have not converted to UTF-8?

No, this doesn't break other translations, eg. zh_CN. But zh_CN has never worked for me (tested on Ubuntu 18.04, Windows 7). This only works if I rename `translations_zh_CN.properties` to `translations_zh_cn.properties` (i.e.  in lower case). Has anyone tested this?

vonnieda:
@markmaker I was not very clear - I was trying not to suggest a solution, but now I will, to explain how I think it could work. I think you and I have have the same goals, and I think @VSSavin's PR definitely is a step towards those.

If there is no clear standard and better way to do it, then I think doing it our own way and the best we can is fine.

So here are some ideas that I think could build on top this PR to make it more future proof:

- Use annotations to define the keys and defaults. This makes them easily available to tooling.
- Write an annotation tool that gathers up all the keys and defaults and does two things:
    - Put all the keys and defaults in the default language file. Regenerate this every time. Translators can use this to create translation files.
    - Generate one or more Java enums containing the keys.
- Reference the annotated key in the getStringOrDefault(), instead of the text key. This will make it type safe and typos in key names will be a compile time error.

So, it would look something like this:
```java
// User code
public class MyUi {
    @Translation(key = "hello message", default = "Hello, World!")
    @Translation(key = "goodbye message", default = "Goodbye, Take Care!")

    public void sayHello() {
        System.out.format(Translations.getStringOrDefault(Translations.MyUi.hello_message));
    }

    public void sayGoodbye() {
        System.out.format(Translations.getStringOrDefault(Translations.MyUi.goodbye_message));
    }
}

// Translations.java auto-generated by annotation processor
public class Translations {
    public enum MyUi {
        hello_message,
        goodbye_message,
    }
}

// i18n_defaults.properties auto-generated by annotation processor
MyUi.hello_message="Hello, World!"
MyUi.goodbye_message="Goodbye, Take Care!"
```

Regardless of all that - I think this PR is worth merging **as it is** as long as it doesn't break anything and we can get @VSSavin and @tonyluken synced up. Getting everything in UTF-8 is a great first step and adding some more structure and automation on top of it can be a future task.

VSSavin:
I don't understand why the build failed in the workflow... Is this a problem due to a multithreaded build? Everything is built manually without any problems.

ian-arkver:
zh_CN doesn't work because the Locale argument is wrong.

This fixes it:
```diff --git a/src/main/java/org/openpnp/gui/MainFrame.java b/src/main/java/org/openpnp/gui/MainFrame.java
index 05172da73f..4d51cffec9 100644
--- a/src/main/java/org/openpnp/gui/MainFrame.java
+++ b/src/main/java/org/openpnp/gui/MainFrame.java
@@ -431,7 +431,7 @@ public class MainFrame extends JFrame {
         buttonGroup.add(menuItem);
         mnLanguage.add(menuItem);
 
-        menuItem = new JCheckBoxMenuItem(new LanguageSelectionAction(new Locale("zh_CN")));
+        menuItem = new JCheckBoxMenuItem(new LanguageSelectionAction(new Locale("zh", "CN")));
         buttonGroup.add(menuItem);
         mnLanguage.add(menuItem);
```

image

markmaker:
Regarding Window Builder translation support: 

If you open a UI class that already contains translations, it transparently uses them, and you can edit texts interactively:

Translation support

You can also switch the active language, and do the translation in the Window Builder:

image

New entries are automatically added ... (just for fun I added a new tooltip for the Z **P** button)...

image

... in the code, it automatically adds the translation lookup, even generating the key for you ...

image

... and it adds the text to the active translation file ...

image

So there seems to be some sort of "canonical Java translation support". 

I would expect there to be additional tools and support that I have yet to discover. The `//$NON-NLS-1$` clearly serves the purpose to mark strings that should _not_ be translated, which sort of _implies_ that there are tools to parse the source code for yet untranslated strings and automatically refactor them into translation lookups. I also expect there to be tools to sync the different language files for missing entries, so translations can be done offline. 

I'm just saying we should be careful not to break that built-in support. 

_Mark

ian-arkver:
There's discussion of Eclipse's NON-NLS markers [here for example](https://stackoverflow.com/a/40266605/14569828)

ian-arkver:
Personal opinion below on the "future task" bit.

I think the source code needs to remain easily readable and the changes in this PR and existing OpenPnP code seem very verbose to my eye.

The only localisation framework I've brushed across in the past was Qt's where you mark text to be translated with a lightweight tr(string)...
```
        titleLabel->setText(tr("Document Title"));
        ...
        okPushButton->setText(tr("&OK"));
```
There's a lot of tooling available for Qt, eg. Qt Linguist, however that's not very relevant here of course.

A brief Google rummage turns up [GNU gettext i18n](https://www.gnu.org/software/gettext/manual/html_node/Java.html) for Java which feels similarly lightweight...

```System.out.println(i18n("Operation completed."));```

Gettext uses [PO format](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html), but there are tools to convert between PO and properties. PO files seem to be standard to various frameworks. Being [GNU of course it has a lot of documentation](https://www.gnu.org/software/gettext/manual/gettext.html).

There are also tools to extract the strings which need localisation and generate template PO files (POT files). The i18n method returns the original string if no translation is available. Templates can be used to update existing translation files with new/removed strings.

There are a few [editors for PO files, including a couple of online](https://www.drupal.org/community/contributor-guide/reference-information/localize-drupal-org/working-with-offline/editors-for-po-files) ones. 

IMHO marking UI strings to translate is cleaner than marking all the strings which **don't** need localisation. As someone who has given up on Eclipse, I'm not sure that relying on Eclipse for internationalisation tooling is a good thing.

One potential downside to using gettext with Java is that the Java classes seem to be quite old. I'd hesitate to bring in more unmaintained code to this codebase.

VSSavin:
I don't have much experience with Window Builder.
To keep the Window Builder support, I can rename the `Translations.getStringorDefault(String key, String defaultValue)` method to `Translations.getString(String key, String defaultValue)`. Window Builder then understands all the keys.

But there are some things in Window Builder that it currently cannot handle:
1 defaultValue in this case can only be one string. String cannot be compound
2 Can't handle Action. Some of the values are in the code like `putValue(SHORT_DESCRIPTION, "Some description")`
3 Can't handle TitledBorder (or i don't know how to do it)

Perhaps there is something else...

markmaker:
It seems OpenPnP was previously modeled after the "Classic Eclipse messages class" pattern, look for section "Messages class and properties files", here:  

https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.wb.doc.user%2Fhtml%2Ffeatures%2Finternationalization.html&cp%3D89_4_1

So this is not just a Window Builder feature, but a generic Eclipse/Java facility. And sure enough, you can select menu "Source > Externalize Strings..." and get automatic support for all strings:

Externalize Strings...

You can configure the accessor class and properties file:

image

There is also help documentation available if you press the `?` button. 

I hope you can make this compatible with your PR, it seems possible. 👍 

_Mark

VSSavin:
Thanks for the info. I will try this later

VSSavin:
Hi all! 

I've made some changes to the code to support Ecpipse strings externalization tools:
*  The `Translations.getString(String key)` method is used everywhere everywhere instead of Translations.getStringOrDefault(String key, String defaultValue)`;
* All default values moved to the default translation bundle;
* All `Translations.getString(String key)` methods are marked with `//$NON-NLS-...$` comment;
* zh_CN the Locale argument bugfix

This PR is now fully compatible with Ecpipse strings externalization tools.

markmaker:
Hi @VSSavin 

great work! 

Unfortunately, I only just now saw that you targeted this at the `develop` branch, not the `test` branch. For historical reasons (and despite its name), the `develop` branch is the _stable_ branch for OpenPnP 2.0 (while `master` is for the quite outdated OpenPnP 1.0). The `test` branch is the actual branch where development must take place. 

I will be retargeting this now, anticipating quite some conflicts and new/missing translations. Sorry I haven't seen this earlier. 😭

_Mark

markmaker:
Ouch, I hoped there would be less... 😭

Please holler if you need help with git (I'm no expert, but I guess I could help with this).

VSSavin:
How can i fix this?
Should I merge the branch 'develop' with the branch 'test' in my forked repository? And then create a new Pull Request?

markmaker:
Hi @VSSavin,

Your patience is much apprecieated...

Like I said, I'm no `git` expert and still sometimes wrestle with it. So take the following with a grain of salt:

First, I guess it would be best to branch a proper "feature branch" from your local `develop` branch, because it should no longer be associated with the repo `develop` branch. Search for "feature branch" online to learn about those. 

Then you can merge `test` into your feature branch, which will recreate the conflicts. You then need to resolve those conflicts. Then commit the resolved state to the feature branch.

Because the branch name changed, the resolved feature branch must be submitted as a _new_ Pull Request (we can then close _this old_ PR). I guess that's the cleanest way forward now.

Again, sorry that I did not spot the wrong target and base branches before. **Your valuable work to resolve all this is much appreciated!**

I also see that the [Your First Pull Request](https://github.com/openpnp/openpnp/wiki/Your-First-Pull-Request) Wiki page is very much outdated and misleading. 

_Mark

ian-arkver:
You can use something like `git checkout -b ui_translation` to name your feature branch, then `git rebase --onto` to move the chain of commits over onto the upstream test branch. You'll have to fix up the commits as the rebase progresses.

[Docs are here.](https://git-scm.com/docs/git-rebase)

It's possible that this is what github is doing under the hood anyway, but I prefer the command line tools.

markmaker:
See #1498

PR #1510, opened 2022-12-29
Fix/Camera Properties

markmaker:
# Description
Fixes/improvements contained in this PR:

- New option to freeze OpenPnpCaptureCamera properties, so they will be reapplied to the camera after reopening the device. Properties not persisted correctly, or changed in other OpenPnP configurations, or other web cam applications will be restored. Connecting the wrong cameras will no longer overwrite configured property settings.
- Fixes some layout inconsistencies on the camera Device Properties form. 
- Improves the stepping through static exposure probing by Issues & Solutions. A maximum of 64 probes will be done, probes are properly stretched over the full range (no need to be a integer step size multiple).
- Fix a threading issue where the Issues & Solutions apply method would compete for the camera lighting/switching actuator, which will trigger an error message ("Machine still busy after timeout expired, task rejected"). 
- Stowaway: Support a new detection method for Smoothieware on CHMT controllers using the `X-HARDWARE` property.

# Justification
Various user reports and own tests on Linux.
- https://groups.google.com/g/openpnp/c/SpkC2XxB_jk/m/wy-fmdb9AwAJ
- https://groups.google.com/g/openpnp/c/RT9uaZO82w0/m/dqEozDd-BQAJ

# Instructions for Use

Camera Properties, Frozen

The new **Freeze Properties?** determines how camera properties are handled: 

- If **disabled** (default), the camera device driver will store the properties (**Auto** and **Value**). They are queried from the camera whenever it is reopened. 
- If **enabled** (new), the OpenPnP configuration (`machine.xml`) will store all the currently set properties (**Auto** and **Value**). They are reapplied to the camera whenever it is later reopened. They can still be changed through the UI, but settings will never be overwritten from the camera device. 

Use the **Reapply to Camera** button, to reapply properties to the camera at any time. If this is required, i.e. if the camera is arbitrarily not accepting or losing its properties, please report to the [discussion group](http://groups.google.com/group/openpnp).

Enable **Freeze Properties?** when you want repeatable settings stored with the OpenPnP configuration. You can then switch between configurations or even between different applications using the web cam with no harm. The configured  properties will always be restored.

Likewise if you mistakenly reconnect the two cameras to the wrong/swapped USB port, it will no longer overwrite or reset the configured settings. Just connect to the right USB ports or select the right USB device from the drop-down and press **Apply**. The configured settings will be restored to the camera (an OpenPnP restart might be needed for a fresh USB device enumeration).

Finally, if your camera device driver or OS does not properly store settings between OpenPnP sessions or OS reboots, they are now restored as soon OpenPnP reopens the camera.

Notes: 
- If you first switch the  **Freeze Properties?** on, press **Apply** for the properties to be initially frozen. 
- Once you want to change the camera device for a different make or model, you will likely need to disable **Freeze Properties?** and press **Apply** to query reasonable settings from the new camera (availability and ranges of properties may change between makes and models). 

# Implementation Details
1. Tested under Linux with ELP and built-in web cam.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/camera/OpenPnpCaptureCamera.java
+++ b/src/main/java/org/openpnp/machine/reference/camera/OpenPnpCaptureCamera.java
@@ -90,6 +90,9 @@ public class OpenPnpCaptureCamera extends ReferenceCamera implements Runnable {
     @Element(required = false)
     private CapturePropertyHolder zoom = new CapturePropertyHolder(CaptureProperty.Zoom);
 
+    @Attribute(required = false)
+    private boolean freezeProperties = false;
+
     public List<CaptureDevice> getCaptureDevices() {
         return capture.getDevices();
     }
@@ -427,6 +430,7 @@ public class OpenPnpCaptureCamera extends ReferenceCamera implements Runnable {
         @Attribute(required = false)
         private Boolean auto;
 
+        private OpenPnpCaptureCamera camera;
         private CaptureStream stream;
 
         public CapturePropertyHolder(CaptureProperty property) {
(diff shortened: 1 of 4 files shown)


markmaker:
Added Wiki:
https://github.com/openpnp/openpnp/wiki/OpenPnpCaptureCamera#freezing-camera-properties

PR #1513, opened 2022-12-31
Diptrace import rotation fix

sovcharenko:
DipTrace exports rotation of bottom parts 'mirrored'. This PR fixes this by mirroring parts rotation during import.

Tested on my machine. See below. All diodes should point to the center.

[rotation-30x30.csv](https://github.com/openpnp/openpnp/files/10327785/rotation-30x30.csv)

**Before the fix**

Top:
Screenshot of Screen Sharing (31-12-2022, 12-59-21)

Bottom:

Screenshot of Screen Sharing (31-12-2022, 12-59-51)

**After the fix**

Top:

Screenshot of Screen Sharing (31-12-2022, 12-59-21)

Bottom:

Screenshot of Screen Sharing (31-12-2022, 12-58-55)

### DIFF
--- a/src/main/java/org/openpnp/gui/importer/DipTraceImporter.java
+++ b/src/main/java/org/openpnp/gui/importer/DipTraceImporter.java
@@ -119,11 +119,20 @@ public class DipTraceImporter implements BoardImporter {
             double placementX = Double.parseDouble(tokens[2]);   		// X (mm) in Diptrace export
             double placementY = Double.parseDouble(tokens[3]);   		// Y (mm) in Diptrace export
             double placementRotation = Double.parseDouble(tokens[5]); 	// Rotate in Diptrace export
-            String placementLayer = tokens[4];    						// Side in Diptrace export
+            Side placementLayer = tokens[4].charAt(0) == 'T' ? Side.Top : Side.Bottom;    	// Side in Diptrace export
+
+            if(placementLayer == Side.Bottom){
+                // Bottom parts need to be mirrored
+                placementRotation = 180 - placementRotation;
+
+                // ensure rotation is positive for better UX
+                if(placementRotation < 0){
+                    placementRotation = placementRotation + 360;
+                }
+            }
 
             Placement placement = new Placement(placementId);
-            placement.setLocation(new Location(LengthUnit.Millimeters, placementX, placementY, 0,
-                    placementRotation));
+            placement.setLocation(new Location(LengthUnit.Millimeters, placementX, placementY, 0, placementRotation));
             Configuration cfg = Configuration.get();
             if (cfg != null && createMissingParts) {
                 String partId = pkgName + "-" + partValue;
@@ -143,7 +152,7 @@ public class DipTraceImporter implements BoardImporter {
 
             }
 
-            placement.setSide(placementLayer.charAt(0) == 'T' ? Side.Top : Side.Bottom);
+            placement.setSide(placementLayer);
             placements.add(placement);
         }
         reader.close();


tonyluken:
Is this a work-around for the problem reported in this [DipTrace Bug Report](https://diptrace.com/forum/viewtopic.php?f=9&t=13070)?

sovcharenko:
Doesn't sound like it

From the linked thread:

> Sorry Alex but that is not true. Having the component flip vertically** when you change sides but the angle staying the same is definitely wrong. Specifically, the cathode of a diode pointing up and calling that 90 degrees rotation is 180 degrees away from the industry standard, and this bug is causing problems with automated assembly, requiring manual inspection and correction of all XY data. This really is a problem! Manually checking 50 diodes in a layout against the XY output is very time consuming and error prone. Tom also showed earlier in this thread that the angle in the "Component Properties" window does not always match the "Properties" tab on the right side of the main window. There is obviously an underlying issue here.
> 
> Thank you for your time and help!
> 
> **edited to clarify vertical flipping. horizontal flipping with the angle staying the same is correct, but not vertical flipping.

The bug as I read it is about 

1. Different rotation data in the properties panel, properties dialog for the same component
2. Vertical flipping

While as you can see from the screenshots I provided there is also horizontal flipping. Also I checked a dozen of my boards in diptrace and rotation data is consistent and 'mirrored'

Im don't what's the industry standard for rotation data on bottom side but I can see that both [KiCad](https://github.com/openpnp/openpnp/blob/03b0ab14eae6f593e9c444df01fff542be8d9364/src/main/java/org/openpnp/gui/importer/KicadPosImporter.java#L123) and  DipTrace export rotation data in the same way

tonyluken:
But if you read further into that bug report, you will find:

> Thanks, Serg.
> 
> I want to clarify that this isn't an issue of practicality or my preference, but the **exported pick and place data is actually wrong, showing the opposite angle for some parts placed on the bottom of a PCB.**

My only real concern is that if DipTrace decides to fix the above-mentioned bug (if in fact it is a bug), then OpenPnP may need to be changed again. That's ok, but then there is the problem of what to do if someone tries to import from an older version of DipTrace.  They don't by any chance include any version information in the export do they? If so, that could be used as a key to determine whether to apply the fix or not.

sovcharenko:
it doesn't include the version. i'll try to reach out to developers and confirm what's is the expected bahaviour

sovcharenko:
Sorry for delay. I spoke to the DipTrace developers in Telegram.

They're saying that indeed rotation logic was changed in version 4.0 and this logic is aligned with KiCad now - bottom rotation is mirrored.

So yes while this PR fixes imports from version 4.x it also breaks imports from version 3.x.

There are a few options given that DipTrace exports do not include version information:

1. Add a checkbox (ticked by default) to the import dialog for the user to confirm whether export is from DipTrace 4.x or later
2. Have 2 import options in OpenPNP: `Diptrace`  and `Diptrace 3.x or prior`

tonyluken:
Hey @markmaker, what do you think?

I'm fine with the first option @sovcharenko suggested or simply just supporting DipTrace 4.0 and later only (since earlier versions actually were doing it wrong).

ian-arkver:
Since DipTrace costs money and the upgrade from 3.x to 4.x isn't free either I think a checkbox would be better than just changing the behaviour and breaking import for the pre-4.x users.

Just my $0.02 - I'm a Kicad user. :-)

tonyluken:
Yeah, I wasn't aware DipTrace users had to pay to upgrade so I'm good with a checkbox.

markmaker:
Hi,

Is there really nothing in that file that identifies the version? More columns in the header, for instance?

I'm just speaking out of experience from my professional life: such version "switches" must be automatic, or you will have constant hassle. If the DipTrace makers are pros, they added such (implicit) info. Or maybe the file extension changed? Surely, they cannot just change their export files' semantics without any way to determine the difference. 

There is a comment in the source that shows the format:

https://github.com/openpnp/openpnp/blob/7c0ee50fd5bb0d4b36db4414de5bec4cb82d9d8c/src/main/java/org/openpnp/gui/importer/DipTraceImporter.java#L100-L104

Maybe that can be a clue as to how the column header looked originally, and you can compare to how it looks today, then switch the version based on the header, here:

https://github.com/openpnp/openpnp/blob/7c0ee50fd5bb0d4b36db4414de5bec4cb82d9d8c/src/main/java/org/openpnp/gui/importer/DipTraceImporter.java#L108

_Mark

markmaker:
@sovcharenko, 

any feedback to my question?

_Mark

markmaker:
Hi @sovcharenko, @ian-arkver, @tonyluken,

What should I do? 

I'm reluctant to break 3.x users' workflows. 

_Mark

tonyluken:
I'd say the checkbox option mentioned above would be the way to go. I think I'd have it default to the 3.x version so only those that have upgraded to 4.x and beyond would need to change the setting. It would be nice if the setting of the check box was saved (in the preferences?) so that an operator doesn't have to remember to check/uncheck each time they do an import.

Tony

ian-arkver:
2nd what Tony said.

hydra:
> Is there really nothing in that file that identifies the version? More columns in the header, for instance?

Hi Mark,

Just reading this and getting up-to-speed.

Yes, you should be able to detect diptrace 4.x (4.3 at least) by checking the CSV headers, as detailed here:

https://github.com/openpnp/openpnp/issues/1625

So for older versions of diptrace you should be able to apply the rotation fix automatically.  If there's a version of diptrace which has the old headers but has the rotation bug fixed then you'll need to prompt the user or have a checkbox when > 4.0 (fixed rotation) but < 4.3 (new headers+fixed rotation) is detected.

Suggest checking the headers that diptrace 4.x writes.  One of the things they changed was the name of the column from `rotate` to `rotation` so we might be in luck here...

PR #1522, opened 2023-02-13
Fixed propagation of Set Enabled and Set Error Handling context menus

tonyluken:
# Description
Fixes context menu selections for Set Enabled and Set Error Handling on the Boards tab so that they properly propagate to all instances of the Placement.

# Justification
Bug fix for all users.

# Instructions for Use
No special instructions - this just makes it work as expected.

# Implementation Details
1. How did you test the change? **Setup a job with multiple instances of a board, changed the Enabled status and the Error Handling for some placements on one of the boards, and then used the context menu on the Boards tab to set them to their default state. Verified that all instances of the placements in the job were now back at their default settings.**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **No changes to any of these packages.**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,29 @@
 This file lists major or notable changes to OpenPnP in chronological order. This is not
 a complete change list, only those that may directly interest or affect users.
 
+# 2023-02-14
+
+## Panelization and other UI changes/improvements
+
+Panels are now stand-alone entities much like boards. They are now stored in *.panel.xml files
+rather than being "built-into" the job file. Panels can now have arbitrary layouts and can consist
+of any number of different boards and/or subpanels. Many of the issues issues that have been
+reported with the legacy panelization method have been fixed.
+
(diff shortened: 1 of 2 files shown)


markmaker:
Thanks @tonyluken, do you want to add an entry in `CHANGES.md` right now, or later?

tonyluken:
@markmaker,

>do you want to add an entry in CHANGES.md right now, or later?

I don't know. When do you usually update it - when something goes into test or into develop? BTW - I'm writing up an announcement for the Google Group right now.

Tony

markmaker:
Because it is such a great addition, I think it should be added right away. People use the About dialog to check for such a feature. Learned that with the multi shot vision😏
_Mark

tonyluken:
Ok, I'll add something.

markmaker:
Great, thanks.

PR #1527, opened 2023-02-25
Capture tool's Z coordinate and use it to set board's Z

tonyluken:
# Description
Changed the Capture Tool Location button on the Job table to only update the Z and not the X, Y, or Rotation of the selected board.  Also added the capability to update multiple boards to the same Z value. 

# Justification
Somewhere along the line, this was broken. It used to work but now, prior to this change, there is no way to capture the Z coordinate of the tool and use it to set the board's Z.

# Instructions for Use
Select one or more boards and/or panels in the Job table, jog the nozzle tip over one of them and lower it until it is just touching the board or panel, click the Capture Tool Location button to set all selected board Z coordinates to the tools Z. 

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **Tested it by selecting one or more boards in the list, lowering the nozzle tip, and clicking the Capture tool Location button - observed that all of the selected boards' Z coordinate changed to the Z coordinate of the tool and that their X, Y, and Rotations did not change**.
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **No changes.**
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. **All maven tests were run and passed**.


ian-arkver:
Looks good.

What's the benefit of the change to getLengthZ and deriveLengths over getZ and derive?

tonyluken:
>What's the benefit of the change to getLengthZ and deriveLengths over getZ and derive?

They automatically handle any unit conversions that may be necessary.

ian-arkver:
OK, makes sense. So should there be a similar change for captureCameraBoardLocationAction where camera.getLocation is applied to the board with setLocation? Maybe those units are always consistent (and the tool ones aren't)?

markmaker:
Thanks, @tonyluken !

PR #1530, opened 2023-03-13
Fixed in GcodeDriver:

futurelink:
# Description
I've encounted a few bugs in GCodeDriver:
- when connection error appears (e.g. serial port was busy) it was unable to reconnect once again. In my case this was a terminal program that occupied the serial port. So when I closed it I was still unable to turn on the machine.
- when parse error raises because of no command was confirmed (i.e. Grbl answered with 'error:xxx') it was not handled correctly. This actually was a strange bug - when the controller cannot process a request for some reason OpenPNP treats it as a 'timeout' exception. To me, when the controller answered (no matter if it was an error message) then this should be considered as cofirmed request.

# Justification
These fixes really made my life easier (at least on machine setting up and debugging). As well, I believe, they can do for other users.

# Instructions for Use
Not a new feature, just a bug fix.

# Implementation Details

Automatic test passed. Manually tested in particular circumstances:

1st case:
1) turn on terminal program to keep the serial port busy, try to turn on machine. You'll get an error message.
2) close the terminal and try again. You'll get the same error. 
2.1) **When fixed** - it should connect successfully (unless something else is wrong).

2nd case:
1) define both _COMMAND_CONFIRM_REGEX_ and _COMMAND_ERROR_REGEX_
2) leliberately set _MOVE_TO_COMMAND_ ot faulty value (e.g. add some random characters at the end so that controller reacted with error code)
3) try to move an axis. You'll get a 'timeout' error. 
3.1) **When fixed** - this will be a normal error message.

Yes. I guess.

PS. Probably errorResponse is no longer needed, but I left it in place not to make this change too large.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/GcodeAsyncDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/GcodeAsyncDriver.java
@@ -278,7 +278,7 @@ public class GcodeAsyncDriver extends GcodeDriver {
                 catch (Exception e) {
                     // We probably got a timeout exception. We can't throw from the writer thread. Therefore, set 
                     // the exception as an error response, it will be reported when the driver wants to do the next step. 
-                    errorResponse = new Line(e.getMessage());
+                    errorResponse = new Line(e.getMessage(), true);
                     //Logger.error("[{}] {}", getCommunications().getConnectionName(), e);
                 }
             }
--- a/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
@@ -293,11 +293,18 @@ public class GcodeDriver extends AbstractReferenceDriver implements Named {
     static public class Line {
         final String line;
         final double transmissionTime;
+        boolean isError;
 
         public Line(String line) {
             super();
             this.line = line;
             this.transmissionTime = NanosecondTime.getRuntimeSeconds();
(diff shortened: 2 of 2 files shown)


markmaker:
Hi @futurelink, nice addition! Thanks.

The confirmation flow control for `GcodeDriver` looks good. 

It should probably also be checked in `receiveResponses()` to be complete, otherwise it will still hang on an actuator read or any other wait for a specific regex in the responses:

https://github.com/openpnp/openpnp/blob/68a055eaa8c156311d18a58b6b250c921eab5662/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java#L1190-L1194

@futurelink have you also tested this with `GcodeAsyncDriver`? 

Background, see here:
https://github.com/openpnp/openpnp/wiki/GcodeAsyncDriver#settings

There are some overrides where the change might matter too. Confirmation flow control is handled in the reader thread, and errors might also matter when using **Location Confirmation**, though I guess this is fine if you add the error handling to `receiveResponses()`. I haven't checked myself, though.

Use Issues & Solutions to replace the `GcodeDriver` with a  `GcodeAsyncDriver` to test. You have to reach the Advanced milestone. Or as a hack, replace the class name in the `machine.xml`.

As @vonnieda suggested, this should go into `test`.

_Mark

markmaker:
Hi @futurelink,

can you please resolve the conflicts?

Or should I close this?

_Mark

markmaker:
We can reopen this when author resurfaces.

PR #1536, opened 2023-03-19
Fix pseudo-placement rotations and display correct footprint during multi-placement alignment process

tonyluken:
# Description
Fixes the rotation angles for pseudo-placements and pseudo-fiducials on the bottom of panels. Also changes the Multi-Placement Alignment Process to individually select each placement in the Job Placements table as it runs so that the correct footprint will be overlayed on the top camera view.

# Justification
Will help anyone who uses the Multi-Placement Alignment Process to align their boards and panels. Note, the incorrect pseudo-placement and pseudo-fiducial rotation angles only affected how the footprints were overlayed on the top camera view - they had no affect on the actual alignment results since their rotations are not used in computing the placement affine transforms.

# Instructions for Use
No changes to usage instructions.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **Added all the real placements and fiducials from all the boards on the test panel as pseudo-placements/fiducials and verified that they all now all have the same rotation as the real placements/fiducials (including those on the bottom of the panel).**

### DIFF
--- a/src/main/java/org/openpnp/gui/panelization/ChildFiducialSelectorDialog.java
+++ b/src/main/java/org/openpnp/gui/panelization/ChildFiducialSelectorDialog.java
@@ -72,6 +72,7 @@ import org.openpnp.model.Part;
 import org.openpnp.model.Placement;
 import org.openpnp.model.PlacementsHolderLocation;
 import org.openpnp.model.Point;
+import org.openpnp.model.PseudoPlacement;
 import org.openpnp.util.Collect;
 import org.openpnp.util.QuickHull;
 import org.openpnp.util.Utils2D;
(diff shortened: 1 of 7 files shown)


markmaker:
Hi @tonyluken 

I see two times 

https://github.com/openpnp/openpnp/blob/ef5e8009a0db4362db30ddfe0f2509c8281ca284/src/main/java/org/openpnp/gui/panelization/ChildFiducialSelectorDialog.java#L416

https://github.com/openpnp/openpnp/blob/ef5e8009a0db4362db30ddfe0f2509c8281ca284/src/main/java/org/openpnp/gui/panelization/ChildFiducialSelectorDialog.java#L372

(the second is actually from before)

Could you please replace them with something like this?

```java
    Logger.error(e, "What function failed, maybe why, and what to do about it.");
```

_Mark

tonyluken:
Yes - ready to be merged.

PR #1539, opened 2023-03-27
Replace index -> photon and add more full support for photon firmware based feeders

Jnesselr:
# Description
The Index machine was renamed to Lumen PnP, but we decided to make the feeder its own entity called photon. That's why this diff removes the index namespace and adds one called photon. Any feeder that runs the photon firmware and/or uses the photon protocol should be able to work with this. See https://github.com/photonfirmware/photon for firmware.

# Justification
Since it's 99.9% under a specific machine package and doesn't touch anything else, I'm hoping it can just be accepted. This is a large diff that we could have pushed into multiple smaller diffs, but based on past conversations we didn't feel this was the right move. I also know that without hardware to test on, it may be difficult to give suggestions.

# Instructions for Use
The Photon feeder efforts started a couple of years ago with one simple goal in mind: It should just work.

Feeder setup is one of the most time consuming and costly aspects of setting up a job for a PnP machine. We wanted to make setup as easy as possible. After you've set up your machine, gotten it moving around and such, add a PhotonFeeder in the feeders tab. It will automatically set up a data actuator that it needs on your existing G-Code driver.

If you've bought a kit from Opulo, the feeder slots all came pre-programmed. But seeing as building it yourself is just as valid, a wizard is included to program feeder slot addresses. Simply click on any PhotonFeeder in your list, click the "Global Config" tab, and start the wizard. The feeder slots only need to be programmed once.

Load up your machine with feeders, however many you need. Under the "Global Config" tab, click "Search". The maximum feeder address can be changed if you have more than 50 feeders. The protocol supports up to 254 feeders on a single machine but it will take longer to search for that many feeders.

Once the search is complete, the feeders list will be populated with all of the feeders that are on your machine. You don't have to manually add a PhotonFeeder for every physical feeder, we handle that. The only time you ever would need to hit Search again is if you've added a feeder to your machine that OpenPnP has never seen before. You can click on any of the existing PhotonFeeder entries in the feeders tab, click "Global Config" again and search for it.

# Implementation Details
There are 166 unit tests in the photon namespace alone, out of 234 unit tests in all of OpenPnP. We've also conducted manual testing both by using simulated and actual hardware.

👍 

👎 

mvn test has never worked well on my system but IntelliJ shows all unit tests passing so I think we're good. Bless you, GitHub CI action.


vonnieda:
This looks good to me. The only thing it modifies outside of it's own tree is ReferenceMachine, and just to add the feeder. As long as @markmaker is okay with it I think we can merge.

Jnesselr:
> I guess this must first go into test first, I changed the base branch. Please resolve the conflicts.

Done!

> If the machine.xml already contains any Index classes, this will fail on load. However, as the class was not registered before, this will only happen for users that hand-tweaked their machine.xml, right?

That is correct. But even our beta users won't be affected, only @sphawes and I and we've already made the requisite changes. Beta users have been testing the core functionality of this code for at least a month.

> Then, after resolving the conflicts, this is okay for me.

Awesome, thanks for the sign off! CI looks good, so I think it's safe to land?

vonnieda:
Looks good! Thanks for the PR, and to Mark for reviewing. Good luck with the feeders!

PR #1542, opened 2023-04-05
Feature/Script to sort Translation Properties

markmaker:
# Description
WindowBuilder seems to shuffle the `./src/main/resources/org/openpnp/translations.properties` file whenever the slightest text change is made in the interactive editor. Despite intensive search I have not found a way to stop that.

This PR adds a simple bash script to sort the `translations.properties` file (more or less) the same way the normal externalization seems to do. By using a canonical sort order, the changes to the file become minimal in terms of versioning.

# Justification
The diff in the  `translations.properties` file should be kept minimal.

# Instructions for Use
After each change that affects the  `translations.properties` file, execute the `./src/main/resources/org/openpnp/sort-properties.sh` script in a bash class shell. The script creates backups for safety.

# Implementation Details
1. Tested in several past PRs. Now published.
2. The [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style) does not apply.
4. `mvn test` does not apply.

### DIFF
--- a/src/main/resources/org/openpnp/sort-properties.sh
+++ b/src/main/resources/org/openpnp/sort-properties.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+cp translations.properties translations.properties.bak \
+&& LC_ALL=C sort -t$'=' -k1,1 translations.properties > translations.properties.new \
+&& mv translations.properties.new translations.properties


markmaker:
Hi @vonnieda,

please see the PR. 

I just wanted to confirm with you that you likewise do not know another/proper way to ensure this. 

If you know a way to ensure the running of the script before `git commit`, it would even be better.

Thanks,
_Mark

ian-arkver:
Hi @markmaker 

Using git hooks is the canonical way to do something like this...
https://git-scm.com/docs/githooks

I don't know how you would tie this in with github though, nor how you would make it work in general for windows git users. Mingw git shell users might be ok with a /bin/bash shebang.

Regards,
Ian

markmaker:
Hi @ian-arkver,

thanks, I finally had a look.

According to these links, they are bash scripts more or less _by definition_.

> By default, Git for Windows executes hook scripts using its own Windows port of the bash shell.

- https://stackoverflow.com/a/18278072
- https://www.howtogeek.com/devops/how-to-use-git-hooks-for-commit-automation/
- https://gitforwindows.org/

So that should be okay. I guess MacOS is also not a problem.

I guess, we would have to test whether that shell also supports the `sort` command and behaves exactly the same way. I'll investigate. 

_Mark

markmaker:
I guess we can add hooks later. Merging this now.
_Mark

PR #1546, opened 2023-04-15
Feature/Whole Body Bottom Vision, Footprint Cleanup

markmaker:
# Description

Adds stock whole-part-body bottom vision using a color-keyed background/shade. 

Notable example: the Espressif ESP modules that defy normal bottom vision due to white markings on underside in addition to having off-center contacts.

image
_Photo by Fabian Kropp_

## Features

- Adds "Whole Part Body Bottom Vision Settings" for parts where the whole body rather than contacts are to be aligned. 
- Vision Compositing (a.k.a. "Multi-shot") **Body** option to take body rather than pads for corner candidates.
- Adds `SimulatedUpCamera` nozzle tip and background color schemes for simulation of color-keyed backgrounds (testing).
- Refactored `Footprint`s to use their native right-handed coordinate system. 
- Use proper affine transform in all 2D contexts to transform right-handed machine coordinates and units into left-handed 2D pixel coordinates.
- Fix `SimulatedUpCamera` to draw asymmetric footprints correctly (formerly mirrored). 
- Fix tests that actually _compensated_ for the bug in `SimulatedUpCamera`, i.e., remove the compensation.
- Extend the footprint templates drawing to support rotation, offsets and crop.
- Add these capabilities to the `CreateFootprintTemplateImage` stage. 
  **Note**, this is still not a fully operational setup for footprint template matching (we still can't detect rotation offsets). However, this did show that footprint template drawing is complete and working correctly, even with Vision Compositing multi-shot offsets. 
- Bottom Vision Size Check: improve diagnostic log and error messages.
- Bottom Vision Test Align: improve diagnostic error messages regarding selected package.
- Nozzle tip calibration: improve diagnostic error messages regarding distance threshold.
- Some speed-up for simulated machine tests (yes, I did many test runs).

# Justification
See the user report here:
https://groups.google.com/g/openpnp/c/ucHeolNaPuk/m/FSrvE9JdEgAJ

# Instructions for Use

Use a key-color shade behind the nozzle tip. The key-color is usually green, but any vivid color should work.

For parts smaller than the nozzle tip, you also need a matching key-color nozzle tip (e.g. green Juki nozzle tips). For parts where the package covers the whole nozzle tip, a shade will suffice. For these cases you can use CP40 and other black nozzle tips too. 

**Note,** this won't work without color-keying!

Select the "Whole Part Body Bottom Vision Settings" on packages/parts where you need it.

Select the **Body** method on the **Vision Compositing** tab, if you need multi-shots because the part is too large for the camera view. 

image

Note, this will only work, if the body dimensions (e.g. the module PCB outline milling) and offsets to embedded contacts are manufactured with precision adequate for the needed placement accuracy. As these modules often have large pitch (1.5mm in the case of the ESP module shown above), this should usually be the case.

Multi-shot in front of magenta shade

# Implementation Details
1. Tested in simulation.
3. Change in the  `org.openpnp.model` package: `Footprint` now uses right-handed coordinates (as does everything else in OpenPnP).


vonnieda:
Sounds awesome Mark! Both the feature and the additional improvements along the way. Nice work!

PR #1547, opened 2023-04-17
Default board location

janm012012:
# Description
This PR add a default board location property to the ReferenceMachine Locations section. This location will be used as default when a board is added to a job. 

# Justification
On machines with a dedicated board registration system like CHM-T36 or 48, the location of boards on the machine is always the same. With this preconfigured location there is almost no board position setup required for new boards anymore.

# Instructions for Use
The default default board location is the same as the hardware coded location currently in place. So without specifying a default board location nothing will change. Once the default board location is specified boards added to jobs will automatically have this location as reference. 

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. _I've set the default board location and added new and existing boards to jobs and verified, that the default location is set as expected. I also verified, that the default board location is stored in machine.xml._
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? _I copied some code and kept the format as it was._
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. _I modified spi by coping DiscardLocation related code to get the default board location in the config file and to provide getter and setter methods._
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. _I executed "Run As -> Marvin test" from within Eclipse without errors._

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/wizards/ReferenceMachineConfigurationWizard.java
+++ b/src/main/java/org/openpnp/machine/reference/wizards/ReferenceMachineConfigurationWizard.java
@@ -35,6 +35,10 @@ public class ReferenceMachineConfigurationWizard extends AbstractConfigurationWi
     private JTextField discardYTf;
     private JTextField discardZTf;
     private JTextField discardCTf;
+    private JTextField defaultBoardXTf;
+    private JTextField defaultBoardYTf;
+    private JTextField defaultBoardZTf;
+    private JTextField defaultBoardCTf;
     private JComboBox motionPlannerClass;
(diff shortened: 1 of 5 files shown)


markmaker:
Hi @janm012012,

looks good to me from the code side. But the GUI stuff is hard to judge, could you please add a screen-shot? 

_Mark

janm012012:
Hi Mark! 
Here it comes: 

DefaultBoardLocation

Jan

Issue #1101, opened 2021-01-06
Altium import heading decode


DavidG-Aus:
# Read This First
To submit an issue for OpenPnP you must use this template or it will be deleted. Please fill out either the Problem Report or Feature Request section below and delete the other section.

-----------------------------------------------------------------------

# Problem Report
## Describe the Issue
The board import function does not recognise the current Altium AD20 CSV headings

## Steps to Reproduce
How can a developer reproduce the issue you are experiencing? Use a step by step format like this:
This is a sample CSV file from AD20 (AD21 has just been released but I don't have it yet).

Altium Designer Pick and Place Locations
D:\PCB\AAAA\Aligner\.\Pick Place\Pick Place for Aligner.csv

========================================================================================================================
File Design Information:

Date:       07/01/21
Time:       10:17
Revision:   Not in VersionControl
Variant:    No variations
Units used: mm

"Designator","Comment","Layer","Footprint","Center-X(mm)","Center-Y(mm)","Rotation","Description","Height(mm)"
"R21","33k","TopLayer","1206","54.4576","64.4428","270","Resistor 1206 1/4W","0.0000"
"J2","LINK6","TopLayer","IDC6","50.0126","17.3990","180","6 pin link","0.0000"
"R20","4k7","TopLayer","1206","52.9336","59.4136","90","Resistor 1206 1/4W","0.0000"
"R19","22k","TopLayer","1206","52.8858","78.4352","360","Resistor 1206 1/4W","0.0000"

## Expected Result
The X & Y headings are not being decoded.

## Actual Result
Import fails.

## Notes, Log Files, Screen Captures, Videos, etc. to Show the Issue
From OpenPNP log:

2021-01-07 10:20:52.225 NamedCSVImporter TRACE: checkLine: "Designator","Comment","Layer","Footprint","Center-X(mm)","Center-Y(mm)","Rotation","Description","Height(mm)"
2021-01-07 10:20:52.237 NamedCSVImporter TRACE: checkCSV: DESIGNATOR = 0
2021-01-07 10:20:52.238 NamedCSVImporter TRACE: checkCSV: COMMENT = 1
2021-01-07 10:20:52.238 NamedCSVImporter TRACE: checkCSV: FOOTPRINT = 3
2021-01-07 10:20:52.238 NamedCSVImporter TRACE: checkCSV: Ref = 0
2021-01-07 10:20:52.238 NamedCSVImporter TRACE: checkCSV: Val = 1
[...]

PR #1549, opened 2023-04-27
Altium CSV importer

janm012012:
# Description
This PR adds a CSV importer for Altium generate pick and place files. It reuses the Named CSV importer with updated pattern and applies the currently missing correction of the rotation of bottom side components. The pattern used to decode the data now only contains pattern used by Altium, includes "center-x" and "center-y" as requested and imports comments.
In addition this PR fixes the issues with the Named CSV importer that on a second pass bad data is not detected any more (flags used to detect successful header line detection where not reset before importing).

# Justification
This PR adds "center-x" and "center-y" to the X and Y pattern which are the defaults by Altium and which has been requested a few times. It adds the required pattern to import comments. It also fixes the rotation of components on the bottom side.

# Instructions for Use
The Altium CSV importer shows up in the list of importers and can be used just like the Named CSV importer. As it reuses its code, it behaves the same.

# Implementation Details
1. How did you test the change? _I've take multiple jobs where I manually fixed the rotation and verified that the data imported using the Named CSV importer and the Altium CSV importer has the expected rotation._
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? _Yes_
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. _No_
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. _Yes_


markmaker:
Hi @janm012012, 

Can you shed some light on what is new and what is just renamed/refactored and how? 

Frankly, I don't want to review the full diff.

Btw. I'm no expert in theses matters, but shouldn't you use your own name in the header copyright of new files?

You can of course still credit Jason, like I usually do:

https://github.com/openpnp/openpnp/blob/460d8aa4bbfa2a59af1afd2c6e03487d4ba63406/src/main/java/org/openpnp/machine/reference/driver/AbstractMotionPlanner.java#L2-L4

_Mark

janm012012:
Hi @markmaker!

Many thanks for taking the time to look into this PR. I'm happy to discuss and modify it as needed to make it nice.

The design goal was to reuse as much of the existing code as possible. Basically, for Altium imports, the rotation of bottom side components needs to be corrected. In order to apply that only when needed, one has to make sure the imported data is really from Altium. Therefore I favored a separate importer over modifying the existing one. So here is what I did:

1. Generalize the Named CSV importer so that it can be specialized for its current purpose and for the new Altium importer. This applies manly to the pattern used to decode the header line and to the way the mil to mm conversion detection is done.
2. Based on the generalized CSV importer I created two specialized version that actually serve as importers.

For the first step I renamed NamedCSVImporter to CSVImporter, removed all references to the BoardImporter class and added dynamic pattern loading using abstract getter methods.

The second step is then to regenerate the NamedCSVImporter as child of the new CSVImporter. It then only has to provide the methods required by the BoardImporter class and the (old/current) pattern to decode the data.

The actual Altium CSV importer is finally a copy of the new NamedCSVImporter with modified pattern (Altium specific only) and a modified version of the import function which, after a successful import, corrects the rotation of components on the bottom side.

Finally, while debugging, I found a few more issues with the original NamedCSVImporter that I corrected in order to provide a more robust user experience.

Actually I'm not very happy with the amount of getter methods required to get the pattern into the importer. It probably would be cleaner to group all configuration data into a new class and provide that as part of the constructor to the CSVImporter. Unfortunately I'm not an expert in Java and hence don't know if that would be possible and/or would make it cleaner.

Jan

markmaker:
Hi @janm012012, 

Thanks for the quick response. 

So there is much more new code than I thought. I (wrongly) assumed you just put existing code into a base class and the slightly specialized per CAD.

OK, so I has a quick look and there are some issues. Numbered for easier referencing.

1. **Note**, I did not analyze whether you took this from the previous implementation, so I'll just comment what I see, and hope you agree this is the opportunity to improve some things. In any case, there is no offense or blame or whatever intended. 🤠

1. Strictly speaking, `CSVImporter` should be named `CsvImporter`. See the [Coding Style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style). 

1. Ditto the derived classes. 

1. I know the `NamedCSVImporter` is historical, but given it now changes so drastically, I guess the class name can now be fixed. I wonder if there is a better name. I mean "named" are all of them, right? Perhaps `ReferenceCsvImporter`, assuming I understand correctly that OpenPnP actually _sets the reference_ for how the columns must be named.

1. In  `CsvImporter` the column header names and many other variables are `static`. Unless I'm missing something, these should be non-static members, assigned to this **instance** of the importer. Granted, we will hardly ever run two (different) importers at the same time (and thus create a race condition), but this is just bad design that should be fixed. 

1. In fact, proper non-static design would have prevented the bug you fixed, i.e. you can remove the bugfix (resetting flags in the constructor) and just let the the flags initialize themselves with each new importer instance. It shows once again, globals are 👿! 

1. All these member variables should be lower case, this is regardless of whether they are `static `or not. See the [Coding Style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style).

1. Similarly, the methods should be non-static. The compiler will force you, once you change the variables.

1. Personally, I don't think you need declare constants like `cRefs`, `cVals`, etc. and then also override all the getters. IMHO, these getters already represent the symbols and can return these Strings as literals, directly, in their respective classes, including in the base class `CsvImporter`.

1. I find the design is good like that, it allows sub-classing and overriding the columns _individually_, which may be useful if some informal "industry standard" with variants, or subsequent versions of an E-CAD are to be supported, where only few things change between variants/versions.

1. Also, please check what I said about the copyright header.

_Mark

markmaker:
Brilliant! 

The renaming triggered `git` to compare the right classes, so I now see what you all improved. Great work!

Thank you so far and tell me when it is ready.

_Mark

janm012012:
Hi @markmaker!

Many thanks!

Please have a look at the new code. I hope I did not miss any of your recommendations. I tried to move the dialog into a separate file, but failed. It shares to many variables with the importer itself.

Jan

markmaker:
Thanks, @janm012012 , great work!

janm012012:
Thanks @markmaker for merging!

Is there anything else for me to be done to integrate the changes and/or make them available for others?

There are also issues #1101 that might be solved with this PR. How is that handled?

Jan

markmaker:
You need to make a new PR (maybe there's another way, but I don't know how).
_Mark

PR #1553, opened 2023-05-02
Actuator signaler remove duplicate calls

janm012012:
# Description
This PR changes the way ActuatorSignaler works be filtering out duplicated calls to the actuator if the job state did not change.

# Justification
When using an ActuatorSignaler to signal a finished job, the actuator is called for each step of the job. This slows the job down dramatically even if no GCode is omitted. (Unfortunately with the NullDriver and the pnp-test.job I was not able to measure a difference in CPH.) One can see this behavior by filtering the log (at Trace level) for "Signaler". Without this PR there a many calls to the actuator with state false. With this PR applied, there are two calls with state false and one with state true.
In addition, this PR allows to use machine coordination with the actuator without a significant performance penalty. 

# Instructions for Use
There is no change in usage. Its just that the actuator and its driver is not called that frequent anymore.

# Implementation Details
1. How did you test the change? _I've used the pnp-test.job and filtered the log for "signaler". This shows the expected reduction in calls to the actuator._
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? _Yes_
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. _No_
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. _Yes_

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 This file lists major or notable changes to OpenPnP in chronological order. This is not
 a complete change list, only those that may directly interest or affect users.
 
+# 2023-05-03
+
+Behaviour of ActuatorSignaler changed to only call the actuator if the job state has changed.
+
 # 2023-05-02
 
 Named CSV importer renamed to Reference CSV importer
--- a/src/main/java/org/openpnp/machine/reference/signaler/ActuatorSignaler.java
+++ b/src/main/java/org/openpnp/machine/reference/signaler/ActuatorSignaler.java
@@ -19,7 +19,6 @@ public class ActuatorSignaler extends AbstractSignaler {
 
     protected Machine machine;
     protected Actuator actuator;
-    protected AbstractJobProcessor.State lastJobState;	// last job state, used to only call the actuator if the job state has changed
 
     @Attribute(required = false)
     protected String actuatorId;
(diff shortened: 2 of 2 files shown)


janm012012:
Hi @markmaker!

Many thanks for your explanation. It makes sense to use the actual state of the actuator as reference to detect if an actuation is needed. I've changed the code accordingly and applied the same check to the machine state related path as well. The new code only generates two calls to the driver, one with state false on job start and a second with state true when the job has finished. Thats exactly what I expected. (The actuator is setup with state after enable "AssumeUnknown" that likely cases the false actuation on job start.)

Jan

markmaker:
Hi @janm012012,

thanks!

Because of parallel #1555 there is now a conflict that you must resolve.

Also, it seems there are whitespace/indentation issues. Note that we want all spaces, not tabs in OpenPnP. See
https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style
specifically
https://google.github.io/styleguide/javaguide.html#s2.3.1-whitespace-characters

**Just as a side note:** if I'm not mistaken, the machine state signalling is **defunct**. There is only _one_ machine state defined, namely `ERROR`, and the signalling is never actually called. I guess it could be removed from the actuator and Wizard.

We have the _newer_ machine state logic on _each_ actuator, not only the signalling actuators, which IMHO makes more sense, i.e., it is useful to be able to initialize and finalize _any_ actuator when the machine is enabled/disabled etc. 

So if you are feeling energetic, you could rip that out too. Just make sure to keep the `machineState` member as it is declared `@Attribute` and we need to keep it, so the `machine.xml` is not  broken. You can mark it as `@Deprecated`. If you don't feel energetic, that's fine too... 😁

_Mark

markmaker:
Thanks a lot, @janm012012!

janm012012:
Hi @markmaker!

I've submitted PR #1556 to entirely remove AbstractMachine.state.

Jan

PR #1556, opened 2023-05-03
Remove abstract machine state

janm012012:
# Description
As discussed as part of PR #1553, AbstractMachine.state is not used and shall be removed. The removal is done in two steps

1. remove all reference from signaler except in AcutatorSignaler to prevent xml parser errors and
2. remove it entirely from AbstractMachine. 

The second step might break user setups in the unlikely event, that they configured a signaler of type ActuatorSignaler and altered the Machine state drop down to ERROR. To recover they would either have to change the Machine state back to null before the update or remove "MachineState" from the signalers section of their machine.xml manually. As PR #1553 fixes a performance penalty with ActuatorSignalers, its unlikely that users use this signalers widely. In addition as machine state is unused, it might just be for testing purposes that they switch the machine state. Therefore I would consider it acceptable to remove it silently. However, merging just the first step, will cleanup the ActuatorSignaler wizard without any other impact to the user.

# Implementation Details
1. How did you test the change? _I loaded my signaler test setup I used when testing PR #1553._
5. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? _Yes_
6. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. _Yes_
7. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.


markmaker:
Thanks, @janm012012 ! 👏

PR #1534, opened 2023-03-14
JDK17 and MacOS Support

vonnieda:
This finishes up JDK17 support #1493 by @lags.

In addition to @lags work on https://github.com/openpnp/openpnp/issues/1361 and https://github.com/openpnp/openpnp/issues/1363, this fixes MacOS support for x86 and Apple Silicon, and fixes Github Actions for all platforms.

PR #1562, opened 2023-05-25
Fix / Upgrade to openpnp-capture-java 0.0.28-0.

markmaker:
# Description
Upgrade to openpnp-capture-java 0.0.28-0.

# Justification
See 
- https://github.com/openpnp/openpnp-capture/pull/68
- https://github.com/openpnp/openpnp-capture-java/commit/52374b9730a8d3d78163bbad8c4b8761469743ab

# Instructions for Use
Upgrade `test` version.

# Implementation Details
1. Tested local develop on Windows, using Java 11. Real test will be after deploy, using the installer-packaged JVM. 
2. The [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style) does not apply.
4. Used `mvn -DskipTests clean package` to build and then test the faulty functions under Windows.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -183,7 +183,7 @@
 		<dependency>
 			<groupId>org.openpnp</groupId>
 			<artifactId>openpnp-capture-java</artifactId>
-			<version>0.0.27-0</version>
+			<version>0.0.28-0</version>
 		</dependency>
 		<dependency>
 		    <groupId>javax.xml.bind</groupId>


markmaker:
Confirmed it worked with the deployed installer and its packaged Java 17.

lags:
Good Job @markmaker 🎉🎉
I noticed that the js scripts are no longer listed (although there are), only the python ones.
I'm not sure if it's related to this PR but I will investigate and if anything open an issue

neliooliveira:
Same problem with me.
I'm on Linux.

markmaker:
Are you sure this changed with **that last** upgrade? (this would be very strange)

Or could it have happened with the Java 17 upgrade that was introduced through https://github.com/openpnp/openpnp/pull/1534?

I vaguely remember that the Nashorn js script interpreter had a Java version compatibility issue:

https://github.com/openpnp/openpnp/pull/1475#issuecomment-1236712135

@alexanderwiller, do you happen to know how to bump Nashorn now that we left Java 8 behind? The lowest supported Java version is now 11.

_Mark

lags:
@markmaker maybe you are right. I can't check now.
As owner mention [here](https://stackoverflow.com/a/69865070/3081112), Nashorn is compatible with Java 11 (and 17) from 15.3 version. Now openpnp use v15.0

markmaker:
Guess we should proceed as described here:
https://github.com/szegedi/nashorn/wiki/Using-Nashorn-with-different-Java-versions#using-nashorn-on-java-11-to-14

markmaker:
I've made #1563 

Please upgrade to `test`, allow some minutes to deploy. Thanks.

_Mark

PR #1565, opened 2023-06-04
Feature / Fiducial vision with parallax view-points, and optimization.

markmaker:
# Description
This PR adds a new parallax view-point mode to the Fiducial Vision, and generally overhauls the system.

- Introduce parallax view-point fiducial vision to avoid problems with highly reflective fiducials such as HASL surface-finished ones.
- Move the FiducialVisionSettings properties to the actual VisionSettings object, so they can be overridden by package, part.
- Introduce the new **Max. Linear Offset** property to optimize multi-pass fiducial locator iteration: if the offset is already smaller, the iteration is stopped early.
- Optimize fiducial vision to not unnecessarily move to the final detection location.
- Bugfix: VisionSettingsPanel now takes the selected type into consideration, both Bottom and Fiducial vision settings can be created.
- VisionSettings `resetToDefault()` refactored to reset vision settings to their stock settings and pipeline.
- The newly created vision settings are now also initialized using `setToDefault()`.
- Bugfix: proper exception handling in SlotSchulzFeeder's use of the fiducial locator.
- Bugfix: SimulationModeMachine must not insist on the Pick & Place check, i.e. detecting a feeder pocket or PCB footprint at the pick/place location, when doing the confetti test.

# Justification

# Instructions for Use

## Use case

Highly reflective fiducials, such as HASL (Hot Air Solder Leveling) surface-finished fiducials, are very hard to properly light. They will likely reflect the camera peeking through the diffuser, therefore, the fiducial might appear (partially) dark and/or not well-formed for detection. Only very elaborate lighting systems, such as co-axial (half-way mirror) lights can overcome the problem. Those lights are expensive/difficult to make, and have down-sides like very narrow viewing angles, high weight and bulky size.

By detecting fiducials from a parallax view-point, the fiducial appears at an angle, and if properly spaced, will reflect the bright diffuser of the camera light. Doing this from two opposite, symmetrical viewpoints, the mid-point can be taken to cancel-out any errors introduced by the angled view.

fiducial parallax vision

## Fiducial Vision Settings

Fiducial Vision Settings

The **Max. Vision Passes** determines how many times the Fiducial Locator will try to improve the fiducial detection, iteratively centering-in on the detected fiducial.

The **Max. Linear Offset** determines the acceptable residual offset for the fiducial detection. If the residual offset is smaller, then the iteration is stopped early.

When the **Parallax Diameter** is given, the Fiduial Locator will automatically switch to the parallax mode and perform its detection from two parallax camera view-points on both sides of the expected location of the fiducial. The distance between the two is set by the **Parallax Diameter**.

The **Parallax Angle** determines the angle at which the opposite view-points are spaced from each other. Choose an angle which favors the faster axis of your machine: 0° favors X, 90° favors Y. Other angles might be used to target a specific LED pair in a LED ring reflection.

# Implementation Details
1. Tested in simulation with and without parallax, and with the two stock pipelines, Circular Symmetry and Footprint Template Image. 
3. Added the `VisionSettings.resetToDefault()` method in the `org.openpnp.model` package.


markmaker:
Wiki was added
https://github.com/openpnp/openpnp/wiki/Fiducial-Locator#parallax-operating-principle

PR #1568, opened 2023-07-04
Fix UI bug in Job Placements Table

tonyluken:
# Description
Fixes a user interface problem first described [here](https://groups.google.com/g/openpnp/c/N2sTPcGlPgI) in the user forum. Basically, the context menus for setting the Enabled, Side, Type, and Error Handling states of placements in the Job tab's Placements Table were not correctly handling the case where the board's definition should have been being updated versus just the board instance in the job.  This only occurs when a single instance of any particular board design is at the top level of the job, i.e., there are not multiple copies of the board in the job nor is it a member of a panel in the job.

# Justification
Fixes a user interface bug that impacts all users.

# Instructions for Use
No real instructions for use - the context menus now work as they should.

# Implementation Details
1. How did you test the change? **Was able to duplicate the problem and verified after the changes were made, the problem no longer occurred.**
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? **Yes**.

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
@@ -118,6 +118,7 @@ public class JobPlacementsPanel extends JPanel {
     private ActionGroup editFeederActionGroup;
     private Preferences prefs = Preferences.userNodeForPackage(JobPlacementsPanel.class);
     private Configuration configuration;
+    private boolean editDefinition;
 
 
     private static Color typeColorFiducial = new Color(157, 188, 255);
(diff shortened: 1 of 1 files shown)


markmaker:
Thanks Tony,

some questions:

- I'm not entirely familiar with the "board or panel" nature of things, so take this with a grain of salt.
- I understand that when a panel has just one board, it is in `editDefinition`  mode, correct?
- But is that `editDefinition` mode being updated, when users add a second board to the panel on the definition side, but without changing the panel location? Does this still call `setBoardOrPanelLocation()` and update the mode?

_Mark

tonyluken:
Hi @markmaker,

>I understand that when a panel has just one board, it is in editDefinition mode, correct?

No, editDefinition is set to false whenever a panel is selected in the Job table (regardless of how many boards it may have).  It is only set true when a board is selected in the Job table and the job has no other boards using that same definition and the selected board is a direct child of the job (it is not a member of a panel).

Tony

markmaker:
Thanks, Tony.

PR #1378, opened 2022-02-03
 Remove serial port read timeout. Close port before reader thread join. (resubmit)

ian-arkver:
# Description
Avoid loss of characters due to serial port read timeouts causing a partially read line to be discarded.

# Justification
As discussed in #1235 and rebased onto test branch as mentioned in #1377.

# Instructions for Use
N/A

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted.
Only tested with maven built in tests. I've no serial comms to check this.

2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?
Yes
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.
N/A
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.
Tests passed.

Issue #1235, opened 2021-08-04
ReferenceDriverCommunications can lose lines if a character read timeout occurs


Jnesselr:
# Problem Report
## Describe the Issue
In the `org.openpnp.machine.reference.driver.ReferenceDriverCommunications` class, there's a `readUntil` method. The way it's written now, if the read method gets a timeout exception, there might still be data in the StringBuffer that just gets thrown away. Now supposedly, this read timeout only happens after 500 milliseconds for the serial port, and in my testing, it was MUCH shorter than that. I'm running on Mac OS X 11.3.1.

In either case, I think the bug here is still a thing even at the 500 ms timeout point.

## Steps to Reproduce
(Sorry, not a step by step format as it's a little harder to reproduce than that)

I noticed it at first when my M115 command wasn't returning the firmware information correctly. When I started to debug it, there'd be points where it'd get part of the message, throw it away, then get the rest of the message and it wouldn't be able to parse it. It's even more noticeable on some of my code that has a timeout, but my timeout is only like... idk 100 ms? So I don't know why it's so trigger happy.

## Expected Result
I expected the entire response line to be read in.

## Actual Result
It read part of the line, discarded it with a timeout, then read the rest of the line which didn't match some regex and that made OpenPnP mad.

## Notes, Log Files, Screen Captures, Videos, etc. to Show the Issue
I can't do a video atm, and I don't have the logs from before I tested this code, but here's the code I've been using to avoid this issue entirely. It basically will NOT return if it has started reading some data UNTIL it hits one of the characters it's looking for. I wanted to open an issue first because I'm sure this approach has other consequences that should be thought about.

```java
public String readUntil(String characters) throws TimeoutException, IOException {
        StringBuilder line = new StringBuilder();
        while (true) {
            try {
                int ch = read();
                if (ch == -1) {
                    return null;
                }
                else if (characters.indexOf((char)ch) >= 0) {
                    if (line.length() > 0) {
                        return line.toString();
                    }
                }
                else {
                    line.append((char) ch);
                }
            } catch (TimeoutException ex) {
                if (line.length() == 0) {
                    throw ex;
                }
            }
        }
    }
```

markmaker:
@Jnesselr,

Frankly, I have never investigated communications, so I'm a bit out of my depth here. 

I'm astonished, why there is a timeout configured at all. 

[...]

PR #1569, opened 2023-07-06
Fix / Serial read hang

markmaker:
# Description

Users suddenly reported UI hangs with TinyG. After extensive debugging, revisions retracing and half a day of hair pulling, it turned out that reinstating a serial port read timeout solved the issue. Strangely the that abolished the timeout is over a year old. I can only speculate that the recent change from Java 8 to Java 11/17/19 somehow exposed the issue.

See PR #1378 and specifically, [this line-change](https://github.com/openpnp/openpnp/pull/1378/files#diff-dec1db26d6c2776bbec8768f001ed16c73d22f02ef6a455149992cdb83d227edL113), where the timeout of 500ms was removed:

```java
SerialPort.TIMEOUT_READ_SEMI_BLOCKING | SerialPort.TIMEOUT_WRITE_BLOCKING, 500, 0);
```

## Debugging

When the hang happens, I see no obvious OpenPnP threads being blocked in deadlock or similar. Strangely the AWT thread is blocked drawing a simple `FillRect` of the last button being pressed:

image

The only other OpenPnP thread being actively blocked was a `SerialPort` read, which brought me to inspect it closer:

image

The serial reader thread being blocked is actually supposed to be okay (that's what the thread is for). Still, I remembered #1378 and just out of sheer desperation tried to reinstate the timeout, and yes, it helped. 

## Fixes

- Reinstate the serial port read blocking timeout.

During debugging some other minor issues were fixed: 

- Better diagnostics for missing axis letter, name driver.
- Make timeouts more generous in multi-driver firmware detection (the next driver needs to wait for the previous to complete).
- Make sure to disconnect after a firmware detection ad-hoc connection fails late (connection may already be established).

# Justification
See
https://groups.google.com/g/openpnp/c/TW_xEbuD01Y/m/of_qtzcPAgAJ

# Instructions for Use
No change in use.

# Implementation Details
1. Tested with three controllers Smoothieware, Duet, TinyG under Windows.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
@@ -1350,7 +1350,7 @@ public class GcodeDriver extends AbstractReferenceDriver implements Named {
                 }
                 catch (IOException e) {
                     if (disconnectRequested) {
-                        Logger.trace("Read error while disconnecting (normal)");
+                        Logger.trace("Read error while disconnecting (expected)");
                         return;
                     }
                     else {
@@ -1420,10 +1420,10 @@ public class GcodeDriver extends AbstractReferenceDriver implements Named {
             }
             catch (IllegalArgumentException e) {
                 // Axis is not present in pattern. That's a warning, but might not be supported by controller, so we let it go. 
-                Logger.warn("Axis {} letter {} missing in POSITION_REPORT_REGEX groups.", axis.getName(), axis.getLetter());
+                Logger.warn("{}: Axis {} letter {} missing in POSITION_REPORT_REGEX groups.", getName(), axis.getName(), axis.getLetter());
             }
             catch (Exception e) {
-                Logger.warn("Error processing position report for axis {}: {}", axis.getName(), e);
+                Logger.warn("{}: Error processing position report for axis {}: {}", getName(), axis.getName(), e);
             }
         }
         // Store the latest momentary position.
@@ -1681,15 +1681,15 @@ public class GcodeDriver extends AbstractReferenceDriver implements Named {
                 detectFirmwareSequence();
                 return null;
             }, 
-                    false, 100, 1000);
+                    false, 2000, 2000);
         }
         else {
             // If the machine is not enabled, use an ad hoc connection.
             boolean wasConnected = connected;
-            if (!wasConnected) {
-                connect();
-            }
             try {
+                if (!wasConnected) {
+                    connect();
+                }
                 detectFirmwareSequence();
             }
             finally {
--- a/src/main/java/org/openpnp/machine/reference/driver/SerialPortCommunications.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/SerialPortCommunications.java
@@ -107,7 +107,7 @@ public class SerialPortCommunications extends ReferenceDriverCommunications {
             serialPort.setRTS();
         }
         serialPort.setComPortTimeouts(
-                SerialPort.TIMEOUT_READ_SEMI_BLOCKING | SerialPort.TIMEOUT_WRITE_BLOCKING, 0, 0);
+                SerialPort.TIMEOUT_READ_SEMI_BLOCKING | SerialPort.TIMEOUT_WRITE_BLOCKING, 500, 0);
     }
 
     public synchronized void disconnect() throws Exception {


ian-arkver:
Hmm. There was a good reason why we made that change back then, but I can't remember what it was now.

Jnesselr:
Correct, so you just re-instated that bug with this PR. Thanks for the tag @ian-arkver

You also need to check the result of that function call to see if it worked or not for the thread timeout because sometimes if you get the wrong baud rate, it'll adjust under the hood but fail that call, for example.

We could go with my original solution which is to make the reader ignore the -1 timeout if there's data there and never return until data is read, but it honestly REALLY depends on the reason for the hang here. If it hangs the entire application because a thread inside hangs, something else is VERY wrong.

markmaker:
> so you just re-instated that bug with this PR

Yeah, if the serial takes more than 500ms to deliver the second part of the _same_ text-line, it will happen. But frankly, I guess this would also point to something wrong on the controller side of things, i.e. no proper flushing on the line output. If I'm not mistaken it has worked that way (with the bug) for years with the common controllers.

I agree that the serial reads should in theory work when fully blocking. That's what the reader thread is for. But in practice it just does not. 

I guess there is some deadlock taking place in the native code side of things. Not sure if it only affects Windows, but knowing the GUI stuff is all based on the original 16bit single threaded event driven design, it is all too possible. Maybe we could also look at updating [jSerialComm](https://fazecast.github.io/jSerialComm/).

Regardless of the above, the actual bug must be solved in `readUntil()` as @Jnesselr said. I'd say don't worry about the endless loop, because `readUntil()` is already an endless loop if the wanted delimiters never come. The separate reader thread will be woken up for nothing, but will take care of it.

PR very welcome.

_Mark

Jnesselr:
You don't KNOW what's causing this bug or why this change fixed it, so ANYTHING you do here is suspect. You need to figure out why that change caused the issue in the first place. The code is available, and yes the serial library hits some C code but I went through it at one point to figure out why another issue was occurring. It's not impossible, but right now you're doing the equivalent of "it works on my machine".

> But frankly, I guess this would also point to something wrong on the controller side of things, i.e. no proper flushing on the line output

That's not true in any way you can actually guarantee. There's so many buffer layers here that it's near impossible to say it's the controller's fault. It's very possible it's flushed its entire buffer through the USB cable and into the host computer but the host computer's driver hasn't picked up the remainder of the bytes and put them in user land for you.

>  If I'm not mistaken it has worked that way (with the bug) for years with the common controllers

The original failure case is a silent failure and is affected more by longer response lines. 99.99% of the time if you're just getting "ok" as a response with a newline, you would not see this issue. We're only starting to see it because we're replying with longer lines that have a packet involved, instead of the 4 bytes you're seeing with "ok\r\n". So it's not at all a surprise that you didn't see any issues before.

> I guess there is some deadlock taking place in the native code side of things.

How though? That's a really important question here. How does a deadlock in a serial port library affect your whole application?

> I'd say don't worry about the endless loop, because readUntil() is already an endless loop if the wanted delimiters never come.

This is why you need to know what's causing your deadlock. This right here. Because if you're wrong about why the deadlock occurs and it's some non-native thread interaction in your driver code, then causing this thread to block for sure until it hits a newline may cause other unintended interactions or may make the same bug appear again as you supposedly solved here.

markmaker:
>  It's not impossible, but right now you're doing the equivalent of "it works on my machine".

I agree. But it's a no-brainer when the alternative is a 100% reproducible "it **doesn't** work on my machine". 

> That's not true in any way you can actually guarantee.

Nobody said "guarantee" I just said "points to", which in my non-native grasp of English means there is "circumstantial evidence" .

Really, 500ms is a long time for a USB serial connection, regardless of how many buffers, and drivers, and operating system schedulers there are in between. It is very unexpected that sending a line would not be finished in that time, given the receiving end is already dry. The only reason I can see is that the controller uses a buffer and fails to flush it, or that it starts writing a response, then does something lengthy (like waiting for the robot), then finally finishes the line. In any case, dealing with the timeout (like you suggested) would even work with that.

> This is why you need to know what's causing your deadlock. This right here. Because if you're wrong about why the deadlock occurs and it's some non-native thread interaction in your driver code, then causing this thread to block for sure until it hits a newline may cause other unintended interactions or may make the same bug appear again as you supposedly solved here.

Look. What I needed to do "right here" is to make it work again, so users can use the **test** version, as it solves various bugs for them elsewhere. And yes, this is still the **test** version, so it should be considered as such.

But this does not mean I'm happy with the situation or that I won't embrace any effort to solve it properly. And if it turns out to be some problem that I introduced elsewhere (which is perfectly possible), I will gladly admit the mistake.

This is an Open Source project. Nobody is paid for their work. Although I think I've proven to care very much about what I contribute, and to support users _extensively_ in using the features, in the end it is still "take it or leave it". Everybody is free to go back to an older version, if they have problems with the current situation.

> You don't KNOW what's causing this bug or why this change fixed it, so ANYTHING you do here is suspect.

Based on the stuff I assessed for five hours yesterday, the probability is high that the underlying cause was already there before. It just didn't crop up. 

Instead of telling me in capital letters what I don't _know_, or that _anything_ I do is suspect, perhaps you could come forward with a better explanation and solution. Perhaps start by checking out these... 

https://github.com/Fazecast/jSerialComm/issues/147
https://github.com/Fazecast/jSerialComm/issues/177

I gotta go now...

_Mark

ian-arkver:
Those two jSerialComm issues are quite old (2018 vintage). OpenPnP's pom has quite a broad range for jSerialComm, so we get a recent version. My test build pulled 2.10.0 dated last week.

I'd be more interested to see why the AWT EDT is allegedly stuck in a FillRect, if indeed it is. It's interesting that googling for AWT and/or Swing and "deadlock" gives many hits involving the EDT and invokeAndWait vs. invokeLater.

There's clearly some odd interaction going on here and I've no idea why changing serial timeouts would make things apparently work. Perhaps it's a racy deadlock and this just shifts the timing a little.

As I said on the original PR, I don't use serial comms (I use TCP), so replicating and investigating is a little tricky for me.

ian-arkver:
There were a lot of changes in jSerialComm 2.10.0 which came out last week. Would that be about the time that this problem appeared? Smoking gun?

https://github.com/Fazecast/jSerialComm/releases

Perhaps try pinning the version at 2.9.3 in the pom which was the previous release from the end of 2022.

markmaker:
> Those two jSerialComm issues are quite old (2018 vintage). 

True, at least it still shows that this is an area where bugs keep cropping up.

> I'd be more interested to see why the AWT EDT is allegedly stuck in a FillRect, if indeed it is.

Yeah, I was wondering that too. Not sure how reliable Eclipse global thread "Suspend" is. Maybe that's just a "residual" image of the last stack trace, not the actual state. 

When I was hunting another race condition a year or so back, it behaved erratically too.


>  Smoking gun? ... Perhaps try pinning the version at 2.9.3 in the pom which was the previous release from the end of 2022.

Good find. Will try.

_Mark

markmaker:
Yes the `test` version where this started to crop up was from #1568 which in itself has nothing to do with serial comm, but Maven must _for the first time_ have picked up jSerialComm 2.10.0 or 2.10.1. 

Testing with 2.10.1 _and_ the timeout removed, clearly shows the hangs. It seems the simultaneous `read()` and `write()` from our two different threads (reader and writer) causes this. Remember: the `Gcode`**`Async`**`Driver` is completely free to send queued command as fast as it can, regardless of any reads for responses from earlier commands taking place or being waited for in parallel (true full-duplex).

Somehow on 2.10.1 simultaneous `read()` and `write()` now deadlock each other, that's how it appears in the Eclipse debugger when hanging: 

reader and writer both blocked

It explains how the driver is deadlocked. But it does still not explain how this hangs the GUI. The debugger shows no other OpenPnP threads being blocked, most specifically, the machine task thread pool is idle. Just the AWT thread blocked on `FillRect`, as before. 

Forcing it back to 2.9.3 _and_ the timeout removed, did still exhibit some problems. No longer hanging the GUI, but sometimes timing out Smoothieboard reads. We had user reports of this before. This does not happen with the read timeout reinstated.

Then I remembered #1508. OpenPnP was sending the content of a command, and then the newline separately, in two `write()` calls. Vespaman reported that in deed it is sent over USB separately, which of course is wasteful (see the PR for more details).

Reverting these #1508 changes, the timeout issue with Smoothie disappears (still with 2.9.3 and without read timeout). Again, _what the heck?!_ 😧 

I know, some USB serial implementations finish up and send a packet, as soon as they receive a newline, i.e. not waiting for the packet to be full as they otherwise do, there is a grace period too, of course. The newline handling obviously reduces latency, when command-lines are thought as datagrams. So something there is fishy.

This newline thing does not help with 2.10.1, it still hangs the GUI. 

So after this round of investigations I'm frankly completely confused. Any ideas how to proceed?

IMHO, the most sensible approach would be:

1. Force `jSerialComm` back to 2.9.3.
2. Reintroduce the timeout.
3. But fix the bug of #1235 in `readUntil()` by just _continuing_ the loop in case of `TimeoutException`.
4. Keep sending line+LF in one `write()` (#1508), as I think this is actually important for speed.

With that I wasn't able to cause an issues, using Duet, Smoothie, and TinyG in wild mixed motion.

But I agree, this is still not really solved. Mysteries remain and it is to be expected that sometimes OpenPnP will still hang for up to 500ms in the same lock, just not a _deadlock_.

What do you think? 

I will likely make a PR to that effect in a few hours, if I don't hear objections.

_Mark

ian-arkver:
I reckon if you've found a combination that fixes all 3 problems (hang, #1235 and #1508) then go ahead and roll that out on test so we can see what else (if anything) is broken. As you say, that's what "test" is for.

It's interesting that there may be a write/read lock taken within the serial library. Perhaps this is exercised at this point, eg. something in the UI thread may be causing a command to be sent? This was a "connect" button click that triggered the deadlock, right?

Perhaps we need to synchronise/serialise the reads and writes at a higher level if the locking within jSerialComm is not working as we expect.

Anyhow, thanks for persevering with this Mark.

markmaker:
> Perhaps this is exercised at this point, eg. something in the UI thread may be causing a command to be sent? This was a "connect" button click that triggered the deadlock, right?

Everything talking to the drivers must be run in the machine task. `UiUtils.submitUiMachineTask()` et al must always be used. 

I removed many calls that weren't properly tasked, mostly in the feeder sub-classes, but of course I cannot guarantee there aren't any left, somewhere. However in my own testing with just some fast and wild jogging and "position to feeder pick location" clicking, I'm quite sure, none were involved. 

> Perhaps we need to synchronise/serialise the reads and writes at a higher level if the locking within jSerialComm is not working as we expect.

That would defy much of the purpose of `GcodeAsyncDriver`. 😭

But `jSerialComm` has event based operation. 

> If you would like to use the jSerialComm library in a completely asynchronous fashion, you can enable event-based callbacks via the [addDataListener()](http://fazecast.github.io/jSerialComm/javadoc/com/fazecast/jSerialComm/SerialPort.html#addDataListener-com.fazecast.jSerialComm.SerialPortDataListener-) method.

https://github.com/Fazecast/jSerialComm/wiki/Event-Based-Reading-Usage-Example

Maybe it is optimized for this, and not for multi-threading. And maybe the assumption of using events instead of threads explains why it locks itself out.

_Mark

markmaker:
Reported my suspicions here:
https://github.com/Fazecast/jSerialComm/issues/505#issue-1795000612

_Mark

Jnesselr:
I also found some code that I thought may have been a copy/paste bug in the Posix implementation but I haven't tested it yet. https://github.com/Fazecast/jSerialComm/issues/504

Overall the changes made in 2.10.0 definitely have a lot of evidence suggesting they are the issue. Thanks @ian-arkver for that find!

> IMHO, the most sensible approach would be

This feels like a reasonable approach and if we do identify bugs in jSerialComm that warrant us to upgrade to a version that has those fixes, it can be re-pinned to be above the bad version.

I will say, given the nature of serial and serial drivers, the 500ms is fine but sometimes you just get things where read is not immediately available but is very very soon after. With a thread interrupt for timeout like we're doing, it just all depends on how often we want to check and get that exception thrown up the stack when there's no data to be had. This is why doing the actual block is nice, because you know you're only blocking the one thread. Theoretically.

The event callbacks could also probably be tested with a little refactoring to start with and may be the best overall approach while combining parts of the other approaches such as "don't discard any data you've read".

ian-arkver:
>>  Perhaps we need to synchronise/serialise the reads and writes at a higher level if the locking within jSerialComm is not working as we expect.
> That would defy much of the purpose of GcodeAsyncDriver.

I didn't mean as high a level as the gcode command/response level which would indeed break the asyncness. All I meant was that we ensure there's only one call to jSerialComm read or write active at a time with a mutex (or java equivalent) around those calls.

markmaker:
> All I meant was that we ensure there's only one call to jSerialComm read or write active at a time with a mutex (or java equivalent) around those calls.

The problem is that `readBytes()` is practically _always_ pending, wanting to read whatever the controller has to say. So we could practically _never_ spontaneously call `writeBytes()`, i.e., send new commands.

After I think I found the culprit (see [here](https://github.com/Fazecast/jSerialComm/issues/505#issue-1795000612)) I'm confident we can resolve this. @hedgecrw seems very responsive, e.g. [here](https://github.com/Fazecast/jSerialComm/issues/502).

In the meantime, I will proceed as laid out [earlier](https://github.com/openpnp/openpnp/pull/1569#issuecomment-1627268331).

_Mark

markmaker:
See also https://github.com/Fazecast/jSerialComm/issues/505#issuecomment-1627665026

PR #1570, opened 2023-07-08
Fix / jSerialComm temporary workaround

markmaker:
# Description
This provides a temporary workaround for the current issue with `jSerialComm`.

* Block jSerialComm at 2.9.3 until https://github.com/Fazecast/jSerialComm/issues/505 is resolved.
* Keep serial read timeout as introduced by https://github.com/openpnp/openpnp/pull/1569.
* Fix the bug in https://github.com/openpnp/openpnp/issues/1235, resume reading on timeout.
* Keep https://github.com/openpnp/openpnp/pull/1508 (which was found to cause problems too), it is equally solved by the read timeout.
* Add a debug log when a read timeout happens during a write, which would indicate that the writer was blocked until the reader timed out, which may or may not point to an internal read/write/full-duplex deadlock. Which might cause the problems through https://github.com/openpnp/openpnp/pull/1508.
* As an afterthought: In the reader/writer thread, keep copies of the comms and name, so changing them through the GUI while connected, is okay.

# Justification
See the follow up of:
https://github.com/openpnp/openpnp/pull/1569#issuecomment-1627268331

# Instructions for Use
No change in use.

# Implementation Details
1. Tested extensively with three test controllers Duet, Smoothie and TinyG in parallel.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
 		<dependency>
   			<groupId>com.fazecast</groupId>
   			<artifactId>jSerialComm</artifactId>
-  			<version>[2.0.0,3.0.0)</version>
+  			<version>2.9.3</version>
 		</dependency>
 		<dependency>
 			<groupId>org.swinglabs.swingx</groupId>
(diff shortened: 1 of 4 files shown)


markmaker:
Disregard the latest commit. 

I believe it is an entirely unrelated issue, the USB hub being powered off when idle:

USB power saving

Seems that Smoothie does not like that. Another topic...

_Mark

PR #1572, opened 2023-07-11
- Fixed line lenght check if a comment column is present

janm012012:
# Description
This PR fixes the import of comment columns if they are the last column in the file. It also fixes the behavior of an unchecked _Create Missing Parts_ checkbox. Previously no part was assigned to any placement in that case. Now an existing part is assigned and the placement is skipped if the part does not exists. A warning is logged in the later case.

In addition this PR changes placements whose ID starts with "FID" or "REF" followed by a digit to type fiducial.

# Justification
Both problems where reported at https://groups.google.com/g/openpnp/c/JjAjSdWhUkA/m/BLhpk9dyAwAJ.

# Instructions for Use
There is no change in usage.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **The code was tested using the supplied .csv file.**

### DIFF
--- a/src/main/java/org/openpnp/gui/importer/CsvImporter.java
+++ b/src/main/java/org/openpnp/gui/importer/CsvImporter.java
@@ -61,6 +61,7 @@ import org.openpnp.model.Location;
 import org.openpnp.model.Package;
 import org.openpnp.model.Part;
 import org.openpnp.model.Placement;
+import org.openpnp.util.Utils2D;
 import org.pmw.tinylog.Logger;
 
 import com.Ostermiller.util.CSVParser;
@@ -340,14 +341,7 @@ public abstract class CsvImporter {
 
                 double placementRotation = convert(as[rotationIndex]);
                 // convert rotation to [-180 .. 180]
-                // FIXME: this range is invalid as -180° == 180°. Whats the OpenPnP convention?
-                // FIXME: does OpenPnP provide a unified method to limit angles?
-                while (placementRotation > 180.0) {
-                    placementRotation -= 360.0;
-                }
-                while (placementRotation < -180.0) {
-                    placementRotation += 360.0;
-                }
+                placementRotation = Utils2D.angleNorm(placementRotation, 180);
                 
                 String partId = as[packageIndex] + "-" + as[valueIndex]; //$NON-NLS-1$
                 Part part = cfg.getPart(partId);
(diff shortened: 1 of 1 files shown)


markmaker:
Hi @janm012012,

wow, that was fast. :rocket:

Can I interest you in some optimizations? Even though the same flaws were more or less already present before? :grin:

1. I don't think you need to test `Configuration.get()` for null. I use it all the time without. If this is for a "naked" unit test environment, then the test would hopefully fail, because it now imports _nothing_ but warns as many times as there are placements.
1. You should move the `cfg` assignment to the beginning of the method, it makes no sense to get it each time / for each placement.
1. Just to make the code clearer: please move the whole `Placement` creation and property setting to _after_ you figured out whether the part is present. Now it is half-created and then left to the Garbage Collector, which works, but is wasteful and hard to understand for future developers. Plus the setters might actually fire notifications etc. for objects that are "zombies".

Thanks! :coin: 

_Mark

janm012012:
Hi @markmaker!

I always like to take opportunities to learn from experts and increase code quality. Makes life easier tomorrow...

Do you also have a solution of the FIXME, that is still present? The placement angle as read from file is unfolded into the range [-180..180°] which is ambiguous because -180 == 180°. Whats the OpenPnP convention and does OpenPnP provide a unified method for unfolding the angle?

Jan

markmaker:
> Do you also have a solution of the FIXME, that is still present? The placement angle as read from file is unfolded into the range [-180..180°] which is ambiguous because -180 == 180°. Whats the OpenPnP convention and does OpenPnP provide a unified method for unfolding the angle?

OpenPnP does not care about how many "wind-ups" the rotation angle has. For physical motion, it will optimize nozzle rotation according to **Nozzle Rotation Mode** and **Rotational Axis Limits**: 

https://github.com/openpnp/openpnp/wiki/Nozzle-Rotation-Mode

And  **Issues & Solutions** will properly suggest the right and optimized settings.

You can (if you like) remove all the following code, so the original E-CAD angle is preserved, which users might actually expect.

https://github.com/openpnp/openpnp/blob/6967220770a2d03b4e56e6d6991d113d803ad4d1/src/main/java/org/openpnp/gui/importer/CsvImporter.java#L342-L350

Or you can use 

```java 
placementRotation = Utils2D.angleNorm(placementRotation, 180);
```

_Mark

janm012012:
Its was there, so I'd like to keep it there.

Many thanks for the help! I've changed the code to use angelNorm().

Jan

markmaker:
angelNorm?

I asked NightCafé what the "angel norm" is. :smirk: 

image

Thanks, @janm012012 !

PR #1574, opened 2023-07-14
Optimize placements of multi nozzle machines

janm012012:
# Description
This PR adds sorting of placements of multi nozzle machines to the pick, align and place steps of the reference PnP job processor to optimize the travel path the head has to make. The optimization uses the locations where the different steps will take place for the different placements and nozzles and in combination with the heads current location and the location where the next step will move the head to optimizes the order the nozzles are operated to minimize the travel distance.

# Justification
For multi nozzle machines there is a good potential for optimizing the job when nozzles/placements can be executed in any order. This PR will make this machines faster by minimizing the path the head will travel.

# Instructions for Use
The optimization can be disabled using the reference PnP job processor configuration wizard.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **yes. I've modified the default machine.xml to a double nozzle machine and checked that the pnp-test.job runs faster**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **yes: I've added a Location thats used for the optimization**


markmaker:
Hi @janm012012 ,

thanks for tackling this. :+1:

Just so you know, I'm on a vacation, so I can't really look into this in depth, so I have to ask for your patience (two, three weeks).

Hope I got it right:

This is strictly an optimization **inside** one **set** of nozzle-assigned placements, right?

It is **not** an optimization of shuffling placements **between** these **sets**, notably for proximity of feeders, and/or for proximity of placements. For example, the planner will still take a part from a feeder all the way in the left front corner of the machine, and cluelessly pair it with a part from a feeder all the way on the right rear corner of the machine, and therefore it will still be traveling all over the machine, more or less aimlessly, right? 

I guess it is evident, that the **"between"** optimization has a lot of potential, so what I'd like to make sure off, before I merge, is that this  **"inside"** optimization is not in any way obstructing a future **"between"** optimization. 

Like I said, I likely won't be able to look into that in the next weeks. Hope you understand. :sweat_smile:  

_Mark

janm012012:
Hi @markmaker!

Enjoy your vacation. There is no need to hurry.

Yes, it's that **inside** I'm targeting with this PR. It has lots of potential for me, as on my double-nozzle machine the alignment step is almost always in the wrong order.

As you will see when you come back and take a closer look into the code, I've modified the constructor of the PlannedPlacementStep class that is called when ever a new step is to be executed. Each step is executed for all nozzles the Planner has assigned a placement to and the constructor is called with a list of this placements. I've inserted an optimizer before the list is stored for processing. The optimizer uses location information, the step provides for each placement. At present the steps Pick, Align and Place are optimized. The optimization is done by feeding the location information into a traveling salesman and using the suggested order. In addition steps can provide a location where the next step might move the head to. If available, this information is feed into the traveling salesman as well.

As long as you're on vacation, I'll use the time to update this PR with more optimizations. As discussed by e-mail, a linear distance is likely not a good measure for the traveling salesman optimization and should be replaced by something, that takes the Cartesian nature of pnp machines into account as well as the potential differences in x and y axis. I'll try to add that next. There might me also more discussions required on how to optimize the Pick step as feeding and picking is done in one step and I'm not sure what locations to use for optimization here. At present I'm using the current feeders pick location.

For the **between** optimization I'll probably prepare an other PR. Yes, I agree, that the Planner may select the placement for the first nozzle freely but should then favor placements for the other nozzles that have a short distance for pick and place locations.

Jan

markmaker:
Hi @janm012012,

just a quick reply from vacation... :sun_with_face: 

> I'll use the time to update this PR with more optimizations. As discussed by e-mail, a linear distance is likely not a good measure for the traveling salesman optimization and should be replaced by something, that takes the Cartesian nature of pnp machines into account as well as the potential differences in x and y axis.

Don't overthink this. When I mentioned the non-linear cost issue in the email, I was under the impression you wanted to implement the **"between"** optimization, where this might in-deed matter. :wink: 

For the **"inner"** optimization, it only shuffles the order between two or maybe four nozzles. Because the **"between"** optimization is missing, the cost differences will very likely be large, the best solution will win with a lot of margin. It would take a very, very unlikely and exotic situation where the non-linear cost would _ever_ change the solution. And even then, the true cost difference would very likely be negligible.

> At present I'm using the current feeders pick location.

Yes, that's one of the concerns I have. Some feeders perform calibration and  part/feeder OCR/QR-code identification inside the `getPickLocation()`. So just by asking for the location, the machine would move there, and perform some vision op (but see further below).

One Example;

https://github.com/openpnp/openpnp/blob/460d8aa4bbfa2a59af1afd2c6e03487d4ba63406/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L373-L374

The reason for this, is that these feeders want to do the calibration _and_ the OCR/QR-code part/feeder identification with the _same_ camera position and _same_ image capture to save time. Therefore, we can't move the operation to `feed()`, as other feeders do. In `feed()` it would obviously be too late to realize it is the _wrong_ part/feeder after all. :smile: 

On the other hand, as you do the **"inner"** optimization _after_ the job preparation, it should be fine, because the feeder was already calibrated in the `prepareForJob()` and it won't do it a second time. We just need to make sure that assumption is _always_ correct, and you perhaps need to test your PR with some `ReferencePushPullFeeders` and `BlindsFeeders` specifically. 

Or we need to change the interface of `Feeder` to decouple the `getPickLocation()` from calibration and OCR/QR-code detection. 

_Mark

janm012012:
Hi @markmaker 

> just a quick reply from vacation... 🌞

Please don't let me or this PR spoil your vacation!

> Don't overthink this. When I mentioned the non-linear cost issue in the email, I was under the impression you wanted to implement the **"between"** optimization, where this might in-deed matter. 😉

I'll keep it as is and wait for user experience. Could be changed later anyhow.

> On the other hand, as you do the **"inner"** optimization _after_ the job preparation, it should be fine, because the feeder was already calibrated in the `prepareForJob()` and it won't do it a second time. We just need to make sure that assumption is _always_ correct, and you perhaps need to test your PR with some `ReferencePushPullFeeders` and `BlindsFeeders` specifically.

I'll try to convert my real machine.xml into a simulation machine and see what my ReferencePushPullFeeders do on a real job.

> Or we need to change the interface of `Feeder` to decouple the `getPickLocation()` from calibration and OCR/QR-code detection.

Yes, that would be something, I'd prefer. However, I guess it's a lot of work with all the different feeder types around and it's very close to the real world so any mistake will kick back hard. For optimization I need location information. Ideally I can read a feed and pick location. I'm planning to use pick(1) -> feed(2) + place(1) -> place(2) as measure for good second part selection. If feed location is not available, as e.g. for automatic feeders, I'd fallback to pick location. At the end, reading location information shall be independent of any movement or processing. As this PR is likely the first to just read feed/pick locations, one could also move it to a new interface However, this would again involve lots of changes to lots of feeder implementations.

Jan

markmaker:
Hi @janm012012,

Some questions and concerns (please read all, before starting to fix stuff):

1. How do we proceed with the above-mentioned `getPickLocation()` decoupling? If I'm right, you could just remove the `assertCalibration();` from the `BlindsFeeder` and `ReferencePushPullFeeder` overrides. Then you would have to make sure, it is calibrated in the `prepareForJob()` overrides. According to a cursory glance, it _might_ already be the case (maybe configuration dependent). So extra careful to no duplicate calibration!
1. If I'm not missing something, you should add `nozzle.toHeadLocation()`  in  `getRawPlaceLocation()`, so you get _all_ head locations.
1. Generally think about using more descriptive name: `getPlanningLocation()` instead of `getLocation()` would prevent [future] misunderstandings.
1. You refactored `getPickLocation()`, `getAlignLocation()`, `getRawPlaceLocation()` into the `PlannedPlacementStep` base class. This looks like a violation of [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns). It should be the business of the `getLocation()` overrides in the `PlannedPlacementStep` sub-classes. If you really need to extract them, do it in the respective classes and make them at least `protected`, they should not be public.
1. You added `getHeadLocation(Nozzle nozzle)` but it is only used once, and then more or less duplicated a few more times. Please roll it into a unified method, make it take a `HeadMountable` parameter, so you can pass nozzles _and_ cameras, and add a `Location` parameter.  Make it `protected`. 
1. Just to document, but okay to ignore for now: the `Feed` step might also have motion. But that is very feeder specific. Some feeders move the head into the vicinity of the feeder, some the camera, some the nozzle, some an actuator.
1. To be honest, I don't like the `getNextLocation()` method, at all. You currently hard-wire assumptions about what the next step does. But these assumption might be wrong, for instance, some (clunkier) parts will not be aligned at all, they move straight from pick to placement. What makes this worse: it might actually vary among parts on the nozzles!
1. Instead there should be a way to generically get the _next_ `PlannedPlacementStep` and ask it `getLocation()` in turn. This would be more elegant and also restore separation of concerns. But you would need to iterate steps until one returns a non-null location. But that's not possible due to the simplistic `removeFromList` flag you added in `planDo()`, which can only peek ahead one set of placements. Instead you'd need a way to look ahead further.
1. Needless to say, you currently do not optimize the next step, your `plannedPlacements.get(0)` guess might be completely on the wrong side of the machine (yes, because the "between" optimization is missing :wink:).

The last three really bother me. Hope you find a way around it. 

_Mark

janm012012:
Hi @markmaker!

Many thanks for reviewing my code.

>     1. How do we proceed with the above-mentioned `getPickLocation()` decoupling? If I'm right, you could just remove the `assertCalibration();` from the `BlindsFeeder` and `ReferencePushPullFeeder` overrides. Then you would have to make sure, it is calibrated in the `prepareForJob()` overrides. According to a cursory glance, it _might_ already be the case (maybe configuration dependent). So extra careful to no duplicate calibration!
> 
I'm not sure either yet: I've run real jobs with ReferencePushPull feeders (with vision features disabled) and did not had any problems. I guess an assertion can not be overlooked. However, I'd love to have a method which just returns the or at least a rough location and which does not cause any movements. If more optimization will be applied, we might have to call them as part of the job creation process and probably more the once. For now, this PR only uses the location when the job is already running, when the Planner has selected placements and when the initialization has been done. 

>     2. If I'm not missing something, you should add `nozzle.toHeadLocation()`  in  `getRawPlaceLocation()`, so you get _all_ head locations.
> 
Good catch. I'll fix that!

>     3. Generally think about using more descriptive name: `getPlanningLocation()` instead of `getLocation()` would prevent [future] misunderstandings.
> 
Good point, changed.

>     4. You refactored `getPickLocation()`, `getAlignLocation()`, `getRawPlaceLocation()` into the `PlannedPlacementStep` base class. This looks like a violation of [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns). It should be the business of the `getLocation()` overrides in the `PlannedPlacementStep` sub-classes. If you really need to extract them, do it in the respective classes and make them at least `protected`, they should not be public.
> 
I guess I understand what you mean: the logic of getAlignLocation() should be moved into the getPlanningLocation() of the Align class and wise versa. However, I don't know how to code it because I would like to have the same location information as getNextPlanningLocation() in the Pick class to tell the optimizer, as part of the pick optimization where the alignment will take place. How can I do that?

>     5. You added `getHeadLocation(Nozzle nozzle)` but it is only used once, and then more or less duplicated a few more times. Please roll it into a unified method, make it take a `HeadMountable` parameter, so you can pass nozzles _and_ cameras, and add a `Location` parameter.  Make it `protected`.
> 
Changed.

>     6. Just to document, but okay to ignore for now: the `Feed` step might also have motion. But that is very feeder specific. Some feeders move the head into the vicinity of the feeder, some the camera, some the nozzle, some an actuator.
> 
I guess that's a show stopper for this optimization because the feed location will be asked for at different locations and hence will likely case extra movements then. That's pointing back to point 1 that I'd like to have a location information without side effects...

>     7. To be honest, I don't like the `getNextLocation()` method, at all. You currently hard-wire assumptions about what the next step does. But these assumption might be wrong, for instance, some (clunkier) parts will not be aligned at all, they move straight from pick to placement. What makes this worse: it might actually vary among parts on the nozzles!
> 
You're right. That's a misconception in my design. getNextLocation() is only used to give the traveling salesman an endpoint. If there is no endpoint, the optimization might be not as good but it should at least be better then without. Removing getNextLocation() entirely would be an option and would also solve point 3.

>     8. Instead there should be a way to generically get the _next_ `PlannedPlacementStep` and ask it `getLocation()` in turn. This would be more elegant and also restore separation of concerns. But you would need to iterate steps until one returns a non-null location. But that's not possible due to the simplistic `removeFromList` flag you added in `planDo()`, which can only peek ahead one set of placements. Instead you'd need a way to look ahead further.
> 
I fear things are getting mixed up here: running a job - to my understanding - consists of two nested loops of linked steps. The outer loop is the JobProcessor which runs a linked list of steps initialized with PreFlight(). Each step does something or returns an other step until Cleanup(), Finish() or Abort() return null. One of the steps is Plan() which calls the selected JobPlanner. This planner is the second/inner loop that selects/returns the next placements, one per nozzle. 

I tried to use the getNextPlanningLocation() method to provide a location where to go after the placement has been done. In order to extract a placement, that the planner will provide next, I had to reuse the former plan() (now planDo()) but make sure, that this placement will not be marked as done.

I guess by removing the getNextPlanningLocation() this concern will also be solved.

>     9. Needless to say, you currently do not optimize the next step, your `plannedPlacements.get(0)` guess might be completely on the wrong side of the machine (yes, because the "between" optimization is missing 😉).
> 
Sorry, I don't understand this point. Searching the code, I only see, that I use this sequence to provide a rough location where the next action will take place. Yes, that might be imprecise (I only take the location of one arbitrary (the first) placement) and even wrong (if no part on any nozzle will be aligned), but I'm convinced, that it's better then nothing.

Writing all this down, its likely best to remove the getNextPlanningLocation() feature and let the traveling salesman run without an endpoint. We could later add move optimizations and make this smarter. (I guess one would have to put all steps for this pick, align, place cycle into a list and optimize that.)

I've now removed getNextPlanningLocation(), moved getPickLocation(), getAlignLocation() and getPlaceLocation() to their getPlanningLocation() and reverted the job planner and the BasicJobTest.

Please have an other look. Thank You!

Jan

janm012012:
Hi @markmaker!

>This will be wrong with 50% probability (dual nozzle) or 75% probability (quad nozzle). And due to lack of "between" optimization, it will likely be "significantly wrong".

Ok, now I understand what you mean and you're right. If there are feeders all around the machine and feeders on opposite sides are used for the next placements, then this will likely be a bad hint.

>If you're happy too, I can merge this 😄!

No, please wait a little longer. It seems, that there is a risk for an unhandled exception when a pick location is not available and the traveling salesman gets a null location. I'll investigate that and provide a fix.

And, no, I actually a little unhappy because I removed the optimization, that after a pick, the parts are - usually - aligned and that after the alignment, parts are for sure placed.  Do you see any way to get end location for pick and align steps into the optimization?

Jan

markmaker:
Can you provide examples, _how/when_ it does it wrong now that the next location is unavailable? Do you test that in simulation only?

_Mark

janm012012:
As far as I can see the ReferencePushPull feeder is causing the problem: if Calibration Trigger is set to UntilConfident, getPickLocation() calls assertCalibrated() which calls obtainCalibratedVisionOffset() and generates a Null Pointer exception for me: *Exception caught, executing pending motion: java.lang.NullPointerException
	at org.openpnp.machine.reference.ReferencePnpJobProcessor$PlannedPlacementStep.step(Unknown Source)
	at org.openpnp.machine.reference.ReferencePnpJobProcessor.next(Unknown Source)*
This exception causes an empty "Job Error" for me.

If I understand the debugger correctly, the exception originated in "FindFeatures().invoke()" in line 2480 in ReferencePushPullFeeder.java. (I'm debugging using simulated cameras without a picture. So its likely that vision operation will fail).

I'll change the behavior and escalate all exceptions so that users get the intended error messages for the cause.

Maybe we should add an interface to read the pick location as raw value making sure no movements will take place. Small inaccuracies due to missing calibration are acceptable and will/shall not change the result of the optimization.

janm012012:
> Can you provide examples, _how/when_ it does it wrong now that the next location is unavailable? Do you test that in simulation only?

I'm debugging again and try to create significant use cases.  At present, its just a feeling (please keep my CHM-T in mind): with the PCB and the camera in the front, a pick sequence with one (drag-)feeder close to the PCB and the second in the back will likely trigger it because the overall path will be shorter to pick the feeder in the back second.  The same is true for the alignment step: with some strip feeders in the back of the PCB, when the head arrives picking from one of them. Then its shorter to align the left nozzle first, but bad with respect to the place location as the head has to go right and back next.

Anyhow, could be solved if there would be a legal way to call getPlanningLocation() of the next step. Unfortunately I don't know how...

Jan

markmaker:
> As far as I can see the ReferencePushPull feeder is causing the problem: if Calibration Trigger is set to UntilConfident, getPickLocation() calls assertCalibrated() which calls obtainCalibratedVisionOffset() and generates a Null Pointer exception 

Yeah that's exactly one of the concerns I had. :frowning_face: 

I guess you need to bite the bullet here, and implement what I wrote in point _"1. How do we proceed with the above-mentioned `getPickLocation()` decoupling?"_

> At present, its just a feeling (please keep my CHM-T in mind): with the PCB and the camera in the front, a pick sequence with one (drag-)feeder close to the PCB and the second in the back will likely trigger it because the overall path will be shorter to pick the feeder in the back second. 

I understand. But I think inefficiency is limited to some unfortunate cases, and when these happen, the difference is small. Dual nozzle "inner" optimization can only swap two locations, they _still_ need to be visited _both_. Making the "start" leg shorter will almost always make the "end" leg longer and vice versa. Unless there is significant parallel "directionality" between the two locations and the "start" and (unknown) "end" location, e.g., when the second feeder is _more to the front_ than the PCB (I made a few hand-sketched travel graphs...).

And even then, remember, the distance cost is not actually linear. On the long move all the way from the back of the machine to the front/camera, the machine can nicely accelerate and cruise at higher speed. I would expect this effect to cause some surprises as to which path is actually faster.

> The same is true for the alignment step: with some strip feeders in the back of the PCB, when the head arrives picking from one of them. Then its shorter to align the left nozzle first, but bad with respect to the place location as the head has to go right and back next.

I don't understand this second case. Both Alignment and Place are independently "inner"-optimized too. If the left/right nozzle predisposition favors going to the left placement first, it will be done in that order, there is no "go right and back". Again, any remaining difference will be small, with the camera location being the same and the placements being quite near together (unless you're doing huge panels). 

In technical terms this is [so-called greedy optimization](https://en.wikipedia.org/wiki/Greedy_algorithm). True, it is not optimal, but compared to what is wasted through the lack of "between" optimization, it is likely negligible.

_Mark

janm012012:
Hi @markmaker!

I rebased this branch to test again to test it using the latest ImageCamera updates you merged. Therefor the list of changes looks longer then it actually is.

I added getPickLocationWithoutMovement() to the Feeder class to collect location information without movements and vision operations. As first step, I added it as abstract member to catch all feeder implementations. I only found, that ReferencePushPullFeeder and BlindsFeeder actually do more then just returning a static location. For this two I implemented getPickLocationWithoutMovement(). Then I changed the member to default and return getPickLocation(). This way only the two feeders are modified by this PR.

Finally I changed the error handling so, that only in case of no sortLocation is uninitialized (due to exception or no information available) the sorting/optimization takes place. This shall safely handle all exception and situations where no location information is available.

I tested the implementation using a real job on my simulation machine and found that most movements are better (optimized) now but some are not. I do see the cases I mentioned earlier. Due to missing end location, some pick operations could be done better. I think for now, this shall be ok. (I don't see a nice way to implement the look-ahead to collect more location information.)

I'm planning to add more optimizations probably in an other PR: at present I would like to change the planner into some sort of list of placements, that is optimized once (at job start) and then just iterated. In parallel I'd change the processor into an other list of all steps required to handle current placements. This list could then be better optimized because the next step is always available (except for the end of the sequence).

Jan

markmaker:
Hi @janm012012,

just quickly (I gotta go):

I don't like to have two `getPickLocation()` methods. Instead, please proceed as discussed earlier, [here in point 1](https://github.com/openpnp/openpnp/pull/1574#issuecomment-1657175823). And even [earlier, why this is necessary](https://github.com/openpnp/openpnp/pull/1574#issuecomment-1637670098).

_Mark

janm012012:
Hi  @markmaker!

I'm sorry for not being patient enough with all your help!

I've now reverted the commit that modifies the feeder interface and tried to debug the vision operation: as far as I can see, at least the ReferencePushPullFeeder moves the head as part of the very first getPickLocation() in the job. The call to ensureCameraZ() (line 2455) generates a first moveTo() in the log and there are a lot more Driver related messages until performVisionOperations() (line 535) has finished. It seems that obtainCalibratedVisionOffset() (line 350) would not be called if visionOffset != null.

prepareForJob() (line 2370) is called as part of the preflight step, but does not do anything for me: the logic seems to be, that only ocr operates are to be performed there. visit is false, visionOffset null so isOcrDiscoverOnJobStart() has to return false.

I'm a little bit lost here: getPickLocation() performances vision operation via assertCalibrated() as long as visionOffset is null but if visionOffset is not null, prepareForJob() will not do anything anymore. Maybe one can add assertCalibrated() in the else path of prepareForJob()... However, while debugging this does not help: with the camera between the sprocket holes Preview Vision shows good results, starting the job the feeder complains "No line of sprocket holes can be recognized". (The head does not move)

The next two weeks I'll be on vacation. When I'm back, I'll continue.

Jan

markmaker:
@janm012012, thanks for bearing with me!

> It seems that obtainCalibratedVisionOffset() (line 350) would not be called if visionOffset != null.

I had a look and I think this is OK.

The OCR stuff is separate. Calibration needs also to be done **before** and **after** feeding (at least "until confident"). See the `feed()` method and two calls to `assertCalibrated()` there.

So unless I'm mistaken, it is okay to just remove the `assertCalibrated()` from the `getPickLocation()`. Like I said before, back when I wrote the `ReferencePushPullFeeder`, the pre-flight step wasn't yet emulated in the GUI feed&pick action, so I needed to include it, but now it is obsolete.

It doesn't hurt if you double-check this :smile: 

The `BlindsFeeder` needs to be looked at separately. Report back if in doubt there.

_Mark

janm012012:
Hi  @markmaker!

I've now removed assertCalibration() from ReferencePushPullFeeder and BlindsFeeder and fixed a merge conflict in the translate file by rebasing the branche.

According to my tests, this works. But unfortunately I do not have BlindsFeeder and I don't know how a failure would manifest itself. Furthermore the vision feature of the ReferencePushPullFeeder seems not to work anymore with the CHM-T drag feeders. "Preview Vision" works reliable, but Auto-Setup does not. I have the suspicion, that the two holes visible are not enough anymore.

So, in fact, I did not tested all the vision features in depth. What I do see is that there is no movement anymore when getPickLocation() is called and that all automated tests are passed.

Jan

markmaker:
Hi @janm012012 

looking better an better,  thanks.

I think the `else` here...

https://github.com/openpnp/openpnp/blob/aeb986c0d5c07dcc2cb74e825eca1e61ea736482/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L2376-L2378

...should be remodeled like that, i,.e., we must _only_ move if the `visit` flag is `true`:

```java
    @Override
    public void prepareForJob(boolean visit) throws Exception {
        super.prepareForJob(visit);
        if (visit && visionOffset == null) {
            if (isOcrDiscoverOnJobStart()) {
                // Check the part in the feeder using OCR, this also calibrates the feeder.
                // Note, we cannot change the parts at this point, it is too late in the Job Process, so we always stop.
                performOcr(OcrWrongPartAction.None, true, null);
            }
            else {
                assertCalibrated(false);
            }
        }
    }
```

And to trigger the `visit == true` call in the first place, we need to take the `calibrationTrigger` into consideration inside [`getJobPreparationLocation()`](https://github.com/openpnp/openpnp/blob/aeb986c0d5c07dcc2cb74e825eca1e61ea736482/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L2358C5-L2367C1), so it will add the feeder location to the visiting list when calibration is not switched off, i.e. when `calibrationTrigger != CalibrationTrigger.None`. So this should look like this:

```java
    @Override
    public Location getJobPreparationLocation() {
        if (visionOffset == null 
          && (isOcrDiscoverOnJobStart() || calibrationTrigger != CalibrationTrigger.None)) {
            return getPickLocation(0, null);
        }
        else {
            return null;
        }
    }
```

Please double-check this critically, it is just drafted from quickly looking at the code :smile: 

**Question:** how much **real machine testing** has gone into this latest iteration? have you ever tested with Ocr switched off? With calibration switched off?

Thanks for your continuing endurance :muscle: :muscle: :muscle: 

_Mark

janm012012:
Hi @markmaker!

I'm wondering weather it would be better/cleaner to refactor this PR into a second one that prepares/updates the two conflicting feeders. This new PR could then be considered a prerquirement for this.

I'm willing to make the proposed changes. I'll also carefully check the code. However, I'm not an expert on how the feeder is supposed to work. I can only make assumptions on how I think it should work.

I've tested one of the earlier revisions using a real job. That worked very well with OCR and calibration switched off. I have captured the state of that job and converted it into a simulation machine (take picture with PCB mounted, upscale it, add it to the simulation camera) and used that for more tests. It was obvious, that with calibration enabled, getPickLocation() caused the machine to move, which it should not. With the latest one or two iterations this was fixed. I never tried to run with OCR enabled. Do you by change have a picture of some feeders with OCR that I can use for simulation? Same for the Blinds feeder, do you have any pictures to debug with? (I've just take a picture with my mobile (carefully aligned to avoid tilting). The distortions are ok. You can still capture/update the feeder locations and see the machine in action. However, picking and placing itself is not directly visible as long as there is an offset between camera and nozzle.)

Jan

janm012012:
Hi @markmaker!

I did some more tests and here is what I found:

**PushPullFeeder**

- If _Calibration Trigger_ is configured, calibration takes place as part of the _Preparing feeders_ step.
- If _OCR Wrong Part Action_ is configured, it takes place as part of the _Preparing feeders_ step as well, however, the results are unexpected, because nothing actually happens. prepareForJob() calls performOcr() with OcrWrongPartAction.None which is then in performVisionOperations() translated into ocrPass = false bypassing all Ocr operations. I would have expected, that I can at least verify, that all feeders are available. However, the _CheckCorrect_ option the BlindsFeeders provides is not available here so hence its probably ok as is.
- There is a "calibrating sprocket holes pass" log entry (and some movements) even though _Calibration Trigger_ is disabled but _OCR Wrong Part Action_ enabled. I would say that with the above Ocr related actions shall be removed from the job preparation step.

**BlindsFeeder**

Both, calibration and OCR checking now takes place in the _Prepering feeders_ step.

markmaker:
Hi @janm012012,

just wanted to let you know I'm currently very busy otherwise (job), so all I can do is trying to give very quick answers.

> I'm wondering weather it would be better/cleaner to refactor this PR into a second one that prepares/updates the two conflicting feeders. This new PR could then be considered a prerquirement for this.

Strictly speaking it would be more beautiful, however (at least in my book) these two things depend on each other mutually, i.e., due to the risk, it would not make sense to change the way the feeders operate, without a very good reason ... which only the PR provides. So all I ask is to make (at least) two commits to separate the respective changes. 

> If OCR Wrong Part Action is configured, it takes place as part of the Preparing feeders step as well, however, the results are unexpected, because nothing actually happens. prepareForJob() calls performOcr() with OcrWrongPartAction.None which is then in performVisionOperations() translated into ocrPass = false bypassing all Ocr operations. I would have expected, that I can at least verify, that all feeders are available. 

Then that is a bug. It seems to have slipped in through in https://github.com/openpnp/openpnp/pull/1468. Although @crashmatt is the author, it was probably my [code snipped, suggested here,](https://github.com/openpnp/openpnp/pull/1468#issuecomment-1214107908) to blame. See the commit here: https://github.com/openpnp/openpnp/commit/ce4f97cf469c61b0a57e54ab6ef2562f5342c8b6.

I guess `performVisionOperations()` should also check for `ocrStop` which is passed `true` in `prepareForJob()`. 

So `performVisionOperations()` should be as such:

 ```java
    protected void performVisionOperations(Camera camera, CvPipeline pipeline, 
            boolean storeHoles, boolean storePickLocation, boolean storeVisionOffset, OcrWrongPartAction ocrAction, boolean ocrStop,
            StringBuilder report) throws Exception {
        Location runningHole1Location = getHole1Location();
        Location runningHole2Location = getHole2Location();
        Location runningPickLocation = getLocation();
        Location runningVisionOffset = getVisionOffset();
        ensureCameraZ(camera, true);
        
        FindFeatures feature = null;

        final boolean ocrPass = ((ocrAction != OcrWrongPartAction.None || ocrStop) 
            && getOcrRegion() != null);
        ...
```
Please double-check my analysis, and if found okay, fix with separate commit.

>  I would say that with the above Ocr related actions shall be removed from the job preparation step.

No! Instead fix the bug. :sweat_smile: 

> BlindsFeeder

I haven't checked BlindsFeeder yet. And unfortunately, I will be away professionally, traveling and with full days. So I will only be able to do that next week. Hope you understand.

_Mark

**Comment by janm012012** [comment]
Hi @markmaker!

Take your time!

I've now prepared a fix.

Jan

**Comment by markmaker** [comment]
Hi @janm012012,

still quite overloaded, sorry.

In your https://github.com/openpnp/openpnp/pull/1574/commits/dca7dd70ec6b73ed87e0d12d676e513a273fd59e you changed two things. Why the `&& calibrationTrigger != CalibrationTrigger.None`?  It seems wrong to me. 

The `CalibrationTrigger` only says when _calibration_ should happen, nothing about OCR per se. Conversely, when `ocrStop` is set `true`, then we want OCR, no matter the calibration. Or am I missing something?

_Mark

**Comment by janm012012** [comment]
Hi @markmaker!

I'm also still completely overloaded too.

Yes, at first glance it looks wrong. However, I'm not entirely convinced and can't just see the reason why I did it this way. The commend itself is very likely misleading: calibrationTrigger only covers location calibration.

Lets go two steps back and check the hole picture again: all this modification, we're currently discussing, are required because this PR wont's to read the pick location without causing any physical movements. As result to my analysis in https://github.com/openpnp/openpnp/pull/1574#issuecomment-1705712238 you suggested the ocrStop dependency. Maybe I added the calibrationTrigger dependency before the ocrStop change or to fix the unexpected sprocket hole operation.

Maybe I should revert the entire commit and recheck whats unexpected and how to fix that: most obvious performVisionOperations() shall be capable of handling ocr with and without location calibration. But how can I detect that? Is it ok to assume, that if storeHoles || storePickLocation || storeVisionOffset == false, location calibration can be safely skipped? Without such a modification each call to performVisionOperations() will perform a movement (line 2478).

In addition the ocrStop modification likely already fixed, that no ocr was performed in prepareForJob(). However, I'd suggest to double check that does not cause new issues by forcing an ocr check if that was not intended by the user. In prepareForJob() performceOcr() is encapsulated using isOcrDiscoverOnJobStart(), which makes it safe at this location.

Please tell me what you expect the feeder to do then I can recheck that.

I addition I'd like to you to tell me if you now see any way to get the location of the next movement into the optimizer. According to my experience this PR currently covers about 2/3 of the badly sorted operations and fails for the rest because it just does not take into account, where the next movement will go to. This is most notable when picking from feeder close the the camera and close to the current location followed by a pick from a feeder far away. Then the sorting is always "wrong" because its "cheaper" to go to the closer feeder first. With the addition knowledge, that the head will have to go to the camera after picking, it would always choose the other way.

Jan

**Comment by markmaker** [comment]
> Lets go two steps back and check the hole picture again: all this modification, we're currently discussing, are required because this PR wont's to read the pick location without causing any physical movements.

That always helps! :smile:

But no need to go back to square one. The bug you discovered was there before, as I said [here](https://github.com/openpnp/openpnp/pull/1574#issuecomment-1708717185) it slipped in through #1468. This means it is unrelated to all this discussion and we should not let it distract us.

Some confusing factors:

- Calibration and OCR are done together, simply as an optimization, saving time (only _one_ head positioning and _one_ camera settling and vision op). This is nice but also makes the logic (and our life) much harder.
- If either of them is due, a vision op is required. Then it becomes a _secondary_ question, whether the vision op  is done with or without OCR.
- Consequently, `CalibrationTrigger` must not have any impact on that _secondary_ OCR question altogether. It is logically an _independent_ operation, that merely happens to be _optimized_ into one vision op. 
- Even `OcrWrongPartAction.None` does not fully switch off OCR. If the `ocrStop` switch is enabled, it still needs OCR to check if the right part is loaded and then _stop_ the job preparation. 
- Why both the `OcrWrongPartAction` and `ocrStop`, why not make an `OcrWrongPartAction.Stop`? Because each `OcrWrongPartAction` can be  _combined_ with or without the `ocrStop`, e.g., you can let the machine automatically discover the part/feeder for you, but you might still want to stop the job preparation after that happens, simply to double-check.

Hoping this makes it clearer. 

If I did not make a mistake in that logic, all that remains to be done is this: 

- In your https://github.com/openpnp/openpnp/commit/dca7dd70ec6b73ed87e0d12d676e513a273fd59e simply remove the `&& calibrationTrigger != CalibrationTrigger.None`.

_Mark

**Comment by janm012012** [comment]
Hi @markmaker!

I removed the calibrationTrigger dependency as you suggested. 

There might be a missed optimization now if calibration trigger is none but ocr checking enabled. performOcr() will then move the camera to the location of the sprocket holes (line 2476), which would be not needed (all stores are false).

Jan

**Comment by markmaker** [comment]
> There might be a missed optimization now if calibration trigger is none but ocr checking enabled. performOcr() will then move the camera to the location of the sprocket holes (line 2476), which would be not needed (all stores are false).

You're right. 

But first, just to make sure we're on the same page and for others reading this, let me make the following clear: The ideal ReferencePushPullFeeder always detects the OCR _together_ with the centered sprocket holes, i.e. in _one_ camera shot. 

image

See also the [video](https://youtu.be/5QcJ2ziIJ14?feature=shared&t=42) where the OCR region is setup, and note how the process positions the camera cross-hairs between the sprocket holes it detected before.

In this _optimized_ case, there won't be any extra move.

In fact that was the only option, until @crashmatt added the ability to have a second camera shot for the OCR label at an offset in #1468. Therefore, only _if_ you use the new offset feature, then yes, an essentially unneeded extra move will occur.

To remedy this, we would need to eliminate the `for` loop with an extra `if {}` , like so:

```java
        if ((ocrPass && ocrZeroOffset) ||  storeHoles || storePickLocation || storeVisionOffset) {
            // Calibrate the exact hole locations by obtaining a mid-point lock on them,
            // assuming that any camera lens and Z parallax distortion is symmetric.
            for (int i = 0; i < calibrateMaxPasses; i++) {
                // move the camera to the mid-point 
                ...
            }
        }
    
        // setup OCR if wanted
        if (ocrPass) {
            //If ROI offset is not null then use the existing capture.
            if (!ocrZeroOffset) {
                Location ocrLocation = getOcrLocation();
                MovableUtils.moveToLocationAtSafeZ(camera, ocrLocation);
                ...
```

The OCR would still be done at the offset, after the `if {}` (careful with the nesting).

_Mark

janm012012:
Hi  @markmaker!

I understand the intention and fully agree. I also agree, that this is a minor corner case. As the code is some time old, it's likely that it was never noticed by anyone.

I added your proposed fix. To my understanding it fixes the unintended movement and it shall not case any new issues.

Jan

markmaker:
Thank you @janm012012 

for the first time I've now also looked at the BlindsFeeder. The following looks odd. The `&& ! isCoverOpen()` should IMHO only be combined with the `getCoverActuation() == CoverActuation.OpenOnJobStart`, i.e., it has nothing to do with calibration or OCR:

https://github.com/openpnp/openpnp/blob/39a830c92f813ee2d0a5207fb65448d35614b2ea/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java#L1510-L1511

The bracketing changed from before, but embarrassingly I believe it was _otherwise_ wrong before.

That's all I've seen. 

Can you also quickly describe how you tested it? Before/after these modifications? With which feeder classes? And how did the new cool optimization work out? Just a few words. :1st_place_medal: 

_Mark

janm012012:
Hi  @markmaker!

Many thanks for reviewing! I'm not sure if the code sequence is correct or not. I suspect its not ok and the intention suggests the same: if the cover is to be opened at job start AND its not open yet, the feeder has to be visited to open it. If that's the intention of the code, then extra braces would be required.

I've reviewed all feeders getPickLocation() implementation and found that only the ReferencePushPullFeeder and the BlindsFeeder do more then just returning a static location. To find out if that interferes somehow with my sorting modification, I cloned my physical machine into a simulation machine (remember the upscaling discussions we had) and run a real job in it. By using a few breakpoints I checked, if any movement is generated as part of the new optimization procedure.

I'm using ReferencePushPullFeeders regularly and it was therefore easy, to verify its operation despite the fact, that I do not have any actual vision configuration. However, if the head is moving or not or if the code enters performVisionOperations() was easy to detect. As part if this analysis I found, that a feeder is prepared multiple times if its part is used multiple times. That was also the situation where I detected, that OCR scanning without position calibration causes extra movements.

For the BlindsFeeder test I was only able to instantiate a feeder replacing one of my ReferencePushPullFeeders. I repeated the mentioned test with different configurations. However, as I do not have any vision features of a physical feeder, I could only detect if a head is moving or if the code enters assertCalibration() or its decedents.

Concerning the performance. I've added a metric, that calculates the path advantage after optimizing the order. However, this is not always a very good number because the optimization can not take the next location into account. That results on my machine in the situation, that picking is not very good optimized (feeders close the camera might still be visited first even if the other feeder is far away.), but vision is always (second nozzle is always closer to the camera then first). For the place step its the same as for the pick step. It could likely be better but that's even more difficult to implement because the next location is part of the planner. From my experience, the performance gets only better. I can't think of a situation where the optimization would choose a worth path compared to no optimization. To get the performance even better, I would like to add the cameras location as next/end location when optimizing the pick step. That's the only thing for me this PR is missing. (In an early stage it was in, but the code was not very clean. If you have a good idea, I'll try to implement it.) For more optimization one would optimize the placements next. But that's part of the planner and shall be done in a different PR.

Jan

markmaker:
Hi @janm012012, 

Happy new Year! :clinking_glasses:  

If you quickly fix the braces, I guess we could merge this. It's the **test** version after all.

I assume you will add instructions and principle of operation, including a screenshot with the new switch to the Wiki, and link+announce it on the discussion list, right?

_Mark

janm012012:
Hi  @markmaker!

You too, a happy new year!

I've fixed the braces/logic.

Yes, I'll update the wiki and put an announcement/warning to the list.

Jan

janm012012:
@markmaker. can you please hint me where the job processor and its configuration is discussed on the wiki? I can't find the page to update yet...

markmaker:
> @markmaker. can you please hint me where the job processor and its configuration is discussed on the wiki? I can't find the page to update yet...

There is no obvious page. You can always add a new one, and link it. A new page can either be added to the TOC, or linked as an "in-depth" page, perhaps from here:

https://github.com/openpnp/openpnp/wiki/User-Manual#running-a-job

Perhaps move some of the existing related content over to the new page. I've done it a few times.

Note: the [User Manual](https://github.com/openpnp/openpnp/wiki/User-Manual) page would ideally be split up like that. But I see that would be a lot of work. :crossed_fingers: :smirk:  

_Mark

markmaker:
And big thanks! :100: :1st_place_medal:

PR #1576, opened 2023-07-22
Job Placements UI enhancement

janm012012:
# Description
This PR enhances the Job Placements table by immediately updating the Status if a placement is enabled or disabled using the mouse.

# Justification
It's inconvenient to change the row after a placement has been enabled or disabled until the status is updated. By studding the code I learned, that the space bar could be used for that purpose as well and that there the Status is immediately updated.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **using the debugger and test job**

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPlacementsPanel.java
@@ -265,11 +265,18 @@ public class JobPlacementsPanel extends JPanel {
         });
         table.addMouseListener(new MouseAdapter() {
             public void mouseClicked(MouseEvent mouseEvent) {
+                int row = table.rowAtPoint(new Point(mouseEvent.getX(), mouseEvent.getY()));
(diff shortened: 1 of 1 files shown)


markmaker:
I don't understand this PR. The way I interpret the description, it already works in the current version.

Similarly, I can't make any sense of the code change.

_Mark

markmaker:
Hi @janm012012,

duh, now I understand. with "status" you mean the **Status** column cell, not the enabled status (checkbox) itself. 

I guess the way it is done here is a bit of a hack, right? This should go through some property change event listener 🤓

But frankly, Java's / Swing's property change system is not clear to me in relation to tables. What I'm missing is a way listen to certain `firePropertyChange()` events _by class_ , and then update the table accordingly (Note, the event itself is already fired in the` setEnabled()` setter). But tables seem to have their own/unrelated event system... Short of registering a separate listener for each and every object in a table, I don't see a way.

So I guess this hack is okay. 🤭

_Mark

PR #1577, opened 2023-07-27
job exception handling enhancement

janm012012:
# Description
This PR enhances the job panel by selecting the board/placement/feeder/part that has thrown the exception.

# Justification
Its more convenient if the part/feeder/placement/board that cased an exception is already selected when acknowledging the error message.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **I've used the test job and let it run into all sorts of errors**

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -952,17 +952,16 @@ public class JobPanel extends JPanel {
              * both dispense and PnP this is not possible. Once dispense is removed we can include
              * the current placement in the thrown error and add this feature.
              */
-            if (t.getClass() == JobProcessorException.class) {
+            if (t instanceof JobProcessorException) {
                 JobProcessorException jpe = (JobProcessorException)t;
-                Object s = jpe.getSource();
-                Class<?> c = s.getClass();
+                Object source = jpe.getSource();
                 
                 // decode the source of the exception and try to select as much as possible
-                if (c == BoardLocation.class) {
-                    BoardLocation b = (BoardLocation)s;
+                if (source instanceof BoardLocation) {
+                    BoardLocation b = (BoardLocation)source;
                     Helpers.selectObjectTableRow(jobTable, b);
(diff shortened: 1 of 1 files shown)


markmaker:
Hi @janm012012,

good idea, just a short comment from vacation: please use `instanceof` instead of comparing classes. So objects can also be derivatives (like you already did with the feeders).

_Mark

markmaker:
Thanks @janm012012,

another small concern: For parts and feeders you switch the tab. But not for BoardLocations and Placements. Shouldn't this be symmetric?

_Mark

janm012012:
Hi @markmaker!

I had the same thoughts but decided against because for this PR handles job exceptions which are thrown as part of jobs and one can - as far as I know - only start jobs from from the job panel. So a board or placement related exception shall only happen with the jobs tab in view.

As this is just my view and my workflow, I'd be happy to change it, if you think it would be advantageous to do so.

Jan

markmaker:
> So a board or placement related exception shall only happen with the jobs tab in view.

Initially, yes, but if you have to troubleshoot repeatedly, then you might be elsewhere. For instance: the last error might have switched over to the Feeders tab. It could be very confusing if the actual error now points to a placement, but the Feeder tab is still on top and the feeder with the last error still selected. 

>  I'd be happy to change it, if you think it would be advantageous to do so.

Yes, please :+1: 

_Mark

janm012012:
Changed!

At some point you've to tell me how to start/resume jobs without focusing the jobs tab first...

Jan

markmaker:
Thanks, @janm012012 .

> At some point you've to tell me how to start/resume jobs without focusing the jobs tab first...

Well, good point but...

https://github.com/openpnp/openpnp/wiki/User-Manual#keyboard-shortcuts

Plus the user is always free to navigate tabs during the job. Maybe you want to look at camera settling diagnostics, or vacuum diagnostics, or the log. There's always some potential to tune :wink: 

Plus in software design it's just good style to _not_ build on such _external_ assumptions. :nerd_face: 

_Mark

PR #1578, opened 2023-07-30
Fix bug introduced by PR1547

tonyluken:
# Description
This change moves the initialization of board and panel locations from the BoardLocation default constructor to the addBoard and addPanel methods of JobPanel.

# Justification
Initialization of board locations in the BoardLocation default constructor can fail due to accessing of the machine configuration (to obtain the default board location) before the machine configuration is completely loaded. The consequence of this is that the Panels tab is always empty after a restart of OpenPnP and there will be erroneous errors reported in the log implying the *.panel.xml files are corrupted when in fact they are fine. This change moves the initialization to its proper place, namely where a new board or panel is being added to the job.

# Instructions for Use
No special instructions for operators.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **Verified the Panels tab is now correctly populated on a restart and no errors are reported in the log. Also verified that boards and panels that are added to a job still have their location initialized to the Default Board Location as desired.**
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. **Maven tests were run and all passed**

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -1136,6 +1136,7 @@ public class JobPanel extends JPanel {
     protected void addBoard(File file) throws Exception {
         Board board = new Board(configuration.getBoard(file));
         BoardLocation boardLocation = new BoardLocation(board);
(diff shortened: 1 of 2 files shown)


markmaker:
Thanks, @tonyluken

PR #1579, opened 2023-08-09
ImageCamera update

janm012012:
# Description
This PR adds an Offset feature to the ImageCamera to shift the image and added automatic scaling of the cameras default Units per Pixel setting with respect to the Viewing Scale parameter.

The offset can be used to align the image with a serious feature like the homing fiducial.

When Viewing Scale is changed, the cameras default Units per Pixel are scaled likewise to keep the camera scaling in sync with the image scaling. 

# Justification
When trying to use a picture of a real machine in simulation, the offset feature is required to align the image with the simulation machine. To further adjust the resolution of the simulated camera view, the Viewing Scale parameter can be used. If this parameter is changed, the cameras default Units per Pixel is scaled automatically to reflect the new image scale. This keeps all overlays and camera jogging intact with respect to the new scaling. Calibration or other simulation use cases are not effected because the adjustment is only done in the moment the Viewing Scale parameter is changed.

# Instructions for Use
To align the image with a simulation machine, move the camera to a serious location in a serious part of the image and use the Offset feature to shift the image as required. Assuming the pictures Units per Pixel where correctly set, other location in the image shall align now as well with the simulation machine.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **I've used a picture of my machine for testing**


markmaker:
Thanks, @janm012012,

Looks mostly good, just two things:

1. You use `imageOffset` as pixel coordinates, but it is typed as `Location`, i.e. storing length units internally. This will cause problems if a user switches system units, e.g. as soon as the `ImageCameraConfigurationWizard` is saved after that, the numeric values will jump and break the ImageCamera unexpectedly. This is all the more unfortunate, as I actually believe length unit offsets would be better, as the offsets would be easier to measure in real life, from the corner to the position seen at `0, 0`, referenced via photo. So why not just subtract the `imageOffset` from the location, _before_ transforming to pixels? This would also implicitly take care of system unit changes and subsequent unit conversions.
2. In `setSimulatedScale()` you now automatically change the Units per Pixel. As mentioned earlier, this is not the idea here: this setting is intended to test the calibration of the Units per Pixel, so setting them automatically would sort of defy the purpose. Furthermore, there are other dependencies, that you do not update (or invalidate), like the secondary Z units per pixel (3D calibration) and Advanced Camera Calibration, so this "half update" would be very confusing, and lead to malfunction. Regarding external up-scaling as the recommended alternative, [I will answer in the discussion group](https://groups.google.com/g/openpnp/c/ShlsMBApN5E/m/lHDbpaZyAAAJ).

Thanks :rocket:!

_Mark

janm012012:
Hi @markmaker!

Many thanks for the hint with the units. The offset is now applied onto the location before the conversion to pixels. That should provide a physical and comprehensible behavior now.

I don't like inconsistencies either and therefore reverted the units per pixel rescaling commit.

Jan

markmaker:
Thanks,  @janm012012,

I should have been clearer, `getX()` and `getY()` simply get the numeric coordinate value, ignoring the units. You should be using the `Location.subtract()` method, it guarantees units conversion. Like so:

```java
  Location imageLocation = location.subtract(imageOffset);
  double locationX = imageLocation.getX();
  double locationY = imageLocation.getY();
```

_Mark

janm012012:
Changed. I prefer the add() the offset...

markmaker:
@janm012012 you must use a different Location variable (like in the snippet I proposed), the `location` itself is still used for other purposes, where it must not be offset.

janm012012:
Changed. Was to late yesterday...

markmaker:
Thanks @janm012012, much appreciated!
:100: :1st_place_medal:

Issue #1451, opened 2022-07-23
ReferenceRotatedTrayFeeder not picking from correct location


infinitemachinery:
[machinesettings.zip](https://github.com/openpnp/openpnp/files/9174242/machinesettings.zip)
[OpenPnP.log](https://github.com/openpnp/openpnp/files/9174244/OpenPnP.log)
# Read This First
Issues must only be opened once they are confirmed.
Please [visit the discussion group](https://groups.google.com/group/openpnp) first, so bugs or missing features can be separated from other problems.

To submit an issue for OpenPnP you must use this template or it will be deleted. Please fill out either the Problem Report or Feature Request section below and delete the other section.

-----------------------------------------------------------------------

# Problem Report
## Describe the Issue
I have a ReferenceRotatedTrayFeeder defined, 2 rows and 20 columns. I have defined the beginning position, last position on row 1 and last position on row 2, as required by the tray definition.

The problem I'm having is that when I pick the first part from the tray, rather than picking from row 1 from column 1 it picks from column 1 row 2. I have the current feed count set to 0.

![Screenshot 2022-07-23 100942 RotatedTray](https://user-images.githubusercontent.com/71993878/180615531-49da459d-5888-47b1-bcce-96479f8ec9e4.png)

## Steps to Reproduce

1. Start OpenPnP
2. Connect to machine and home.
3. Load a job
4. In the Feeders tab create a ReferenceRotatedTrayFeeder with the necessary rows and columns. Establish the A, B and C points of tray using camera vision.
5. Establish the part for the feeder.
6. Set Feed Count to 0 
7. Place a part in position row 1, column 1 of tray.
8. Perform feed and pick on the feeder, or alternatively place a part from feeder using the job.

## Expected Result
Part is picked from beginning row and column of the feeder, row 1, column 1.

## Actual Result
Part is picked from row 2, column 1 of feeder.

-----------------------------------------------------------------------

tonyluken:
@infinitemachinery,

Can you try the latest test branch? This problem should be fixed - more info available in the PR mentioned above.

Tony

tonyluken:
Since I haven't heard anything, I'm going to close this issue. If you are still having problems, you can always reopen it or open a new issue.

PR #1581, opened 2023-08-17
Feature/fix rotated tray feeder

tonyluken:
# Description
This change fixes the issue documented below.
- [x] #1451

Other deficiencies corrected and other changes:
- Trays only worked within +/-90 degrees of the machine's X-axis.  They now work correctly at any arbitrary angle.
- Trays only worked if points A, B and C went clockwise around the tray. They now work correctly if the points go counter-clockwise around the tray as well.
- Trays can now be defined with a single row and single column.
- GUI fields now update automatically during feed operations.
- Additional/better error checking and validation was added.
- All GUI strings were externalized.
 
# Justification
Corrects a known bug and makes the Rotated Tray Feeder work as expected.

# Instructions for Use
One of the fields in the GUI has changed from _Component Rotation_ to _Component Rotation in Tray_. Prior to this change, the angle shown in this field was the actual pick angle for the component. Now this field displays the rotation of the component relative to the tray (which will almost always be one of 0, +/-90, or 180). The pick angle is now computed from the _Component Rotation in Tray_ plus the _Tray Rotation_. Note - existing Rotated Tray Feeders will automatically update the _Component Rotation in Tray_ field so that the pick angle remains the same as it was prior to this change.

Note that when the feed count is 0 or 1, the current pick location is the first component in the tray - this is now consistent with the behavior of other feeders.

# Implementation Details
1. How did you test the change? **Setup various different trays at various different angles on the machine and observed feed and pick operations.**


markmaker:
Thanks, @tonyluken,

This looks mostly good, just some remarks/questions:

- Nice upgrade handler. :+1: 
- Github shows inconsistent indentation in the diff, I guess there are tabs dragged in from earlier versions, please replace with all-spaces in changed sections, as per [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style).
-  Could be an old bug: `calculatePickLocation()` uses Millimeters without conversion, but I see you use system units when determining the `location` in `calculateOffsetsAndRotation()`. Needless to say, this would better use `Location.rotateXy()` and other `Location` arithmetic instead of messing with `sin()` and `cos()` and unit conversions itself.
- I wonder if you could move the "divide and modulo" thing into `calculatePickLocation()` to avoid code duplication. Instead of `partX`, `partY`, just pass the feed count.
- I see you rightfully wrinkle your nose where the rows/columns are swapped depending on which is shorter. I wonder if you could remove that and instead _convert_ the stored feed count in the upgrade handler. Now that the rotation of the tray has become meaningful, and `componentRotationInTray` must be understood by the user, similar to "rotation in tape", i.e., in relation to a clear "up" direction (which I guess is looking at a rows/columns in "reading order"), this would IMHO be the right thing to do. See also [here](https://github.com/openpnp/openpnp/wiki/ReferenceStripFeeder#rotation-in-tape).

_Mark

markmaker:
Oh, I was too quick to send this. :sweat_smile: 

The upgrade handler cannot just convert the feed count, as parts were already picked from the tray in the wrong pattern, of course. Even rotating `offsets`  in the upgrade handler does not work, as the swapped order is left-handed. It's not that easy. 

Perhaps a legacy mode boolean?

_Mark

tonyluken:
Ok, I think I addressed all of your concerns. Note that Component Rotation in Tray is relative to the A->B axis of the tray and there is a tool tip that states that. Hopefully that's clear enough.

markmaker:
Very neat, thanks! :100:

PR #1583, opened 2023-08-19
Save raw cropped image size and use it for diagnostics display instead of checking the image size

tonyluken:
# Description
This change adds two new attributes to AdvancedCameraCalibration to save the image width and height that were in effect when calibration was run. This allows the Calibration Results/Diagnostics to display correctly even if the camera is no longer connected to OpenPnP.

Also moved setting of the camera's defaultZ out of applyCalibrationToMachine to CalibrationSolutions just ahead of the call to applyCalibrationToMachine. This still sets defaultZ automatically when calibration is run via Issues&Solutions but allows it to be changed by the operator when running calibration manually.

Additionally, added calls to firePropertyChange in all AbstractCamera setters so that any GUI fields bound to those will properly update.

# Justification
Aids in off-line debugging.

# Instructions for Use

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **Verified that the camera Calibration Results/Diagnostics display correctly regardless of the state of the camera connection.**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **AbstractCamera was touched but only to add calls to firePropertyChange in all existing setters.**

### DIFF
--- a/src/main/java/org/openpnp/gui/processes/CalibrateCameraProcess.java
+++ b/src/main/java/org/openpnp/gui/processes/CalibrateCameraProcess.java
@@ -182,6 +182,9 @@ public abstract class CalibrateCameraProcess {
 
         advCal = ((ReferenceCamera)camera).getAdvancedCalibration();
         
+        advCal.setRawCroppedImageWidth(pixelsX);
+        advCal.setRawCroppedImageHeight(pixelsY);
+        
         testPattern3dPointsList = new ArrayList<>();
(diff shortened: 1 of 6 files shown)


markmaker:
Very fast :rocket: and cool :sunglasses:, @tonyluken! Thanks a lot.

PR #1597, opened 2023-12-22
Fix erroneous reset of placement transform when cells are selected for editing but are not changed

tonyluken:
# Description
This change fixes a bug where the placement affine transform was erroneously being cleared when the operator selected cells in the Job table for editing but did not actually change anything. It also makes the shading of the location cells visible (which indicates the status of the affine transform) even when the row in the Job table is selected.

# Justification
Fixes a bug and makes the affine transform status visible even when the row is selected.

# Instructions for Use

# Implementation Details
1. How did you test the change? **Duplicated the issue and after the code changes observed that the affine transform no longer gets reset when one of the location cells is selected for editing, but nothing is changed. Also checked that the affine status shading is now visible even when the row is selected.**

### DIFF
--- a/src/main/java/org/openpnp/gui/support/LengthCellValue.java
+++ b/src/main/java/org/openpnp/gui/support/LengthCellValue.java
@@ -71,6 +71,14 @@ public class LengthCellValue implements Comparable<LengthCellValue> {
         this.length = length;
     }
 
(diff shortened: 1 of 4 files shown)


markmaker:
Thank you @tonyluken !

PR #1598, opened 2024-01-08
Allow children of panels to be replaced

tonyluken:
# Description
This change allows the children of panels to be replaced thus making it easier to setup panels with different board variants.

# Justification
Anyone who uses panels can make use of this change.

# Instructions for Use
On the Panels tab, select a panel to edit, select one or more of its children (if more than one is selected, they must all be defined by the same board or subpanel), right-click on one of the selected children, and select the Replace child(ren)... option from the pop-up menu. This will open a dialog box where a different board or subpanel can be selected as a replacement. Select the appropriate replacement board or panel and click Ok. The selected children will be replaced by the new selection. All the children's locations and other settings will remain the same.

# Implementation Details
1. How did you test the change? **Created various different board variants and made copies of a panel using one of the board variants and then used the instructions above to change the board variant on each of the different copies of the panel.** 
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **Changes were made to PlacementsHolder.java and Panel.java in the `org.openpnp.model` package to support the change.**
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. **All Maven tests were run and passed**


markmaker:
Thanks, @tonyluken !

PR #1600, opened 2024-02-17
Send FeedRate and Acceleration on change only

janm012012:
# Description
This PR changes the GCodeDriver to only send Acceleration and FeedRate if their values change by more then 0.1%.

# Justification
This significantly reduces the amount of unnecessary GCode send to the driver and hence increases the overall performance.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. - I've verified the operation using a simulation machine and carefully inspecting the logs.
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. - No
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. - Yes


markmaker:
Hi @janm012012 

thanks for taking a swing at that. 

Some feed-back:

1.  Jerk should be handled the same way.
1. Please introduce new Variables `FeedRateF` and `AccelerationF` and `JerkF` that are always set, analog to the "forced" coordinates. Some users may still want to always set rates for some reason (controllers that don't reliably remember rates, canned cycles etc. that may have change the rates behind the back of OpenPnP). 
1. The `currentFeedRate` and `currentAcceleration` and `currentJerk` should be reset on homing, and on connect.
1. The absolute difference should not be hard-coded (`1e-3`). There can be very different driver units involved and correspondingly very different absolute numbers. You could use a fraction of `vMin`, `aMin`, `jMin` that are already obtained there. You could refactor the calculation used here... use something like `pow(10, -digitsToExpress(vMin))`?:
    https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/machine/reference/solutions/GcodeDriverSolutions.java#L886
1. While you're at it: please check if there is actually an old bug present in GcodeDriver [here](https://github.com/openpnp/openpnp/blob/ace6dd799d56c255d7750ca45678743e0715242d/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java#L762). `Feedrate` should be per minute, but `Acceleration`, `Jerk` should be per second (squared, cubed). I guess `vMin` should be applied `vMin*60`.

_Mark

janm012012:
Hi @markmaker!

I've extended the implementation by moving all send-on-change related code into a new class. This class is now also used for Jerk and addressed in the entire GCodeDriver class. It contains a configurable enable/disable/force option that can be controlled using the UI now to actually enable the send-on-change behavior. The default is the old behavior to always send FeedRate Acceleration and Jerk.

The detection what "change" is, is and was a relative change between new and last/old value. The default relative change is 0.001. This delta can be configured using machine.xml. Its not linked to the UI at present. IMHO 0.001 is a good value with respect to physics.

Unfortunately this implementation does not work yet: I'm initializing sendOnChangeFeedRate, sendOnChangeAcceleration and sendOnChangeJerk using a string that represents the key to be replaced in commands. On first pass, this works. However, when importing the machine.xml a new instance is created internally without this string argument. As a result the substitution later on does not work anymore. How can I fix that? (adding a method to set the string later is the only option I currently see, but I don't like it because the string can not be final then.)

Concerning the code you wont me to check I'm a little bit lost because I don't know what units the different variables are in. IMHO using SI units inside would be a good choice converting them as required just before generating the command send to the driver. "feedRate" is loaded using a method that contains "PerMinute" suggesting it something like x/min. In the same line feedRate is converted using driverUnitsFactor, which has no units as well. vMin is basically from ControllAxis.getMotionLimit() which is documented as "AxesLocation units (mm) per second^n". So I think that there is indeed a problem with the units. The minimum modification I'd suggest is to comment the intended units the variables shall have, convert vMin to the units of feedRate, apply Math.max and finally convert the result to driverUnits.

Jan

markmaker:
Hi @janm012012,

looks better and better.

> The detection what "change" is, is and was a relative change between new and last/old value. The default relative change is 0.001. This delta can be configured using machine.xml. Its not linked to the UI at present. IMHO 0.001 is a good value with respect to physics.

That's valid reasoning and okay for me. 

Another solution would be to allow the user to _specify_ the relative change on the UI, _instead_ of the checkbox. Setting the relative change to 0.0000 then disables the optimization.

> ... extended the implementation by moving all send-on-change related code into a new class...

That's nice, but any reason why you did not implement the `"F"` variables I mentioned in my first response (point 2)? It can still be useful, for instance in HeadMountable specific G-code.

> Unfortunately this implementation does not work yet...

That's always difficult. Especially when you need to cover all eventualities, i.e. loading from XML, creating a new driver from the UI, and creating/converting a driver through I&S. It is very important this still works. 

I recommend strictly using getters for the members and then initialize them on first use when null (i.e. don't initialize in the member declaration).

> IMHO using SI units inside would be a good choice converting them as required...

That's exactly how it works.

>  I'm a little bit lost ...

No. Looking at your code changes, I see you found all the bugs, including two I did not see before. :sweat_smile: 

Please make these changes:

- Do the reset only _after_ connecting, and _after_ homing. Why? The `CONNECT_COMMAND` and `HOME_COMMAND` might contain G-code that changes rates (implicitly or explicitly). 
- To be safe, also add the reset to `processPositionReport()` here:
   https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java#L1430
    You might only do the reset, if the position actually _changed_ visavis what OpenPnP thinks it is. It is a signal that something happened behind the back of OpenPnP, for instance Contact Probing, which might also have changed the rates, or the user having used a hand wheel etc.

_Mark

janm012012:
Hi @markmaker!
> > The detection what "change" is, is and was a relative change between new and last/old value. The default relative change is 0.001. This delta can be configured using machine.xml. Its not linked to the UI at present. IMHO 0.001 is a good value with respect to physics.
> 
> That's valid reasoning and okay for me.
> 
> Another solution would be to allow the user to _specify_ the relative change on the UI, _instead_ of the checkbox. Setting the relative change to 0.0000 then disables the optimization.
> 
Ok, then keep it as is for now. I'll be happy to change it if requested.

> > ... extended the implementation by moving all send-on-change related code into a new class...
> 
> That's nice, but any reason why you did not implement the `"F"` variables I mentioned in my first response (point 2)? It can still be useful, for instance in HeadMountable specific G-code.
> 
I'm not sure I understand what you meant: add variables like "FeedRateF" that would be always replaced in G-Code commands? I've added that now.

> > Unfortunately this implementation does not work yet...
> 
> That's always difficult. Especially when you need to cover all eventualities, i.e. loading from XML, creating a new driver from the UI, and creating/converting a driver through I&S. It is very important this still works.
> 
> I recommend strictly using getters for the members and then initialize them on first use when null (i.e. don't initialize in the member declaration).
> 
I've added a setter for the string and used it at connect() to initialize the send-on-change class.

> > IMHO using SI units inside would be a good choice converting them as required...
> 
> That's exactly how it works.
> 
Unfortunately that seems to be not the case. mm/sec as documented is not SI. m/s would be the one to use. (https://www.nist.gov/pml/owm/metric-si/si-units)

> Please make these changes:
> 
>     * Do the reset only _after_ connecting, and _after_ homing. Why? The `CONNECT_COMMAND` and `HOME_COMMAND` might contain G-code that changes rates (implicitly or explicitly).
> 
changed.

>     * To be safe, also add the reset to `processPositionReport()` here:
>       https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java#L1430
>       
>       You might only do the reset, if the position actually _changed_ visavis what OpenPnP thinks it is. It is a signal that something happened behind the back of OpenPnP, for instance Contact Probing, which might also have changed the rates, or the user having used a hand wheel etc.
> 
I've added it unconditionally for now. How would I detect if the position is not in sync anymore?

After this changes I'm faced with test error. How can I find out whats causing this errors. The console just report

[ERROR] [1;31mErrors: [m
[ERROR] [1;31m  GcodeDriverTest.before:82 » NullPointer[m
[ERROR] [1;31m  GcodeDriverTest.before:82 » NullPointer[m
[ERROR] [1;31m  GcodeDriverTest.before:82 » NullPointer[m
[ERROR] [1;31m  GcodeDriverTest.before:82 » NullPointer[m
[ERROR] [1;31m  SampleJobTest.testSampleJob:119 » NullPointer[m
[ERROR] [1;31m  SamplePanelizedJobTest.testSampleJob:106 » NullPointer[m

I thought it might be related to the initialization of the strings to replace variables, but can't find out how the test is executed without connecting and homing the machine. Adding the initialization to both of them does not fix it.

Jan

markmaker:
> Unfortunately that seems to be not the case. mm/sec as documented is not SI. 

True, but still SI _base_ unit. Note: [NIST RS274NGC section 2.1.2.9](https://www.nist.gov/publications/nist-rs274ngc-interpreter-version-3) says either millimeters or inches. That's where this comes from.

> I've added it unconditionally for now. How would I detect if the position is not in sync anymore?

Something like this:

   ```java
for (ControllerAxis axis : position.getAxes(driver)) {
      if (!axis.coordinatesMatch(
            position.getLengthCoordinate(axis), 
            axis.getDriverLengthCoordinate())) {
            // reset
            ...
            break;
       }
}
```

> After this changes I'm faced with test error

You removed initialization of the variables altogether, the whole `SendOnChange` objects are now missing, not just the "strings" inside them. Actually, that's good but you then need getters for these and on-demand-initialize them, which also means you can set the right "string". Then also remove the `@Element` annotation, no serialization needed/wanted. This getter design is much more robust and in turn you can get rid of the separate "string" setters.

Btw. could you rename `string` to `variable` to be more consistent with `substituteVariable()` nomenclature?

_Mark

janm012012:
Hi @markmaker!

With getters for sendOnChange<x> and initialing the object if needed, all tests are passed again.

I'm merged your suggestion on detecting position changes.

Removing the @Element decoration is - to my limited understanding - not possible, because there are two members insight the SendOnChange class that need to be saved to and loaded from machine.xml. Unfortunately the machin.xml-loading still requires that the variable has to be set before first use (in connect). Any suggestion on that? (I only see that the configuration could be stored in the GCodeDriver class and hand over to the SenOnChange class on creation and if changed.)

Jan

markmaker:
> Removing the `@Element` decoration is - to my limited understanding - not possible, because there are two members insight the `SendOnChange` class that need to be saved

Ah yes.

But what happens if you now annotate `variable` with `@Attribute`? 

Do you still need clunky `sendOnChangeSetAllVariables()` and  `SendOnChange.setVariable()`?

_Mark

janm012012:
> But what happens if you now annotate `variable` with `@Attribute`?
> 
Then the code does not work anymore. When trying to write machine.xml an error pops up that the elements can not be saved.

As fast as I can see the problem is twofold:

1. The test code runs without the serializer (simpleXML) leaving the send-on-change objects uninitialized requiring this getter-initilize-on-first-use magic to work.
2. The main code initializes the send-on-change objects using the serializer which calls a constructor without arguments requiring the setVariable() magic to set required variables.

Both are "Krücken" (crutches) to me.

I guess one would have to change the test code to use the serializer to test the same way the user sees OpenPnP and the serializer needs some method to use a specific constructor. I've looked around in the simpleXML documentation but could not find a way to define the default value.

Jan

janm012012:
The point is, that sendOnChange requires - as I've design it - the variable string to replace it in commands. This string is given either using a constructor or via a separate setter. It is not marked as @Attribute or @Element and therefore not serialized/de-serialized and hence has a default value after loading machine.xml.

My tests have shown, that OpenPnP can run fine without the getter/initilize-on-first-use magic because the de-serializer has created sendOnChangeFeedRate and its brothers. Unfortunately it has used a no-arg-constructor so that the required variable is unset. To fix that, the code calls the setVariable() method from connect().

If I run the test code (mvn test) I've seen, that the code does not work without the getter/initialize-on-first-use magic but without the no-arg-constructor. This to me looks like the test code is not de-serializing and creating the sendOnChangeFeedRate and brothers members. This can either happen because the de-serialializer did not run at all or that the machine.xml used for testing does not contain any reference to send-on-change and hence does not create the object. (My machine.xml I debugged with, does contain it.)

I actually don't like the getter/initialize-on-first-use magic and I also don't like that the variables have to be set/fixed later on. However, you said that you favor initialize-on-first-use over initialization as part of the creation of the GCodeDriver object.

Unfortunately I don't have a good solution for the missing variable initialization the de-serializer leaves behind.

Jan

janm012012:
> But what happens if you now annotate `variable` with `@Attribute`?
> 
> Do you still need clunky `sendOnChangeSetAllVariables()` and `SendOnChange.setVariable()`?
> 
I'm sorry, I did not read this carefully enough.
 
Annotating `variable` as `@Attribute` works. Now its value is stored in machine.xml (not so nice) but the setVariable() method is not needed anymore (nice).

I think, thats a good compromise.

Jan

janm012012:
>     1. When updating from an earlier version of OpenPnP (from a `machine.xml` where it was not present yet).

Check using my simulation machine.xml by explicitly removing all send-on-change related stuff. Works.

>     2. When you manually create a new `GcodeDriver` from the UI (that's a trap I fell into more than once 😅) .

I tried by removing the existing driver in my simulation machine an adding a new GCodeAsyncDriver manually. After a lot of fiddling like linking the axis with the driver and assigning all the commands again, the machine is moving and never faced any problems.

>     3. The last one should also cover when I&S converts GcodeDriver into GcodeAsyncDriver, and vice versa. I hope 😇. If you want to be extra sure, you could test that too, from a fresh install.

Hm.. I don't know how to change back from GcodeAsyncDriver to GcodeDriver in my simulation machine...

Using a fresh installation, I can click around and replace the NullDriver with a GcodeDriver still talking to the Tcp-Server I can request to move and do not get any errors. Maybe there is no MOVE_TO_COMMAND set yet, but I&S does not yet suggest it.

> And one last wish: can you add a `Logger.trace()` in the `sendOnChangeResetAll()` I guess it would be useful for debugging, when users start to adopt that cool feature, and may at times wonder why unchanged rates are still sent.

Added.

Jan

markmaker:
Thanks, there's a typo in the log message. As an irregular verb it should be "reset" instead of "resetted". :nerd_face: 

Fun fact: ["resetted" seemed to exist](https://en.wiktionary.org/wiki/reset#Verb_2) in some obscure Scot law lingo and means "To receive and hide (stolen goods, or a criminal, etc.)" :smile:

markmaker:
Tests are good!

markmaker:
Thanks, @janm012012, please announce on the list! :100: :1st_place_medal:

PR #1604, opened 2024-03-04
Linux support for Neoden 4 cameras

jsphpl:
# Description
This PR adds Linux support for the Neoden 4 cameras using [libneodencam](https://github.com/jsphpl/libneodencam).

# Justification
Always good to support Linux, right?

# Instructions for Use
In order to use OpenPNP with the Neoden 4 camera driver on Linux, you need to build and install [libneodencam](https://github.com/jsphpl/libneodencam) as described in the linked repository. Everything else works the same as on Windows. I will update the Wiki with Linux instructions for Neoden 4 when this PR is merged. Until then, instructions can be found in the libneodencam repo.

# Implementation Details
The Neoden4CamDll interface is renamed to Neoden4CameraDriver because libneodencam.so exposes the same interface as the Windows dll.

Neoden4CameraHandler now has a `driver` property that is populated with the respective native library on initialization. If the OS is identified as Windows, `NeodenCamera.dll` is loaded. Otherwise we load `libneodencam.1.so`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -244,7 +244,7 @@
 	</dependencies>
 	<build>
 		<plugins>
-			<!-- Used to specify that the Maven Eclipse task should download sources 
+			<!-- Used to specify that the Maven Eclipse task should download sources
 				and JavaDocs and attach them to the Eclipse project. -->
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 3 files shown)


markmaker:
Thanks @jsphpl,

this seems great for Neoden4 users, be sure to announce this in the user group too! And thanks for updating the Wiki.

_Mark

PR #1606, opened 2024-03-08
Fix getPickLocation of ReferenceTrayFeeder

janm012012:
# Description
This PR fixes the getPickLocation method of the ReferenceTrayFeeder by always calculating it based on current UI values. In the previous implementation the location was static and only updated on feed operations. This is not ideal as there is no direct relation to the UI and its not obvious to the user.

# Justification
I tried to use a ReferenceTrayFeeder and failed because I was not able to verify its configuration prior use. Changing feedCount or offsets did not help to check the locations a part would be picked up from.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **I used a simulation machine and verified that UI changes are immediately reflected as well as feed and pick operations work as before.**

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceTrayFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceTrayFeeder.java
@@ -84,11 +84,7 @@ public class ReferenceTrayFeeder extends ReferenceFeeder {
         // and then add them to the location to get the final pickLocation.
         pickLocation = location.add(offsets.multiply(partX, partY, 0.0, 0.0));
         
-        Logger.debug(String.format("Location of part # %d, x %d, y %d, xPos %f, yPos %f, rPos %f",
-                feedCount, partX, partY, pickLocation.getX(), pickLocation.getY(),
-                pickLocation.getRotation()));
-
-        Logger.debug("{}.getPickLocation => {}", getName(), pickLocation);
+        Logger.debug("{}.getPickLocation => {}, part # {}, x {}, y {}", getName(), pickLocation, feedCount, partX, partY);
         return pickLocation;
     }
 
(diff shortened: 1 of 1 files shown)


markmaker:
Hi @janm012012,

functionality looks good. 

As always, one small thing left :smile: 

Logging (2 x) in a getter like `getPickLocation()` is probably too chatty. Move the logging to  `setFeedCount()` (which is also called by `feed()` where logging was before), so changing the feed count from elsewhere (e.g. the UI) would _also_ generate a helpful debug log. Just eval `getPickLocation()` there to do it.

_Mark

janm012012:
Changed! Thanks for reviewing!

markmaker:
Hmmm... please read again, in detail. I hope this was understandable...

janm012012:
Hi @markmaker!

I'm sorry for being not patient enough! Calling getPickLocation() from setFeedCount() requires to handle the exception, that getPickLocation can throw, which is "Tray empty".  I think its better to keep the exception in feed() (as it was initially) and do not generate it in getPickLocation. I've replaced it by a warning and limited feedCount to the tray size. What do you think?

Jan

markmaker:
The exception change is a good one.

But please move the logging to `setFeedCount()` to avoid logging on each getter evaluation. By calling `getPickLocation()` in the log message expression, you'll get a fresh calculation.

janm012012:
Do you mean like this?

markmaker:
Yes!

I reviewed the whole again, and I think you should name the feeder and part id in the "Tray empty" exception. Otherwise the user might not know which tray it refers to. 

Sorry, I haven't seen this before :sweat_smile:, this should now be my last remark...

_Mark

janm012012:
Changed!

While testing I found, that getName() for the part I used is null. Instead I now use getId() and changed the exception in takeBackPart() where I copied getName() from as well. What is getName() of parts used for? I don't see any reference in the UI explaining the difference between ID and Name.

If there is anything else you think should be changed, please let me know.

Jan

markmaker:
The `getName()` is just an unwanted class inheritance artifact, I guess.

Thanks for the good work! :100:

PR #1607, opened 2024-03-08
Fix getPickLocation of ReferenceRotatedTrayFeeder

janm012012:
# Description
This PR fixes the getPickLocation method of the ReferenceRotatedTrayFeeder by always calculating it based on current UI values. In the previous implementation the location was static and only updated on feed operations. This is not ideal as there is no direct relation to the UI and its not obvious to the user.

# Justification
I tried to use a ReferenceRotatedTrayFeeder and failed because I was not able to verify its configuration prior use. Changing feedCount or offsets did not help to check the locations a part would be picked up from.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **I've used a simulation machine and tried manual validation and feed/pick operations to work. Unfortunately the value entered in the Feed Count field never reaches the code and hence no update on that is visible. feed operations correctly increment it**

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceRotatedTrayFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceRotatedTrayFeeder.java
@@ -49,7 +49,7 @@ public class ReferenceRotatedTrayFeeder extends ReferenceFeeder {
     @Element
     private Location offsets = new Location(LengthUnit.Millimeters);
     @Attribute
-    private int feedCount = 0;
+    private int feedCount = 0;  // UI is base 1, 0 is ok because a pick operation always preceded by a feed, which increments feedCount to 1
 
     @Attribute(required=false)
     @Deprecated
@@ -65,8 +65,6 @@ public class ReferenceRotatedTrayFeeder extends ReferenceFeeder {
     protected Location lastComponentLocation = new Location(LengthUnit.Millimeters);
     @Element
     protected Location firstRowLastComponentLocation = new Location(LengthUnit.Millimeters);
-
-    private Location pickLocation;
 
     @Commit
     public void commit() {
@@ -97,31 +95,21 @@ public class ReferenceRotatedTrayFeeder extends ReferenceFeeder {
     }
 
     @Override
-    public Location getPickLocation() throws Exception {
-        if (pickLocation == null || feedCount == 0) {
-            //Normally the pick location is only valid after a feed operation. But in the case of no
-            //prior feed operation (feed count is zero), we set the pick location to the first part
-            //(which is also where it will be after the first feed operation).
-            int partX, partY;
-            int fc = feedCount;
-            if (pickLocation == null && feedCount > 0) {
-                //Since the pick location is null, the machine configuration must have just been
-                //loaded. And since the feed count is greater than 0, this feeder must have already
-                //performed a feed operation prior to the configuration being saved. This means we
-                //must back-up one feed count to point to the pick location for that prior feed 
-                //operation.
-                fc = fc - 1;
-            }
+    public Location getPickLocation() {
+        int feedCountBase0 = feedCount -1; // UI uses feedCount base 1, the following calculations are base 0
 
-            calculatePickLocation(fc);
+        // if feedCound is currently zero, assume its one
+        // this can happen if the pickLocation is requested before any feed operation
+        // return first location in that case
+        if (feedCount == 0) {
+            feedCountBase0 = 0;
         }
-
-        Logger.debug("{}.getPickLocation => {}", getName(), pickLocation);
-
-        return pickLocation;
-    }
-
-    private void calculatePickLocation(int feedCount) throws Exception {
+        // limit feed count to tray size
+        else if (feedCount > (trayCountCols * trayCountRows)) {
+            feedCountBase0 = trayCountCols * trayCountRows -1;
+            Logger.warn("{}.getPickLocation: feedCount larger then tray, limiting to maximum.", getName());
+        }
+        
         //The original version of this feeder fed along either the rows or columns depending on
         //which was shorter. This version now feeds along a row until it is empty and then it moves
         //to the next row. However, if an old version of the feeder was just loaded and it was
@@ -131,13 +119,13 @@ public class ReferenceRotatedTrayFeeder extends ReferenceFeeder {
         if (legacyPickingInProgress && (trayCountCols >= trayCountRows)) {
             //Pick parts along a column (stepping through all the rows) until it is empty and then 
             //move to the next column
-            rowNum = feedCount % trayCountRows;
-            colNum = feedCount / trayCountRows;
+            rowNum = feedCountBase0 % trayCountRows;
+            colNum = feedCountBase0 / trayCountRows;
         } else {
             //Pick parts along a row (stepping through all the columns) until it is empty and then
             //move to the next row (the new default)
-            rowNum = feedCount / trayCountCols;
-            colNum = feedCount % trayCountCols;
+            rowNum = feedCountBase0 / trayCountCols;
+            colNum = feedCountBase0 % trayCountCols;
         }
 
         //The definition of the tray has row numbers increasing in the negative y direction so that
@@ -145,7 +133,7 @@ public class ReferenceRotatedTrayFeeder extends ReferenceFeeder {
         Location delta = offsets.multiply(colNum, -rowNum, 0, 0).rotateXy(location.getRotation()).
                 derive(null, null, null, componentRotationInTray);
         
-        pickLocation = location.addWithRotation(delta);
+        return location.addWithRotation(delta);
     }
 
     public void feed(Nozzle nozzle) throws Exception {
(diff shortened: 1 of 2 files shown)


janm012012:
As mentioned in the description, the _Feed Count_ field still behaves erratic: values entered are immediately used in the UI to update the _Components remaining_ field but _Reset_ and _Apply_ are not enabled and the value is never send to the code. Any hints how to fix that are appreciated. Thank You!

markmaker:
Hi @janm012012,

regarding the `pickLocation()`, are you sure this wasn't already working before? 

Unlike the non-rotated `ReferenceTrayFeeder`, it _always_ seems to recalculate.

If it is only about the UI not updating, use proper 

```java
firePropertyChange("propName", oldValue, prop);
```

See `setFeedCount()` for an example.

This will likely need multiple firings of dependent calculated properties.

_Mark

tonyluken:
Hi @janm012012,

Can you explain precisely what steps you were performing and what it was doing versus what you expected it to do? Perhaps a video would help me understand the problem better. I thought I had pretty thoroughly tested the ReferenceRotatedTrayFeeder and found it to be working as I expected.

Thanks,
Tony

janm012012:
Hi Tony!

Very well! I run out of MCUs on tape and decided to use some left overs in a strip I taped to an edge on the machine (covering the sprocket holes). While trying to set that up as tray, I configured the first and last part and tried to verify my setup using the "move the camera to the selected feeder's current pick location." button (second to right in the row above the feeder table on the feeder tab). I then noticed, that I did not setup the rotation in tray. (I verify that using a footprint overlay). Changing the rotation was not reflected by the mentioned button. I then wondered if I have setup the correct spacing. To verify that, I changed Feed Count and pressed the button again. What ever value I entered, the location did not change.
Looking into the code I found, that the getPickLocation() returns a static variable, that is only updated on feed (and once when the feeder was never feed before). After changing the code to always calculate the pickLocation, almost all UI changes are immediately reflected by the button. "Almost" because changing the Feed Count is still not reflected. I just checked it again, and today it works. What ever went wrong yesterday... When I enter a digit, I immediately see the setFeedCount() method called. Thats a little unexpected, because for all other values I've to press "Apply" first.
From todays perspective, my objection is invalid and I don't see any pending issues anymore.
Please have a look if you can reproduce my findings. If you think it should stay as it is or it should be done different, please let me know.

Jan

tonyluken:
@janm012012,

I looked at this some more and this is how I'd fix it:

1) Undo all your other changes.
2) In ReferenceRotatedTrayFeeder remove the throws Exception from the calculatePickLocation method (nothing in that method throws an exception).
3) Add a new public method to the RefererenceRotatedTrayFeeder called setFeedCountWithLocationUpdate like this:

```
    public void setFeedCountWithLocationUpdate(int feedCount) {
        int oldValue = this.feedCount;
        this.feedCount = feedCount;
        if (feedCount == 0) {
            legacyPickingInProgress = false;
        }
        firePropertyChange("remainingCount", trayCountRows*trayCountCols - oldValue, 
                trayCountRows*trayCountCols - feedCount);
        calculatePickLocation(Math.max(0, feedCount - 1));
    }
```

Note: It is important to not call firePropertyChange("feedCount", ...) in this new method or otherwise the property change events will attempt to modify the GUI at the same time as the operator is typing in the feed count field.

4) In ReferenceRotatedTrayConfigurationWizard, change the bindings from:

```
        bind(UpdateStrategy.READ_WRITE, feeder, "feedCount", textFieldFeedCount, "text", intConverter); //$NON-NLS-1$ //$NON-NLS-2$

```
to:

```
        bind(UpdateStrategy.READ, feeder, "feedCount", textFieldFeedCount, "text", intConverter); //$NON-NLS-1$ //$NON-NLS-2$
        bind(UpdateStrategy.READ_WRITE, feeder, "feedCountWithLocationUpdate", textFieldFeedCount, "text", intConverter); //$NON-NLS-1$ //$NON-NLS-2$

```

That will result in the proper updating of the feed count in the GUI when the feeder feeds (due to the first binding) and also lets the operator set the feed count in the GUI and have the pick location automatically update to correspond with it (due to the second binding).

Tony

janm012012:
Hi @tonyluken!

Many thanks for your review and suggestions. I've implemented and tested them. The pickLocation is now updated on every change of feedCount. However, it not updated anymore if any other parameter like column or row offset or count is changed. Would you please advice how that could be solved? (I still believe that the pickLocation should always point to the current location based on the UI. If you don't like that, we might add a remark to the UI indicating that the pickLocation in invalid until the next feed operation.)

Jan

tonyluken:
I don't have too much time to spend on this, but I think you could just add to the wizard's validateInput method a call to 

`feeder.setFeedCountWithLocationUpdate(feeder.getFeedCount());`

Tony

markmaker:
Can somebody quickly tell me what the difference is vs. the `ReferenceTrayFeeder`, i.e., why so complicated? 
(apart from rotation)

tonyluken:
I think the main complication is that the ReferenceRotatedTrayFeeder has fields that are computed based on other fields. If you want to preserve the functionality of the Revert button, it can't just set those fields in the feeder in order to get a correct pick location calculation. The pick location should only be calculated after the Apply button is clicked. Now that I think about it, I think my suggestion of adding

`feeder.setFeedCountWithLocationUpdate(feeder.getFeedCount());
`

to the wizard's validateInput method is not correct. It needs to be added to the wizard's updateModel method (I think that is what it is called) after super.updateModel() is called instead.

Tony

markmaker:
Thanks @tonyluken,

> The pick location should only be calculated after the Apply button is clicked. 

Not sure I understand correctly. Nothing should be applied or be _effective_ on the feeder, before the Apply button is pressed. So if you press the Feed&Pick toolbar button before pressing Apply, it should still go to the old location. Not very intuitive, but that's how OpenPnP works since ages.

If there are function buttons directly on the Wizard, so the user assumes the latest settings are effective, then in their Action, the Apply should be done implicitly, as the first thing. In my feeder classes, I'm doing this all the time.

```java
applyAction.actionPerformed(e);
// now do the feeder action
...
```            

_Mark

janm012012:
> [...] Nothing should be applied or be _effective_ on the feeder, before the Apply button is pressed. So if you press the Feed&Pick toolbar button before pressing Apply, it should still go to the old location. Not very intuitive, but that's how OpenPnP works since ages.
> 
Thats the issue I came across and the reason why I'm trying to change the behavior with this PR. In the old design nothing is updated until a feed-action is performed. This is/was not indicated anywhere.

Now, with the latest changes, feedCount is immediately (without Apply) effective and all other parameters are only effective after apply and a feed action.

I'd be happy to help making this feeder's UI more consistent.  Please just indicate how to do that. My proposal would be to change the binding for feedCount such, that apply/reset are effective. Then _components remaining_ would be updated only after apply. Finally I'd add the suggested updateModel() behavior to trigger a recalculation of the pickLocation on apply. I think that will provide the consistent behavior of the UI and limit the update of pickLocation when needed.

Jan

markmaker:
>  Please just indicate how to do that. 

I guess the `ReferenceTrayFeeder` and `ReferenceTrayFeederConfigurationWizard` can serve as examples.

The `feedCount` is simply bound like everything else:

https://github.com/openpnp/openpnp/blob/694f0b89a016a9bbf570fec212204891142863f8/src/main/java/org/openpnp/machine/reference/feeder/wizards/ReferenceTrayFeederConfigurationWizard.java#L149

and the `setFeedCount()` fires a property change event:

https://github.com/openpnp/openpnp/blob/694f0b89a016a9bbf570fec212204891142863f8/src/main/java/org/openpnp/machine/reference/feeder/ReferenceTrayFeeder.java#L165-L170

It doesn't even store the pick location, it just calculates it whenever needed, guaranteed to take fresh data into consideration. That's the proper design in my eyes.

_Mark

tonyluken:
Don't get me wrong, I'm all for simplifying and making it consistent with the other feeders.

janm012012:
Ok, I've now synced the modifications with thus of PR #1606 I submitted to fix the same getPickLocation issue on the ReferenceTrayFeeder. According to my tests, the UI now behaves almost as expected: any modification has first to be applied and still generates errors in case of inconsistencies. The pick location is always correct. However, feedCount is not always valid. It seems that there is a race condition between UI -> setFeedCount() and getFeedCount() -> validateInput() -> setFeedCount(). Sometimes feedCound is limited by nCols*nRows (wizard, line 572). I think it would be nice to have this fixed as well. Do you know why/how this race condition could be solved?

Jan

tonyluken:
I think the problem is the call to feeder.setFeedCount in the wizard's validation method. I think the way to fix is to add a new int variable to the wizard to hold the feedCount that is displayed in the wizard (see how it is done with nRows and/or nCols, don't forget to add a getter and setter for the new variable). Then add a binding that binds that int variable to the feed count text field (again see how it is done with nRows and/or nCols, the order here is important, the new binding must occur before the wrapped binding of the feeder's feedCount to the text field). Then change the validation method to use the wizard's setFeedCount method rather than the feeder's setFeedCount method.

Tony

janm012012:
Many thanks @tonyluken! That works very well now!

Jan

tonyluken (approved):
All these latest changes look good to me.

markmaker:
Thanks, @janm012012 !

:100:

markmaker:
Oh I see it is still marked as WIP, I can't merge it.

Is it ready?

janm012012:
I just changed the status back to open.

PR #1609, opened 2024-03-09
Skip interrupting to unload a nozzle tip while loading a new tip

janm012012:
# Description
This PR enhances the usability for machines with manual nozzle tip changing enabled by skipping the interruption to load the nozzle tip if a load will immediately follow.

# Justification
At present the job processor generates two interrupts to change one nozzle tip, one for the unload and a second for the load. The unload is useless and a waste of time if a load is known to immediately follow.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **using a simulation machine**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **yes: I added a new version of unloadNozzleTip() which can taken an argument to signal that a load will follow.**

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -995,14 +995,16 @@ public class JobPanel extends JPanel {
                 }
             }
             
-            MessageBoxes.errorBox(getTopLevelAncestor(), 
-                    Translations.getString("JobPanel.JobRun.Error.ErrorBox.Title"), t.getMessage()); //$NON-NLS-1$
+            // update the state before showing the error to allow exceptions with continuations to change it
             if (state == State.Running || state == State.Pausing) {
                 setState(State.Paused);
             }
             else if (state == State.Stopping) {
                 setState(State.Stopped);
             }
+            
+            // call showError() to support exceptions with continuation
+            UiUtils.showError(getTopLevelAncestor(), Translations.getString("JobPanel.JobRun.Error.ErrorBox.Title"), t); //$NON-NLS-1$
         });
     }
 
--- a/src/main/java/org/openpnp/gui/support/MessageBoxes.java
+++ b/src/main/java/org/openpnp/gui/support/MessageBoxes.java
@@ -50,8 +50,9 @@ public class MessageBoxes {
         Logger.debug("{}: {}", title, cause);
         message = message.replaceAll("\n", "<br/>");
         message = message.replaceAll("\r", "");
-        message = message.replaceAll("<", "&lt;");
-        message = message.replaceAll(">", "&gt;");
+        // FIXME: why are this replacements here? They make html in message void and even the <br> above.
+        //message = message.replaceAll("<", "&lt;");
+        //message = message.replaceAll(">", "&gt;");
         message = "<html><body width=\"400\">" + message + "</body></html>";
 
         // if this errorBox shall ask for Continuation, show a ConfirmDialog and return if the user selected YES
--- a/src/main/java/org/openpnp/gui/support/MessageBoxes.java
+++ b/src/main/java/org/openpnp/gui/support/MessageBoxes.java
@@ -31,8 +31,9 @@ import org.pmw.tinylog.Logger;
 public class MessageBoxes {
 
 
-    public static void errorBox(Component parent, String title, Throwable cause) {
+    public static boolean errorBox(Component parent, String title, Throwable cause, boolean withContinuation) {
         String message = null;
+        boolean ret = false;
         if (cause != null) {
             message = cause.getMessage();
             if (message == null || message.trim().isEmpty()) {
@@ -52,7 +53,19 @@ public class MessageBoxes {
         message = message.replaceAll("<", "&lt;");
         message = message.replaceAll(">", "&gt;");
         message = "<html><body width=\"400\">" + message + "</body></html>";
-        JOptionPane.showMessageDialog(parent, message, title, JOptionPane.ERROR_MESSAGE);
+
+        // if this errorBox shall ask for Continuation, show a ConfirmDialog and return if the user selected YES
+        if (withContinuation) {
+            ret = JOptionPane.showConfirmDialog(parent, message, title, JOptionPane.OK_CANCEL_OPTION, JOptionPane.ERROR_MESSAGE) == JOptionPane.OK_OPTION;
+        } else {
+            JOptionPane.showMessageDialog(parent, message, title, JOptionPane.ERROR_MESSAGE);
+        }
+        
+        return ret;
+    }
+
+    public static void errorBox(Component parent, String title, Throwable cause) {
+        errorBox(parent, title, cause, false);
     }
 
     public static void errorBox(Component parent, String title, String message) {
--- a/src/main/java/org/openpnp/gui/support/MessageBoxes.java
+++ b/src/main/java/org/openpnp/gui/support/MessageBoxes.java
@@ -30,6 +30,16 @@ import org.pmw.tinylog.Logger;
 
 public class MessageBoxes {
 
+    // prepare message for use in a message box
+    static String prepareMessage(String message) {
+        if (message == null) {
+            message = "";
+        }
+        message = message.replaceAll("\n", "<br/>");
+        message = message.replaceAll("\r", "");
+        message = "<html><body width=\"400\">" + message + "</body></html>";
+        return message;
+    }    
 
     public static boolean errorBox(Component parent, String title, Throwable cause, boolean withContinuation) {
         String message = null;
@@ -50,9 +60,7 @@ public class MessageBoxes {
         Logger.debug("{}: {}", title, cause);
         message = message.replaceAll("<", "&lt;");
         message = message.replaceAll(">", "&gt;");
-        message = message.replaceAll("\n", "<br/>");
-        message = message.replaceAll("\r", "");
-        message = "<html><body width=\"400\">" + message + "</body></html>";
+        message = prepareMessage(message);
 
         // if this errorBox shall ask for Continuation, show a ConfirmDialog and return if the user selected YES
         if (withContinuation) {
@@ -73,9 +81,7 @@ public class MessageBoxes {
             message = "";
         }
         Logger.debug("{}: {}", title, message);
-        message = message.replaceAll("\n", "<br/>");
-        message = message.replaceAll("\r", "");
-        message = "<html><body width=\"400\">" + message + "</body></html>";
+        message = prepareMessage(message);
         JOptionPane.showMessageDialog(parent, message, title, JOptionPane.ERROR_MESSAGE);
     }
 
@@ -84,9 +90,7 @@ public class MessageBoxes {
             message = "";
         }
         Logger.debug("{}: {}", title, message);
-        message = message.replaceAll("\n", "<br/>");
-        message = message.replaceAll("\r", "");
-        message = "<html><body width=\"400\">" + message + "</body></html>";
+        message = prepareMessage(message);
         return JOptionPane.showConfirmDialog(parent, message, title, JOptionPane.YES_NO_OPTION) == JOptionPane.YES_OPTION;
     }
 
@@ -94,9 +98,7 @@ public class MessageBoxes {
         if (message == null) {
             message = "";
         }
-        message = message.replaceAll("\n", "<br/>");
-        message = message.replaceAll("\r", "");
-        message = "<html><body width=\"400\">" + message + "</body></html>";
+        message = prepareMessage(message);
         JOptionPane.showMessageDialog(MainFrame.get(), message, title, JOptionPane.INFORMATION_MESSAGE);
     }
 
--- a/src/main/java/org/openpnp/gui/support/MessageBoxes.java
+++ b/src/main/java/org/openpnp/gui/support/MessageBoxes.java
@@ -48,11 +48,10 @@ public class MessageBoxes {
             message = "No message supplied.";
         }
         Logger.debug("{}: {}", title, cause);
+        message = message.replaceAll("<", "&lt;");
+        message = message.replaceAll(">", "&gt;");
         message = message.replaceAll("\n", "<br/>");
         message = message.replaceAll("\r", "");
-        // FIXME: why are this replacements here? They make html in message void and even the <br> above.
-        //message = message.replaceAll("<", "&lt;");
-        //message = message.replaceAll(">", "&gt;");
         message = "<html><body width=\"400\">" + message + "</body></html>";
 
         // if this errorBox shall ask for Continuation, show a ConfirmDialog and return if the user selected YES
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
@@ -38,9 +38,7 @@ import org.openpnp.spi.Camera.Looking;
 import org.openpnp.spi.CoordinateAxis;
 import org.openpnp.spi.Feeder;
 import org.openpnp.spi.HeadMountable;
-import org.openpnp.spi.JobProcessor.JobProcessorException;
-import org.openpnp.spi.JobProcessor.ManualLoadException;
-import org.openpnp.spi.JobProcessor.ManualUnloadException;
+import org.openpnp.spi.JobProcessor;
 import org.openpnp.spi.MotionPlanner.CompletionType;
 import org.openpnp.spi.Nozzle;
 import org.openpnp.spi.NozzleTip;
@@ -50,13 +48,28 @@ import org.openpnp.spi.base.AbstractNozzle;
 import org.openpnp.util.MovableUtils;
 import org.openpnp.util.SimpleGraph;
 import org.openpnp.util.UiUtils;
-import org.openpnp.util.UiUtils.ExceptionWithContinuation;
 import org.pmw.tinylog.Logger;
 import org.simpleframework.xml.Attribute;
 import org.simpleframework.xml.Element;
 import org.simpleframework.xml.core.Persist;
 
 public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMountable {
+    public static class ManualUnloadException extends JobProcessor.JobProcessorException {
+        private static final long serialVersionUID = 1L;
+    
+        public ManualUnloadException(Object source, Throwable throwable) {
+            super(source, throwable, true);
+        }
+    }
+
+    public static class ManualLoadException extends JobProcessor.JobProcessorException {
+        private static final long serialVersionUID = 1L;
+    
+        public ManualLoadException(Object source, Throwable throwable) {
+            super(source, throwable, true);
+        }
+    }
+
     @Element
     private Location headOffsets = new Location(LengthUnit.Millimeters);
 
@@ -605,6 +618,9 @@ public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMoun
             throw new Exception("Can't load incompatible nozzle tip.");
         }
 
+        // detect if an unload is required:
+        // either if the current nozzle has a different tip loaded
+        // or if the tip to be loaded is on a different nozzle.
         ReferenceNozzle unloadNozzle = this;
         if (nt.getNozzleWhereLoaded() != null) {
             // Nozzle tip is on different nozzle - unload it from there first.  
@@ -613,14 +629,15 @@ public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMoun
         ntUnload = unloadNozzle.getNozzleTip();
 
         boolean unloadSwallowed = false;
-        try {
-            unloadNozzle.unloadNozzleTip();
-        } catch (JobProcessorException e) {
-            // if the exception is a ManualUnloadException, remember and swallow it.
-            if (e instanceof ManualUnloadException) {
+        
+        // if there is any nozzle tip requiring unload, do it now
+        if (ntUnload != null) {
+            try {
+                unloadNozzle.unloadNozzleTip();
+            } catch (ManualUnloadException e) {
+                // if the exception indicates a manual unload, remember and swallow it.
+                // we'll later throw a combined unload and load exception.
                 unloadSwallowed = true;
-            } else {
-                throw e;
             }
         }
 
@@ -708,8 +725,7 @@ public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMoun
                 message += " a manual nozzle tip " + nt.getName()+" load on nozzle "+getName()+" now. ";
                 message += "When you press OK, the task will continue.";
                 throw new ManualLoadException(this, 
-                        new UiUtils.ExceptionWithContinuation(message,  () -> { loadNozzleTipFinish(nozzleTip); }),
-                        true);
+                        new UiUtils.ExceptionWithContinuation(message,  () -> { loadNozzleTipFinish(nozzleTip); }));
             }
         }
 
@@ -810,16 +826,15 @@ public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMoun
             waitForCompletion(CompletionType.WaitForStillstand);
             throw new ManualUnloadException(this, 
                     new UiUtils.ExceptionWithContinuation("Task interrupted: Please perform a manual nozzle tip "+nt.getName()+" unload from nozzle "+getName()+" now. "
-                            + "You can then resume/restart the interrupted task.", () -> { loadNozzleFinal(); }),
-                    true);
+                            + "You can then resume/restart the interrupted task.", () -> { unloadNozzleTipFinish(); }));
         }
 
-        loadNozzleFinal();
+        unloadNozzleTipFinish();
     }
     
     // this is the final part of the nozzle tip unload procedure and a separate method 
     // to be called either directly or as continuation of a manual nozzle tip change
-    private void loadNozzleFinal() throws Exception {
+    private void unloadNozzleTipFinish() throws Exception {
         ensureZCalibrated(true);
 
         // May need to calibrate the "unloaded" nozzle tip stand-in i.e. the naked nozzle tip holder. 
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
@@ -57,8 +57,8 @@ public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMoun
     public static class ManualUnloadException extends JobProcessor.JobProcessorException {
         private static final long serialVersionUID = 1L;
     
-        public ManualUnloadException(Object source, Throwable throwable) {
-            super(source, throwable, true);
+        public ManualUnloadException(Object source, String message) {
+            super(source, message, true);
         }
     }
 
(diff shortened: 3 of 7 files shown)


markmaker:
Hi @janm012012, 

I'm a it reluctant here. :sweat_smile: I hope this long text does not completely scare you away...

The message box tells a user _which_ nozzle tip is being unloaded, so they can store the tip in its properly labelled storage thingy. Users might not be able to reliably tell the tips apart just from looking at them, especially when using the machine only sporadically. A confusion could really cause havoc down the road (for _both_ of the subsequently confused tips).

It is clear that this manual loading should be improved, but I propose a completely different, more general approach:

There is a worse problem when loading the tip: there is no after-loading calibration possible, as the exception interrupts the load operation here:

https://github.com/openpnp/openpnp/blob/c70d2e4a5f9ec3d0bf74fa72dc7fe207c05b9e81/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java#L688-L704
 
In a job it is not so bad, as the tip _should_ re-calibrated when the job is resumed. But for manual testing (during machine setup) this is not possible. And it might lead to problems and much frustrated hair-pulling when suddenly operations are failing due to missing run-out and/or bottom camera offsets and/or bottom vision background calibration. 

A better feature would be to somehow daisy-chain an operation after _confirming_ the exception message box that was used to prompt for a manual user action. 

Like so: 

1. When deliberately throwing the exception prompting the user to unload/load the nozzle tip, one could register a follow-up action (`Thrunnable`) on the exception. 

1. Make an Exception sub-class e.g. `ExceptionWithContinuation` with a `Thrunnable` as payload.

1. Then in  [`UiUtils.showError()`](https://github.com/openpnp/openpnp/blob/c70d2e4a5f9ec3d0bf74fa72dc7fe207c05b9e81/src/main/java/org/openpnp/util/UiUtils.java#L69C24-L69C33) you could detect that `Throwable t` is an instance of `ExceptionWithContinuation` and execute the stored `Thrunnable` using `submitUiMachineTask()`. 

1. The thrown exception text must then obviously warn the user that pressing `OK` will execute the continuation action.

1. In such a case, the `UiUtils.showError()` message box must have a `Cancel` option too, so the user can opt out of the continuation action.

1. Then to use this new facility for nozzle tip loading:

1. Remove the `nozzle.unloadNozzleTip();` in the JobProcessor, instead directly call `nozzle.loadNozzleTip(nozzleTip);`. 
1. This will automatically call the unload from the `loadNozzleTip()`, as you can [see there](https://github.com/openpnp/openpnp/blob/c70d2e4a5f9ec3d0bf74fa72dc7fe207c05b9e81/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java#L613). 
1. Inside the `unloadNozzleTip()`, on manual unload throw a specific Exception sub-class e.g. `ManualUnloadException`.
1. Back in the `loadNozzleTip()` catch (only) the `ManualUnloadException` and re-throw with a continuation to load the nozzle tip.
1. Likewise, at the end of the `loadNozzleTip()`, throw a `ManualLoadException` with a continuation to calibrate the nozzle tip (as configured). 

So with a manual nozzle tip change machine you would get the full sequence: unload, load, calibrate. 

Inside a job one would then still need to resume the job, but this is probably fine and consistent with the "play/pause/stop" paradigm. 

Alternatively, if you want full automatism, catch the `ManualUnloadException` and/or `ManualLoadException` inside the JobProcessor and rethrow with a daisy-chained continuation of a) the (optional) Thrunnable that was already registered and b) executing a Job resume.

Phew... :smile: 

_Mark

janm012012:
Many thanks Mark! Thats a lot of input. I need to get familiar with exception handling and will then try to implement your suggestions. Continuing the job immediately after loading the nozzle sounds promising too, maybe with a little configuration option...

Jan

janm012012:
Hi @markmaker!

I started implementing your suggestions. Would you please have a look what I did so far?

- ExceptionWithContinuation is a sub-class of JobProcessorException because when loading/unloading a nozzleTip interrupting = true is given.

In principle the continuation works. I can already unload, load and calibrate a nozzleTip.

The Cancel-Option for the continuation is missing and unloading from a different nozzle also. And the integration is not done. There are still multiple interrupts for a single nozzleTip change.

Jan

markmaker:
Hi @janm012012,

this is going in the right direction. :+1: 

Can you explain why you made this a sub-class of the `JobProcessorException`? 

Personally, I would rather keep this [concern separate](https://en.wikipedia.org/wiki/Separation_of_concerns) in `UiUtils` and define a `UiUtils.ExceptionWithContinuation` class. These continuations will also work and be useful outside the `JobProcessor`.

Also I would rename the `thrunnable` member to `continuation`, so it is clearer what purpose it has. Also make it `protected`, so it can be set in any subclass.

I so see some entanglement with the `JobProcessorException` when exceptions are repackaged, here:

https://github.com/openpnp/openpnp/blob/694f0b89a016a9bbf570fec212204891142863f8/src/main/java/org/openpnp/spi/JobProcessor.java#L43-L49

So the continuation would be lost. You need to somehow propagate the continuation, like it is already done with the `interrupting` flag.

But to solve this, I would make `JobProcessorException` a subclass of `UiUtils.ExceptionWithContinuation`, not the other way around. The continuation being `null` means there is none (you already handle that case correctly). 

Then change the repackaging constructor to something like the following, so the continuation does not get lost:

```java

    public class JobProcessorException extends UiUtils.ExceptionWithContinuation {
        ...
        public JobProcessorException(Object source, Throwable throwable) {
            super(getThrowableMessage(throwable), throwable);
            this.source = source;
            if (throwable instanceof UiUtils.ExceptionWithContinuation) {
                this.continuation = ((UiUtils.ExceptionWithContinuation) throwable).getContinuation();
            }
            if (throwable instanceof JobProcessorException) {
                this.interrupting = ((JobProcessorException) throwable).isInterrupting();
            }
        }
```

To construct `JobProcessorException` with `continuation`s directly, you probably need one more constructor.

_Mark

markmaker:
Oh, I just realized that these exceptions are being linked when repackaged. It calls the `super` constructor with the original `throwable`. This goes into the `cause` of the super class `Throwable`.

So one can simply call `getCause()` on a `Throwable` to get original cause.

So forget what I said above about propagating in the constructor and inheriting from `UiUtils.ExceptionWithContinuation`. The other things I said about keeping the concern separate in `UiUtils`, still apply.

Instead of propagating, do this in `UiUtils`:

```java
        // FIXME: add Cancel button and  only execute on OK
        // Go through all causes, creating a combined continuation
        Thrunnable combinedContinutation = null;
        for (Throwable cause = t; cause != null; cause = cause.getCause()) {
            if (cause instanceof ExceptionWithContinuation) {
                Thrunnable continuation = ((ExceptionWithContinuation)cause).getContinuation();
                if (continuation != null) {
                    if (combinedContinutation = null) {
                        combinedContinutation = continuation; // just take the first one
                    } else {
                        combinedContinutation = (() -> { 
                            // combine the two
                            continuation.thrun(); // inner first
                            combinedContinutation.thrun(); // outer last
                            return null;
                        });
                    }
                }
            }
        }
        if (combinedContinutation != null) {
            submitUiMachineTask(combinedContinutation);
        }
```

This way, one can easily combine multiple continuations, like we spoke about:

    Unload > Load > Calibrate > Continue Job

Just repackage the exceptions inside each other. Note, an exception does not need to be thrown and caught in order to be repackaged. You could do something like this:

```java
throw new JobProcessorException(nozzle, 
    new UiUtils.ExceptionWithContinuation("Please flip the omelette in the pan. "
      + "When you press OK, the frying will go one", 
         () -> { 
            continueFrying(pan); 
        }));

```

This way the `JobProcessorException` can remain a completely [separate concern](https://en.wikipedia.org/wiki/Separation_of_concerns), and you don't need inheritance and yet another constructor :smile: 

True to the [Composition over Inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance) principle.

_Mark

markmaker:
Sent too quickly, sorry. :sweat_smile: 

This will not work:

```java
// combine the two
continuation.thrun(); // inner first
combinedContinutation.thrun(); // outer last
```
it would have to catch the exceptions all over again. Hmmm...

Interesting but difficult problem in-deed!

_mark

janm012012:
Hi @markmaker!

I continued implementing your suggestions. Now the unload/load/calibrate cycle is working with a single exception. The combine unload/load message is also implemented. Missing is the linking of continuations (currently that was not needed yet) and the automatic job resume. Would you please have an other look? Thank You!

Jan

markmaker:
Looking better and better!

Just had a quick look (not saying I caught everything):

- the `loadNozzleFinal()` example seems good, as far as I see, this can become the pattern.
- in the case of the (comment) `// Nozzle tip is on different nozzle - unload it from there first`, the continuation would just call to `loadNozzleTip()` again. But do not catch JobProcessorException, catch the `ManualUnloadException` directly, and no `instanceof` needed.
- move the `ManualUnloadException`, and `ManualLoadException` to the `ReferenceNozzle` where they belong regarding _concern_.

_Mark

janm012012:
Hi @markmaker!

Many thanks for reviewing! I've added you suggestions.

What do you think, shall be put the automatic job resume feature (which shall be possible now) into a separate PR or add it here? (at present,  don't know how to continue a job from machine task...)

Jan

markmaker:
Hi @janm012012.

> What do you think, shall be put the automatic job resume feature (which shall be possible now) into a separate PR

Yes, I guess it is good enough to support _one_ continuation for now. This also changes the focus and goal of this PR slightly (see point 3, below). 

Some feed-back:

1. A comment says:
    `// detect if an unload is required:`
    ` // either if the current nozzle has a different tip loaded`
    ` // or if the tip to be loaded is on a different nozzle.`
    This is wrong. These condition can **both** be true at the same time, and **two** unloads may be needed.

2. This means you must revert that part of the changes. Simply catch the `ManualUnloadException` on both unloads, and allow both to continue. No need to restructure the original code otherwise, just a `try{} catch {}` around the original `unloadNozzleTip()` call. See point 6, how this is then recorded.

3. If I understand your latest changes correctly, the goal has slightly changed. You want to tell the user **the whole list of required manual interventions** for a nozzle tip change **at once**. So this could be up to two nozzle tip unloads (see point 1) and one nozzle tip load for them to do. That makes sense. 

4. But we need to be aware of one caveat. An _automatic_ nozzle tip changer needs accurate motion of the bare nozzle, so it can gently and accurately connect with the nozzle tip, changer slot rails etc. Therefore, we need `unloadNozzleTipFinish()` to make sure the bare nozzle is run-out and Z calibrated. But for a _manual_ changer, we can assume that is **not** needed, i.e., the user can do it by eye-hand coordination. 

5. Therefore we don't need the `unloadNozzleTipFinish()` and it does not matter, if it is being _suppressed_ (as your current code does). But you should add a note of this (comment) where you catch the `ManualUnloadException` (both places). And also in `unloadNozzleTipFinish()`, so when somebody adds a crucial function there later, they are aware of this caveat.

6.  I propose, instead of setting `unloadSwallowed` and  `unloadNozzle` (now 2 x) you should instead just compose the exception `message` from top to bottom, initialize with basic prompt at the top and then add user instructions in the `ManualUnloadException` handlers where you also have direct access to the nozzle and nozzle tip names to be specified  in the message. Rename `message` into `manualChangeInstructions` or similar and it becomes very clear what you are doing. If no such condition arises, the string just never gets used.

7. I suggest you are more specific in the messages: instead of `"When you press OK, the task will continue."` and  `"You can then resume/restart the interrupted task."`, say what it really does e.g. `"When you press OK, the nozzle tip will be calibrated (if enabled)."`. and  `"When you press OK, the bare nozzle will be calibrated (if enabled)."` With the limitation of not queuing continuations, this is all we can promise for now.

8. There is a pre-existing bug I just noted (mine :sweat_smile:): after `setNozzleTip(nt)` it calls `ensureZCalibrated(true)` directly. This is wrong with manual changing. Physically, there is still the _old_ nozzle tip loaded when this happens. Please move  `ensureZCalibrated(true);` to `loadNozzleTipFinish()`.

I guess that is a simple and yet solid implementation. We still get asserted `loadNozzleTipFinish()` out of it and that's a very good thing.

_Mark

janm012012:
Hi @markmaker!

Many thanks for the details feedback. Thats very much appreciated.

> Some feed-back:
> 
>     1. A comment says:
>        `// detect if an unload is required:`
>        ` // either if the current nozzle has a different tip loaded`
>        ` // or if the tip to be loaded is on a different nozzle.`
>        This is wrong. These condition can **both** be true at the same time, and **two** unloads may be needed.
> 
You're right. Now I understand why there might be two unload required.

>     2. This means you must revert that part of the changes. Simply catch the `ManualUnloadException` on both unloads, and allow both to continue. No need to restructure the original code otherwise, just a `try{} catch {}` around the original `unloadNozzleTip()` call. See point 6, how this is then recorded.
>
done.

>     3. If I understand your latest changes correctly, the goal has slightly changed. You want to tell the user **the whole list of required manual interventions** for a nozzle tip change **at once**. So this could be up to two nozzle tip unloads (see point 1) and one nozzle tip load for them to do. That makes sense.
> 
I'm sorry if it sounds this way. I did not intended to make more/other changes then just combining the unload with the load exception. That was most annoying on my last runs. Actually, we already have extended the focus by fixing problems in the current design by adding the continuation stuff. If you think, we shall split this PR into two to separate the issues or if there is any other change, you think does not belong here, please tell me and I'll try to adapt that.

>     4. But we need to be aware of one caveat. An _automatic_ nozzle tip changer needs accurate motion of the bare nozzle, so it can gently and accurately connect with the nozzle tip, changer slot rails etc. Therefore, we need `unloadNozzleTipFinish()` to make sure the bare nozzle is run-out and Z calibrated. But for a _manual_ changer, we can assume that is **not** needed, i.e., the user can do it by eye-hand coordination.
> 
The Unload exception has this as continuation. Maybe this did never work due to lack of combined continuations, but I at least intended to fix that together with the calibrate-after-load issue.

>     5. Therefore we don't need the `unloadNozzleTipFinish()` and it does not matter, if it is being _suppressed_ (as your current code does). But you should add a note of this (comment) where you catch the `ManualUnloadException` (both places). And also in `unloadNozzleTipFinish()`, so when somebody adds a crucial function there later, they are aware of this caveat.
> 
I've reverted this part as well so that unloadNozzleTipFinish() is not present anymore and the code is again part of unloadNozzleTip() as before.

>     6. I propose, instead of setting `unloadSwallowed` and  `unloadNozzle` (now 2 x) you should instead just compose the exception `message` from top to bottom, initialize with basic prompt at the top and then add user instructions in the `ManualUnloadException` handlers where you also have direct access to the nozzle and nozzle tip names to be specified  in the message. Rename `message` into `manualChangeInstructions` or similar and it becomes very clear what you are doing. If no such condition arises, the string just never gets used.
> 
changed.

>     7. I suggest you are more specific in the messages: instead of `"When you press OK, the task will continue."` and  `"You can then resume/restart the interrupted task."`, say what it really does e.g. `"When you press OK, the nozzle tip will be calibrated (if enabled)."`. and  `"When you press OK, the bare nozzle will be calibrated (if enabled)."` With the limitation of not queuing continuations, this is all we can promise for now.
> 
The first part is changed now, the second part not: As I've just removed the continuation (5) the bare nozzle calibration code will never be executed. Em is missing anything?

>     8. There is a pre-existing bug I just noted (mine 😅): after `setNozzleTip(nt)` it calls `ensureZCalibrated(true)` directly. This is wrong with manual changing. Physically, there is still the _old_ nozzle tip loaded when this happens. Please move  `ensureZCalibrated(true);` to `loadNozzleTipFinish()`.
> 
Moved.

Jan

janm012012:
Hi @markmaker!

> This will not work:
> 
> ```java
> // combine the two
> continuation.thrun(); // inner first
> combinedContinutation.thrun(); // outer last
> ```

I may have an idea how to implement that: We could use a heap (queue operated as last-in-fist-out), iterate through the throwable inserting all continuations and drain the queue executing them all. If any of them throws a new exception with continuation, it gets inserted and executed next.
To implement that (in a new PR focusing on automatic job resume after manual nozzle tip change) I'd need to know if you think this could be the way to go, which class would be appropriate providing the heap feature and where to instantiate this one unique heap for the entire application.

Jan

markmaker:
Hi @janm012012,

regarding your first comment (about the points).

I get the feeling that my response came across  too "negative" to you, especially point 3. Please read it carefully, and note how I **agree** with your change of focus in the end, I specifically said _"That makes sense."_ :smile: 

Similar for points 4 and 5, I get the feeling your reaction seems not to reflect, that I actually **agree** with you :smile: , and just want you (and future developers) to be aware of the caveat of suppressing the unload continuation.

In the end, I find your code changes result in the desired solution, and I also agree with your conclusion of always _consistently_ suppressing the bare nozzle calibration when automatic changing is disabled.

There is one issue with the `// FIXME: why are this replacements here? They make html in message void and even the <br> above.` Clearly the _order_ of the `replaceAll` calls is wrong (a pre-existing bug), but the replacing is needed ([Java generics](https://en.wikipedia.org/wiki/Generics_in_Java) in stack traces). Please put it back, just replace the angle brackets first.

Re your last comment (combining continuations), I fear it would become very complicated, very quickly.

I propose the following: instead of splitting functions endlessly up in "before" and "after" and trying to keep track of the current work still left to do, just make it one function that is "interruptible-restartable" and ultimately "[idempotent](https://en.wikipedia.org/wiki/Idempotence)".

Example: 

The `loadNozzleTip()` already (mostly) works that way. It looks at the _state_ of the nozzles/nozzle tips and then works towards loading the wanted nozzle tip. It will unload up to two nozzle tips and then load the wanted tip. If it is interrupted, you can still call it again, it will continue where it left of, not repeating the already done steps.

The only thing it currently does not handle the right way, is calibration. At the top it checks whether the nozzle tip is already the desired one and immediately returns. Instead it should a) reset the nozzle calibration immediately after calling `setNozzleTip()` if recalibration is wanted on change, and then b) do the recalibration if not yet done, when calling `loadNozzleTip()` again, even if the tip is already the right one.

This way the `loadNozzleTip()` would be fully "interruptible-restartable" and ultimately "idempotent". This means that calling it again and again will always do the right thing (including ultimately nothing). 

Once we have established this "interruptible-restartable" / "idempotent" pattern, we can always just register the whole function as the continuation (e.g. no need for a separate `loadNozzleTipFinish()`). 

More generally speaking, using that pattern you can make any function call  "interruptible-restartable" and ultimately "idempotent". 

The same is already the case in the JobProcessor. It will always go through the nozzle tip loading and calibration step after a pause. This is also "interruptible-restartable" and ultimately "idempotent". So you could actually just wrap a `try {} catch {}` around `loadNozzleTip()` in the JobProcessor and register a "resume job" as the continuation. But you would have to make sure the exception message, which has the recorded `manualChangeInstructions`, is properly propagated.

I leave it to you whether you want to do this in this PR. I'm happy either way.


Apart from these, as far as I can tell, this PR can be merged. 

I assume you tested it thoroughly, also that it does _not_ change anything for users with the automatic changer _enabled_. Simulation with some log analysis is sufficient.

I suggest you make a test with the case where there are two unloads and one load. It would be nice to post a screen-shot of such a MessageBox giving these three instructions as a showcase for how this PR helps users without automatic changer. Instead of interrupting the user three times, there is now only one interruption!

Plus it also calibrates the nozzle tip afterwards, which is very important, not only for bottom vision and placement accuracy, but also to ensure fresh background calibration for auto-tuning bottom vision, which can be decisive whether vision actually works or not. Early in machine setup, Users typically play around with the GUI buttons, i.e., _outside jobs_, and your PR now ensures calibration is done, and some frustrating "hair pulling" will be a thing of the past.

All-in-all this PR a big gain in my eyes! You should then also announce it on the group with that screen-shot. :100: 


_Mark

janm012012:
Hi @markmaker!

> Hi @janm012012,
> 
> regarding your first comment (about the points).
> 
> I get the feeling that my response came across too "negative" to you, especially point 3. Please read it carefully, and note how I **agree** with your change of focus in the end, I specifically said _"That makes sense."_ 😄
> 
It's alright with me! I've to say sorry for not reading your commends carefully enough. I'm not a native speaker and I started shifting work into off-peek hours (to keep the ball rolling). This in combination makes me likely not a good negotiation partner. However, I very much appreciate all your support and reviewing and continuous discussions. There is tremendous amount to learn for me. And at then end, there is likely always a little dispute between us two: you have to focus more on code quality and general usability while I'm trying to get features in, I'm interested in the most. Joining both is IMHO a challenge...

> Similar for points 4 and 5, I get the feeling your reaction seems not to reflect, that I actually **agree** with you 😄 , and just want you (and future developers) to be aware of the caveat of suppressing the unload continuation.
> 
> In the end, I find your code changes result in the desired solution, and I also agree with your conclusion of always _consistently_ suppressing the bare nozzle calibration when automatic changing is disabled.
> 
I thinks, once combined continuations are supported, we shall make it take the same path whatever changer was used. I don't know what features will follow and if there might be the need to add stuff there. It shall be very easy to add then. For now - as a user of a manual nozzle tip changer - I agree that this can be safely skipped. (I also added code that shall put a warning in the logs in the unlikely event that someone relies on that even thought this feature never worked.)

> There is one issue with the `// FIXME: why are this replacements here? They make html in message void and even the <br> above.` Clearly the _order_ of the `replaceAll` calls is wrong (a pre-existing bug), but the replacing is needed ([Java generics](https://en.wikipedia.org/wiki/Generics_in_Java) in stack traces). Please put it back, just replace the angle brackets first.
> 
I temporarily disabled them while experiment to display the list of manual change instructions as unordered list. The indention and separation is nice, however, the bullet is rendered very badly and it's size and distance to the text does not match.

I've fixed and pushed that already.

> Re your last comment (combining continuations), I fear it would become very complicated, very quickly.
> 
I'll put that on my list and try it. Maybe, with some code to review, a clean and easy solution pops up...

> I propose the following: instead of splitting functions endlessly up in "before" and "after" and trying to keep track of the current work still left to do, just make it one function that is "interruptible-restartable" and ultimately "[idempotent](https://en.wikipedia.org/wiki/Idempotence)".
> 
> Example:
> 
> The `loadNozzleTip()` already (mostly) works that way. It looks at the _state_ of the nozzles/nozzle tips and then works towards loading the wanted nozzle tip. It will unload up to two nozzle tips and then load the wanted tip. If it is interrupted, you can still call it again, it will continue where it left of, not repeating the already done steps.
> 
I'll change that! That sounds very good. We'll then continue loadNozzleTip() with itself...

[...]
> The same is already the case in the JobProcessor. It will always go through the nozzle tip loading and calibration step after a pause. This is also "interruptible-restartable" and ultimately "idempotent". So you could actually just wrap a `try {} catch {}` around `loadNozzleTip()` in the JobProcessor and register a "resume job" as the continuation. But you would have to make sure the exception message, which has the recorded `manualChangeInstructions`, is properly propagated.
> 
Concerning the interruptable JobProcessor: That is a very handy feature that can not be mentioned often enough! It makes recovering from errors like feeder, pick or alignment so easy... (I recently added a new feeder and just resumed the job. Unfortunately I don't remember if the second nozzle even already had a part loaded while doing so...)

> I leave it to you whether you want to do this in this PR. I'm happy either way.
> 
I'll provide an other PR as we'll need continuation chaining for the calibration-after-load to co-exist with the job resume. Otherwise one or the other will be swallowed. In addition, the exception messages might require joining to inform the user that the job will continue if acknowledged. But ok, maybe the general remark thats now already in there, would be sufficient here too.

> Apart from these, as far as I can tell, this PR can be merged.
> 
That's good to hear! However, I'd like to make load/unloadNozzleTip "interruptible-restartable" first as it would make the code cleaner and simplify the continuation request in the exception.

I'm also considering refactoring the sequences for automatic tip loading/unloading into separate functions so that the loadNozzleTip() would either call automaticUnloadNozzleTip() or manualUnloadNozzleTip() and do the required work there. This shall increase the readability and group common items closer together. If you don't like that, please feel free to merge straight way what we have or tell me and I'll revert/stop that.

> I assume you tested it thoroughly, also that it does _not_ change anything for users with the automatic changer _enabled_. Simulation with some log analysis is sufficient.
> 
Unfortunately I've never used automatic nozzle tip changing and therefore did not tested that. While designing the code, I thoroughly made sure to preserve thus sections. I'll try to setup some automatic nozzle changers and test that.

> I suggest you make a test with the case where there are two unloads and one load. It would be nice to post a screen-shot of such a MessageBox giving these three instructions as a showcase for how this PR helps users without automatic changer. Instead of interrupting the user three times, there is now only one interruption!
> 
This is how it currently looks like:
UnloadLoadNozzleTips

> Plus it also calibrates the nozzle tip afterwards, which is very important, not only for bottom vision and placement accuracy, but also to ensure fresh background calibration for auto-tuning bottom vision, which can be decisive whether vision actually works or not. Early in machine setup, Users typically play around with the GUI buttons, i.e., _outside jobs_, and your PR now ensures calibration is done, and some frustrating "hair pulling" will be a thing of the past.
> 
> All-in-all this PR a big gain in my eyes! You should then also announce it on the group with that screen-shot. 💯
> 
Thank You very much! Yes, I'll announce it with the screen-shot.

Jan

janm012012:
Hi @markmaker!

> > Apart from these, as far as I can tell, this PR can be merged.
> 
> That's good to hear! However, I'd like to make load/unloadNozzleTip "interruptible-restartable" first as it would make the code cleaner and simplify the continuation request in the exception.
> 
> I'm also considering refactoring the sequences for automatic tip loading/unloading into separate functions so that the loadNozzleTip() would either call automaticUnloadNozzleTip() or manualUnloadNozzleTip() and do the required work there. This shall increase the readability and group common items closer together. If you don't like that, please feel free to merge straight way what we have or tell me and I'll revert/stop that.

I've now fixed the replace problem of the errorBox() and tried to make load- and unloadNozzleTip "interruptible-restartable". This works according to my tests. A nozzle tip load request now calibrates an uncalibrated nozzle tip if the requested tip is already loaded (suppose nozzle tip calibration is enabled). However, I found, that there are now two ensureZCalibrated() close by (line 714 and 734) of which I think one may be removed. I've marked them with a "FIXME".

I also added fixmes to implement the continuation chaining and to continue a manual nozzle tip unload with itself (for completeness and symmetry with loadNozzleTip()) . I'm also find if you decide to skip the effort for the later...

I've setup an automatic nozzle tip changer and verified, that the code works I would expect it to do. However, while setting this up, I was presented with a strange error message "can't move to X -33mm" or something which had nothing in common with the coordinates I used for setup. Turns out that I only used three of the four locations. There would be room for improvements... 

I skipped the idea to refactor the automatic nozzle change motion because I was not aware, that there is also motion for manual nozzle tip changing in the same section. Afterword both take the same path to set/unset the nozzle tip before - in case of manual nozzle tip change - the exception is thrown. I don't see much gain now in refactoring that.

From my perspective, the goal to combine all exception for manual unload/load into one is now almost completed. Please advice how to handle the double ensureZCalibrated() calls and if let me know if you see anything else that shall be changed/implemented.

Thanks You again for your continuous support!

Jan

markmaker:
Thanks, @janm012012,

isn't this message box a cool thing? :+1: 
image


Re 

> I thinks, once combined continuations are supported, we shall make it take the same path whatever changer was used.

and

>` // FIXME: once combined continuation is implemented, continue unloading with itself to eventually calibrate the bare nozzle`

No! Combining the manual instructions like shown in the message box, means this is no longer possible or desirable. We want the exceptions to be collected as manual instructions and when the user unloads and reloads in the same interruption, we no longer _want_ the "calibrate bare nozzle" continuation, i.e., the nozzle is no longer bare when the task resumes!

Re 

```java
            // force the nozzle tip now installed to be uncalibrated
            this.nozzleTip.getCalibration().reset(this);
```

This should **not** be done unconditionally. Keep an automatic changer in mind. The user might have disabled re-calibration on each change, assuming the nozzle tip is unloaded/reloaded at exactly the same rotation angle, i.e., runout remains the same. [See the Wiki about this](https://github.com/openpnp/openpnp/wiki/Nozzle-Tip-Calibration-Setup#automatic-recalibration). Yes, with a manual changer you should of course set it to "On Each Nozzle Tip Change".

Code should look like follows ... but now I see that's exactly what already happened inside the `if (!changerEnabled) {  }`. So please move it up from there and slightly adjust the comment:

```java
            if (this.nozzleTip.getCalibration().isRecalibrateOnNozzleTipChangeNeeded(this) 
                || this.nozzleTip.getCalibration().isRecalibrateOnNozzleTipChangeInJobNeeded(this)) {
                Logger.debug("{}.loadNozzleTip() nozzle tip {} calibration reset", getName(), this.nozzleTip.getName());
                // in case we can't automatically recalibrate (with manual change) - reset() for now
                this.nozzleTip.getCalibration().reset(this);
             }
```

Note, I believe it is fine that this reset is then also done for the `isUnloadedNozzleTipStandin()` case.

The reset means we now need to change the last bit of `loadNozzleTip()` from line 736, i.e., when already calibrated, we no longer want to calibrate on extra calls. To be truly "idempotent", `loadNozzleTip()` must ultimately do nothing. Also the `isRecalibrateOnNozzleTipChangeInJobNeeded()` reset was already done earlier, and can now be removed.

So it should look like so:

```java
        if (this.nozzleTip.getCalibration().isRecalibrateOnNozzleTipChangeNeeded(this)
            && !this.nozzleTip.isCalibrated(this)) {
            Logger.debug("{}.loadNozzleTip() nozzle tip {} calibration needed", getName(), this.nozzleTip.getName());
            this.nozzleTip.getCalibration().calibrate(this);
        }
```

Please also retest this, thanks.

> Please advice how to handle the double ensureZCalibrated() calls

Good catch. Remove the first `ensureZCalibrated(true)`.

Cosmetic: you could move the actuator-name-resolves inside the `if (changerEnabled) { }` both in load and unload.

> There would be room for improvements...

Yes. It could skip a location when not `isInitialized()`.

Thanks for all the hard work :1st_place_medal: 

_Mark

janm012012:
Hi  @markmaker!

I've made all changes you mentioned including the improvement to the automatic nozzle tip changer and some cleanup. Please have an other look if I have correctly adapted your suggestions.

Yes, I've tested the code again and it still works as expected.

Jan

janm012012:
Hi @markmaker!

I've to switch back to draft because I just noticed, that the continuation handling does not work anymore if the nozzle tip change is requested by the JobProcessors. The error is routed via JobPanel.java jobRun() and generates in line 998 an errorBox() bypassing the showError() method we've extended for continuation handling. It is safe/ok to remove the second errorBox() method and modify jobRun() to call showError() and modify showError() to accept a title for the errorBox() to preserve the current behavior?

Jan

janm012012:
Hi @markmaker!

I've now made an implementation that fixes the exception handling in the JobPanel by calling showError() to handle exceptions with continuations. Would you please have a look? Thank You!

Jan

markmaker:
Hi @janm012012,

this looks good, and I assume you tested it.

Can this be merged now?

_Mark

janm012012:
Yes, I've tested it using manual load and unload, as well as within a job. With my latest commit - fixing JobPanel to use showError() as well - it's ready for merge, IMHO.

markmaker:
Many thanks, @janm012012, don't forget to announce it! :smile:

PR #1615, opened 2024-03-19
New feeder type ReferenceAutoVisionFeeder based on CV implemented by PushPullFeeder

prusk:
# Description
This PR adds a new feeder type named _ReferenceAutoVisionFeeder_. It combines the vision calibration initially implemented by the _ReferencePushPullFeeder_ with the standard 2 actuators in the _ReferenceAutoFeeder_.
This is extremely useful for auto feeders that have a larger pick window which exposes at least 2 tape holes.

In addition:
* Creates a base abstract class `ReferenceVisionFeeder` that implements the CV calibration and can be reused in multiple feeders without code duplication. This is a stripped down version of the PushPull feeder.
* Creates the new `ReferenceAutoVisionFeeder` to allow for CV calibration and standard 2 actuators
* Updates the `ReferencePushPullFeeder` to inherit the new `ReferenceVisionFeeder` abstract so that CV code is not duplicated unnecessarily and remains just as easy to maintain moving forward

# Justification
Multiple auto feeders expose a larger pick window where multiple tape holes are visible.
The new feeder type allows better calibration of the pick location.

# Instructions for Use
Create a new feeder of type `ReferenceAutoVisionFeeder` and configure as follows:
* Vision calibration parameters - same as `ReferencePushPullFeeder`
* Actuators - same as `ReferenceAutoFeeder`

# Implementation Details
* Ran all the automated tests
* Compared `machine.xml` for `ReferencePushPullFeeder` before and after the change noticing no content difference
* Tested the new feeder type on my personal hardware
* Yes
* No changes to `org.openpnp.spi` or `org.openpnp.model` packages
* All tests passed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ hs_err*
 
 .vscode/
 /src/main/resources/org/openpnp/translations.properties.bak
+.DS_Store
--- a/src/main/java/org/openpnp/machine/reference/ReferenceMachine.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceMachine.java
@@ -68,6 +68,7 @@ import org.openpnp.machine.reference.feeder.ReferenceHeapFeeder;
 import org.openpnp.machine.reference.feeder.ReferenceLeverFeeder;
 import org.openpnp.machine.reference.feeder.ReferenceLoosePartFeeder;
 import org.openpnp.machine.reference.feeder.ReferencePushPullFeeder;
+import org.openpnp.machine.reference.feeder.ReferenceAutoVisionFeeder;
 import org.openpnp.machine.reference.feeder.ReferenceRotatedTrayFeeder;
 import org.openpnp.machine.reference.feeder.ReferenceSlotAutoFeeder;
 import org.openpnp.machine.reference.feeder.ReferenceStripFeeder;
@@ -427,6 +428,7 @@ public class ReferenceMachine extends AbstractMachine {
         l.add(ReferencePushPullFeeder.class);
         l.add(ReferenceTubeFeeder.class);
         l.add(ReferenceAutoFeeder.class);
+        l.add(ReferenceAutoVisionFeeder.class);
         l.add(ReferenceSlotAutoFeeder.class);
         l.add(ReferenceLoosePartFeeder.class);
         l.add(AdvancedLoosePartFeeder.class);
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceAutoVisionFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceAutoVisionFeeder.java
@@ -16,18 +16,19 @@ import org.openpnp.spi.PropertySheetHolder;
 import org.openpnp.util.MovableUtils;
 import org.pmw.tinylog.Logger;
 import org.simpleframework.xml.Attribute;
+import org.simpleframework.xml.core.Persist;
 
 public class ReferenceAutoVisionFeeder extends ReferenceVisionFeeder {
 
 	@Attribute(required=false)
-    protected String feedActuatorName;
+    private String feedActuatorName;
 	protected Actuator feedActuator;
     
     @Attribute(required=false)
     protected double feedActuatorValue;
 
     @Attribute(required=false)
-    protected String postPickActuatorName;
+    private String postPickActuatorName;
 	protected Actuator postPickActuator;
 
     @Attribute(required=false)
@@ -40,8 +41,18 @@ public class ReferenceAutoVisionFeeder extends ReferenceVisionFeeder {
         return feedActuatorName;
     }
 
-    public void setFeedActuatorName(String feedActuatorName) {
-        this.feedActuatorName = feedActuatorName;
+    public Actuator getFeedActuator() {
+        return this.feedActuator;
+    }
+
+    public void setFeedActuator(Actuator feedActuator) {
+        Object oldValue = this.feedActuator;
+        this.feedActuator = feedActuator;
+        firePropertyChange("feedActuator", oldValue, feedActuator);
+        
+//        Object oldValue2 = this.feedActuatorName;
+        feedActuatorName = (feedActuator == null ? null : feedActuator.getName()); 
+//        firePropertyChange("feedActuatorName", oldValue2, feedActuatorName);
     }
 
     public double getFeedActuatorValue() {
@@ -56,8 +67,18 @@ public class ReferenceAutoVisionFeeder extends ReferenceVisionFeeder {
         return postPickActuatorName;
     }
 
-    public void setPostPickActuatorName(String postPickActuatorName) {
-        this.postPickActuatorName = postPickActuatorName;
+    public Actuator getPostPickActuator() {
+        return this.postPickActuator;
+    }
+
+    public void setPostPickActuator(Actuator postPickActuator) {
+        Object oldValue = this.postPickActuator;
+        this.postPickActuator = postPickActuator;
+        firePropertyChange("postPickActuator", oldValue, postPickActuator);
+        
+//        Object oldValue2 = this.postPickActuatorName;
+        postPickActuatorName = (postPickActuator == null ? null : postPickActuator.getName()); 
+//        firePropertyChange("postPickActuatorName", oldValue2, postPickActuatorName);
     }
 
     public double getPostPickActuatorValue() {
@@ -113,6 +134,13 @@ public class ReferenceAutoVisionFeeder extends ReferenceVisionFeeder {
         });
     	
 	}
+
+    @Persist
+    private void persist() {
+        // Make sure the newest names are persisted (legacy way).
+        feedActuatorName = (feedActuator == null ? null : feedActuator.getName()); 
+        postPickActuatorName = (postPickActuator == null ? null : postPickActuator.getName()); 
+    }
 
 	// inherited from ReferenceFeeder
 	@Override
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferenceAutoVisionFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferenceAutoVisionFeeder.java
@@ -1,0 +1,215 @@
+package org.openpnp.machine.reference.feeder;
+
+import javax.swing.Action;
+
+import org.openpnp.ConfigurationListener;
+import org.openpnp.gui.support.PropertySheetWizardAdapter;
+import org.openpnp.gui.support.Wizard;
+import org.openpnp.machine.reference.feeder.wizards.ReferenceAutoVisionFeederConfigurationWizard;
+import org.openpnp.model.Configuration;
+import org.openpnp.spi.Actuator;
+import org.openpnp.spi.Head;
+import org.openpnp.spi.Machine;
+import org.openpnp.spi.MachineListener;
+import org.openpnp.spi.Nozzle;
+import org.openpnp.spi.PropertySheetHolder;
+import org.openpnp.util.MovableUtils;
+import org.pmw.tinylog.Logger;
+import org.simpleframework.xml.Attribute;
+
+public class ReferenceAutoVisionFeeder extends ReferenceVisionFeeder {
+
+	@Attribute(required=false)
+    protected String feedActuatorName;
+	protected Actuator feedActuator;
+    
+    @Attribute(required=false)
+    protected double feedActuatorValue;
+
+    @Attribute(required=false)
+    protected String postPickActuatorName;
+	protected Actuator postPickActuator;
+
+    @Attribute(required=false)
+    protected double postPickActuatorValue;
+    
+    @Attribute(required=false)
+    protected boolean moveBeforeFeed = false;
+	
+    public String getFeedActuatorName() {
+        return feedActuatorName;
+    }
+
+    public void setFeedActuatorName(String feedActuatorName) {
+        this.feedActuatorName = feedActuatorName;
+    }
+
+    public double getFeedActuatorValue() {
+        return feedActuatorValue;
+    }
+
+    public void setFeedActuatorValue(double feedActuatorValue) {
+        this.feedActuatorValue = feedActuatorValue;
+    }
+
+    public String getPostPickActuatorName() {
+        return postPickActuatorName;
+    }
+
+    public void setPostPickActuatorName(String postPickActuatorName) {
+        this.postPickActuatorName = postPickActuatorName;
+    }
+
+    public double getPostPickActuatorValue() {
+        return postPickActuatorValue;
+    }
+
+    public void setPostPickActuatorValue(double postPickActuatorValue) {
+        this.postPickActuatorValue = postPickActuatorValue;
+    }
+
+    public boolean isMoveBeforeFeed() {
+		return moveBeforeFeed;
+	}
+
+	public void setMoveBeforeFeed(boolean moveBeforeFeed) {
+		this.moveBeforeFeed = moveBeforeFeed;
+	}
+
+	
+	public ReferenceAutoVisionFeeder() {
+    	
+        Configuration.get().addListener(new ConfigurationListener.Adapter() {
+            @Override
+            public void configurationComplete(Configuration configuration) throws Exception {
+                // Resolve the actuators by name (legacy way).
+                Machine machine = Configuration.get().getMachine();
+                try {
+                    feedActuator = machine.getActuatorByName(feedActuatorName);
+                }
+                catch (Exception e) {
+                }
+                try {
+                	postPickActuator = machine.getActuatorByName(postPickActuatorName);
+                }
+                catch (Exception e) {
+                }
+                // Listen to the machine become unhomed to invalidate feeder calibration.
+                // Note that home()  first switches the machine isHomed() state off, then on again, 
+                // so we also catch re-homing.
+                Configuration.get().getMachine().addListener(new MachineListener.Adapter() {
+
+                    @Override
+                    public void machineHeadActivity(Machine machine, Head head) {
+                        checkHomedState(machine);
+                    }
+
+                    @Override
+                    public void machineEnabled(Machine machine) {
+                        checkHomedState(machine);
+                    }
+                });
+            }
+        });
+    	
+	}
+
+	// inherited from ReferenceFeeder
+	@Override
+    public void feed(Nozzle nozzle) throws Exception {
+        if (isMoveBeforeFeed()) {
+            MovableUtils.moveToLocationAtSafeZ(nozzle, getPickLocation().derive(null, null, Double.NaN, null));
+        }
+
+        Logger.debug("feed({})", nozzle);
+
+        Head head = nozzle.getHead();
+
+        if (getFeedCount() % getPartsPerFeedCycle() == 0) {
+            // Modulo of feed count is zero - no more parts there to pick, must feed
+
+            // Make sure we're calibrated
+            assertCalibrated(false);
+
+            long feedsPerPart = (long)Math.ceil(getPartPitch().divide(getFeedPitch()));
+            long n = feedsPerPart;
+            for (long i = 0; i < n; i++) {  // perform multiple feed actuations if required
+
+                // Actuate the tape once for the length equal to Tape Pitch
+            	feedTape(nozzle, getFeedPitch().getValue());
+            }
+
+            head.moveToSafeZ();
+            // Make sure we're calibrated after type feed
+            assertCalibrated(true);
+        }
+        else {
+            Logger.debug("Multi parts feed: skipping tape feed at feed count " + getFeedCount());
+        }
+
+        // increment feed count
+        setFeedCount(getFeedCount()+1);
+
+    }
+	
+	private void feedTape(Nozzle nozzle, Double feedPitch) throws Exception {
+		if (feedActuatorName == null || feedActuatorName.equals("")) {
+			throw new Exception("No actuatorName specified for feeder " + getName());
+        }
+        Actuator actuator = nozzle.getHead().getActuatorByName(feedActuatorName);
+        if (actuator == null) {
+            actuator = Configuration.get().getMachine().getActuatorByName(feedActuatorName);
+        }
+        if (actuator == null) {
+            throw new Exception("Feed failed. Unable to find an actuator named " + feedActuatorName);
+        }
+     // Note by using the Object generic method, the value will be properly interpreted according to actuator.valueType.
+        actuator.actuate((Object)feedActuatorValue);
+	}
+	
+	@Override
+    public void postPick(Nozzle nozzle) throws Exception {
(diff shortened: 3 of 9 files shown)


markmaker:
Hi @prusk,

just so you know up front, I currently have limited time to look into this in-depth right now. With that in mind, I only had a cursory glance, please answer a few questions.

Note, I am deliberately asking critical/sceptical questions, but there is no offense intended, I just want to hear your reasoning. As you've surely seen, an [upfront discussion would have been required](https://github.com/openpnp/openpnp/wiki/Developers-Guide#discussion-first), so I hope you understand, it still comes _after_ you've invested all that work.

- The description and justification is very terse. Why exactly was this necessary?
- The  `ReferencePushPullFeeder` can already drive two Actuators, and users already use it for electric/automatic feeders. Why not simply add the missing Post Pick Actuator? See also the Wiki:
    https://github.com/openpnp/openpnp/wiki/ReferencePushPullFeeder#actuator-boolean-command
    https://github.com/openpnp/openpnp/wiki/ReferencePushPullFeeder#assigning-actuators
- Also note that the "move before feed" effect can be already achieved in the `ReferencePushPullFeeder` simply by setting the **Start Location** above the pick location. But I would not oppose adding the boolean anyway.
- Please explain why the OCR options are not in the base class, where all the OCR detection and pipeline is present anyways? This means users have to decide which feature-set they want (OCR+motion **XOR** post pick actuator), they can't have both. Why should users of automatic feeders not want OCR/QR-Codes?
- Same for the template logic. Why make it harder for users of your `ReferenceAutoVisionFeeder`?
- Shouldn't we have discussed other options, like integrating missing features (I only see the Post Pick Actuator) into `ReferencePushPullFeeder`, so all can benefit?
- For Actuators in general: in new code they should bound by id and not by name. Alternatively use the `@Persist` legacy handling as in the `ReferencePushPullFeeder`.

Like I said, I only had a cursory look into the sources, I might have misinterpreted one thing or two. :sweat_smile: 

The refactoring actually looks very good. But until there is much better justification that I missed, I'm very reluctant to adopt it. It makes me sad we haven't discussed this **before** you started  :cry:! So even if this does not get integrated as is, I hope this will not chase you away from the project. We could use your skills. :sunglasses: 

_Mark

prusk:
Hi @markmaker,

This will be a long reply so bear with me please.
A few upfront things:
1. Thank you for your questions and notes. They are very much appreciated and in software engineering I personally believe all changes should be treated with skepticism. I expected nothing less.
2. Please consider my notes below as constructive observations and rest assured no offence is intended.
3. Context - this comes from a new version of feeders for the Pandaplacer A1 machine but actually can apply to many others according to the research I have done. I explored the option of creating a class for BambooFeeder but the functionality provided is too generic and many other people could benefit.
4. Background - last time I did desktop development was probably 20 years ago so for the last little while I worked on client-server solutions mostly with web UIs. However some principles apply to everything. I'm saying this as it might make it easier to understand why I'm bringing some things forward.

I am sorry but I kind of missed the "discussion first" item. However I don't fully agree with the approach as it sometimes wastes time without having some code to look at and the conversation can easily become fuzzy. Having discussions after a first cut of the code is available is quite productive and quick so I would suggest changing that guide item to "discussion required", weather first or second it becomes less important as long as it happens.

About "Reference" things: I was actually surprised to find concrete classes in this area/folder as usually what I encountered in reference architecture were just templates for actual concrete implementations. I would suggest this term to be revisited as the only true reference thing is the ReferenceFeeder that is not even in the same folder. I did abide by the naming conventions though and it is the only reason some of the classes have "reference" in the name.
Considering the above I would argue these changes are not for "reference functionality" therefore the "discussion first" does not really apply. Don't get me wrong, I welcome and appreciate discussions anyways and I find them always needed, reason why PRs allow and encourage them by design.

About ReferencePushPullFeeder: wonderful piece of code that does a lot of things, too many for this use case.
I actually tried extending this class and it kind of worked but due to the overwhelming quantity of attributes and mandatory xml elements, the configuration for the extended class was too much and it became too easy to break the desired functionality by changing xml values, even if the UI wizard did not expose those items. I applied the "hope for the best, prepare for the worst" software engineering rule so I specifically was looking for ways to break it and unfortunately I found them.
As a principle I've been taught that no object should have more than it needs and under no circumstances should expose configuration that can break its functionality. It is simply impossible to create an extended class that removes parent attributes and methods by reducing the visibility, therefore removing un-needed items. Because of this I abandoned extending the PushPull class.
When thinking about enriching the existing ReferencePushPullFeeder class, I didn't feel the risk of breaking existing functionality is worth taking as I'm sure many people use it in production.

Realistically, the ReferencePushPullFeeder, while doing a lot of things perfectly suitable for the associated 3D-printed object, it is quite monolithic and not well designed for reuse in other different physical feeders. It was obvious to me this was not designed to be the base of anything but only a concrete final class.
What I found were a few deficiencies that made me want to extract some functionality in a base abstract class:
* It contains vision calibration, OCR, head-mounted boolean actuators, nozzle movements and templating/cloning all in one, very tightly coupled between each other. Once could say the separation of concerns principle is simply ignored.
* Makes quite a few assumptions, one of them being users want both vision calibration and OCR, also relying completely on the assumption that the OCR area is captured in the same image as the tape holes. This assumption is incorrect as for other designs the OCR area is further away, towards the "nose" of the feeder. Maybe other feeders will have the OCR area towards the back. Overall there is no guarantee this interest area is in the same picture.
* Vision and OCR are too intertwined to the point you simply cannot have vision calibration without defining an interest area for the OCR, even if you don't want OCR. For some reason the code made me feel that OCR was more important than the vision calibration...
* Disabling OCR is quite awkward, a head scratcher, and still one needs some basic configuration for OCR even if later they want to disable it. I personally don't agree with this approach.
* Movement does includes 2 actuators that are limited to head-mounted ones with Boolean values exclusively. As you know this is not compatible with 0816-type Arduino feeder controllers that require Double actuators. I could in theory create individual Boolean actuators for each of my feeder ports but then I end up with 56 of them when there could be a better implementation offering superior user experience.
* It is impossible to remove the "StartLocation" and "EndLocation" xml elements as they are mandatory, making them available in the config file and therefore when changed breaking the desired functionality of a simple vision-calibrated auto feeder that requires no nozzle movement for the tape advance. Same goes for other attributes that do not support null values.

So, for the use-case where one requires only an auto-feeder with vision calibration, using the PushPull feeder is unnecessary complex with too many options available and too easy to miss-configure. I found the user experience suboptimal when trying to use this for something that was not designed to support. For an extended class, the only option to reduce complexity was to hide things in the UI therefore injecting business logic in that layer.

With regard to client-server architecture: I understand this is a desktop application with many things reinforced in the UI. However, one day I hope OpenPnP will introduce support for other UIs as well like a web UI and even a touch-screen UI, in which case this approach is faulty.
What I was taught about software architecture was that no business logic belongs in the UIs or databases therefore all rules and restrictions must be implemented in the functional layer, BLL. More so, the exposed configuration should only be minimal and should never contain items that can break the designed functionality.

Side note: I think serialization / deserialization needs a second look as the mapping between class attributes and XML content is too tight, 1:1. It should probably be allowed for the XML config to omit items with default values or contain attributes implemented only in newer versions, not mapped yet to any class properties.

Because of everything above, the conclusion was that the best way of moving forward was to create something new. Applying the principle of "no code duplication", there was no other way of going about it except to extract some core functionality in a base class that later 2 feeders inherit.
The only common ground between ReferencePushPullFeeder and ReferenceAutoVisionFeeder is the CV calibration as the OCR, head-mounted boolean actuators, nozzle movements and templating/cloning are specific to the PushPull concept and not required in a simple AutoVision feeder.
I did consider adding support for OCR as well but realistically that is optional and the interest area sometimes not in the same picture as the tape. Maybe in the future...

Finally, to answer your specific questions:
* I wrote the description and justification as succinct and clear as I could. Sorry they were found lacking.
* Both actuators for ReferencePushPullFeeder are Boolean and head-mounted. Auto feeders work with Double machine-level actuators.
* Correct, move before feed can be achieved with the StartLocation but when all of the nozzle movement section is stripped, that option is not available. More so, StartLocation is specific to PushPull feeder, not necessarily to the AutoVision feeder.
* OCR was not included in the base class because it is a different thing, separate concern when compared to vision calibration. In my opinion OCR can be considered an optional feature and really doesn't even probably need to be present in the XML-defined pipeline (like implemented by the RapidFeeder). I did not include it in the base class because of the core assumption that the OCR area is in the same picture as the tape, false assumption for the feeders I have on hand.
* Template logic was not included because I found it too specific to a concrete final class feeder considering the amount of configuration required for cloning. I simply could not build an abstract class for it. Maybe in the future but considering Java doesn't allow inheritance from multiple classes, I'm puzzled of how to go about it and find a way to make it available to any kind of feeder, not just PushPull and AutoVision. This is a bigger discussion that probably needs to happen. Specifically for the AutoVisionFeeder, the configuration is minimal so not that complicated to create new feeders therefore diminishing the value of templating/cloning.
* I agree ReferencePushPullFeeder can be enriched to include additional functionality, however I'm not personally a fan of building a tool for everything, a universal hammer for all things nail-shaped (sometimes you actually need a screwdriver). It's already a monolith and 17 years as a software architect didn't let me go in this direction.
* Noted about actuators bound by ID. I will see what I can do about it.

Don't be sad we haven't discussed this before, I see nothing wrong with talking with the code in front of us. It is a perfectly valid approach proven by lots of projects, open source or not.
No worries, I won't go away from the project as I have Production needs and I'm already quite vested in Pandaplacer (I'm PP.ca on the google group).

TODO for me: bind the actuators by ID. I'll push a new commit soon.

markmaker:
Hi @prusk 

very good discussion.

> Having discussions after a first cut of the code is available is quite productive and quick

Completely agree, if there are no hard feelings if the first attempt is rejected.

> About "Reference" things: I was actually surprised to find concrete classes in this area

Why not? As opposed to the "Abstract" classes? I always assumed these "Reference" classes are like [reference implementations](https://en.wikipedia.org/wiki/Reference_implementation) elsewhere. We would have to ask @vonnieda how he originally envisioned things. :smile: 

> As a principle I've been taught that no object should have more than it needs and under no circumstances should expose configuration that can break its functionality.

Agree on a theoretical level. But very hard to realize in a practical _and_ usable manner. See below.

>  It is simply impossible to create an extended class that removes parent attributes and methods by reducing the visibility, therefore removing un-needed items. Because of this I abandoned extending the PushPull class.

Yes. In a similar spirit you could name Java's lack of multiple inheritance. You can't inherit and combine features (mix-ins), like say "Sprocket hole vision", "OCR", "Motion actuation", "Actuators" etc. And even if it had multiple inheritance like C/C++ it would not be practical, because developers then would need to make _every combination_ of feature-set available for users to instantiate, i.e. there would be a SprocketVisionOCRMotionFeeder class, and a  SprocketVisionMotionActuatorFeeder class and 13 others with just this choice of feature sets (2<sup>n</sup> - 1).

And this is even further encumbered by the much more restrictive Wizard classes. 

So increasingly, the [Composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance) is being used. Instead of inheriting feature sets, we compose them as optional components. These components then even can have their own class hierarchy. 

Each feature set also gets its own Wizard tab(s), that only appear when the feature is enabled. Enabling such features can then also be dynamic, i.e., during the life-time of an object. So if a user later say, "hey I want to add OCR", they can do it.

Take the [Camera Auto Focus](https://github.com/openpnp/openpnp/wiki/Up-looking-Camera-Auto-Focus) as such an example. I wanted to add it to all Camera classes. But I did not want to not make the camera more complex for users who don't want to use it. So this wasn't possible through inheritance.  

If you [enable AutoFocus](https://github.com/openpnp/openpnp/wiki/Up-looking-Camera-Auto-Focus#camera-configuration) on the base camera class by selecting a method, another Auto Focus tab will appear. 

image

image

This is reflected by setting the `@Element` member `ReferenceCamera.focusProvider` in the camera base class. It is extensible with other **Focus Sensing Methods** and corresponding component classes in the future.

But this is not perfect either. If you mix two features, you can't optimize a _merged_ operation, as is now done in sprocket vision + OCR. Doing both visions in one shot, saves a lot of time! Many such inter-feature-dependencies are very hard to get right. Code tends to get very complex, and please don't propose the [Visitor Pattern](https://en.wikipedia.org/wiki/Visitor_pattern) as being "simpler" :vomiting_face:. 

Maybe if we talked this through before you started :smirk:, I would have suggested you extract the unwanted feature sets (Motion, OCR) as such component classes. But then we would quickly have seen this is impossible to migrate using SimpleXML. Therefore I would have suggested you could still make inner component classes but leave the` @Attribute`s out in the main class, still easy to access. Move all UI of such feature-sets onto separate Wizard tabs. Dynamically enable the tabs using their "method" drop-downs (or similar) as shown with Auto Focus above. Perhaps add a new "Auto Feeder"-Tab with the unified and extended Actuator features and indexing/addressing (Double) added.

>  I didn't feel the risk of breaking existing functionality is worth taking as I'm sure many people use it in production.

Appreciate. But are you sure there is less risk with this remodeling?

> Once could say the separation of concerns principle is simply ignored.

Yes. And no. Theory and practice. Just look at how complex the smart cloning is. It needs "business" knowledge of the other concerns, to know how to work correctly. OCR and vision are also tightly integrated for efficiency, as explained above. Push-pull motion coordinates need to be calibrated by vision, both translated and rotated Push-pull motion needs to be _coordinated_ with Actuator actuation, and more. It would be interesting to see an _actual practical attempt_ to _really_ separate these concerns here :rofl:. 

Regarding features, I would like to mention that this is still an Open Source project. I agree that it is a completely different situation than having a proprietary commercial product, where software and hardware are tightly tuned to each other. In contrast, it is not a concern for us, that "surplus features" that are not supported by the hardware might make that hardware look "less complete". We _want_ our users ask that nasty question: "Why does feeder XY not support labeling with OCR/QR-codes?" It's what drives innovation.

> also relying completely on the assumption that the OCR area is captured in the same image as the tape holes

No longer true thanks to @crashmatt's #1468. But it is still very much recommended to make it so for performance considerations.

> Vision and OCR are too intertwined to the point you simply cannot have vision calibration without defining an interest area for the OCR

Huh? If this were true, it would be a bug. Maybe the Wiki/UI is not clear enough. Please elaborate. 

As I laid out above, I would be happy to let you improve this as a dynamically addable component class, and separate tab.

> is not compatible with 0816-type Arduino feeder controllers that require Double actuators

I was not aware of that and agree. But not a major thing to add/change.

> It is impossible to remove the "StartLocation" and "EndLocation" xml elements as they are mandatory,

Huh? And why not just change that? :roll_eyes: 

> So, for the use-case where one requires only an auto-feeder with vision calibration, using the PushPull feeder is unnecessary complex with too many options available and too easy to miss-configure.

After reading all your points, and seeing you sum it up it like that, perhaps you extended/refactored the wrong feeder class? You could simply have copied the few things you wanted from the sprocket vision to be also used in the ReferenceAutoFeeder. But I hope I convinced you of the powerful component approach. 

> With regard to client-server architecture... 

I'm too old for that discussion. :smile: 

> I think serialization / deserialization needs a second look ... 

I don't know enough about SimpleXML to really form a judgement. It seems to support tons of customizations. The only thing I say is that _Reflection_ is the only sane way to do it, IMHO. There should be more, not less of that in OpenPnP.

> however I'm not personally a fan of building a tool for everything, 

I'm Swiss, I love my Swiss Army Knives. Maybe that's an inbred bias. :wink:

image

And yes, they work.

_Mark

prusk:
> I would have suggested you could still make inner component classes but leave the @Attributes out in the main class, still easy to access. Move all UI of such feature-sets onto separate Wizard tabs. Dynamically enable the tabs using their "method" drop-downs (or similar) as shown with Auto Focus above. Perhaps add a new "Auto Feeder"-Tab with the unified and extended Actuator features and indexing/addressing (Double) added.

Very good ideas. Not sure I can tackle them as a first step.

> >I didn't feel the risk of breaking existing functionality is worth taking as I'm sure many people use it in production.

>Appreciate. But are you sure there is less risk with this remodeling?

Yes because as far as ReferencePushPullFeeder is concerned, there is no remodelling. The combination of the abstract and concrete classes obtains the exact same thing.

>OCR and vision are also tightly integrated for efficiency, as explained above. Push-pull motion coordinates need to be calibrated by vision, both translated and rotated Push-pull motion needs to be coordinated with Actuator actuation, and more.

And this is exactly the reason why I extracted only the vision calibration in the base abstract as I found it fundamental, while everything else is on top.

>> Vision and OCR are too intertwined to the point you simply cannot have vision calibration without defining an interest area for the OCR

> Huh? If this were true, it would be a bug. Maybe the Wiki/UI is not clear enough. Please elaborate.

See this please: https://github.com/openpnp/openpnp/blob/test/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L2457
OCR comes first and vision cannot be used without it.


>>It is impossible to remove the "StartLocation" and "EndLocation" xml elements as they are mandatory,

>Huh? And why not just change that? 🙄

I tried and those elements can go away. There is no option though for all the main properties to disappear from the xml configuration. All the speed values & booleans for the push-pull motion cannot get nulls in a child class so they go away from the configuration and prevent breaking the functionality by doing direct file changes.

> But I hope I convinced you of the powerful component approach.

No issue there and probably viable as a second step / further refinement.

>You could simply have copied the few things you wanted from the sprocket vision to be also used in the ReferenceAutoFeeder.

I thought of that but something didn't let me duplicate code as I know it is painful to maintain. Just the other day you had a PR for debugging the vision calibration and it was just 5-6 changes in a single place. I would not like to have to do that more than once having the overhead of keeping various code sections in sync.

> I'm too old for that discussion. 😄

You and me both.

>>is not compatible with 0816-type Arduino feeder controllers that require Double actuators

>I was not aware of that and agree. But not a major thing to add/change.

For me it might be bigger than for you. Also it increases usage complexity while defining a different type of feeder that works with these seems to be cleaner.

> I'm Swiss, I love my Swiss Army Knives. Maybe that's an inbred bias. 😉

You might be biased 😉. Here in Canada we have a special tool for every different type of nail and screw 🤣.

PS: I kept my word and pushed the commit for persistence.


Moving forward, what options do I have for implementing the AutoVision feeder?
I see the following:
1. Merge the change as is and later add more component-based improvements
2. Remove the changes to ReferencePushPullFeeder and merge the others
3. Create a brand new non-reference feeder just for Pandaplacer under openpnp/machine - a BambooFeeder
4. Go nuts and create a special Pandaplacer OpenPnP repo - similar to what I heard other PnP builders did (unverified).

I still think there is value in this code and while I don't disagree with you, this is how far I can take it right now. I'm pretty sure in the future I will do more as time allows me.

prusk:
Hi @markmaker,

Could you please let us know what is the decision about this PR?

Thank you.

markmaker:
Hi @prusk,

as I said before, in the current form, this looks too disruptive and risky vis-a-vis existing ReferencePushPullFeeder users to me. It creates a second user base that cannot later combine OCR. All the things I laid out earlier. 

Furthermore, I also start to wonder if you are aware of the "Slot" variety of the auto-feeders. If you don't have OCR, don't you want that?

If you want a quick solution, I propose to just add the missing post-pick actuator to ReferencePushPullFeeder, I'm also open for a UI redesign where OCR can be moved to a different Wizard tab and disabled or even hidden with one checkbox. Plus one more Wizard tab for all the actuators including the new post pick, properly extracted and more control around those.

_Mark

markmaker:
Oh, and again, I would like to remind you of what I said at the beginning:

>  It makes me sad we haven't discussed this before you started 😢! So even if this does not get integrated as is, I hope this will not chase you away from the project. We could use your skills. 😎

_Mark

vonnieda:
Hi @prusk and @markmaker,

TL;DR: I suggest submitting a new PR where your changes are limited to a new openpnp/machine/pandaplacer directory, without the refactoring in Reference* classes. Those can be a separate PR that should be discussed beforehand.

Some ancient history:

I originally envisioned OpenPnP as being "non-useful" without making one's own machine classes that could extend or copy from the Reference classes. The Reference classes were meant to be purely for reference, not every day use.

In fact, everything under the machine.reference package was meant to be "the part you should replace/extend/copy to make OpenPnP work for your machine." The Nozzles, the Head, the Feeders, all of it.

The problem was, while there were people who were building machines, or wanted to, there were no Java programmers. You won't believe the flaming I took for writing it in Java. The mailing list is chock full of people refusing to even think about learning Java.

So, when people would ask for small changes, I just added them to the Reference classes. Because a request like "I could use this with Whatever if it would just send a [whatever] on connect" was easier to just make generic rather than creating a whole new WhateverDriver hierarchy.

And then people starting submitting PRs to Reference classes, and some were small and some were huge, but, like, it's really hard to ask brand new committers to resubmit their tiny three line feature PR as a giant thing with new classes, and learn enough Java to know how to extend classes and implement interfaces. So sometimes I just merged them.

That is all to say a few things:

Changes like this, to me, are an easy merge if they come in a new machine sub-package. We have several groups that maintain a machine sub-package and overall the bar for merge is low on those.

I would recommend creating a new pandaplacer directory in https://github.com/openpnp/openpnp/tree/develop/src/main/java/org/openpnp/machine and add a BambooFeeder there. And in that sub-tree, add anything you like as long as it's generally useful and maintainable.

As for the refactoring you've done, I have not reviewed it, but as Mark said, we generally ask that people get in touch before making big changes to core code, so perhaps you could demonstrate those ideas in your sub-package, and a discussion could happen there, without adding any additional risk to existing code. 

Thanks,
Jason

prusk:
Thank you both for your input.

I will prepare a BambooFeeder under a Pandaplacer machine folder and take it from there.
I'll try as much as time allows me to reuse as much as possible from the code to avoid maintenance overhead.

Would it be acceptable in terms of  breaking changes risk to start from the ReferencePushPullFeeder class and either extend it, use an instance of it or reference the FindFeatures sub-class?
Alternatively I'm considering implementing a copy of the vision and later OCR and keeping an eye on changes to the base reference classes... not a huge fan of this approach.

markmaker:
Thank you @vonnieda, very interesting to learn these backgrounds :smile: 

I think there are two possible approaches laid out now:

1. Create your own "brand" feeder class. Starts as copy&paste. Independent. No risk to other users. Easy to merge. But also your own responsibility to keep up with future developments. Contribution remains tied to your "brand" feeder.
2. Extend and refactor an existing feeder class, along the "Composition over inheritance" pattern perhaps. Contribution benefits a wider audience. Favors free combination of the features a user wants. Evolves with contributions from other users, i.e. your "brand" feeder users might one day benefit too. But needs discussion of everything first. Needs you to thread lightly. Risk guided, i.e. rather conservative merge policy.

(I gotta go now, will be offline quite a bit)

_Mark

prusk:
Hello @markmaker @vonnieda ,

I followed up with a dedicated Pandaplacer feeder here: https://github.com/openpnp/openpnp/pull/1622

Note there is a class there named `TapeUtils` that could be for general use and can be easily moved to `org.openpnp.util`.

PR #1616, opened 2024-03-22
add an option to automatically load the most recent job

janm012012:
# Description
This PR provides auto-loading of the most recent job on startup on request.

If the new configuration option "Auto-load most recent job?" on the Configuration tab of the machine is enabled and there are known recent jobs, the most recent one will be loaded automatically. Per default this option is disabled.

# Justification
At least with debugging is tedious to always load the same job to test with. And if OpenPnP starts with the "Untitled.job.xml" anyhow, why not load the most recent one, which is known to have some meaning for the user anyhow.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **I used it while developing other code**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **yes: isAutoLoadMostRecenetJob() was added to Machine.java to read the state of the configuration option from within JobPanel.java**

### DIFF
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -461,6 +461,14 @@ public class JobPanel extends JPanel {
 
                 machine.getPnpJobProcessor().addTextStatusListener(textStatusListener);
 
+                if (machine.isAutoLoadMostRecentJob()) {
+                    // try to load the most recent job
+                    if (recentJobs.size() > 0) {
+                        File file = recentJobs.get(0);
+                        loadJobExec(file);
+                    }
+                }
+
                 // Create an empty Job if one is not loaded
                 if (getJob() == null) {
                     setJob(new Job());
@@ -837,10 +845,7 @@ public class JobPanel extends JPanel {
                     return;
                 }
                 File file = new File(new File(fileDialog.getDirectory()), fileDialog.getFile());
-                Job job = configuration.loadJob(file);
-                setJob(job);
-                addRecentJob(file);
-                mainFrame.getFeedersTab().updateView();
+                loadJobExec(file);
             }
             catch (Exception e) {
                 e.printStackTrace();
@@ -1641,10 +1646,7 @@ public class JobPanel extends JPanel {
                 return;
             }
             try {
-                Job job = configuration.loadJob(file);
-                setJob(job);
-                addRecentJob(file);
-                mainFrame.getFeedersTab().updateView();
+                loadJobExec(file);
             }
             catch (Exception e) {
                 e.printStackTrace();
@@ -1652,6 +1654,19 @@ public class JobPanel extends JPanel {
                         Translations.getString("JobPanel.Action.Job.RecentJobs.ErrorBox.Title"), e.getMessage()); //$NON-NLS-1$
             }
         }
+    }
+
+    /**
+     * Perform all action required to load a job from a given file
+     * 
+     * @param file job to load
+     * @throws Exception
+     */
+    private void loadJobExec(File file) throws Exception {
+        Job job = configuration.loadJob(file);
+        setJob(job);
+        addRecentJob(file);
+        mainFrame.getFeedersTab().updateView();
     }
 
     private final MachineListener machineListener = new MachineListener.Adapter() {
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -461,17 +461,26 @@ public class JobPanel extends JPanel {
 
                 machine.getPnpJobProcessor().addTextStatusListener(textStatusListener);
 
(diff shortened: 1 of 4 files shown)


markmaker:
Hi @janm012012,

makes sense. One question:

Is the `configurationComplete()` where you put the auto-loading firing _asynchronously_?

(I'm asking the following, because I know some of these `ConfigurationListener`s surprisingly fire _synchronously_ if the machine config is already loaded, which was surprising at least to me) 

If **asynchronous**, then okay. 

If **synchronous**, I would recommend putting the `addListener()` in a `SwingUtilities.invokeLater()` wrapper, so the GUI can initialize properly, before loading the job. Otherwise OpenPnP startup might look ugly/frozzen/sluggish on slower machines (it still is in the middle of building the GUI when this happens). This also offloads the JIT compiling for all these additional functions.

One could even argue to do it anyways, to make sure.

_Mark

janm012012:
Hi @markmaker!

> Is the `configurationComplete()` where you put the auto-loading firing _asynchronously_?

I'm afraid I'm suffering from a lack of understanding here. I put it there because this seems to be the place where the list of recent jobs is created.

With the previous implementation the UI title bar was already showing the loaded job while the window was still empty. Now, with the invokeLater() wrapping in place, it sometimes shows undefined.job (as before) and sometimes the loaded job. Thats a little be unpredictable and not ideal. Is there a better place to trigger the auto-loading from?

Jan

markmaker:
>      I'm afraid I'm suffering from a lack of understanding here.

Best add two logging lines: 

1. in `loadJobExec()` and 
2. _after_ the `Configuration.get().addListener()` block.

If (1) is logged before (2) it is called _synchronously_, otherwise _asynchronously_. Please report the findings.

You added the `SwingUtilities.invokeLater()` wrapper around the loading only. I proposed to wrap it around the whole `addListener()` block. This way _the whole thing_ is deferred. 

Also you moved the `// Create an empty Job if one is not loaded` above the auto-load. You should not. It should only do that is auto-loading is inactive.

It is possible that I'm missing something. Like some code somewhere expecting a Job to _always_ exist. Then report back. It could mean fixing it _there_ or reconsidering the whole thing.

_Mark

janm012012:
Hi @markmaker!

> If (1) is logged before (2) it is called _synchronously_, otherwise _asynchronously_. Please report the findings.
> 
Here are my findings: (2) is never logged, (1) is. (According to the debugger, (2) is executed in the AWT-EventQueue with only a "Daemon Thread" running.) So its called _asynchronously_.

> You added the `SwingUtilities.invokeLater()` wrapper around the loading only. I proposed to wrap it around the whole `addListener()` block. This way _the whole thing_ is deferred.
> 
I tried it, without notable change to my first try (without any invokeLater()). The window still starts up empty while already showing the loaded jobs name in its title bar.

> Also you moved the `// Create an empty Job if one is not loaded` above the auto-load. You should not. It should only do that is auto-loading is inactive.
> 
Yes, I did. I wonted to keep the changes small compared to the original version. Its in there.

> It is possible that I'm missing something. Like some code somewhere expecting a Job to _always_ exist. Then report back. It could mean fixing it _there_ or reconsidering the whole thing.
> 
Seems like: if I remove that code and start OpenPnP without any job, strange things are happening to the job-control buttons:
On start (power-button) the Run and Step buttons are enabled generating a null-pointer-exception if pressed. After acknowledging it, the Run button is an active Pause button and the Stop button is enabled. I'd say that updateJobActions() in JobPanel.java line 757 shall keep this buttons disables if the machine is disabled (as it already does) _and_ as long as no job is loaded. However, setJob() (line 563) also creates a few PropertyChangeListener(). I also can't jog anymore. There is an other null-pointer-exception generated at checkJogMotionSaftey (MachineControlsPanel.java line 217). After all this exceptions, I'm not able to load a job anymore. I'm not even able to load a job at all after start. There is a third null-pointer exceition only vision on the console of the debugger generated by checkForJobModifications(JobPanel.java:677).

It seems, the code is not prepared at all to run without any job. If you think we/I shall fix that, please let me know. Then I'd try to fix all this exceptions.

Jan

markmaker:
> So its called asynchronously.

in that case and with the clarification that a Job must always exist, it seems the original implementation of the PR was the right one. 

Notably ,` // Create an empty Job if one is not loaded` was _below_ the auto-loading.

Please revert it back to https://github.com/openpnp/openpnp/pull/1616/commits/648bf4bb4c49f60c7ccc69ca538e94abf8f4c10b

Thanks, and sorry for the detour. :sweat_smile: 

_Mark

janm012012:
I've reverted the changes and added the findings as comments for future developers...

markmaker:
Very good! Thanks a lot, @janm012012 :100:

markmaker:
Please announce too!

PR #1617, opened 2024-03-22
Continuation queueing and automatic job resume

janm012012:
# Description
This PR extends PR #1609 by providing continuation queuing and automatic job resume after a manual nozzle tip change.

It adds a global queue in which all continuations requested by any exceptions to be processed after approval of the user are stored. If the user disagrees, all new continuations will be dropped, continuations already in the queue will still be executed.

Executing the continuations is done by removing them one after the other from the queue  and sending them to the machine task.
The underlying implementation seems to the such, that multiple items can be send before they are actually executed and that pending items will be removed/dropped in case an exception is thrown. This implementation accept, that if a continuation throws an error all other continuations will not be executed anymore. I've documented a possible solutions to change that behavior.

# Justification
It makes sense to use more continuations to continue after an exception as much as possible. Hence resuming a job is just the next step.

# Instructions for Use
Users with manual nozzle tip changing configured will notice, that their jobs continue with just a single click once they performed the requested nozzle tip change.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **there is test-code in ReferencePnPJobProcessor.java that chains five continuations which all generate log different log messages and from which the middle one throws an exception. Using counter and logging I found, that at the time the exception is throws, there are no continuations in the global queue anymore.**


janm012012:
Hi @markmaker!

Many thanks for taking a look reviewing my PR. Obviously I did not expressed my intention clearly enough. My goal was and is to continue a job automatically after a single nozzle tip change. With PR #1609 unloading and loading a tip on one nozzle is folded into a single interruption already but there is still room for improvements to automatically resume the job. To make it clear: I'm not trying to fold more unload and load events for multiple nozzles into a single interruption. I agree, that this makes the process overly complicated and difficult to get right.

To recap what PR #1609 does: it catches manual unload events associated with a manual load event and throws a single exception with a list of unload/load items to execute. This works without continuations. It then uses a single continuation to eventually calibrate the nozzle tip after the loading.

With this PR I'm trying to add a second continuation to resume the job after the nozzle tip load interruption. Yes, I've used your proposal for the implementation and yes, its actually WIP because there are a few design decisions that require discussions and your approval. I've added FIXMEs at changes I'm not entirely confident with.

I think the first decision I'd like you to make is whether you prefer a very generic implementation of continuations or if you can life with the situation that this PR adds that a continuations get dropped when any throws an exception.

While debugging I noticed the following:

- Adding multiple continuations is straight forward. The while loop in UiUtils.java line 153 sends them all immediately to the machine task for executions. However, if any throws an exception, the following are not executed anymore. It seems that the continuations are queued up and that the queue is drained as part of the exception processing.
- I think that exceptions in general are serious errors and hence shall be handled carefully. For me, its ok to skip further continuations after one of them has thrown an exception.
- If a continuation throws an exception, it may carry continuations as well. This continuations have to - in my understanding - be executed before the remaining of the first/older exception. I've decided to use a queue, because it would servers this purpose. I think your proposal would also work.

I agree with your suggestion, that one could just throw the continuation to calibrate the nozzle tip away and add one that resumes the job. That would provide the added advantage, that it would speed thinks up by grouping nozzle tip loads and calibrations into separate steps again (as was before PR #1609). (Today a continuation might calibrate the nozzle tip immediately after loading and before the job resumes. If a second nozzle tip change is required, the head goes to the location where the nozzle tip change will take place. For me this locations are the same. Hence this results in extra movements that could either be changed by configuring the nozzle tip calibration to InJob or with your suggestion.)

Please advice who we could add auto-resume after a manual nozzle tip change. I'd be willing to start from scratch implementing your suggestion or make the continuation queuing more robust.

Jan

markmaker:
Hi @janm012012,

thanks for the solid response.

I should perhaps elaborate that for a long time I had "visions" for how to use this feature further.  :smiley: 

I always intended to add auto-bottom-vision-teach-in to OpenPnP. So when a package (or part?) is sent to bottom vision for the very first time, it would do its best to perform stock bottom vision, but then interrupts and ask the user if it is okay. They could perhaps adjust the pipeline parameter (thresholds etc.). 

Or, for difficult parts, they could switch to template vision: OpenPnP asks the user to align by hand, then it takes a template picture. Once teached-in, the next parts can be aligned by the template picture. You no longer need to carefully isolate bright contacts etc. with a bent pin or a speck somewhere throwing the whole thing off. Template matching is super robust, it will just nail it by the sheer force of correlation.

It would be even more useful if combined with multi-shot vision. Each corner of a part could be template-taught-in. This would open it for those impossible asymmetric  connectors etc.

For this I would need a very robust system for multiple continuations. Hence the urge to do it right.

> I think that exceptions in general are serious errors and hence shall be handled carefully. For me, its ok to skip further continuations after one of them has thrown an exception.

Note the careful catching of only `ExceptionWithContinuation` in my proposal. Only for those will there be a "re-continuation".

> I agree with your suggestion, that one could just throw the continuation to calibrate the nozzle tip away and add one that resumes the job.

Good.

> That would provide the added advantage, that it would speed thinks up by grouping nozzle tip loads and calibrations into separate steps again (as was before PR https://github.com/openpnp/openpnp/pull/1609). (Today a continuation might calibrate the nozzle tip immediately after loading and before the job resumes. If a second nozzle tip change is required, the head goes to the location where the nozzle tip change will take place. For me this locations are the same. Hence this results in extra movements that could either be changed by configuring the nozzle tip calibration to InJob or with your suggestion.)

Good catch! I agree.

> Please advice...

Personally, I don't like half-measures that sometimes work, sometimes not. So ...

- Either skip nested continuations altogether and leave it to JobProcessor to resume. No dequeue needed.
- Or implement queuing fully, including re-catching `ExceptionWithContinuation`. Which I think should work with my code (should be easy to test). But still needs figuring out how to properly instruct the user (message text). 

_Mark

janm012012:
Hi  @markmaker!

[...]
> Or, for difficult parts, they could switch to template vision: [...]

I like the idea and we discussed that a few times already. IIRC there is no rotated template matching available yet but there are cleaver public implementations around (unfortunately in C/C++). If there is anything I can help, please let me know.

> For this I would need a very robust system for multiple continuations. Hence the urge to do it right.

I'm with you. Make it right in the first place is better then being forced to fix it later in a hurry.

[...]
> Personally, I don't like half-measures that sometimes work, sometimes not. So ...
> 
>     * Either skip nested continuations altogether and leave it to JobProcessor to resume. No dequeue needed.
> 
>     * Or implement queuing fully, including re-catching `ExceptionWithContinuation`. Which I think should work with my code (should be easy to test). But still needs figuring out how to properly instruct the user (message text).

I'll take the second approach and test your code.  Shall be easy with the test-code I've still in place.

In addition it would be probably advantageous to implement your other suggestion to remove the nozzle-tip-calibration continuation and replace it with just a job-resume.

Jan

janm012012:
Hm... I got stuck at compiler error "Local variable combinedContinuation defined in an enclosing scope must be final or effectively final", which is generate at "combinedContinuation.thrun(); // outer" which is almost the last effective line in your suggestion. It seems that the local variable can not be referenced in the (local) lambda. Do you have any idea how to fix that?

Jan

markmaker:
Okay make a copy of the variable before the lambda:

```java
                    } else {
                        final Thrunnable outerCombinedContinuation = combinedContinutation;
                        combinedContinutation = (() -> { 
                            // combine the two
                            try {
                                continuation.thrun(); // inner first
                            }
                            catch (ExceptionWithContinuation e) {
                                throw new ExceptionWithContinuation(e, () -> {
                                     outerCombinedContinuation.thrun(); // requeue in case of exception
                                });
                            }
                            outerCombinedContinuation.thrun();
                            return null;
                        });
                    }
```
_Mark

janm012012:
Hi @markmaker!

I've now rebased the code onto current test branch and added an option to skip nozzle tip calibration as part of the loading step to calibrate it as step of the job processor. If multiple tips are to be changed at once, this speeds things up as loading and calibration are grouped together requiring only one move from changer to camera.

Jan

markmaker:
> added an option to skip nozzle tip calibration as part of the loading step to calibrate it as step of the job processor.

1. Hmmm... I'm not happy with that addition of another user option (`calibrateNozzleTipsInJobProcessor`), completely removed from the calibration settings, which otherwise are all to be found on the NozzleTip. Who will expect to find that option there on the Job Processor? Plus there is already a similar sounding distinction in the **Automatic Recalibration** triggers:

   > - **On Each Nozzle Tip Change**: Whenever a nozzle tip is changed, it is immediately calibrated. This also happens when the machine is homed, and a nozzle tip is already loaded.
   > - **On Each Nozzle Tip Change in a Job**: Whenever a nozzle tip is changed inside a job, it is calibrated in its proper Job Processor step. This also happens when a nozzle tip was already loaded when the job starts. However no automatic calibration is done outside the job.

      https://github.com/openpnp/openpnp/wiki/Nozzle-Tip-Calibration-Setup#automatic-recalibration

   The second trigger already kinda does the same thing as the new option (but not quite). In my experience the average new user does not understand such "overlapping" concerns.

   Therefore, I suggest you remove the UI, but keep the `@Attribute`, defaulting to the _optimized_ behavior (but see below about polarity). In the unexpected case of users having a problem with it, they can edit the` machine.xml`. And then you can solve their underlying problem.

2. The polarity of the parameter `skipCalibration` should be positive, i.e., `withCalibration`. It is unexpected to call   `loadNozzleTip(nozzleTip, false);` and then calibration is actually **en**abled.
3. The name `calibrateNozzleTipsInJobProcessor` is misleading. It can mean both polarities at the same time, either interpreted as "immediately" or as "postponed to proper step". I actually misunderstood the polarity first. Assuming you  flip polarity as in (2.) it should be called `allowImmediateNozzleTipCalibration` and it should be defaulting to `false`, to suppress the regular calibration.
4. In `ReferenceNozzle`:

    ```java
    @Override
    public void loadNozzleTip(NozzleTip nozzleTip) throws Exception {
        loadNozzleTip(nozzleTip, false);
    }
    ```
    This can be moved to `Nozzle` as a `default` method implementation. Don't forget to flip the polarity in case you follow the previous advice.

5. Another concern is Z calibration. I always wanted to add auto-focus based Z calibration. Obviously it would best be performed together with nozzle tip calibration, when already above the camera. Needless to say this is not made easier with this PR. But I guess we can cross that bridge when we come to it.

_Mark

janm012012:
Hi @markmaker!

I'm sorry for making things more complicated. It would be easy to remove that part, if you think its better off in a different PR.

Basically, I added the UI to the job processor as I consider it a job processor optimization and not a general nozzle feature.

I've made all changes you mentioned, each in a separate commit so that you can easily follow the changes.

Jan

markmaker:
Excellent, @janm012012 , can this be merged?
_Mark

janm012012:
Hi @markmaker!

I reviewed my changes from the web perspective and found, that I missed to remove loadNozzleTip(NozzleTip nozzleTip) from ReferenceNozzle.java. I also reverted some indentation and import statements to their current state in the test branch. Other then that, I'm still happy with the code and don't see any reasons not to merge it.

 Jan

markmaker:
Thanks, @janm012012, nice improvement!

PR #1618, opened 2024-03-25
add support for grblHAL dialect

heeplr:
# Description
[grblHAL](https://github.com/grblHAL/core) devices are currently not autodetected. This PR:
* adds a new GCode dialect for grblHAL
* detects the dialect from the M115 output
* add solutions for ```COMMAND_ERROR_REGEX``` and ```ACTUATE_BOOLEAN_COMMAND```

It probably isn't complete as we're still in the process of setting up the machine. But in case anyone likes to experiment with grblHAL controllers and because of fine granularity PRs and "release early", follow up PRs could be created if this is welcomed.

# Justification
[grblHAL](https://github.com/grblHAL/core) is a fork/successor of grbl and it's nicer without having to enter all command regexes manually.

# Instructions for Use
e.g. 
* get a Raspberry Pico board
* build & flash [grblHAL](https://github.com/grblHAL/RP2040/) with the [openpnp plugin](https://github.com/grblHAL/Plugin_OpenPNP) (or use my [fork](https://github.com/heeplr/RP2040) which has the plugin added already).
* connect the driver and press "Detect Firmware"
* check issues & solutions to set gcode commands
 

# Implementation Details
1. Tested with physical hardware and "nvm package" run.
2. I followed the prevalent style as close as possible
3. The changes shouldn't be critical but might need changes to make it fit more elegantly into OpenPNP?


markmaker:
Hi @heeplr,

thanks for taking this on. 

Always aiming for optimal usefulness, I have these questions/remarks:

1. Should go into the `test` branch. 
2. Is there a real difference between the already present Grbl and the GrblHal dialect on the G-code level? 
3. Could you perhaps make the `FIRMWARE_NAME` pattern more permissive? 
4. For instance how does Grbl signal errors? Or if there are differences, could you fall-through in the switch-case? 
5. The `ACTUATE_BOOLEAN_COMMAND` cannot be done like that, there is more than one actuator, and not all control simple IOs and want to use an index. You need to address this in `ActuatorSolutions`, where you also see that you need to add the value type specific actuation value variable, like `{True}`, `{False}`, `{DoubleValue}`, etc.

_Mark

heeplr:
Thanks for your remarks.

>     2. Is there a real difference between the already present Grbl and the GrblHal dialect on the G-code level?

I think so since GrblHAL supports setting a [compatibility level](https://github.com/grblHAL/core/wiki/Compatibility-level) to adapt to changes. But I never used grbl so I can't really say much about it. There's also a [list of "changes from Grbl 1.1"](https://github.com/grblHAL/core/wiki/Changes-from-grbl-1.1) I thought a new dialect might be the cleanest solution to avoid interference with the grbl dialect.
 
>     3. Could you perhaps make the `FIRMWARE_NAME` pattern more permissive?

You mean "Grbl" instead of "GrblHAL"? Sure, but I'd worry about "namespace collision" with other (future) grbl flavours that also use uppercase "G".

>     4. For instance how does Grbl signal errors? 

Not sure about original grbl but GrblHAL responds with ```error:x``` instead of ```ok``` while the error strings can be enumerated to a string with ```$EE``` or (```$EEG``` for grbl format). It'd be nice if that could be implemented in OpenPNP to show human readable error messages.

> Or if there are differences, could you fall-through in the switch-case?

I'm not sure what you mean with that.

>     5. The `ACTUATE_BOOLEAN_COMMAND` cannot be done like that, there is more than one actuator, and not all control simple IOs and want to use an index. You need to address this in `ActuatorSolutions`, where you also see that you need to add the value type specific actuation value variable, like `{True}`, `{False}`, `{DoubleValue}`, etc.

I also don't understand this completely. It currently "works for me" with ```M6{True:4}{False:5} P{Index}``` as ACTUATOR_COMMAND. I looked at ActuatorSolutions but I'm not quite sure how to implement it there or how to access the ```FirmwareType``` there. Would you briefly elaborate?

heeplr:
I think I found that at [ActuatorSolutions.java#L86](https://github.com/openpnp/openpnp/blob/test/src/main/java/org/openpnp/machine/reference/solutions/ActuatorSolutions.java#L86) I could ideally add something like:

```java
case Boolean: {
                    HashMap<CommandType, String []> suggestions = new HashMap<>();                    
                    String firmware_name = ((GcodeDriver) actuator.getDriver()).getFirmwareProperty("FIRMWARE_NAME", "");
                    if(firmware_name.contains("grblHAL")) {
                        suggestions.put(CommandType.ACTUATE_BOOLEAN_COMMAND, new String[] {
                                "{True:M64} P{Index} ; actuate "+qualifier+" ON\n"
                                        + "{False:M65} M4 P{Index} ; actuate "+qualifier+" OFF"
                        });
                    }
                    else {
                        suggestions.put(CommandType.ACTUATE_BOOLEAN_COMMAND, new String[] {
                                "{True:M\u00BF ; actuate "+qualifier+" ON}\n"
                                        + "{False:M\u00BF ; actuate "+qualifier+" OFF}"
                        });
                    }
```

But it seems I can't just include the final "{Index}" in the template. I'm stuck understanding how to provide a template for ACTUATE_BOOLEAN_COMMAND, that doesn't need further editing - just accepting of the solution.

heeplr:
> If Grbl and GrblHAL are so different that they have no common ground, then we can forgo points 1 through 4.

I suppose they have a lot of common ground, I'm just neither familiar with Grbl nor OpenPNP :)
I didn't find a lot of Grbl specific stuff in the code. If you think it makes sense to extend that, I'd change that. But since I can't test with real Grbl hardware, it might break things for users currently using Grbl.

> And even if you only want these simple features today, is this true for all users that also want to use GrblHAL in the future?

I thought since most other GCode Regexes are automagically filled in and only need a click on "Accept", it'd be nice to have it for actuators, too. I think all users would like to avoid looking up "M64/5 Px" for boolean actuators.

> OpenPnP might also have input actuators, like for vacuum sensing. What about these? They will now all get a `ACTUATE_BOOLEAN_COMMAND`, possibly with the wrong Index, causing strange behavior, when mistakenly actuated by the user.

Oh, I didn't see that. I thought it's just used when the actuator has "Value Type" == "Boolean" set in the GUI (and a GrblHAL GCode driver). Is that still the case in my previous snippet?

> Accessing the `FirmwareType` might be a problem. It should probably be stored or refactored somehow, so the ActuatorSolutions can either reach, or compute it too.

I think I found it but it's not very nice. Making FirmwareType a public datatype and providing some getter might make sense *IF* there's some clean way to provide dialect-specific solutions for actuators - which I'm not quite so sure now.

heeplr:
I now implemented suggestions based on the qualifier which now should suggest the correct command for GrblHAL firmwares (and only for those) and for each qualifier. I will add probe as soon as I get there hardwarewise.

I left the default since it's probably useful when it's set and it shouldn't interfere with other head mountables.

Does that look ok?

markmaker:
1. Do not filter for the `qualifier`, the suggestion is alright for all the Boolean actuators.
2. Conversely, please remove the default.

Have you tested it? Does it properly appear in the I&S Solution UI? 

(I'm a bit unsure because the default is on two lines)

_Mark

heeplr:
> Have you tested it? Does it properly appear in the I&S Solution UI?

yes






> (I'm a bit unsure because the default is on two lines)

It wouldn't need the ```\n```
For the default template ```java "{True:M\u00BF ; actuate "+qualifier+" ON}\n{False:M\u00BF ; actuate "+qualifier+" OFF}"``` , in ```extractBooleanCommand()``` the call to ```TextUtils.substituteVar()``` will generate two strings:
* ```M\u00BF ; actuate camera light ON\n``` for on, and
* ```\nM\u00BF ; actuate camera light OFF``` for off

as it replaces {x:y} with empty string if x is false.
This can't be used when using {Index} since y cannot include a } character.

> Conversely, please remove the default.

Are there really no cases where a boolean actuator would then go without a working command? 
It seems that a Signaler actuator doesn't appear in I&S and has no command set, unless I'm doing something else wrong.

markmaker:
I'm losely following this... this is getting better and better. 

You might consider switching the PR to `WIP` as long as you're still working on it.

One feed-back:

Please do not set options unconditionally. The user might want to Revert the solution, so it should be undone. Use `(state == Solutions.State.Solved)` as the setting, i.e. set to `true` when solved, `false` (reverted) otherwise.

```java
                                @Override
                                public void setState(Solutions.State state) throws Exception {
                                        gcodeDriver.setBackslashEscapedCharactersEnabled((state == Solutions.State.Solved));
                                        super.setState(state);
                                }
```
See other overrides of `setState()` to understand the principle. :+1: 


_Mark

markmaker:
Is this still WIP? It looks good and could be merged IMHO.

_Mark

heeplr:
Currently our hardware is delayed. Additionally to this, I also added a lot of the relevant ```$x=y``` grblHAL settings to the CONNECT, ENABLE and HOME gcode commands but didn't do a full test, yet. (not sure if that's good practice or if there's a better way - it eases setup and prevents fiddling in the console before firmware detection can work)

Wouldn't hurt to create a new PR for that if you like to merge. I don't have a strong opinion on this and also wouldn't mind to add it here. (I guess I can test in the next 2 weeks)

markmaker:
Okay no pressure :sunglasses: 

> Additionally to this, I also added a lot of the relevant $x=y grblHAL settings to the CONNECT, ENABLE and HOME gcode commands

TinyG (being derived from Grbl) uses `$x=y` settings too. Ultimately it proved best practice to just set a very few select settings. Like when we set the serial flow control mode, we also set the corresponding `$x`. But nothing else.

What I don't understand is why not just the CONNECT? Why also ENABLE and HOME? Aren't these permanent/quasi-constant settings that are stored in the EEPROM?

_Mark

heeplr:
> TinyG (being derived from Grbl) uses $x=y settings too. Ultimately it proved best practice to just set a very few select settings. Like when we set the serial flow control mode, we also set the corresponding $x. But nothing else.

ah, good to know. Thanks.

> What I don't understand is why not just the CONNECT? Why also ENABLE and HOME? 

Since when there's an error or ALARM, the controller needs a ctrl-c followed by an $X to abort/reset alarm. By doing this before every enable/home, no reconnect is needed and disable/enable can recover from most unrecoverable errors.

> Aren't these permanent/quasi-constant settings that are stored in the EEPROM?

yes, that's the disadvantage. Writing them everytime is the price for first-time-setup not needing any console/terminal hacking.
I also send non-necessary settings commented out, added commented descriptions to each setting. Maybe a note in I&S to review those settings might make sense?

I looked for another way to somehow provide setting docs and a sane default config to the user but found none. Maybe a short "Info" text depending on the firmware (or even better, usb vendor/device id) would be useful for other dialects aswell? Not sure...

markmaker:
> yes, that's the disadvantage. Writing them everytime is the price for first-time-setup not needing any console/terminal hacking.

You should check in the firmware source code whether they are actually written when unchanged. TinyG does not.

But setting such values dynamically between two (or more) values might (?) limit the lifetime of the EEPROM and thus MCU. If it is an EEPROM at all, most MCUs nowadays use flash to simulate the EEPROM, methinks. Whether that is better regarding how many times it can be written, I frankly don't know. :thinking: 

_Mark

heeplr:
> But setting such values dynamically between two (or more) values might (?) limit the lifetime of the EEPROM and thus MCU.

yes, even with flash it would wear the NVM unnecessarily.

> You should check in the firmware source code whether they are actually written when unchanged. TinyG does not.

That's actually an excellent point. I will check that soon (and do a grblHAL PR if it's not the case - should be trivial).

Maybe it's best to just merge the current state since it's working, and discuss the details of changed/added default commands in a new PR in a few days to keep PRs tidy?

markmaker:
Still loosely following this...

I'm still not all that happy with the very specific and large `$xxx`-setup in `CONNECT_COMMAND`. If it were an off-the-shelf/kit machine that specifically and exclusively comes with that controller and that many other users buy too, then it would be fine. But I understand your machine is unique, right?

Given that, and with an EEPROM/Flash present to persist the settings, I still think it is best to configure the controller _outside_ OpenPnP, by just using [Printrun](http://www.pronterface.com/index.html) or similar. 

Note, this is actually the standing recommendation. Always get the raw motion control running first (e.g. with Printrun) like that, and only switch to OpenPnP once all the electromechanical and controller configuration issues  (like axis polarities, steps/mm, end-switches etc.) have been figured out and tested. Otherwise, you will be unnecessarily mixing up two problem domains and make it _much_ harder to diagnose and solve issues.

Imagine another GrblHAL user actually follows that recommendation, but then does not pay enough attention to the proposed  `CONNECT_COMMAND`, accepts it, and therefore screws up their carefully crafted config!

I hope you understand. :sweat_smile: 

Note, I'm open to improve the (now very rudimentary) console inside OpenPnP, so it could replace Printrun for that matter. And/or you could make a new `SETUP_COMMAND` as a separate deliberate command to be sent via press of extra button from the console (instead of sending on each `CONNECT_COMMAND`). And/or adding controller config parameters to the OpenPnP Machine Setup tree, with proper UI, so that the  `SETUP_COMMAND`-scripts could be _parametrized_. This would make it more useful and versatile towards different machines and controllers, which in turn justifies building such scripts into OpenPnP and I&S. But that would require quite some extra work. :sunglasses: :rocket:

_Mark

markmaker:
> What do you think about adding a completely new step  or some COMMAND_SETUP or COMMAND_INITIAL_CONFIG ...

Agree! 

But I believe it would be cumbersome to edit via I&S, and hard to revisit later. Instead it should just be that new command, that is not automatically sent on connect (i.e. cannot ruin your day), but can be edited and revised in the G-code tab, like every other G-code command. As always, the blue `[i]` button can link to a Wiki page explaining better how to go there and select the right command (not sure it currently does that).

And there could be a new button on the Console tab, that sends the `COMMAND_SETUP` or whatever it will be called.

_Mark

heeplr:
That sounds very elegant. Is someone currently working on the console able to implement that? 
I'm currently working on another project but will surely come back to this PR eventually. 
I could have a look then but of course it'd be very nice if someone more familiar with the code could add that "Send Controller Config" button.

sham-cyber:
GRBL is a many page software contains c.h. how to make work in Arduino ,pl quote

raymondbh:
Have you checked what the OpenPnP plugin for grblHAL adds? https://github.com/grblHAL/Plugin_OpenPNP
I'm rebuilding a NeoDen YY1 to run OpenPnP, and I have compiled grblHAL to run on the original controller with success.
I have a lot of tweaking to do yet, but it looks good so far...

heeplr:
> Have you checked what the OpenPnP plugin for grblHAL adds? https://github.com/grblHAL/Plugin_OpenPNP

no, i missed that completely. Thanks for referencing. It seems like a more viable solution to use a 3rd party plugin.

This PR isn't complete either and development has pretty much stalled, since I currently don't have access to the hardware and the owner shows little interest to completing it. This could change anytime which is why I leave this open. But I wouldn't bet on it.

Contributing to/Forking that plugin might be the better solution long term and would offload maintenance workload from the OpenPNP project.

raymondbh:
> But it's not complete, yet.

Sorry if I mislead you...
What remains to make it complete you think?

heeplr:
> What remains to make it complete you think?

Beyond the conversation with markmaker above, I'm not sure since I stopped at the point where the cameras need to be configured.
I could move the head from within OpenPNP and error handling until that point worked (for me™).

The chance that it "just works" is not zero if there were no breaking changes in the meantime.

terjeio:
I am the grblHAL maintainer and I am currently working with a user setting up grblHAL for OpenPNP.

grblHAL is higly configurable/customizable due to its HAL and supporting plugins that can extend functionality. The [core](https://github.com/grblHAL/core) is squarely aimed at CNC machine control and adheres to the [NIST RS274NGC gcode specification](https://www.nist.gov/publications/nist-rs274ngc-interpreter-version-3), with most extensions based on the LinuxCNC specification and a few taken from other industry standards in the CNC world. _Plugins_ can and do implement gcode extensions (M-codes only) that are either according to non CNC based gcode "standards", such as for 3D printers, or are grblHAL specific.

The core supports up to 8 axes and up to 254 analog or digital inputs/ouputs thus making grblHAL, IMO, in combination with a suitable controller, a good candidate for OpenPNP. This is the reason for making the [OpenPNP plugin](https://github.com/grblHAL/Plugin_OpenPNP/), which I started in cooperation with a user that since abandoned his project - leaving it in its current WIP state (I do not have a PnP machine).

I am keen to get the plugin completed but are frustrated by the seemingly incompatible, and sometimes badly specified, variations of M-code commands that OpenPNP has to interact with.

I am open to have a discusstion of how to get grblHAL working with OpenPNP in a streamlined and futureproof way. Compliance with clear specifications is, IMO, a must. If these are adopted from previous work or made anew does not matter for me - but should not be in breach of principles laid down in the NIST RS274NGC gcode specification.

PR #1619, opened 2024-03-26
PhotonFeeder: Add vision pipeline for calibration

nbp:
# Description
This change adds an optional vision pipeline to the PhotonFeeder to be used in cases where is a lot of variance, or as a mean to (re-)calibrate at start-up or periodically during a job.

# Justification

This change is limited to PhotonFeeder users. Without this change some users might experience a lot of variance in the picking location.

# Instructions for Use

Using the vision pipeline added in this PR, one can either:
 - Re-calibrate each time the feeder is pugged in: The variance history should be cleared each time the feeder settings are change, and setting a varianceHistory of 1 will run the vision pipeline once and never more after each re-configuration of the settings.
 - Wait for low-variance sections: Using a larger amount of varianceHistory will aggregate more samples using the vision pipeline until the feeder is capable to provide a reliable position, in which case the vision pipeline would no longer be used once a steady state is reached.
 - Always use vision: Set the variance history to 0 and always recalibrate before picking a component. {to be verified again}

In addition an extra parameter to skip some recalibration helps handle cases where a machine would have low variance between successive picks but might accumulate large variance over time.

The vision pipeline relies on the fiducial location of the feeder (located in front of the tape), and the approximate offset of the pocket compared to the fiducial. This is used to determine the tape orientation of the feeder. {back position not tested yet}.
Then the pipeline compensates for the tape dimensions to locate the holes {tested with 8mm feeders, not yet tested with 12mm feeders} and scans for holes using circular symmetry in a narrow band centered on the expected hole location. It scans for 4mm in order to at least detect one hole, and compensate the picking location using the detected hole.

User settings are not modified, and saving the feeder configuration should reset any of the variance history causing the re-calibration process to be executed once more.

# Implementation Details
This PR remains as a draft as many corner cases remain to be tested. Otherwise this changes have been tested successfully across multiple jobs / boards using 8mm tapes. The points to be tested are annotated with curly brackets above.

The coding style might have to be revised due to the fact that the PhotonFeeder classes seems to be using tabs, as opposed to the recommended 2-spaces from the coding style.

`mvn test` pass with the existing PhotonFeeder test, however this PR is opened as a draft as it would be even better to add vision tests, and check the various configurations of vision settings.

This PR also includes additional Nix files such as `flake.nix`, which might be moved to a different PR, as a way to provide a single command approach for others to reproduce custom builds of OpenPnP using `nix run 'github:nbp/openpnp/photon-vision#openpnp'`, which will clone & build & test & cache and execute the custom OpenPnP version in a controlled environment.

(cc @Jnesselr )


markmaker:
Hi @nbp,

thanks. Please keep us posted about progress.

> This PR also includes additional Nix files such as flake.nix, which might be moved to a different PR

Yes please, and in that PR, please explain why this is superior to just using git/mvn. Thanks.

_Mark

PR #1620, opened 2024-04-01
Fix unload actuation as false for automatic nozzle tip changer

janm012012:
# Description
This PR changes the actuator usage for automatic nozzle tip changers to use False for unload and True for load. Previously it was using True for both load and unload. 

# Justification
According to https://groups.google.com/g/openpnp/c/hH1CKmOWX8I/m/TV64xricAgAJ this is considered a bug which shall be fixed ASAP. In addition it adds support for non-symmetric load/unload sequences as required eg. for elevator style changer, which requires a something like open, up, close, down on load and up, open, down, close on unload. 

# Instructions for Use
All users currently using automatic nozzle tip changing, shall verify their setup to make sure they are not relying on actuators being True only.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **not tested, code change is to small**
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? **not applicable**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **on**
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. **not applicable**

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
@@ -776,7 +776,7 @@ public class ReferenceNozzle extends AbstractNozzle implements ReferenceHeadMoun
 
                     Actuator tcPostThreeActuator = getMachine().getActuatorByName(nt.getChangerActuatorPostStepThree());
                     if (tcPostThreeActuator !=null) {
-                        tcPostThreeActuator.actuate(true);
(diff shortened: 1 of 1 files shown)


markmaker:
Brilliant @janm012012 :1st_place_medal: probably the fastest PR ever :rocket: !

PR #1623, opened 2024-04-16
Update ReferencePushPullFeeder to use FeederVisionHelper

prusk:
# Description
Update the `ReferencePushPullFeeder` to remove the FindFeatures inner class and use the `org.openpnp.util.FeederVisionHelper`.

# Justification
This is consolidating the vision detection of tape features in a feeder (holes & part pocket). The PR will reduce code duplication and simplify maintenance of this functionality

# Instructions for Use
The change does not modify any usage instructions.

# Implementation Details
* Ran all the automated tests
* Tested the new feeder type on my personal hardware
* Yes
* No changes to `org.openpnp.spi` or `org.openpnp.model` packages
* All tests passed

PS: @markmaker please take a closer look at the previous line 1478: https://github.com/openpnp/openpnp/blob/test/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L1478
`if (autoSetupMode == FindFeaturesMode.OcrOnly || calibrationTrigger == CalibrationTrigger.None) {`
The first condition is handled correctly but I would appreciate a second set of eyes on the second condition that was not implemented in the FeederVisionHelper due to calibrationTrigger being out of scope for the Utils class.
However, I could not find a scenario where the second condition would be useful and after testing I found no difference in functionality by only having the first condition.

I will leave this PR in draft until this aspect is clarified.


markmaker:
Hi @prusk,

This PR is difficult to assess, as I don't see a diff between the inner class and the helper. Can you please make and post such a fragment diff, with white-spaces ignored?

A developer changing `FeederVisionHelper` in the future must be aware that pipeline XML and the helper class go hand in hand. I propose the facility to get the default pipeline to also be moved to `FeederVisionHelper`, and should be taken from resources, similar to this:

https://github.com/openpnp/openpnp/blob/32dcd338fe69f1daff99908fb629415cfa7694ce/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L1293

Note, pipelines can _still_ be different by feeder class if absolutely necessary (is it?), but code and XML should be noticeably "together", and work the same way via resources, where they can also be edited or copy&pasted much easier by the (future) developer.

> please take a closer look...

As far as I see `calibrationTrigger == CalibrationTrigger.None` can also be implemented on the calling side, wherever the `findFeatures()` is called, by checking the condition there and passing `FindFeaturesMode.OcrOnly` instead of the original mode if true. 

Having calibration disabled must be supported, OCR/QR must still work. One such case was transparent tapes that could not be reliably detected, while OCR still works. This is likely no longer the case with the CircularSymmetry pipeline that can even handle transparent tapes, but I like to retain the possibility to switch calibration off.

Note, it is possible this was not completely optimized, I guess there are calling paths where the vision operation could be cancelled earlier when OCR is _also_ disabled, or not wanted. Feel free to augment this. :smile: 

The `findFeatures()` calls are very repetitive, just passing the same feeder class members again and again. If you add one in the future it becomes a major change. I propose to refactor this by making a subclass of `FeederVisionHelper` (now `AbstractFeederVision`?)  per feeder class, and just passing `this` to the sub-class constructor, where the feeder specific settings are passed. Consider renaming the `findFeatures()` back to parameter-less `invoke()`, it was much more elegant.

_Mark

prusk:
Hi @markmaker,

One step at a time :)

The diff for `FindFeaturesMode` enum is insignificant, just renaming the last item to "CheckOnly" as it made more sense to me. It does the same thing as OcrOnly. Let's put that aside for now, call it open item 1. The name makes no difference.

I specifically left the CV pipeline definition out of this feature-detection class to allow each feeder to chose theirs. There might be value to consolidate the pipelines definitions as well. Let's keep this open as item 2.

The findFeatures() calls with multiple tape-specs parameters indeed look suboptimal. We could make something else and pass it along as `this`. Open item 3.


Back to that condition question...
[FindFeatures.patch](https://github.com/openpnp/openpnp/files/15000792/FindFeatures.patch)

Attached is the diff between the FindFeatures class and the FeederVisionHelper class. I intentionally left the Ocr part inside the utils class with the thought of this present PR, even though I don't need it in BambooFeeder.
As you can see, the differences are very few:
* +FeederVisionHelper has some additional privates for tolerance, EIA specs, internal locations (part, hole1, hole2), tape specs (part & feed pitch) and minimal settings (normalize, snap)
* +FeederVisionHelper - some basic methods to work with internal settings, tape specs and locations
* +FeederVisionHelper - getters and setters for privates
* +FeederVisionHelper required transform methods
* +-FeederVisionHelper - finally, a few get function calls were replaced by local private attributes (location/partLocation, hole1location, hole2Location, etc...)
Essentially other than that specific condition, I don't see any real difference between the two...

Here's what I tracked down in terms of function calls in FindFeatures / PushPullFeeder

FindFeatures.invoke() called from:
-showFeatures -> FindFeaturesMode=null but _calibrationTrigger_ condition doesn’t matter for showFeatures()
-autoSetupPipeline -> FindFeaturesMode.FromPickLocationGetHoles -> _calibrationTrigger_ condition doesn’t matter for autoSetupPipeline that is only called from the wizard UI
-performVisionOperations (Line 2499) -FindFeaturesMode.CalibrateHoles -> _calibrationTrigger_ condition doesn’t matter ? We know the _autoSetupMode_ condition is false, but if _calibrationTrigger_ is true should the CV calibration be skipped?
I guess we could add a check before this call, something like this:
`FindFeaturesMode _findMode_ = (calibrationTrigger == CalibrationTrigger.None) ? FindFeaturesMode.CheckOnly : FindFeaturesMode.CalibrateHoles;
feature = new FindFeatures(camera, pipeline, 2000, _findMode_).invoke();`
-performVisionOperations (Line 2544) -FindFeaturesMode.OcrOnly / CheckOnly -> calibrationTrigger condition doesn’t matter because autoSetupMode condition is true 

performVisionOperations called from:
-autoSetupPipeline [OK] - _calibrationTrigger_ irrelevant for autoSetup
-obtainCalibratedVisionOffset
-performOcr

obtainCalibratedVisionOffset called from:
-assertCalibrated

performOcr called from:
-prepareForJob

assertCalibrated called from:
-prepareForJob
-feed



What do you think about that _calibrationTrigger_ condition? Should I add the if for findMode in the first call to invoke() inside performVisionOperations?

Next we'll tackle the other open items...

markmaker:
Ooops, this one got off my radar... sorry. :see_no_evil: 

Thanks for the patch, it makes sense.

You moved the coordinate transformation to the vision helper. Is this also thought to be universal? Then you should also replace the `ReferencePushPullFeeder`'s own methods, right?

> I guess we could add a check before this call, something like this:

I think that covers it. I came to the same conclusion. 

Obviously, all this has to be announced in the discussion group and hopefully quickly tested by `ReferencePushPullFeeder` users. And I'm assuming you would be ready to fix any bugs that crop up :smile: :lady_beetle: :bug: 

> PS: about switching calibration off - totally agree. ... not easy to work with.

Yes, there has been some "organic growth". Feel free to improve it :sweat_smile: 

_Mark

prusk:
Hi @markmaker,

> You moved the coordinate transformation to the vision helper. Is this also thought to be universal? Then you should also replace the ReferencePushPullFeeder's own methods, right?

You mean these 5 methods should be public in the VisionHelper? That sounds like a good idea.
`forwardTransform(Location location, Location transform)`
`backwardTransform(Location location, Location transform)`
`getTransform(Location visionOffset)`
`transformFeederToMachineLocation(Location feederLocation, Location visionOffset)`
`transformMachineToFeederLocation(Location machineLocation, Location visionOffset)`

For the support related to these changes, of course I'll be ready to fix any bugs.
No idea yet how to improve the switching off of calibration but I'm thinking about it...

I'll prepare the rest of the changes so the PR is ready to merge and then start a thread in the Google group.

prusk:
I have one quick question @markmaker : how important is this check? https://github.com/prusk/openpnp/blob/feature/updated-referencepushpullfeeder/src/main/java/org/openpnp/util/FeederVisionHelper.java#L200
It is making my head hurt while trying to make all 5 transforms as static. 

If we don't really achieve much with that condition and I can remove it, the getTransform can be simplified and I can make it static and it's only declared in the Helper class.

prusk:
Hi @markmaker,

I just pushed another commit that achieves the following:
* makes all the 5 transform methods static so they don't need to be redefined in the feeders
* adds a class for storing the various FeederVisionHelper parameters - this simplifies calling the new() & findFeatures
* solves the question about that condition for CalibrationTrigger
* default CV Pipelines are now included in the FeederVisionHelper -> still have to wrap my head around using these instead of redefining the pipeline in the feeder classes

I think all the open items are addressed now.
Take a look please and let me know what you think.

prusk:
All done, figured out how to make use of the pipeline inside the Helper.
I left the PushPullFeeder without it for now so you can compare between the two approaches.
After that I can do the same for PushPull like I did for the AbstractPandaplacerVisionFeeder

prusk:
Hi @markmaker,

Did you have a chance to review the latest updates?

Thanks.

markmaker:
Sorry, completely swamped at work. I sometimes find time to answer a question or two in the group, but not to delve deep into this review... Sorry. :sweat:

prusk:
Hi @markmaker,

Any idea when we should be expecting the review?

Thank you

markmaker:
Hi @prusk 

sorry about this. But not looking good (even worse at work).

But this is an Open Source project, and I'm merely one contributor, so maybe someone else could have at least a a _first look_. Maybe @janm012012 or @crashmatt or @tonyluken etc.

_Mark

prusk:
Hi @markmaker ,  @janm012012 , @crashmatt , @tonyluken 

I just force-pushed small commits to make the changes much easier to follow by examining one commit at a time.
This should simplify the review process and hopefully we can merge this PR.

For the record, I applied all the changes to ReferencePushPullFeeder, including the switch to the Helper's pipelines. I think the PR is ready for merge once reviewed, therefore I'm taking it out of draft.

Thank you.

janm012012:
Hi @prusk!

Many thanks for your contribution! I'm a fan of joining double code instances into one and I'm a fan of running feeders without runtime calibration. As far as I can see, the later is preserved by your PR.

However, I don't understand your construction of the FeederVisionHelperParams that is than passed over and over again to the actual helper. It looks just like procedural programming. IMHO code and data shall be grouped into one class and this could then replace the vision features in the ReferencePushPullFeeder. Just my 5 cents...

Jan

prusk:
Hi @janm012012 

> However, I don't understand your construction of the FeederVisionHelperParams that is than passed over and over again to the actual helper. It looks just like procedural programming. IMHO code and data shall be grouped into one class and this could then replace the vision features in the ReferencePushPullFeeder. Just my 5 cents...

As Mark suggested above, it is better to just use a small class for passing parameters to the FeederVisionHelper instead of a long list of individual values. This is simply because I wanted to preserve as much as possible the existing implementation from ReferencePushPullFeeder where each time there's some need for CV calibration, a new instance for the FindFeatures class is created.
With that in mind, the new FeederVisionHelper needs the correct parameters every time it is instantiated but there's many of them for maximum flexibility.
So rather than having a list of 12 params in the constructor, building this FeederVisionHelperParams makes the feeder code more readable.
Essentially the params class is there just for convenience so things look nice in the feeders code with a small 2-line call.

Thanks

janm012012:
> As Mark suggested above, it is better to just use a small class for passing parameters to the FeederVisionHelper instead of a long list of individual values.

Agreed. A long list of arguments is difficult to maintain.

However, then I've to revise my question: why is it _created_ multiple times? If a feeder has vision features and the feeder is created once, why is it's vision created more then once? I'm wondering, why, after all your work to group/abstract vision features and use it multiple times, feeders can't "just" inherit it, Then they would all have default vision features and some might overwrite specific aspects if needed. (I think that's not possible at present except for the configuration stored in the param class.) Then also the UI could be abstracted and inherited (maybe later).

Jan

prusk:
I think it makes more sense to have the vision class instantiated only when needed.
Once vision calibration is done, memory is released and on next runs of OpenPnP it’s likely vision is not even loaded if calibration is not desired when parts are picked.
I also considered a permanent feeder property to avoid multiple instances of vision class but the above logic made me abandon the idea.

markmaker:
@janm012012,

just a quick side remark:

> code and data shall be grouped into one class

You are right. But the params cannot directly stored inside the helper class, because we need to preserve `machine.xml` compatibility. If @prusk were to move them there, it would break existing feeders.

_Mark

janm012012:
Hi @markmaker, @prusk!

> You are right. But

Ok, design decision. It's a pity that Niclas never managed to finish replacing the xml parser...

Anyhow, other then

> Consider renaming the `findFeatures()` back to parameter-less `invoke()`, it was much more elegant.

this PR looks good to me.

Jan

prusk:
Thanks @janm012012 

original `FindFeatures.invoke()` - made sense

`FeederVisionHelper.invoke()` - weird
`FeederVisionHelper.findFeatues` - nice

crashmatt:
Wrong season to get any meaningful attention from me.  The snow has gone
and the sun is up.   Try again in November when its dark and raining.

prusk:
Can we merge?

markmaker:
Hi @prusk,

sorry about the delay. I've finally had a look. 

Needless to say this is a large and hard to assess change, as code portions were moved around between files. It is impossible to catch issues just from this diff. I hope you tested it thoroughly. :smiley: 

As far as I can tell, it looks good!

The only thing that I don't yet understand is the duplication in the helper's `getPartLocation()` and both feeder's respective `getPickLocation()`, containing the pitch logic duplicated. Couldn't the latter simply be a call to the former? Am I missing something?

Caution: if you de-duplicate the two, you'd have to include the rotation in tape!

_Mark

andvd:
Hi @markmaker,

Thank you for taking a look.


> The only thing that I don't yet understand is the duplication in the helper's `getPartLocation()` and both feeder's respective `getPickLocation()`, containing the pitch logic duplicated. Couldn't the latter simply be a call to the former? Am I missing something?

The reason for this is: the PushPullFeeder implements something called "multiplier" but the helper class doesn't. I consider this to be an edge case, not something most feeders would implement, therefore I did not add it in the `getPartsPerFeedCycle` in neither the Helper class or in the Abstract one.
This concept of "multiplier" changes the `getPartLocation()` behaviour but because it's specific to PushPull feeder, I couldn't implement it as a public method in the Helper class.

prusk:
Hi @markmaker,

Thank you for taking a look.


> The only thing that I don't yet understand is the duplication in the helper's `getPartLocation()` and both feeder's respective `getPickLocation()`, containing the pitch logic duplicated. Couldn't the latter simply be a call to the former? Am I missing something?

The reason for this is: the PushPullFeeder implements something called "multiplier" but the helper class doesn't. I consider this to be an edge case, not something most feeders would implement, therefore I did not add it in the `getPartsPerFeedCycle` in neither the Helper class or in the Abstract one.
This concept of "multiplier" changes the `getPartLocation()` behaviour but because it's specific to PushPull feeder, I couldn't implement it as a public method in the Helper class.

prusk:
Looking at it again, I could probably remove the duplication between `FeederVisionHelper` and `AbstractPandaplacerVisionFeeder` but due to that multiplied edge-case I would leave `ReferencePushPullFeeder` as-is.

Is this something I should do?

prusk:
Hi @markmaker ,

I added a new commit: "_Consolidate getPartsPerFeedCycle & getPartLocation in FeederVisionHelper_"
This simplifies getPickLocation for non-multiplier feeders.

markmaker:
> The reason for this is: the PushPullFeeder implements something called "multiplier" but the helper class doesn't. I consider this to be an edge case, not something most feeders would implement, therefore I did not add it in the `getPartsPerFeedCycle` in neither the Helper class or in the Abstract one. This concept of "multiplier" changes the `getPartLocation()` behaviour but because it's specific to PushPull feeder, I couldn't implement it as a public method in the Helper class.

No this is not an "edge case", its a valuable speed optimization for all kinds of feeders that are not "self propelled". 

[Quoting the Wiki](https://github.com/openpnp/openpnp/wiki/ReferencePushPullFeeder#tape-settings):
> The **Multiplier** allows you to actuate the feeder multiple times to feed more parts per serving, as a speed optimization. This may reduce the feed time per part because the actuator is already at the right place and/or engaged in the mechanics.

Please implement the multiplier in the helper. You also need it for visualization (numbering the parts on the camera image), whenever the `ReferencePushPullFeeder` has a multiplier. This would no longer have worked, glad we caught this.

Just pass `1` for the `AbstractPandaplacerVisionFeeder`.

_Mark

prusk:
Hi @markmaker ,

> No this is not an "edge case", its a valuable speed optimization for all kinds of feeders that are not "self propelled".
This is the very definition of edge-case: a feature that applies to only a subset of feeders, the ones that are not self-propelled.

Anyways, redid the last commit to include the multiplier. I also tested with multiple values and everything works for the PushPullFeeder just as it did before.
With this change I was able to consolidate the getPartsPerFeedCycle and getPartLocation for the PushPullFeeder too.

markmaker:
Nice, thank you. :smile: 

Just one cosmetic thing: there is white-space inconsistency on some latest changes. OpenPnP [wants all spaces, no tabs](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style). :nerd_face:

Otherwise ready to merge?

_Mark

prusk:
Whitespace fixed. I seem to have problems with my editors adding tabs instead of spaces... 
Thanks for reminding to do this check. It also makes the PR nice with 14 commits instead of the dreaded no 13 :)

prusk:
Yes, ready to merge.

markmaker:
Thanks, @prusk :100:

PR #1628, opened 2024-05-27
Fix broken Camera permissions on macOS

prusk:
# Description
Attempt to fix camera entitlement on macOS Monterey - Sonoma.

# Justification
The Test build does not request access to Camera devices at least on macOS Monterey and Sonoma and the app crashes with the message `This app has crashed because it attempted to access privacy-sensitive data without a usage description. The app's Info.plist must contain an com.apple.security.device.camera key with a string value explaining to the user how the app uses this data.`

However, the `NSCameraUsageDescription` key is present and while testing I noticed a difference between having whitespaces vs not in Info.plist. There is also a difference about these carriage returns between the Test build and the Develop build, where Develop doesn't have them and works as expected.

The whitespace was introduced in https://github.com/openpnp/openpnp/pull/1534

# Instructions for Use
No special usage instructions.

# Implementation Details

- Tested on local macOS Monterey machine

- Yes

- no changes in this space

- tests passed

### DIFF
--- a/OpenPnP.install4j
+++ b/OpenPnP.install4j
@@ -69,11 +69,8 @@
           <directory name="native/${compiler:native_path}" />
         </nativeLibraryDirectories>
       </java>
-      <infoPlist>
-&lt;key&gt;NSCameraUsageDescription&lt;/key&gt;
-&lt;string&gt;openpnp-capture needs permission to access the camera to capture images.&lt;/string&gt;
-
-</infoPlist>
+      <infoPlist>&lt;key&gt;NSCameraUsageDescription&lt;/key&gt;
+&lt;string&gt;openpnp-capture needs permission to access the camera to capture images.&lt;/string&gt;</infoPlist>
       <iconImageFiles>
         <file path="./src/main/resources/icons/OpenPnP-logo.iconset/icon_32x32.png" />
         <file path="./src/main/resources/icons/OpenPnP-logo.iconset/icon_128x128.png" />


prusk:
@vonnieda FYI

ian-arkver:
LGTM +2

prusk:
> LGTM +2

🤞

Hope it works. I might be wrong but the single-file app doesn't work right now on neither of my Macs, Monterey or Sonoma...
Weird enough, the develop installer has no issues.

ian-arkver:
Ah, I thought you'd tested it. :-)
Not having a Mac, I can't. Visually you're just undoing the newlines Jason added, so seems safe enough?

prusk:
I tested what I could. I get no crash if I install the app, approve it to run and then modify the Info.plist.
However, it is impossible to test it with the Info.plist modified before the first run because the file gets corrupted and the OS doesn't let me run it anymore at all.
I'm waiting to get the output from the build so I can try to run it again and hopefully observe how it's requesting camera permissions...

prusk:
PS: am I the only one running OpenPnP exclusively on macOS?

vonnieda:
Looks good, thank you @prusk!

prusk:
that didn't work @vonnieda .
I think there's actually a signing problem for macOS. I was able to do a build with install4j trial without signing and everything was working great, camera permissions and all.
Now that I tried the latest build, something is seriously wrong:

Screenshot 2024-05-29 at 12 29 26 AM

STREGA:
I can confirm the signing is broken. 
Running the current test build gives me the unidentified developer gatekeeper dialog.

Bypassing gatekeeper to launch anyway results in an empty camera device list.

I'm on 14.5 (Sonoma, Latest)

vonnieda:
I confirm the signing is broken:

```
> spctl --assess --type execute --verbose=4 /Volumes/openpnp/OpenPnP.app
/Volumes/openpnp/OpenPnP.app: rejected
source=Unnotarized Developer ID
```

I went back to 2023-03-20 and get the same thing. It appears that Notarization become required at some point, when previously it was optional. 

Install4J has the option to Notarize when building but it needs my AppleID and password, which obviously I cannot put on Github. 

I'm looking into https://github.com/marketplace/actions/apple-xcode-notarize currently to see how we can automate this.

prusk:
I found GitHub-hosted runners that offer the possibility of running an install4j build on macOS to solve the notarization issue.
Pandaplacer can sponsor this if needed.
Let us know…

dehret:
Are there any updates on this?

Before setting up my 30+ feeders I wanted to switch to the test build to use the new Vision Feeder feature. This would make setting up the pick position for each feeder much easier.

I was able to open the app anyway (right clicking the app, then open, then confirm), but the camera device list stays empty. Which means I can't use it right now at all.

Is there a workaround available? I do have an Apple Developer account, in case I'd need to self-sign the app somehow.

prusk:
You can do an unsigned build on your Mac with install4j and it will work, it will ask for camera permissions.
That’s what I do until this problem is solved.

vonnieda:
This is still on my radar, but I've been busy.

If someone wants to do the legwork and let me know what steps I need to take to get this fixed, that would be a great help!

@prusk Thank you for the offer, I'll reach out if needed, but I think we can do this with the free Arm runners, and I already pay for an Apple developer account. 

This should be a pretty straightforward fix - I'm just very busy with another project at the moment.

dehret:
@prusk @vonnieda Thank you so much for your quick reply. I created an unsigned build using install4j. It asked for camera permission and the devices now appear again in the device list. Great!

@vonnieda I'm willing to put in some work to get this fixed. But I am not familiar with the project so far, so I would need to get a least pointed into the right direction to begin with.

prusk:
I opened this PR for the permanent fix: https://github.com/openpnp/openpnp/pull/1653

PR #1633, opened 2024-06-01
Modernize macOS app icon

STREGA:
# Description
Modernizes the application icon on macOS to more closely resemble the current system design.

# Justification
The original icon is a bit out of place and feels low quality in the macOS Dock next to other modern app icons.

# Instructions for Use
None

# Implementation Details
This is a purely cosmetic change.

New Icon:
OpenPNP Icon small@2x
icon_512x512@2x


STREGA:
Another option. Used the "O" icon as the orientation silkscreen marker and gave the pads a Copper/ENIG look.
macOS still doesn't allow alternative icons for dark mode so it's one or the other unfortunately.
OpenPNP Icon 2

prusk:
I like the white one.

LGTM.

STREGA:
I agree. It's virtually identical to the original, but with some slight changes Mac users will appreciate.

White is the one in The commit.
So this is ready to merge when someone gets a chance.

vonnieda:
Love it, nice work!

tonyluken:
Hey, does anyone here know why we need to have two different folders that only differ by case in one letter position: OpenPNP-logo.iconset and OpenPnP-logo.iconset. It plays havoc with Windows which doesn't recognize case so it merges the two folders into one. Can we just move all the logo icons into one folder and be done with it? If we do that, does the OpenPnP-logo.icns file need to be updated in anyway?

See the forum discussion [here](https://groups.google.com/g/openpnp/c/0n9vXko6QNQ).

STREGA:
> @STREGA, @vonnieda, Hey, does anyone here know why we need to have two different folders that only differ by case in one letter position: OpenPNP-logo.iconset and OpenPnP-logo.iconset. It causes problems on Windows machines which don't recognize case so it merges the two folders into one. Can we just move all the logo icons into one folder and be done with it? If we do that, does the OpenPnP-logo.icns file need to be updated in anyway?
> 
> 
> 
> See the forum discussion [here](https://groups.google.com/g/openpnp/c/0n9vXko6QNQ).

An icns file on macOS is a container that has all the ordinal files in it. There is no need or benefit to having them in a separate folder.

I only updated it because it was already there. The macOS inconset folder can just be deleted.

tonyluken:
@STREGA, @vonnieda,

Thanks for the info. I think we will just move the files from the OpenPNP-logo.iconset folder (the one that was created by this PR) into the OpenPnP-logo.iconset (the original folder) and then get rid of the OpenPNP-logo.iconset folder. That way, if anybody in the future needs to do something with the individual files, they will still be there.

vonnieda:
Sounds good to me. I can't remember why they are duplicated. It was probably just a merge error at some point.

PR #1641, opened 2024-06-15
Feeder copy paste

hydra:
# Description
Add copy/paste support to feeders.

# Justification
Very useful when setting up a machine with many feeders, regardless of the feeder type and whether the feeder implementation has it's own auto-setup/template.

The reason I added this, is that for my machine, I want to define one feeder, then jog the head 12.5mm on the Y axis to the next pick position and have a new feeder with the exact same settings as the first one, but where only the Y values need changing.  I have to do this 60+ times, once for each feeder.  This PR is phase one of my plan, the next phase of my plan is to add some X/Y offset functionality which can be applied to selected feeders, more on that in another PR.

# Instructions for Use
Select one or more feeders, observe that 'Copy' become available in the toolbar.  Press 'Paste'. Observe that the feeders copied are added to the feeders list, with ` (Copy)` appended to their names.

# Implementation Details
There are no automated tests for the GUI (that I could find), manual testing.
Yes.
N/A
Done

The implementation follows the same approach used to copy/paste vision settings, i.e. copy serializes selections to XML, paste deserializes into a holder, then adds the deserialized objects to the machine.

# Observations/Notes
* This PR also includes cleanup commits that a) delete some unused code, b) fixes some naming, d) removes a spurious separator from the vision panel. See individual commits.


markmaker:
Hi @hydra,

nice idea and I appreciate the initiative! Not wanting to curb that in any way, I still want to urge you to respect the "[Discussion first](https://github.com/openpnp/openpnp/wiki/Developers-Guide#discussion-first)" rule in the future. It would help both sides a lot. :smile: :+1: 

This PR must be proposed together with #1643, because it is IMHO _only_ acceptable in combination. Copy&paste of a feeder on top of the other, without moving it at the same time, is extremely confusing to the user, almost sure to invite bad mistakes, i.e. forgetting to offset the copied feeder and then assembling whole batches of PCBs with the wrong parts, or offsetting the original instead of the copy, effectively confusing the two parts.

Furthermore, the PR combo must fully respect the feeder's business logic. 

For instance, you need to understand the logic and functionality already in place in `BlindsFeeder`. It can already "one-click"-create subsequent blinds feeders in the array. It can then also move and rotate around _the whole array_ of connected feeders (as they are 3D printed together), and your code in #1643 currently breaks that. See the video here: 

move and rotate BlindFeeder

https://youtu.be/dGde59Iv6eY?t=619
https://github.com/openpnp/openpnp/blob/612f2d013955e7821d95ae8d9b6a1fd5a8ee0a70/src/main/java/org/openpnp/machine/reference/feeder/BlindsFeeder.java#L1977

Your copy&paste tool would have to respect and plug into that existing logic, best by copying and displacing the whole array, and by making sure the copy is entirely located _outside_ the current array's fiducials.

The other example is the `ReferencePushPullFeeder`. It can already "one-click"-create feeders in a row with proper auto-learning offsetting (so it could address your needs, btw. :wink: ). This is including OCR/QR and re-calibration etc:

https://youtu.be/5QcJ2ziIJ14?t=296

Plus as you might see in the code, it is a bit more involved (as compared to your current code):

https://github.com/openpnp/openpnp/blob/612f2d013955e7821d95ae8d9b6a1fd5a8ee0a70/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L1758
https://github.com/openpnp/openpnp/blob/612f2d013955e7821d95ae8d9b6a1fd5a8ee0a70/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L1635

I guess your copy&paste&offset should handle  `ReferencePushPullFeeder` in a special way, i.e. using the built-in functionality instead of its own generic one.

Those are the two feeder classes I intimately know. 

I suspect there are many more such business logic pitfalls lurking in other classes. Many feeders have state like "feed count" etc. that is persisted (XML) but must **not** be copied over. Feeders with dedicated actuators invite further user mistakes, and bad assembly SNAFUs, when these actuators are not _all_ exchanged with a new one on the _copy_ or when they are exchanged on the _original_, by mistake. Maybe such actuators should not be copied over, so users know which they still need to add, and where.

Ideally, your PR would also support rotation in addition to mere translation, so one can copy a feeder from "west" to "east" for instance. See #943 by @tonyluken, that wanted to group and move around feeders. It may contain code you could reuse. Ideally, you create a universal `Feeder.transformLocation()` method that covers the "translate and rotate" function for various other applications.

Finally, I would like to see a variant, where the new feeder location is given by the current cross-hairs (or similar), not by a coordinate input box. This would be the very first such input box, AFAIK, OpenPnP users are not accustomed to such a thing. Plus, for coordinate parsing you need to use the `LengthConverter` respecting the user's system units and supporting parsed unit conversions.

On a side note, please follow the ["Do not modify the formatting or white-space of other existing code, i.e., make sure that the source code differences (as tracked by git) will be closely restricted to the code regions that you actually changed." ](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style) directive (in #1643).

"[Discussion first](https://github.com/openpnp/openpnp/wiki/Developers-Guide#discussion-first)" would have revealed all this earlier, and perhaps shaped your goal.

_Mark

hydra:
> nice idea and I appreciate the initiative! Not wanting to curb that in any way, I still want to urge you to respect the "[Discussion first](https://github.com/openpnp/openpnp/wiki/Developers-Guide#discussion-first)" rule in the future. It would help both sides a lot. 😄 👍

Hi, yes I read the 'discussion first', however the latency between a proposal, resulting discussion and technical investigation by everyone is too long and does not fit my time-frame needs.  I need to get my PnP machine up and running again so that I can continue actually making PCBs.   If I had gone the discussion first route I would not have a machine with 60 usable feeders that I'm able to maintain in a usable way as likely neither feature would exist.

> This PR must be proposed together with #1643, because it is IMHO _only_ acceptable in combination. Copy&paste of a feeder on top of the other, without moving it at the same time, is extremely confusing to the user, almost sure to invite bad mistakes, i.e. forgetting to offset the copied feeder and then assembling whole batches of PCBs with the wrong parts, or offsetting the original instead of the copy, effectively confusing the two parts.

It would be possible to add an issues & solutions check for feeders with identical pick coordinates.

Copy/paste is useful in at least these scenarios, regardless of using an offset tool:
1. setting up feeders.
2. experimenting with feeder changes, without having to resort to backing up/restoring configs. i.e. copy, adjust the copy, test, accept or delete the copy.

> Furthermore, the PR combo must fully respect the feeder's business logic.

yes, there's probably stuff I missed for sure, I'm not yet familiar with all the feeders and their operation, additional commits can be added as required.  I'll digest the  info on the feeders you posted, as time permits.

> The other example is the `ReferencePushPullFeeder`. It can already "one-click"-create feeders in a row with proper auto-learning offsetting (so it could address your needs, btw. 😉 ). This is including OCR/QR and re-calibration etc:

yes, however in practice that doesn't work and was a total failure for me.  On the CHMT all the feeders have a single X-co-ordinate, and you don't want to have to actually put a tape in the machine to set them up.  If you use the 'auto-setup' it would fail and never gives consistent X coordinates.  Also, the vision pipeline needs adjusting for every different type of tape in the machine, for me it was failing with the clear tapes and only worked on the white paper tapes, thus preventing me from using it.  The '+' button doesn't work for me, as it uses the auto-setup, and fails, additionally it updates the 'rotation' fields on the push/pull configuration which is not what you want when you use peeler actuators.

> I suspect there are many more such business logic pitfalls lurking in other classes. Many feeders have state like "feed count" etc. that is persisted (XML) but must **not** be copied over. Feeders with dedicated actuators invite further user mistakes, and bad assembly SNAFUs, when these actuators are not _all_ exchanged with a new one on the _copy_ or when they are exchanged on the _original_, by mistake. Maybe such actuators should not be copied over, so users know which they still need to add, and where.

yes, we can investigate and find these issues as we go, this is a first-cut and works for my immedite needs.

it would be entirely possible to give the user error messages for feeders that don't support being offset.

> Ideally, your PR would also support rotation in addition to mere translation, so one can copy a feeder from "west" to "east" for instance. See #943 by @tonyluken, that wanted to group and move around feeders. It may contain code you could reuse. Ideally, you create a universal `Feeder.transformLocation()` method that covers the "translate and rotate" function for various other applications.

it does in code already, I used `rotation.addWithRotation`, but since there's extra testing overhead that doesn't warrant my time (yet) I only supported x/y/z to begin with, gotta start somewhere.  no need for bells and whistles on the first-cut of the idea.

> Finally, I would like to see a variant, where the new feeder location is given by the current cross-hairs (or similar), not by a coordinate input box. This would be the very first such input box, AFAIK, OpenPnP users are not accustomed to such a thing.

yes, using the UI to capture the offset would be a thing that could be added in due course.

> Plus, for coordinate parsing you need to use the `LengthConverter` respecting the user's system units and supporting parsed unit conversions.

I'm not sure where the length conversion might be occurring already or need to be occurring, I'm aware Location has 'lengthUnits' but haven't yet dug into where or why this might be used as I've only ever seen feeders in 'mm', I am an OpenPnP noob so I'm probably missing something, feel free to point me at code examples of where/why this is needed/used and I can look into that.

> On a side note, please follow the ["Do not modify the formatting or white-space of other existing code, i.e., make sure that the source code differences (as tracked by git) will be closely restricted to the code regions that you actually changed." ](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style) directive (in #1643).

yes, I only noticed these after I pushed my changes, I can go back and amend the commits.

Other projects I've worked on have just done blanket codebase reformats at suitable points in development, e..g just after a release, to remove white-space, fix formatting.  Would recommend doing this on the codebase so that you can tell everyone to have their editors trim trailing whitespace by default so there's a consistent and faff-free development experience for all.  This trim trailing whitespace could also be achieved as a pre-merge action or git hook.

> "[Discussion first](https://github.com/openpnp/openpnp/wiki/Developers-Guide#discussion-first)" would have revealed all this earlier, and perhaps shaped your goal.

as noted earlier, these PR's probably just wouldn't even exist and I'd still be frustrated with the lack of progress with my PnP machine conversion.

hydra:
regarding BlindsFeeder:
> Your copy&paste tool would have to respect and plug into that existing logic, best by copying and displacing the whole array, and by making sure the copy is entirely located outside the current array's fiducials

I'm not sure this is true, there is no other code for any of the other feeders that prevents you from using the same machine coordinates for anything else.  OpenPnP is essentially blind and doesn't know what exists in the real world.  Yes, a general 'thing overlaps other thing' in the issues and solutions might work.  i.e. ask all the feeders about what they each know about their bounding boxes and see if any intersect and flag at as an issue if they do.

I looked into binds feeder array definition today and have provided feedback here: https://github.com/openpnp/openpnp/issues/1644

> On a side note, please follow the ["Do not modify the formatting or white-space of other existing code, i.e., make sure that the source code differences (as tracked by git) will be closely restricted to the code regions that you actually changed." ](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)directive (in https://github.com/openpnp/openpnp/pull/1643).

commits updated, changes removed.

regarding ReferencePushPullFeeder:

> This is including OCR/QR and re-calibration

done.

I'll carry on working though the other items you mentioned.

markmaker:
Copying this over from the group discussion:

> Please honestly explore [Calibrate X and Y checkboxes](https://github.com/openpnp/openpnp/wiki/ReferencePushPullFeeder#apply-vision-calibration-not) with stock ReferencePushPullFeeder functionality and report back.

Please do not open up a gazillion new lose ends, before answering this first.

_Mark

hydra:
> > Please honestly explore [Calibrate X and Y checkboxes](https://github.com/openpnp/openpnp/wiki/ReferencePushPullFeeder#apply-vision-calibration-not) with stock ReferencePushPullFeeder functionality and report back.
> 
> Please do not open up a gazillion new lose ends, before answering this first.

yes, I did look at that, and my initial feeder setup uses the drag-pin to pull the tape further than the feed pitch, and then it moves the head back towards the pick position such that the drag pin doesn't touch the tape when the drag pin is retracted from the tape hole, so for my use-case my understanding is that I *do* want the calibrate X/Y to be enabled.  Additionally, for my use-case, I set the 'precision wanted' so that is is less than the distance between the drag pin start location and the edge of the machine's feeder bank which the drag pin should not touch.

I'll dig into the code a bit, as it's not clear yet to me that if calibrate X is enabled, and the hole positions are offset left a little from where they're supposed to be, if all the push/pull locations are offset, or only the start location.  For my use-case it would be helpful if the hole calibration only applied to the pick position and the initial push/pull start location so that errors cannot accumulate.

markmaker:
> yes, I did look at that, and my initial feeder setup uses the drag-pin to pull the tape further than the feed pitch, and then it moves the head back towards the pick position such that the drag pin doesn't touch the tape when the drag pin is retracted from the tape hole, so for my use-case my understanding is that I do want the calibrate X/Y to be enabled. 

As the tooltip and [Wiki says](https://github.com/openpnp/openpnp/wiki/ReferencePushPullFeeder#apply-vision-calibration-not), the **Calibrate X/Y** checkboxes only control whether the offset obtained from calibration is applied to the Push-Pull Motion (per Axis) or not. 

> For my use-case it would be helpful if the hole calibration only applied to the pick position ....

The checkboxes do **not** affect the calibration of the pick location. 

> ... and the initial push/pull start location so that errors cannot accumulate.

It is currently not possible to treat the Start Location differently. 

According to the information from other `ReferencePushPullFeeder` users this is not a real problem, because 

**a)** you can define the start location with the tolerance of the sprocket hole being larger than the drag needle, so even a rough (manual) initial tape position is okay, and 

**b)** even if the tape is initially not positioned right, nothing bad happens, as the needle is solenoid-driven ("elastic"), i.e., the needle will just slide on top of the tape until it snaps into the next sprocket hole. So when you load a tape, just actuate one extra feed to perfectly align it.

_Mark

markmaker:
Next steps:

1. If we have a bug with the rotation [as suggested in #1643](https://github.com/openpnp/openpnp/pull/1643#issuecomment-2178081477), then of course this might also explain the difficulties here.
2. But if not, **the question remains, whether switching one of Calibrate X/Y off** solves your issues, i.e., you can productively use the **`[+]`** button and the **Auto Setup** _while_ the camera crosshairs are on the pick location on a brand-new feeder. In fact, for a feeder with the same orientation, it should work, even _with_ the suspected bug. 

_Mark

PR #1643, opened 2024-06-15
Feeder offset tool

hydra:
# Description
Adds a feeder offset tool.

This tool can be used to offset (X,Y,Z) or more selected feeders.  All the corresponding locations are updated. Works for all feeders.

This PR also fixes some bugs with the existing feeder GUI, see individual commits.  The bugs were found and fixed during testing.

# Justification
It's common to need to make the same change to multiple feeders at the same time, for example:

* adjusting feeder bank offsets on the CHMT36/48 machines which have 2 banks of 30+ feeders. It's extremely time consuming to use the UI as-is to change around 18 fields, per feeder, and make the manual calculations which is also error prone.
* when you adjust the position of a removable feeder, e.g. a bed-mounted 3D printed cut tape feeder with slots for multiple tapes. this can be due to removal, cleaning, etc.
* when initially provisioning a machine, it's common to setup 1 feeder, then duplicating the feeder set for the rest of the machine.  This feature, in combination with Feeder Copy/Paste in PR #1641 makes setting up the 60+ feeders on the CHMT48 take just a few minutes, instead of hours of manual clicking, math calculations and data entry for each feeder.

Here's a video:

[OpenPnP feeder copy/paste and offset tools on a CHMT36/48](http://www.youtube.com/watch?v=iCtfJu0l0uY "OpenPnP feeder copy/paste and offset tools on a CHMT36/48")

# Instructions for Use
* Select one or more feeders in the Feeder tab
* Click the Offset toolbar button
* Enter X/Y/Z offsets (which can also be negative, e.g. to undo an offset)
* Press OK.
* Be amazed as all the selected feeders get updated and rejoice in the fact you don't now have to do this manually on 60+ feeders if you own a CHMT48VB.

image

# Implementation Details
Manually, using the GUI, there's no automated tests.
Yes, and some cleanups were made.
Yes, added `applyLocationOffset` to `org.openpnp.spi.Feeder`
Done

# Notes
* ~My IDE is set to fix indentation and white-space issues by default (IntelliJ Idea) so files that this PR touched also have some whitespace changes, feel free to ignore these whitespace changes when reviewing the code.~ EDIT: whitespace changes removed from PR.
* I used an 'edit feeder' icon that was already present, and seemingly unused, in the codebase for the toolbar button.  We can do another PR/commit if someone can find a better or more suitable icon.


hydra:
@markmaker 

Following on from the issue you raised with the blinds feeder and your reply on https://github.com/openpnp/openpnp/issues/1644#issuecomment-2173972615

> As it is now, each feeder must carry the information of the array too. The info is then synced between the feeders automatically. Yes, I do know that's [not third normal form](https://en.wikipedia.org/wiki/Third_normal_form).

yes, this is where some of the problems start.

I dug into the code some:

BlindsFeeder.java
```java
List<BlindsFeeder> getConnectedFeedersByLocation(Location location, boolean fiducial1MatchOnly)
List<BlindsFeeder> filterFeedersByGroupName(List<BlindsFeeder> feeders, String groupName)
List<BlindsFeeder> getConnectedFeeders()
List<BlindsFeeder> getConnectedFeeders(Location location, boolean fiducial1MatchOnly)
```

This implementation, is flawed when it comes to the UI, UX the API and this offset tool because there is not a real definition of a bank.

SchultzFeeder.java
```java
public static class Bank extends AbstractModelObject implements Identifiable, Named {..}
```

This is probably the approach that should have been taken, then the API calls above could be using the Bank, and then they would have no issue with finding other feeders from another bank that happen to be in the same bounding box defined by the feeder.

Additionally, again, IMHO, the bounding box defined by the fiducials should be stored against the Bank, and not the feeder itself. As currently this duplicate information is leading to many issues.

hydra:
In light of the issues with the BlindsFeeder, the safest thing to do for now, until such time as the BlindsFeeder is updated to have the concept of a Bank, it's probably best that the blinds feeder is excluded from cloning.

I'm thinking adding a 'canOffset' -> `bool` on the feeder API, which can be used by the offset tool to see if the feeder supports being offset, and displaying an error to the user if this is not the case, is probably a good enough solution for now, to mitigate the issues you raised here.  I'll have a crack at that now.

hydra:
@markmaker See https://github.com/openpnp/openpnp/pull/1643/commits/4b5f4d54345787fc24537852ad4724b0bc5adadf

hydra:
@markmaker from https://github.com/openpnp/openpnp/pull/1641#issuecomment-2171684077

> for coordinate parsing you need to use the LengthConverter respecting the user's system units and supporting parsed unit conversions

done, see https://github.com/openpnp/openpnp/pull/1643/commits/82ac8b20137c3a8b7bd5891273576b41277e71b3

markmaker:
@hydra  please, regarding **all** the latest commits: Read what I wrote again. Carefully. Take your time.

Please do it the way I [proposed](https://github.com/openpnp/openpnp/pull/1641#issuecomment-2171684077). 

Its not only about doing it somehow "equal", its also about keeping the [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) intact, provide the full functionality, not just some subset, and to allow future development there.

- `LengthConverter` (which uses [`Length.parse()`](https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/model/Length.java#L176)) can do ad hoc conversions. Say you have metric System units, but your feeder row has imperial slot pitch, then you can enter `1"` and it will be converted into `mm`. 
- and as I said, you should handle `ReferencePushPullFeeder` in a special way, i.e. using the built-in functionality `setFeederLocation()` and/or `createNewAtLocation()` that I linked in my comment. The functionality of these methods will likely grow in the future and we do not want to maintain it in two places, and have two sources of potential bugs.

_Mark

hydra:
>     * `LengthConverter` (which uses [`Length.parse()`](https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/model/Length.java#L176)) can do ad hoc conversions. Say you have metric System units, but your feeder row has imperial slot pitch, then you can enter `1"` and it will be converted into `mm`.

right, now I understand what your meant, I asked here regarding the LengthConverter but you did not answer my question so I had to guess.

https://github.com/openpnp/openpnp/pull/1641#issuecomment-2171788084


>     * and as I said, you should handle `ReferencePushPullFeeder` in a special way, i.e. using the built-in functionality `setFeederLocation()` and/or `createNewAtLocation()` that I linked in my comment. The functionality of these methods will likely grow in the future and we do not want to maintain it in two places, and have two sources of potential bugs.

> Its not only about doing it somehow "equal", its also about keeping the [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) intact, provide the full functionality, not just some subset, and to allow future development there.

these two statements seem to contradict each other.  re-using existing logic meant for swapping feeders is not keeping the concerns of applying an offset to a cloned feeder separate.

Earlier today I had infact tried out using setFeederLocation and generating a new location from the offset and the hole1location, however this doesn't do what I need and still messes up the 'rotation' on the peelers, likely the same bug exists in the existing swap feeders implementation.  try it, define two feeders both with a rotation of `4` on the 'end location' then swap them using the existing code and note the resulting rotation.

It also seems there's other issues regarding the rotation on the holes, as when you capture a location from a tool or the camera the existing code for the UI doesn't reset the rotation, since the `rotation` argument to `copyLocationIntoTextFields` is always `null`.




> 
> 
> _Mark

markmaker:
> these two statements seem to contradict each other. re-using existing logic meant for swapping feeders is not keeping the concerns of applying an offset to a cloned feeder separate.

Not true. Please look at the linked methods, what they can do, how they are built on each other.

> however this doesn't do what I need and still messes up the 'rotation' on the peelers, likely the same bug exists in the existing swap feeders implementation

I said from the very beginning, that I don't exclude the possibility that there is a bug. Document it. We'll fix it at the source. But it needs an isolated step by step reproduction. 

Surely, you don't propose leaving that bug in and instead create your other function as a "band-aid" on top? :rofl: 

> It also seems there's other issues regarding the rotation on the holes, as when you capture a location from a tool or the camera the existing code for the UI doesn't reset the rotation, since the rotation argument to copyLocationIntoTextFields is always null.

The holes don't have a rotation on themselves. The **vector between them** will be evaluated for the feeder rotation. 

https://github.com/openpnp/openpnp/blob/edeb2baaf89593e1813d0e060447119aad4ced8f/src/main/java/org/openpnp/util/FeederVisionHelper.java#L243-L268

As you see there, a certain rotation change check was commented-out in #1632. At the time, I was okay with that, but maybe that introduced the bug you are observing. @prusk, can you explain why it had to be commented out at all?

The essential use of that "unit vector" is also the reason the **order** of these two sprocket holes is important. Using vision it is automatically done right by having the pick location in the camera center, giving you the parts' location relative to the sprocket holes (EIA 481-C gives you the tape orientation). If you capture the holes manually, you must be careful which is which, i.e. hole 1 must be closer to the reel. Read the tooltips.

Sprocket holes

_Mark

prusk:
The HACK section in the FeederVisionHelper was disabled so all the transform functions can be static.
Realistically, does it add any value? The tape rotation is calculated based on holes position and this is what gets back into the feeder location, doesn't really matter what the original input value is.

markmaker:
In the original it set the `ReferenceFeeder.location` with the calculated rotation, and it reset the calibration.

I see it referenced in `ReferencePushPullFeeder.getNominalVisionLocation()` but that's not (currently) relevant.

Let's await confirmation that there is a bug at all.

PR #1651, opened 2024-07-06
AbstractPandaplacerVisionFeeder - Remove automatic cv pipeline reset

prusk:
# Description
In this feeder type, some code inherited from ReferencePushPullFeeder, specific to feeder cloning, was resetting the CV pipeline for each Auto Setup action.
This change removes the automatic reset so a custom vision pipeline can be used when needed for Auto Setup, since BambooFeeders do not implement cloning.

# Justification
Without the cloning functionality, it is not correct to reset the CV pipeline automatically when there is a dedicated button on the UI for it. The reset action should be intentional as the user requires.

# Instructions for Use
In the BambooFeeder objects, the Auto Setup button will no longer automatically reset the vision pipeline, therefore a customized pipeline can be used for both auto calibration and in-job calibration.

# Implementation Details
* Ran all tests and used it on my machine.
* Yes
* No changes
* All tests passed

### DIFF
--- a/src/main/java/org/openpnp/machine/pandaplacer/AbstractPandaplacerVisionFeeder.java
+++ b/src/main/java/org/openpnp/machine/pandaplacer/AbstractPandaplacerVisionFeeder.java
@@ -532,31 +532,16 @@ public abstract class AbstractPandaplacerVisionFeeder extends ReferenceFeeder {
         }
 
         ensureCameraZ(camera, true);
-        // Try with cloned pipeline.
-        Exception e = autoSetupPipeline(camera, null);
+        // Try with the current pipeline.
+        Exception e = autoSetupPipeline(camera, this.pipelineType);
         if (e != null) {
-            Logger.debug(e, "Auto-Setup: exception");
-            // Failed, try with pipeline type defaults.
-            for (PipelineType type : PipelineType.values()) {
-                Logger.debug("Auto-Setup: trying with stock pipeline type "+type);
-                e = autoSetupPipeline(camera, type);
-                if (e == null) {
-                    // Success.
-                    Logger.debug(e, "Auto-Setup: success");
-                    return;
-                }
-                Logger.debug(e, "Auto-Setup: exception");
-            }
-            // Still no luck, throw.
+            // No luck, throw.
             Logger.debug(e, "Auto-Setup: final exception");
             throw e;
         }
     }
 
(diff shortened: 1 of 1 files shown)


markmaker:
>  the code inherited from ReferencePushPullFeeder was resetting the CV pipeline for each Auto Setup action. This change removes the automatic reset so a custom vision pipeline can be used when needed for Auto Setup.

Just to make sure we're on the same page: in ReferencePushPullFeeder the pipeline is cloned from the template feeder, which at this point is preliminarily determined. The preliminary template feeder (without yet having the part detected) is mostly determined by _proximity_, e.g., feeders in a row.

https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L1767-L1771

If all feeders are designed well, their vision geometry i.e. OCR/QR label area relative to sprocket holes, will be equal. This will allow to determine the part by OCR/QR, even if the preliminary template feeder happened to have a different tape width. With the part identified, the right template will be determined (by package and/or tape spec), and it will clone from the template once more to also get the other important settings right:

https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L2352-L2355

So ReferencePushPullFeeder does of course support a customized pipeline, and is not "resetting it". Only when the cloned pipeline _failed_, would it fall back and try with the stock pipelines. With the modern circular symmetry pipelines this should never happen, even when you go from paper to clear plastic carrier tapes, or whatever.

Should OCR fail, you have to choose the assigned part manually, and then you can clone the right setting including the pipeline over. So a few clicks more, but still very automatic.

I just wanted to set the record straight, here. Since your feeder class chose _not_ to adopt all the nice cloning magic, I guess these changes are okay. They do not affect ReferencePushPullFeeder, right? 

Ready to be merged?

_Mark

prusk:
Thank you for the clarification @markmaker .

Indeed, the BambooFeeders do not support cloning therefore the logic is different.

I can confirm this change is contained only to BambooFeeders and does not impact ReferencePushPullFeeder in any way.

markmaker:
Thanks, @prusk!

PR #1654, opened 2024-07-17
Pre-rotate all nozzles

janm012012:
# Description
This PR adds sub-ordinate motion capabilities to the abstract motion planner and uses in in the job planner to rotate all nozzles to their required angle while moving over to the first feed or pick location, the bottom camera or the first place location.

# Justification
This optimization provides a speed enhancement in situations where the rotation takes longer then the actual XY move eg when moving the second nozzle over the bottom camera and reduces the risk of parts slipping on nozzles.

# Instructions for Use
The optimization is enabled automatically and can be disabled via the job processors UI.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **using simulation machine and g-code checking**


markmaker:
Hi @janm012012,

sorry for the long delay in looking at your valuable PR here. My mind wasn't free enough to tackle this difficulty level :exploding_head: 

I'm still very unsure about this, unfortunately. :cry: 

You compare new coordinates against previous ones, and then replace them with the subordinate when they haven't changed. But that's not actually conclusive, is it? In general, it could still be a wanted target coordinate, even when it hasn't changed when compared to before, right? 

This is especially true in rotation (which is the subject here), let me make an example:

- Assume a machine with shared C rotation axis. N1 and N2 have the same C motor, both are simply rotated by the same belt.
- Assume two picks are prepared: N1 at 180°, N2 at 90°.
- Because they share the same C, the second prerotate-merge cannot overwrite the first, so `queue` it is now at 180°. 
- But then due to optimization, N2 is the first to be moved to the pick location. 
- By coincidence, axis C  was already at the N2 target rotation of 90° before, which is not unlikely, as Rotation typically only has the four 90° angles, other angles are rather rare. 
- Because the angle doesn't change, the rotation axis is _not_ considered dominant, and the subordinate rotation is taken. N2 now mistakenly moves to 180° instead of 90°.

With three or more nozzles, I could even make a simpler example when this happens, without the need for shared C axes:

- In the course of moving the previous nozzles, the pre-rotate of the nozzle's C axis could already have taken place, when the nozzle is finally moved to the target.
- So the C coordinate doesn't change _by design_. But now the non-changing C coordinate is no longer considered a dominant one, and might be replaced by a subordinate one that was queued in the meantime.

You might argue that you'll use the _already_ optimized order for pre-rotation, and that pre-rotation only happens in groups of all nozzles and then nothing is queued until all these nozzles are moved. Or whatever argument might hold for this limited use case. But in real life, when you take error conditions with discard and pick retry, or feeder specialties (cover opening, actuation motion, vision motion), or in alignment the part height auto-focus detection, or whatnot into consideration, you'll always find a scenario where this could happen. Either now or with future development.

Even if you can somehow exclude it for the rotation case, e.g. by suppressing the optimization on shared C machines, or by arguing that angles are slightly adjusted by fiducial vision etc. and therefore are unlikely to match exactly, or by the argument in the previous paragraph, it would always only be a house of cards of "current use case assumptions". 

To be more clear and honest: I don't like to make the motion planner rely on shaky assumptions like these. The resulting motion must be deterministic and reliable in all present and future use cases.

I believe we discussed the various problems before: 
https://groups.google.com/g/openpnp/c/fthkARkPBrE/m/JciCztWOAAAJ

My current take is that you'd need to take all the mapped axes of the moved `HeadMountable` (`hm`) as dominant regardless of whether their coordinates change or not.

There is the `Double.NaN` optimization that could perhaps be excluded. Or more generally, the `hm.getLocation().derive()` pattern, but with complications that we discussed too.

Another problem we discussed as well: Pre-rotation must only happen when (all) the intertwined `HeadMountable`s are at Safe Z, i.e. where the subordinate axes are allowed to rotate freely in the air. Again, taking the shared C axis machine into account.

And that's just the stuff that comes to my mind right now. I get a feeling that it could go wrong in many more ways. :sob: 

Note these scenarios have a very high probability of creating false-positive results in testing/simulation. So saying "I tested it, and it works" will not hold much value, I'm afraid. You'd have to include all the feeder classes with special actuation and vision, and alignment with part height auto-focus, placement with contact probing, etc. in the tests, and covering all the various states these facilities can be in, prior to the move in question, unfortunately. And then it would still only be a momentary assessment. The next PR could inadvertently break it.

Like I said: 

# :exploding_head: 

Do you think you could simplify it to the moved-by-`hm` dominance? And solve the Safe Z predicament? Also address your `FIXME` markers, please.

_Mark

janm012012:
Hi @markmaker!

Many thanks for your review! I'm with you, that we shall not add any code, that may be dangerous in any sense. Actually I don't remember why I compared length values. Looking into the code again, Subordinate.queue() is called with newLocation only containing a subset of all AxesLocations and hence this can be merged into the queue in total without making any further assumptions. I've changed the code accordingly and tested it on my simulation machine. Filtering the log for "Parsed Gcode" still indicates, that the subordination works as expected (in this scenario). However, I can not guarantee, that it will work in all situation, but I don't see any situations yet in which it would not because all queued motions are recessive.

Concerning the Safe-Z issue: I'm with you, that merging shall only take place with all axes in Safe-Z and this could likely be added to Subordinate.merge(). (IIRC I once noticed, that in combination with some other feeder I used at that time, feed() finished with the nozzle in down position. The queued subordinate motion was then applied to the motion segment where the nozzle is moved up to Save-Z. Unfortunately I was never able to reproduce that...) My question would then be, how do I verify if all axes in axesLocation in Subordinate.merge() are moving in Safe-Z so that it is safe to merge the queued subordinate motion segments in?

Concerning the fixmes: they are there because I have not found a way how to code the mentioned behavior. I would kindly ask you to provide guidance. In one of our very early discussions on this topic you mentioned, that subordination shall take the affected head into account. So the question is, how can I merge just thus axes from the queue that are on the same head? How would I have to qualify that?

The second fixmes in Subordinatio.merge() is related to the first: if not all queued axes are merged due to different head, only this shall be removed from the queue. If only a single head is taken into account, as currently implemented, there is nothing more to be done. (there are always recessive X, Y motions in the queue that are overwritten by dominate ones in axesLocation, hence this have to be removed from the queue as well to avoid trigger false "queue not empty" messages in Subordinate.drain().

Concerning the "moved-by-hm dominance": isn't that what we now have? The queue takes all axesLocations that are marked as type subordinate and lets the already queued axes take precedence. In merge the queue is recessive added to axesLocation such, that only thus axes from the queue that are not part of axesLocation are added.  For a dual nozzle shared C machine there shall be two queue requests one for each nozzle, effecting the same axis so that there is one rotation motion in the queue. If the queue is then merged into the actual motion, the queued rotation is overwritten by the domination rotation of the segment. I'd say that shall be safe.

Jan

markmaker:
Hi @janm012012,

re the safe Z: 

You need to go through each subordinate axis, then find the mapped head, then go through all its HeadMountables, check whether those are all in the safe zone on _both_ the _current_ and the _target_ location coordinates.  See `org.openpnp.spi.Movable.isInSafeZZone(Length)`. 

It is probably easier to do it the other way around, i.e., go through all the heads of the machine, then through all the head's HeadMountables, check if in safe Z zone, and then collect all `org.openpnp.spi.MovableMountable.getMappedAxes(Machine)`. If that head is okay (all `hm` were in safe Z zone), add all collected mapped axes to a result set of _okay-to-move_ axes. Finally check each subordinate axis against the set of  _okay-to-move_ axes.

If that check is fine, merge the subordinate axis into the motion. If not, keep it in the queue for next time.

For future functionality: if the subordinate axis is a Z axis, it needs kind of  _inverted rules_, namely that all the axes that are moved are Z axes, applicable for both the subordinate and dominant axes. 

I don't know if you want to include that option yet.

> In one of our very early discussions on this topic you mentioned, that subordination shall take the affected head into account. 

From my current perspective it is exactly about the Safe Zone of that head. It does not mean that it cannot move subordinate axes of two or more heads, as long as those other heads are _also_ in the Safe Zone.

> Concerning the "moved-by-hm dominance": isn't that what we now have?

Guess so. Double-check please.

Another concern: did we not conclude that there should be a subordinate option per axis type? 

`MotionOption.SubordinateX`, `MotionOption.SubordinateY`, `MotionOption.SubordinateZ`, `MotionOption.SubordinateRotation`?

Maybe leave out the Z option if ou don't implement the _inverted rules_ (see above).

_Mark

janm012012:
Hi @markmaker!

Many thanks for providing the hints on save-z checking. I'll study them and implement something. May take a few days...

Concerning the subordination per axis type: I don't remember that we talked about that. If you like, I can add that. At present I'd like to rotate multiple nozzle all at once (while moving to first feed and/or place/bottom vision location)., So SubordinateRotation would by the first think I'm aiming for.

Jan

janm012012:
HI @markmaker!

I've implemented the discussed changes but unfortunately don't get them full to work. Would you please help me getting the last bits sorted out?

1. Is is possible to loop over all entries of an AxesLocation as AxesLocation (for the axis wise merging)? At present I only found code that loops as ControllerAxis (using .getAxes(ControllerAxis.class) or .getControllerAxis()) but thats inefficient as I need AxesLocations of this ControllerAxis immediately.
2. How can I remove an axis from an AxesLocation? .put() allows to add axes but I have not found any equivalent to remove an axis (and was not yet able to add that myself).

Concerning splitting the MotionOption into individual axes: I don't remember that we discussed that before. I guess one can give multiple of such options at once so there is no performance penalty. I don't mind to change that, assuming its just a filter then queuing axes. Please advice.

Jan

markmaker:
> I've implemented the discussed changes but unfortunately don't get them full to work. Would you please help me getting the last bits sorted out?

Will try :wink:  


> 1. Is is possible to loop over all entries of an AxesLocation as AxesLocation (for the axis wise merging)? At present I only found code that loops as ControllerAxis (using .getAxes(ControllerAxis.class) or .getControllerAxis()) but thats inefficient as I need AxesLocations of this ControllerAxis immediately.

Not sure I understand. 

Using `org.openpnp.model.AxesLocation.getAxes()` delivers all axes.

Then `new AxesLocation(axis, coordinate)` creates an `AxesLocation` with one `axis` and the `coordinate` either as `double` in Millimeters, or as `Length`.


> 2. How can I remove an axis from an AxesLocation? .put() allows to add axes but I have not found any equivalent to remove an axis (and was not yet able to add that myself).

There's no direct method, I think. Why do you need it?

Filtering can be done by Lambda and setting the null coordinate. Example in `org.openpnp.model.AxesLocation.byType(Type...)`.


> Concerning splitting the MotionOption into individual axes: I don't remember that we discussed that before. 

I believe it is safer and also more versatile to use. For your use case you would only use the `MotionOption.SubordinateRotation` flag and need not care that X, Y, and Z are also specified and changed in the passed `Location`, for instance a pick location or a placement location. This means you don't need to use complex `derive()` shenanigans. Also you don't need to use a complex compare-and-only-use-changed-axes in the queuing code.


> I guess one can give multiple of such options at once so there is no performance penalty. 

Correct.


> I don't mind to change that, assuming its just a filter then queuing axes. 

Correct.

Thanks for your perseverance! :rocket:

_Mark

janm012012:
HI @markmaker!

Many thanks for your help! I've now added the Safe-Z validation. It's quite a lot of looping, so please have a careful look if I followed your guidelines close enough.
Unfortunately the code does not fully work yet. In okToMoveToTargetLocation() something still goes wrong by combining unrelated axes and locations. In my test (with a CAM style head) it validates the Z axis against the nozzles, which do not directly map to Z. There is the CAM related translation step that converts nozzle-Z-millimeter to Z-axis-angle. With the current implement the Z merging fails because the physical Z (in degree) is hand over to the nozzle Z (which expects values in millimeters) and triggers a NOT-ok-to-move. Finally Z is not merged generating a queue-not-empty warning in the logs.

> > Concerning splitting the MotionOption into individual axes: I don't remember that we discussed that before.
> 
> I believe it is safer and also more versatile to use. For your use case you would only use the `MotionOption.SubordinateRotation` flag and need not care that X, Y, and Z are also specified and changed in the passed `Location`, for instance a pick location or a placement location. This means you don't need to use complex `derive()` shenanigans. Also you don't need to use a complex compare-and-only-use-changed-axes in the queuing code.

Thats is actually less of a problem because your recommended put() sequence allows subordinate axis to be overwritten entirely by dominant information. I don't need at present any filtering nor device() logic.

I'll implement the different subordinate types now anyhow.

Jan

janm012012:
I've now added the splitting for MotionOption.Subordinate into SubordinateX, SubordinateY, SubordinateZ and SubordinateRotation. This splitting hides the previously mentioned problem with the okToMoveToTargetLocation() checking because there is not use case for subordinate Z axis and the required filtering by type prevent any Z axis from queuing up.

I'll now rebase this PR onto test to fix the merge conflicts.

Jan

markmaker:
Hi Jan,

> Unfortunately the code does not fully work yet. In okToMoveToTargetLocation() something still goes wrong

Sorry, I should have been more specific, when I said "mapped axes". You need to use `toRaw()` to transform to raw coordinates, i.e., apply the cam transform in your case, or any other transformation on mapped axes.

`org.openpnp.spi.MovableMountable.toRaw(Location, LocationOption...)`

Something like this (this is just written down here, not tested):

```java
        LinkedHashSet<ControllerAxis> getOkToMoveAxes(AxesLocation targetLocation) {
            // start collecting overall axes
            LinkedHashSet<ControllerAxis> okToMoveAxes = AxesLocation.zero;
            // loop over all heads
            for (Head h : machine.getHeads()) {
                // start collecting head axes
                LinkedHashSet<ControllerAxis> okToMoveHeadAxes = AxesLocation.zero;
                // loop over all its head mountables
                for (HeadMountable hm : h.getHeadMountables()) {
                    // take the raw axes location of the current hm location
                    AxesLocation rawLocation = hm
                        .toRaw(hm.getCurrentLocation());
                    // get only controller axes (Note, we don't care if virtual axes are not safe)
                    LinkedHashSet<ControllerAxis> headAxes = rawLocation.getControllerAxes();
                    // optimistically add them to the ok to move head axes
                    okToMoveHeadAxes.addAll(headAxes);
                    // get us the current raw controller head location
                    AxesLocation rawCurrentHeadLocation = new AxesLocation(headAxes, 
                        (axis) -> (getLengthCoordinate(axis)));
                    // get us the target raw controller head location
                    AxesLocation rawTargetHeadLocation = new AxesLocation(headAxes, 
                        (axis) -> (targetLocation.getLengthCoordinate(axis)));
                    // check if the Z axes are safe, both in current and target location
                    if (!(rawCurrentHeadLocation.byType(Type.Z).isInSafeZone()
                           && rawTargetHeadLocation.byType(Type.Z).isInSafeZone()) {
                        // uh oh, not in Safe Z, i.e. it is not safe to move any of the axes of this head
                        // reset to zero (empty)
                        okToMoveHeadAxes = AxesLocation.zero;
                        break;
                    }
                }
                // add the head axes to the ok to move axes (this will be empty if they were not safe)
                okToMoveAxes.addAll(okToMoveHeadAxes);
            }
            return okToMoveAxes;
        }
```
_Mark

markmaker:
Unrelated to the current discussion, I just remembered one additional complication, namely Nozzle Rotation Mode:

https://github.com/openpnp/openpnp/wiki/Nozzle-Rotation-Mode

Your code needs to anticipate that, and I frankly don't quickly see how. :frowning_face: 

_Mark

janm012012:
Hi @markmaker!

Many thanks for your quick response. Unfortunately it left me a little lost again: in my current implementation, I'm collecting all axes, that are safe to move without any target location (the function is currently called colectOkToMoveAxes() and located in line 215). I could hand over the dominant AxesLocation as target, but that would not take any subordinate motion into account, that might be added later. 

I loop over all queued axes and check if they are ok to move based on the above collection. If any is ok to move, I use the separate okToMoveToTargetLocation() in line 171 to verify, that its target is also in Safe-Z.

This way I'm currently only validating the current location of the dominant motion and current and target locations of the subordinate motion.

My current implementation fails when validating the target location of the subordinate motion.

Em I correct, that I shall extend the collection of ok to move axes as you quoted by checking also the target location of the dominant motion AND extend the validation of the subordinate motion by adding toRaw() to fix my problem?

Maybe it would be a good idea to put the ok-to-move check into separate function and use it twice, once for the collection of all axes of all head mountables per head and once for all axes of the subordinate queue that are considered safe to merge to the dominant motion. That would result in the same ok-to-move test for dominant and subordinate motion...

Jan

markmaker:
Hi @janm012012,

the problem is that you are checking this in the `queue` function. Instead, you should only check it in the `merge` function, and use the `axesLocation` from the `moveTo()` as the target location. Remember, the subordinate location is kinda riding "piggy-back" on the dominant move.

And like I said earlier: if the current move starts or ends in a location that is not in Safe Z, the queue should be left as is for later, maybe the next move can be used to hitch a ride. I guess this will almost always be the case: first you record the subordinate moves. Then you call `moveToLocationAtSafeZ()` which in turn creates three moves segments, one up to safe Z, one to the target, one down. Only on the second move segment can the subordinate motion ride.

But don't forget to reset the queue on `waitForCompletion()`. If no move segment happened to allowed subordinate moves, that was back luck.

_Mark

janm012012:
Hi @markmaker!

> if the current move starts or ends in a location that is not in Safe Z, the queue should be left as is for later,

Ok, so we only check the dominant movement for it's current and end location (all in Save-Z) and merge the queued motion unchecked (supposes its axes are effecting only the same head(s) as the dominant move). This will result in rewriting my collectOkToMoveAxes() based on your suggestions and remove my okToMoveToTargetLocation() entirely. This will also remove the code, thats not working yet.

Jan

janm012012:
Hi @markmaker!

> Unrelated to the current discussion, I just remembered one additional complication, namely Nozzle Rotation Mode:
> 
> https://github.com/openpnp/openpnp/wiki/Nozzle-Rotation-Mode
> 
> Your code needs to anticipate that, and I frankly don't quickly see how. ☹️

I don't understand how that is related and how it may cause addition trouble. I created special functions in the JobProcessor (actually long time ago) that return approximate locations of what the JobProcessor will instruct then actually executing the job. This functions are used to gather rotation information which are send to the motion queue as subordinates. If the JobProcessor then does its job, any queued rotation not overwritten by dominator data is merged into the next motion segment. That works well for all steps because the precise rotation angle is always known upfront (the pre-rotation is always done immediate before the actual rotation/motion).

To my understanding the nozzle rotation modes effect the absolute angle of the nozzle when picking a part. But as this angle is calculated as part of the pick step by the JobProcessor in the same way as the locator previously predicted, the result is the same (I verified that and also tuned the locator functions as part of this PR - previously the rotation was not used/needed and hence not calculated).

We also discussed the difficulties, that may arise from machines with shared rotation axes. Again, the pre-rotation step will calculate the required rotation for each nozzle individually and queue them for subordination. Due to the sharing, this will result in queuing the same physical axes multiple times and the dominator move the JobProcessor requests overwrite it as well resulting in probably no pre-rotation at all. (For a machine with four nozzles sharing two rotation motors there shall be four pre-rotation requests generating rotation requests for two axes of which one is dominated by the actual move leaving one axis to rotate in advanced.)

Finally, my intention is to provide a fully recessive way to move addition axes. If there is any problem in the prediction of the location, the queued subordination may move the axes to a wrong location which will later be corrected by the dominant move. This again shall not cause any harm, it's just a missed optimization.

Em I missing anything?

Jan

janm012012:
Hi @markmaker!

I've now pushed the new ok-to-move stuff. I changed my collectOkToMove() to your getOkToMove(AxesLocation targetLocation) and wrapped it by a try/catch to handle the exception getLocation() of HeadMountable can throw. I removed any checks of the queued subordinates motion while merging.

According to my tests, the code is find now. The subordination works with minimum stuff in the queue (due to MotionOption.SubordinateRotation there are only rotation requests now) and there are no unexpected non-empty queues when executing waitForCompletion(). I also verified that MotionOption.SubordinateZ does not generate anything unexpected anymore.

I'm awaiting any further commends.

Jan

markmaker:
> > Unrelated to the current discussion, I just remembered one additional complication, namely Nozzle Rotation Mode:
> > https://github.com/openpnp/openpnp/wiki/Nozzle-Rotation-Mode
> > Your code needs to anticipate that, and I frankly don't quickly see how. ☹️
> 
> I don't understand how that is related and how it may cause addition trouble. ...

The problem is with the call to `prepareForPickAndPlaceArticulation()`, here:

https://github.com/openpnp/openpnp/blob/1dbab2552b02092a93eaa2a0684e808aee8b562a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L792-L796

this changes the `org.openpnp.spi.base.AbstractNozzle.rotationModeOffset`, which in turn changes how the nozzle rotation is converted to head coordinates, here: 

https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java#L478-L488

And the head location is then converted to the raw `AxesLocation` that does the actual `moveTo()` we're talking about. 

As the `rotationModeOffset` is only changed  immediately before `nozzle.moveToPickLocation(feeder)`, it will be different from the offset that was valid when you prepared the rotation in `PrerotateAllNozzlesForPick`.

Maybe you can simply do `prepareForPickAndPlaceArticulation()` earlier, in `PrerotateAllNozzlesForPick`?

But this needs extra thought!

For machines with shared rotation axes, some modes would be wrong for the shared nozzle, because the shared rotation is not done yet, i.e. the wrong rotation taken into consideration. This would break the **MinimalRotation** mode (make  it slow), and worse, the **LimitedArticulation** mode (malfunction!).

But perhaps, you could do to it twice, in  `PrerotateAllNozzlesForPick`, and in `Pick `(as it is today), to remedy that situation, and potentially other situations (even without shared axes), when some [future] code changed the axis rotation between the steps.

This also needs thorough testing, best on a real machine. On a simulation machine, the best you can do is to extremely carefully explore the G-code log, and disregard anything else. Extremely difficult to be sure, these offsets, and the modulo 360°/180° calculations are a pita to follow. All Nozzle Rotation Modes would have to be tested.   :frowning_face:

_Mark

markmaker:
Another issue with my own code (sorry):

```java
                    // get us the target raw controller head location
                    AxesLocation rawTargetHeadLocation = new AxesLocation(headAxes, 
                        (axis) -> (targetLocation.getLengthCoordinate(axis)));
```

This should be...

```java
                    // get us the target raw controller head location
                    AxesLocation rawTargetHeadLocation = new AxesLocation(headAxes, 
                        (axis) -> (targetLocation.contains(axis) ? targetLocation.getLengthCoordinate(axis) : null));
```

... as the `targetLocation` might not actually contain that axis. Otherwise, it would be substituted with a `0` coordinate which is wrong.

I'm not sure, maybe there are always all machine axes in it, but let's be on the safe side.

_Mark

janm012012:
Hi @markmaker!

I've updated the location transformation. Haven't seen any errors yet, but better safe then sorry.

Concerning the rotation mode: In the early state of my development I've switched the rotation mode to AbsolutePartAngle, which is likely the worst to detect errors... Anyhow, I debugged verifying the GCode only. I'll now evaluate the code using different nozzle rotation mode settings. Likely my Locators will require modifications to reflect the nozzle rotation rather then the pick/place rotation. Many thanks for pointing that out!

While playing around with different rotation modes, I already had to notice, that the entire pre-rotation/subordination is working but generating no G-Code at all once a nozzle has rotation limits enabled. Even if the requested angle is well within the limits, the pre-rotation/subordination code does not generate any G-Code while the actual pick/place code does. Any idea why thats happening?

Jan

markmaker:
> While playing around with different rotation modes...  does not generate any G-Code while the actual pick/place code does. Any idea why thats happening?

Well, the **MinimalRotation** does exactly that on purpose (on pick). Other modes are optimized too, if the range is sufficient. Some might be already at the right angle by sheer luck. Otherwise, I don't know. 

You need to test with different feeders with different angles and different **Rotation in Tape**.

You need to test with different placement angles.

You should also test with a shared rotation axis.

I know, its hard...

_Mark

janm012012:
Hi @markmaker!

I've stepped through the code and suspect, that limitRotationAxis (AbstractMotionPlanner, line 802) is the problem. If any rotation axes has limits, it is called with axesLocation set to the entire set of axes. This may includes with pre-rotation/subordination multiple rotation axes. Then hm.toTransformed(axesLocation) (in line 811) strips all but one rotation (don't know if the one surviving is somehow predictable). So at the end limiteRotationAxis returns less rotation axis then it was called with.

I guess the fix would require limitedRotationAxis to respect its axis argument when applying toTransformed() and when applying the limited axis to the set. Does toTransformed() require good (x, y, z) values? if not, one could filter the respective axis for toTransformed() and merge the result back into the input rather then transforming - to me - related (x, y, z) values back and forth.

Jan

janm012012:
Hi @markmaker!

There is an other issue with applying the rotation limits: limitRotationAxis requires a headmountable, which is passed in as first argument. This argument is taken from the moveTo() call further up in the call chain. However, the subordination adds addition axes to the location set, that may not be related/mapped to that headmountable. This results in confusion when applying the rotation limits as the relation between headmountable and axis may be invalid. Would you please tell me why the toTransform() is needed when applying the rotation limits? The commend says "transform to HeadMountable coordinates in order to perform the correct angular math". Does that mean, that headmountables may have rotational offsets and that there is an additional transformation between this moveTo() call and the actual coordinates send by the driver?

Looking into the toTransformed() method, it seems, the only real problem is, that the headmountable and the axes added via subordination do not match. This could be solved by always taking the axis default head mountable. Then there is still the merging with all other axes required to preserve them.

Btw: I've seen, that LocationOption allow to apply soft limits. Is there a reason why toTransformed() is called without applying the limits and the limits are applied later "by hand"?

Jan

markmaker:
Hi @janm012012,

thinking about this, I just noticed, there is another bug in my own code I gave you, so sorry:

```java
                    // take the raw axes location of the current hm location
                    AxesLocation rawCurrentLocation = hm
                        .toRaw(hm.getCurrentLocation());

```

should of course include the head offset, i.e., like so:

```java
                    // take the raw axes location of the current hm location
                    AxesLocation rawCurrentLocation = hm
                        .toRaw(hm.toHeadLocation(hm.getCurrentLocation(), LocationOption.Quiet));

```

I will answer the new questions separately in order not to mix things up.

_Mark

markmaker:
> I've stepped through the code and suspect, that limitRotationAxis (AbstractMotionPlanner, line 802) is the problem. If any rotation axes has limits, it is called with axesLocation set to the entire set of axes. This may includes with pre-rotation/subordination multiple rotation axes. Then hm.toTransformed(axesLocation) (in line 811) strips all but one rotation (don't know if the one surviving is somehow predictable). So at the end limiteRotationAxis returns less rotation axis then it was called with.

You're right. I guess it needs to check whether the `hm` actually has that specific rotation axis. The code already has the...

`hm = axis.getDefaultHeadMountable();`

...fallback. But looking inside it reveals it is not covering transformed axes :frowning_face: 

I propose replacing the code at `getDefaultHeadMountable()` with these two methods:

  ```java
    /**
     * @return The first HeadMountable that has this axis assigned. Used to capture and safely position an axis. 
     * Returns null if the axis is unused.
     */
    public HeadMountable getDefaultHeadMountable() {
        Machine machine = Configuration.get().getMachine();
        for (Head head : machine.getHeads()) {
            // Try cameras with preference.
            for (HeadMountable hm : head.getCameras()) {
                if (isMovingHeadMountable(hm)) {
                    return hm;
                }
            }
            // Then the rest.
            for (HeadMountable hm : head.getHeadMountables()) {
                if (isMovingHeadMountable(hm)) {
                    return hm;
                }
            }
        }
        return null;
    }

    /**
     * @param hm HeadMountable
     * @return true if this axis moves the given HeadMountable.
     */
    public boolean isMovingHeadMountable(HeadMountable hm) {
        if (hm == null) {
            return false;
        }
        try {
            return hm.toRaw(hm.toHeadLocation(hm.getLocation(), LocationOption.Quiet))
                    .contains(this);
        }
        catch (Exception e) {
            return false;
        }
    }
```

Then in `AbstractMotionPlanner`:

```java
    public AxesLocation limitRotationAxis(HeadMountable hm, final AxesLocation axesLocation, ReferenceControllerAxis axis) throws Exception {
        AxesLocation limitedAxesLocation = axesLocation;
        if (!axis.isMovingHeadMountable(hm)) {
            hm = axis.getDefaultHeadMountable();
            if (hm == null) {
                throw new Exception("Rotation axis "+axis.getName()+" cannot be limited when no head-mountable is mapped.");
            }
        }
        ...

```

Note, in case of shared axes, this is still not 100% fool-proof. 

With all nozzles having the same shared rotation axis, I guess it's okay, as we will have the dominant `hm` given, and it should use this one.

With multiple groups of shared rotation axes (e.g. two and two in a four nozzle machine), we don't really know which `hm` the _other_ rotation axes belong to. I'm not sure it will matter in the end, as it only affects the modulo 360° calculation, and it is quite safe to assume any transformation will be the same (or just a negation) for all nozzles on such a machine, and a negative sign does not matter (limits are actually swapped in the code). Right?

_Mark

markmaker:
> Btw: I've seen, that LocationOption allow to apply soft limits. Is there a reason why toTransformed() is called without applying the limits and the limits are applied later "by hand"?

Limiting would just clamp it to the limit. That's wrong for a rotation axis. It needs modulo 360° logic. And for axes with articulation of less than 360°, we require the **LimitedArticulation** rotation mode to _prepare_ the move (as we've seen in the JobProcessor earlier), to make sure it never exceeds the range.

_Mark

markmaker:
> I'm not sure it will matter in the end, as it only affects the modulo 360° calculation

Strike that. :weary: 

It will matter with rotation mode offsets applied. The pre-rotation will likely not work right in that (exotic) use case. It would have to be disabled. 

Speaking of that: is it correct to enable/disable this in the JobProcessor? 

Shouldn't it be globally disabled on the Motion Planner for other/future applications?

_Mark

markmaker:
> I've fixed pickLocator() to respect the nozzle rotation mode by applying the rotation mode again in the locator. This is IMHO safe because the offset does not depend on anything else then pick and place location

I'm uneasy with this. The locator is _not_ supposed to have side effects like that. Maybe at the moment it would work, but these locators are bound to be used elsewhere, maybe much ahead of time. They will be used in your #1658 many steps before in planning, right?

I would prefer when this is specifically done in the `PrerotateAllNozzlesForPick` step, _after_ getting the location from the locator. And then, to make sure it is still applied, again in the `Pick` step, as it is done today.

> Concerning the module 360° calculation ... 

I can't follow you there. What do you mean, they could be "triggered in by manual operation", and how is this related to our case?

> I've carefully tested the current implementation...

Sounds good!

I'll look at the actual code later, gotta go.

Thanks for the stamina! :muscle: 

_Mark

janm012012:
Hi @markmaker!

> I would prefer when this is specifically done in the `PrerotateAllNozzlesForPick` step, _after_ getting the location from the locator. And then, to make sure it is still applied, again in the `Pick` step, as it is done today.

I've changed the implementation by applying the rotation mode _before_ requesting prerotation of all nozzles. The locator uses the nozzles toHeadLocation(), which returns the correct angle if the rotation mode was applied before. If I apply it later, I would have to correct the angle separately...

> > Concerning the module 360° calculation ...
> 
> I can't follow you there. What do you mean, they could be "triggered in by manual operation", and how is this related to our case?

I wonted to say, that if any problem can be triggered by the subordination of rotation, it could also be triggered from the UI by rotating the respective nozzle manually.
All my implementation was designed to repeat all steps in the same order as they will be done later. This way sharing - where multiple derived axes all drive a single physical axis - would result in dropping all but the first motion on the physical axis. Even if different limits/modes are applied to the derived axes, the dropping shall result in the same sequence as will be executed afterwords.
However, maybe this is not true anymore with this latest changes: I request pre-rotation in a separate method, where a loop over all placements is used to request pre-rotation. In the pick-pre-rotation step, I've added a second loop over all placements before this to apply all nozzle rotation modes. So here first the rotation modes for all planned placements is applied and then the pre-rotation is requested. In the actual pick-step a single loop is used that applies the rotation mode and immediately picks the part. So for each placement the rotation mode is applied immediately before the pick and the next rotation mode is applied after the previous pick. I don't know if that matters. All nozzles will be used independently for align and place without reapplying the rotation mode. It just needs to be done after a feeder has been selected because the result depends on the feeders pick location.

Jan

markmaker:
> I wonted to say, that if any problem can be triggered by the subordination of rotation, it could also be triggered from the UI by rotating the respective nozzle manually.

No. The manual rotation always passes the `hm`, and there is no ambiguity between the rotation axis and its `hm`.

However, with groups of shared rotation axes (say N1 and N2 share C1 and N3 and N4 share C2) it can become ambiguous. 

Lets say you pre-rotate all four nozzles in order N1, N2, N3, N4, this will record a pre-rotation on C1 for N1 and C2 for N3, the pre-rotation of N2 and N4 will be suppressed by the queue already having this axis. 

Then later you have a N1.moveTo(), which will move C1 predominantly and C2 from the queue. C1 can be unambiguously associated with the passed `hm`, namely N1, but C2 from the queue is ambiguous. Is it for N3 or N4? With the proposed code it will assume N3 and by sheer luck be right with our example.

But your Travelling Salesman optimization can alter the nozzle order. Consider the same with optimized nozzle order N1, N2, N4, N3, and it will be wrong.

Now what happens? It will run the limit rotation code on C2 with the wrong `hm`. As far as I can see, this will still be okay, as long as the axis transformations of N3 and N4 are the same or simply negated. But if transformations are not the same, it might result in a wrong result. Assume some sort of "connecting rod" drive on the rotation axes, with different rod angles/lengths, and it will be completely wrong. Admittedly a very exotic scenario :rofl: 

Now, this is just pre-rotation, so nothing really bad will happen, limits will be kept (they are on the raw axis), just some time wasted...

_Mark

janm012012:
Hi @markmaker!

> No. The manual rotation always passes the `hm`, and there is no ambiguity between the rotation axis and its `hm`.

First: I now understand where trouble is lurking. 
The more I think about that, the more I have the feeling that there is something between design flow (moveTo only works if the AxesLocation passed are directly reflecting the axes of the head mountable also passed in) and my way of adding subordination (before applying the limits that require the head/axes relations). Looking closer the code does not look that bad to me: its just the addition call to limitAxesLocation in line 357 that is trouble some. At that point, subordinate axes are merged voiding the hm/axesLocation relation. However, the limits have been applied earlier (line 292) before the merging with hm/axesLocation still in valid relation. Then the only change (I can see) to axesLocation happens by merging subrodinate axes, which have been limited before. Therefore I don't see a need to check the limits again and would remove line 357. With that line removed, all the trouble concerning guessing headmountables from axis is gone. 
What do you think, did I overlook something?

Jan

janm012012:
Hi @markmaker!

> Yes, it might matter and that's the reason it must be repeated in the `Pick` step.

I'm sorry to disagree a little bit: the nozzle rotation mode offset is a function of the current location (rotation of the nozzle), the feeder pick location and the place location. Hence, it needs to be update if one of them changes. Insight the job processor everything is related to the placement to be processed. Therefore I would consider it sufficient to update it as part of the feeder selection. That is required for the pre-rotation to guess the pick-angle and in the pick step if not done before and if the feeder changes (due to pick retry).
Back to the topic: do you think my code is ok as it is (applying the rotation mode immediately before the pre-rotation request using the default code path via the nozzles toHeadLocation()) or do you wont me to change it (Then I'd not alter the rotation mode, undo it's effect on toHeadLocation() and model the expected effect separately).

Jan

markmaker:
> the nozzle rotation mode offset is a function of the current location (rotation of the nozzle), the feeder pick location and the place location. 

Correct.

>  Hence, it needs to be update if one of them changes. 

Correct.

>  Therefore I would consider it sufficient to update it as part of the feeder selection.

No. Given two nozzles share the same rotation axis, the pick with the first nozzle will change the nozzle rotation of the second nozzle too. Hence, according to your own rule, the rotation mode offset must be updated.

_Mark

janm012012:
Hi @markmaker!

> > Therefore I would consider it sufficient to update it as part of the feeder selection.
> 
> No. Given two nozzles share the same rotation axis, the pick with the first nozzle will change the nozzle rotation of the second nozzle too. Hence, according to your own rule, the rotation mode offset must be updated.

I'm sorry, I'm lost: the JobProcessor assumes that all nozzles can always be operated completely independently. If changing the nozzle rotation mode would effect any other nozzle, its just a matter of time when a place operation would go completely wrong. However, if two nozzles share one physical axis and both are configured for MinimumRotation, then the rotation offset of the later nozzle will depend on the rotation/pick angle of the earlier. I think this case explains a) why the offset needs to be applied again immediately before each pick (as you always said) and b) that pre-rotation will never entirely work. However, due to the axis sharing, the physical axis will be moved according to the earlier request, while the later request being obviously wrong will be dropped.
My implementation does every pre-rotation activity in exactly the same order as it will be executed later for real. The sub-ordination queuing lets axes in already in the queue take precedence over thus requested to be queued later. Except for not taking effects related to changes of the current location into account, I don't see any other problems yet.

Jan

markmaker:
>  However, if two nozzles share one physical axis and both are configured for MinimumRotation, then the rotation offset of the later nozzle will depend on the rotation/pick angle of the earlier. I think this case explains a) why the offset needs to be applied again immediately before each pick (as you always said) ...

That's basically all I wanted to hear. :+1: 

I thought is takes the current rotation into account in **LimitedArticulation** too, but looking into the code I was mistaken. But such an optimization could be added in the future.

> ... and b) that pre-rotation will never entirely work.

in the most exotic cases of multiple shared rotation axes, yes. But otherwise I think it is quite effective.

We always talked about the Pick step, but Alignment and Placements are valuable optimizations too. And those are with fixed rotation mode. Actually, Alignment might slightly adjust the rotation mode offset for alignment, but that's okay, IMHO, as it happens completely independently.

I hope I find time to check the code soon... still very busy.

_Mark

markmaker:
Just letting you know I haven't forgotten this. This PR is just a few shoe sizes too large for my spare-time brain at the moment, brain is massively loaded at job-time. 😓

_Mark

janm012012:
Hi Mark!

May I ask you for your opinion, shall we merge this or rather not? Do you see any potential issues if we do (into the fresh test branch)? Thank You!

Jan

src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
```diff
@@ -1381,8 +1477,50 @@ protected List<PlannedPlacement> optimizePlacements(Locator sortLocator, Locator
...
+        protected void prerotateAllNozzles(TravellingSalesman.Locator<PlannedPlacement> locator) {
+            // if pre-rotation is disabled, do nothing
+            if (!preRotateAllNozzles) {
+                return;
+            }
+
+            for (PlannedPlacement p : plannedPlacements) {
+                Location l = locator.getLocation(p);
+                if (l != null) {
+                    double speed = p.nozzle.getHead().getMachine().getSpeed();
```
janm012012, line 1498: Hi Mark!

While reviewing my PR, I stumbled upon a possible problem. So I'm seeking for your expertise: do I need to take the max. speed configured per part here? On one hand this is known to just request a subordinate motion, so it is not known if it actually will take place and if the properties are defined by the dominant motion. On the other hand a part configured for slow speed shall always be handle like that... What do you think?

Jan

markmaker:
No, this is always applied later, when the actual move is converted from `hm` move to head move here:

https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/spi/base/AbstractHeadMountable.java#L243

I'm really very busy, so don't expect regular answers... I'm sorry about the situation.

_Mark

janm012012:
Hi @markmaker!

May I ask for your expertise? It seems there is an issue with this PR probably in getOkToMoveAxes(). While running SampeJobTest I noticed some unexpected "Subordinate Motion queue not empty" warnings. Digging deeper, this messages seemed to be caused when a feed() is performed by the JobProcessor and the feeder is using vision. When the camera is send to the feeder, nozzle rotations have been queued and are now tried to be merged but getOkToMoveAxes() returns nothing because the logic thinks that the current location of nozzle N1 is not in Safe-Z. (For the test job both z zone limits are 0 and the current location is 0.25. The nozzle his enableDynamicSafeZ == false) I'm quite sure that the nozzle is at Safe-Z while the camera is moving. However, I don't know why the logic is not seeing that and how to fix it. Can you please help? Thank You!

Jan

PR #1658, opened 2024-07-19
Placement sorting

janm012012:
# Description
This PR provides addition sorting options to the job processor making the order of placements predictable even for panels of identical boards. In addition the sorting options are routed through the translation layer now allowing to give them more user friendly names.

# Justification
At present the sorting of placements on job start claims to provide a unique order but in fact it not always does not because for panels of identical boards there are multiple parts with identical partIds (eg. "C1") which are not handled by the default sorting leaving them in an unpredictable order.

# Instructions for Use
For normal users there is no change because the current - imperfect - sorting is preserved. Addition options are available and had to be selected explicitly.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **using debugger and code review**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **yes: PnpJobProcessor needs to expose boardId as additional sorting option and JobProcessor needs a method to be signaled that the job will be resumed**


janm012012:
I've switched this PR to draft because I'm going to implement the suggested sorting by nozzleTips.

markmaker:
This PR was discussed in the discussion group:
https://groups.google.com/g/openpnp/c/ecWR7XO4vKM/m/B3fI3ZcvBAAJ

The discussion continues here.

markmaker:
Hi @janm012012,

I believe your [last comment in the group](https://groups.google.com/g/openpnp/c/ecWR7XO4vKM/m/7UNDj-JBBAAJ) is somehow answers itself. :wink: 

> And I don't see how the Planner can be instructed to change nozzle tips at a certain placement. 

Is directly answered by

> Even though this might be no real problem as all placements compatible with a nozzle shall be in the same group and hence be planned to use the same nozzle tip anyhow. 

I can only say, "exactly". It will simply (and literally) "sort itself out" in the planned order. As soon as the previously loaded nozzlet tips don't have any compatible parts left on the list, the new nozzle tips will be loaded as planned because the parts appear _next_ on the planned list. And then all the compatible parts will be placed using these loaded nozzle tips as planned, simply by virtue of their compatibility. 

And this will work when there are multiple nozzle tips with overlapping/identical compatibilities. It is well-known that having multiple identical nozzle tips for the most frequent parts is very efficient. It will be the only way to get very frequent co-location picking, i.e. picking with multiple nozzles from the same high volume part feeder. Because all the compatible parts are ordered _together_ on the list, it will automatically load _all_ the identical nozzle tips _together_.

What is more important: this is _resilient_ planning. If a feeder runs out of parts and is disabled (max retries exhausted), the planning will just skip these parts but otherwise continue as planned, dynamically _redistribute_ the subsequent parts to the compatible nozzle tips.

> While debugging the code I found, that the results for multi-nozzle machines might be not perfect yet ... That assignes the same nozzle tip to all placements leaving other nozzle tips required to place the package with other nozzles empty. 

Again, I can only say "exactly". But again, you don't get that problem when just using the planned order loosely, and let the compatibilities play out.

Convinced? Please get rid of the redundant `plannedJobPlacements` list.

_Mark

janm012012:
Hi @markmaker!

Many thanks for your review!

I've rebased the PR now onto the head the test and removed the optimization to remember the list of sorted/optimized placements. (I may put that into an other PR as the optimization for "NozzleTip" takes a significant amount of time now and it is redundant in almost any case. Will see how it works in practice.)

Unfortunately the code does not work anymore with the optimization removed because the relation between jobPlacements is lost resulting in the placement not being marked as Completed correctly anymore.  In ReferencePnPJobProcessor line 1293 the jobPlacement.setStatus(Status.Complete); references a different placement then the global one of the job processor ("jobPlacements" line 127). This results in an endless loop as the placement stays in the "Processing" state it was set to in line 489. I suspect, that line 808 is causing the problem: in this loop the list of placements compatible with a particular nozzle tip is build. I use ".map(j -> {return new JobPlacement(j, nozzleTip); }" to create fresh elements so that the same placement can be independently listed for different nozzle tips. Here each element shall still reference the same placement in the global list of the job processor. I think, the constructor (line 24) might be wrong. How do I have to write it that it only references the existing boardLocation and placement and does not copy them?

Jan

janm012012:
Hi @markmaker!

I dug a little deeper and found the real problem. The job processor uses a list of JobPlacement to store placements with their location, status and error state (line 127). This list is used (filtered for pending placements only) to the sorting/optimization step and the result is stored in a local list (line 470). This local list is then further filtered for one placement per nozzle and hand over to the next step. Finally (in the place step, line 1293) the placements status is updated to Complete. This last step makes the assumption (and only works) that the elements received as part of the constructor of the place step still point to elements of the job processors global list (line 127). If that's not the case, the placement is not marked as completed and will be in the list of placements to be processed on next run. I'm not sure, but it looks a little like a design flow.

My code triggers this problem at present, because I extended JobPlacements by nozzleTip and duplicated the elements to store them (with different nozzle tip information) in the respective list.

Any suggestions how to fix that?

I'd say, that the cleanest would probably be to change the interface to the JobPlanner to except a list of PlannedPlacements and generate such a list as part of the sorting/optimization step. PlannedPlacements would then contain the planned nozzle tip information and just a reference to it JobPlacement.

What do you think?

Jan

janm012012:
Hi @markmaker!

I just noticed, that there is an other potential issue with the Sort-by-Nozzle-Tip option on multi-nozzle machines (at least in the sense I've implemented it): if a part can be processed by more then one nozzle tip, the sorting will always assign one of them even though a second nozzle with a different tip could also be used. For example: suppose there are two identical parts to be placed and both support two nozzle tips then the selection will choose one for both parts. In the current implementation the planner will find, that a nozzle-tip has been selected and use that. He will miss the opportunity that the other nozzle tip could be used on the other nozzle to place both parts in one pass. I hope you get the point: if the sorting/optimizations sets nozzle tips to use, the planer is not and shall not be free to assign any other nozzle tip. This makes multi-nozzle machines with different tips to handle the same part in parallel on multiple nozzles inefficient...

Any thoughts on that?

Jan

markmaker:
Hi @janm012012,

oh no, you must somehow have run the `ReferencePnpJobProcessor.java` through a code formater. Overall whitespace, line-breaks and comments were reformated, in some cases there were lists in comments that are now "melted" together. Needless to say it is very hard to track the actual changes. Please untangle this.

Regarding your three last comments, I don't think we are on the same page. You seem to insist to want to store and enforce specific nozzle tips to be used on individual placements, despite finding problems with that approach. Why?

According to my level of knowledge, this is not necessary, so far I don't see any evidence to the contrary:
https://groups.google.com/g/openpnp/c/ecWR7XO4vKM/m/0-6qO-8_AQAJ

By sorting the placements in the nozzle tip compatibility order, they will be candidates in the intended order. They will automatically be assigned to compatible nozzle tips. No need to store those assignments.
  
The only thing needed in _addition_, is to store and use the order of the nozzle tips in their "largest group of compatibilities" order. 

To be more specific, here (in your code) you have the sorting order:
https://github.com/openpnp/openpnp/blob/347d3daf944c9b2c526095470ea4d9855c82d453/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L813

You should now store a copy of the list of nozzle tips in that order, _before_ eliminating duplicates. Same type nozzle tips (equal compatibilities) should automatically be grouped together that way for most efficient multi nozzle machine usage.

And here (in the old code, where this is still intact) you need to apply that nozzle tip order:

https://github.com/openpnp/openpnp/blob/37f58f7769b32689402f24ef028bc688d9962004/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L1704

The order will ultimately be relevant to preferentially load a new nozzle tip, when the loaded ones have placed all compatible parts in the _planned order_. It takes the first with `.get(0)`, i.e. the _planned order_ of nozzle tips is again guaranteed:

https://github.com/openpnp/openpnp/blob/37f58f7769b32689402f24ef028bc688d9962004/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L1807

Having said all that: this system is not perfect, and we already knew that from the beginning. We already knew that this optimization is only really working when you have the **same number of  equal-compatibility nozzle tips as you have nozzles**. Otherwise, the compatibility groups will be overlapping, and the traveling salesman optimization will not be very effective, because the traveling routes of two compatibility groups are now mixed together, i.e. feeders will likely not be co-located. Overlapping compatibilities will likely also derail the planning in other way, as compatible placements are not exhausted at the same time on all nozzles.

But I still think the optimization is worthwhile: the **same number of  equal-compatibility nozzle tips as you have nozzles**-scenario would really be much, much faster, and assuming that on modern small part projects the first set of nozzle tips would cover ~90% of parts, that will boost speed, even if optimization will then deteriorate in the rest of placements. 

_Mark

janm012012:
Hi @markmaker!

Finally I found the time to complete the open issues. With my last commits the list of plannedNozzleTips is created and hand over to the planner and used there as list of candidates for the next nozzle tip change.
In my test scenarios this works well, even with the Minimize strategy of the plan. As you assumed, the candidate list is accessed from back to front selecting the next best tip. However, for multi nozzle machines, the list is not minimized and hence may still contains tips, that shall not be used. In my test scenarios this only happened once with the job and nozzle tips in an unknown state. I guess thats the known imperfections we discussed earlier. Anyhow, to me its still a good improvement.
Please have a look and tell me what you think now.

Jan

markmaker:
Same brain overload problem here, I'm sorry.  😓

markmaker:
Haven't forgotten you, but still 🤯 at work.

PR #1659, opened 2024-07-25
Start with loaded nozzle tips

janm012012:
# Description
This PR adds an option to the JobProcessor to always start with the first component of the job. At present the planner starts with the current loaded nozzle tips, which void the selected placement order.

# Justification
As discusses on the list, to continue with the current loaded nozzle tip might be efficient but it may cause issues, when small components will be placed close to tall ones. In addition, it voids the expressed wish of the user how parts shall be sorted.

# Instructions for Use
The default is to keep the status quo.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **using simulation**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **yes: the PnPJobPlanner interface needs a new method to signal that a nozzle tip change may be considered on next call**


janm012012:
I changes the UI from tick box to drop down to provide more options:

- Performance: avoid nozzle tip changes (current default)
- Relaxed: place first part of the list using the required nozzle tip and then avoid further nozzle tip changes
- Strict: strictly execute the list of placements potentially executing more nozzle tip change then needed.

**Performance** is the current mode of operation and the default with this PR. **Relaxed** allows to execute a job always in the same order as the first placement is always placed first. **Strict** is required to support more advanced placement sorting at job processor level.

markmaker:
Hi @janm012012,

thanks for this. :+1: 

This seems fine, except for cosmetics. 


- Why do you call the old default strategy `Performance`? It won't be "Performance" at all once you finish your #1658, right? :sunglasses: 

- `Relaxed` also seems not to really pinpoint what it does, right?

- Finally, `Strict` seems unimplemented, or am I missing something?

I suggest you name it for how it behaves. Perhaps `Opportunistic`, `AsPlanned`. 

And remove `Strict`. It doesn't make sense even in the future, assuming my explanation [here](https://github.com/openpnp/openpnp/pull/1658#issuecomment-2368771170) holds true.

_Mark

janm012012:
Hi @markmaker!

Many thanks for reviewing this PR. I'm fine with renaming the settings as you prefer. Actually I had some problems finding reasonable good and self-explaining names for the three possible settings...

With this PR the JobPlanner actually get a configurable setting that has three possible values. The internal logic of the Planner consists of two strategies, a) process a placement with the nozzle tip currently loaded and b) process a placement that can be handled using any nozzle tip compatible with this nozzle. 

At present, the Planner always executes a and - if that was not successful - b. This provides the best performance by minimizing the amount of nozzle-tip changes but violates the requested processing order of the JobProcessor (which is - at present - either PartHeight or Part). I named this setting "Performance".

With this PR I added the option to always use b (named "Strict") which follows the ordering the JobProcessor has planed closely with potentially more nozzle-tip changes (The Planner does not know anything about the additional sorting options the JobProcessor will receive with the other PR). (I've to admit that there is currently no option to only accept exactly the next planed placement, which would likely result in a big performance penalty for multi nozzle machines with distinct nozzle tip for each nozzle and likely no wonted.).

Finally, this PR added the option to start a job with b and then continue with the old a, b behavior. I've named this option "Relaxed" because it shall force a nozzle tip change at job start and continue the job with the same performance as today. This option would be an intermediate step between "Strict" and "Performance" and would let the JobProcessor follow the user configured ordering more closely. (At present, my jobs run with nozzle-tip a->z and then from z->a in a forward backward manner due the a, b strategy. With "Strict" they will only run a->z but with the same nozzle-tip ordering/performance as before.)

I think, we shall keep all three options as I see use cases for all of them. Please let me know which names you'd prefer and I'll change them straight away. It would be also possible to route the names thought the translation layer...

Jan

markmaker:
Hi @janm012012,

maybe the option can be described as "Nozzle tip loading strategy" (tooltip). Then the options could be named:

`None`
`StartAsPlanned`
`FullyAsPlanned`

_Mark

janm012012:
Hi  @markmaker!

I've renamed the options almost as suggested: instead of "None" I used "Minimize" because that's to my understanding the "Nozzle tip loading strategy" that's used: favor to reuse the loaded nozzle tip over following the requested placement sorting.

I also added a tool tip which shall explains the setting a little more detailed.

Jan

markmaker:
Thanks, @janm012012 :rocket: 

Please announce on the list, and perhaps add instructions in the Wiki.

_Mark

janm012012:
Many thanks! I've just updated the wiki and send the announcement.

PR #1660, opened 2024-08-02
ReferenceStripFeeder pick location error

tobydickenson:
# Description

ReferenceStripFeeder is configured with two hole locations. If vision is enabled, it uses the top camera to locate the hole near the component before pick. Ideally the pick location should be entirely determined by the vision output, however there are some scenarios where it is affected by inaccuracy in the configured location.

In practice, I have been seeing that the first component picked from a strip is intermittently picked off-centre.

We can reproduce this in the default configuration simulated driver. The feeder "Lower Strips - 1" has configured hole locations:
x:147.347 y:40.285
x:147.433 y:36.238
Lets introduce a 1mm location error by changing these hole locations to the following. This simulates manual loading of tape into a tape holder with less than perfect alignment.
x:147.347 y:41.285
x:147.433 y:37.238
When performing the first feed and pick we see that the vision correctly identifies the first hole, but the pick location is 1mm out. The simulated machine still reports a successful pick, but we can see that the pick location was wrong.

image

Previously ReferenceStripFeeder has not stored quite enough vision state to constrain the pick location in every scenario. This branch fixes the problem by storing a vision-adjusted reference location (ie location of the first hole), and using that when calculating pick locations.

# Justification
This affects all users of ReferenceStripFeeder.

# Instructions for Use

This is not a new feature - this is getting ReferenceStripFeeder performing with the expected accuracy.

# Implementation Details
A. This branch has been tested by hand-driving the simulated machine, as shown in the screenshot above.
B. It has been tested with [these strip holders](https://www.opulo.io/products/lumenpnp-strip-feeder) on a lumenpnp machine. Those strip holders have no strip alignment feature, so it was pretty easy to trigger this bug by having less than perfect accuracy when inserting the strip into the holder. It has been tested with 0402 and 0603, with picking and recycling. Picking from the start of the tape in order, and picking random locations. 
C. It would be great to have an automated test for this. It looks like the existing tests using the simulated machine do not report an error if the actual pick location is wrong? Feedback is appreciated if I can improve this.

I hope so, but feedback is appreciated. I dont use java very often.

Tests pass

(This is my first openpnp patch - thank you!)


markmaker:
Hi @tobydickenson,

this looks good!

Some questions. Note, I'm just asking, not judging, so if your justification is reasonable, it will be fine _as is_ :smile: 

1. What was the justification to only recalibrate the first hole and not the last hole? Doing both would cover rotation in a shifted/mountable feeder, right?
2. Frankly, from a quick code scan I don't understand the existing (even before your PR) "last hole" semantics in `getIdealLineLocations()`. If it really is the last hole in the strip and the reference hole is the first hole, then the distance `d1` will practically always be larger and `visionLocation` will never be used, i.e. the vision operation preformed for every hole is for nought? Am I missing something? (this point is somewhat mutually exclusive with the previous point). 
3. What is the justification to not make recalibration optional? Recalibrating will take extra time on first feed, and some users might not like that. But then again, the strip feeder is not exactly the most "pro" feeder, so it could still be reasonable. :wink:    
4. As far as I see, there is currently no way to reset vision, other than restarting OpenPnP or settings the feed count to zero. But there might be situations, where a recalibration is wanted, without resetting the feed count. Other feeders (and other calibration item such as nozzle tips) listen to the machine state and reset the calibration when `machine.isHomed()` becomes false, e.g., see the `BlindsFeeder` constructor. So the user could just re-home the machine to make sure everything is recalibrated. What do you think, could this be useful here too?

_Mark

tobydickenson:
Hi, Thanks for your time in reviewing this.

> If it really is the last hole in the strip and the reference hole is the first hole....

Yeah, the lastHoleLocation name is misleading. The automatic configuration wizard sets referenceHoleLocation and lastHoleLocation to the first two hole positions. But fixing this misleading name will break machine.xml compatibility, right?

> What was the justification to only recalibrate the first hole and not the last hole? Doing both would cover rotation in a shifted/mountable feeder, right?

I do not believe this to be significant. The first pick at feedcount zero uses the calibrated position for the first hole, and combines that with uncalibrated strip orientation to determine the pick location relative to the hole. It would take a huge orientation error to cause a picking problem.

All subsequent picks (and first picks at a non-zero feedcount) are entirely based on calibrated data.

> What is the justification to not make recalibration optional? Recalibrating will take extra time on first feed, and some users might not like that. But then again, the strip feeder is not exactly the most "pro" feeder, so it could still be reasonable

This patch introduces zero extra overhead for first feed at zero feedcount. 

There was always an option to disable all vision for a feeder, for anyone who does not want to waste time on vision calibration during the job.

This patch introduces one extra vision operation on the first feed at a non-zero feedcount, but I dont see a way to avoid that without deeper changes to the code structure and changing the referenceHoleLocation/lastHoleLocation system. This overhead feels small, and justified.

My next project after this patch will be performance improvements for ReferenceStripFeeder. Once calibrated, I am pretty sure there are common scenarios where it would be sufficient to only use vision on every Nth hole. For most users this will more than outweigh the overhead of one extra vision operation on first pick.

> e.g., see the BlindsFeeder constructor. So the user could just re-home the machine to make sure everything is recalibrated. What do you think, could this be useful here too?

Yes! Thank you. Arguably it is a bug to not clear any cached vision data during re-home.

> As far as I see, there is currently no way to reset vision, other than restarting OpenPnP or settings the feed count to zero. But there might be situations, where a recalibration is wanted, without resetting the feed count.

Does a "reset vision" button on the wizard make sense? Maybe this is not justified if re-home will have the same effect.

tobydickenson:
> > What was the justification to only recalibrate the first hole and not the last hole? Doing both would cover rotation in a shifted/mountable feeder, right?
> 
> I do not believe this to be significant. The first pick at feedcount zero uses the calibrated position for the first hole, and combines that with uncalibrated strip orientation to determine the pick location relative to the hole. It would take a huge orientation error to cause a picking problem.

There is a practical reason too. When running a job that need N components, we know we can reliably scan the first N holes because they are adjacent to those components. But all other holes are unreliable. The tape might have been cut off, so hole N+1 would be missing. Maybe the film peeled back to expose the first N components is obscuring hole N+1.

So it is good to only scan the first hole when picking the first component.

markmaker:
I see, makes sense. And very nice improvements. 

The only cosmetic issue I have left is with the "Clear Cache" label of the button. Non-developers probably don't know what "cache" means in this context. I recommend calling it as you did in the code "Reset Vision" or perhaps "Reset Calibration". A tooltip explaining the purpose would be nice too.

Apart from that, is it ready to merge?
 
_Mark

markmaker:
Note, I'm currently occupied and only reading email sporadically...

tobydickenson:
All done and ready to merge

markmaker:
Please announce the improvements in the discussion group.

Updating the [Wiki](https://github.com/openpnp/openpnp/wiki/ReferenceStripFeeder) would also be much appreciated. :+1: 

_Mark

PR #1662, opened 2024-08-14
ReferenceStripFeeder performance enhancements

tobydickenson:
# Description

This is the second half of my ReferenceStripFeeder changes. The previous pull request was about correctness and accuracy; this is about performance.

The baseline ReferenceStripFeeder will always perform a vision scan before picking. This patch improves performance by sometimes skipping that vision step. As before it keeps track of two hole locations - the first hole, and the most recently checked hole - and can use that to extrapolate over a short distance. It implements the following logic:

1. It skips vision if this location is the same as the previous location. This occurs in two key scenarios:
  a. For 0402 and smaller, there are two components for each hole. The second component can be picked without re-scanning the same hole.
  b. Following a "recycle", where the component is being picked up from the same location it was just deposited.
2. A fixed **linear limit**. If performs a vision check if the distance between this hole and the previous is greater than a new configuration parameter, named "Extrapolation Distance". For example, setting this to 12mm will allow vision checks to occur for 1 hole in 4. It skips the vision checks for the holes at offset 4mm, 8mm, and 12mm, then checks the location of the hole 16mm distant from the previous check.
3. A **proportional limit**. It performs a vision check if the distance between this hole and the previous is greater than 70% of the span between the two tracked holes. The practical consequence of this is that it will check the first 3 holes, but skip the 4th (if allowed by the linear limit).
4. Otherwise, it can use the extrapolated position.

Note that performing a full vision check can always be forced by homing the machine, resetting the feeder, or pressing the "reset vision" button.

# Justification

The baseline ReferenceStripFeeder with 0402 is gratuitously wasteful in performing unnecessary work. The second vision check on a hole adds no benefit. Skipping that second vision check has no disadvantage.

All users of ReferenceStripFeeder can potentially benefit from a significant improvement in throughput; about 12% improvement in my jobs this week. 

The examples above have the Extrapolation Distance parameter set to 12mm. That is both:
 * Conservative in maintaining pick location accuracy. Strips are pretty straight and predictable over a 12mm span
 * Enough that there are diminishing returns in considering a larger number. Vision has a tiny impact on overall job throughput after being reduced by 75%.

# Instructions for Use

Proposed new wiki text
```
= Vision

ReferenceStripFeeder uses vision to confirm the location of the holes in the strip. This
determines the precise component pick location, and the starting point for the next
vision cycle as the feeder progresses along the strip.

Cached vision data is reset when the feeder is reset or the machine is homed. If you
move the strip at some other time then press the "Reset Vision" button to avoid
inaccurate positioning.

By default ReferenceStripFeeder will vision check every hole. This is safe, but for many
strips this in unnecessarily cautious and the placement cycle time can be reduced if it
can perform fewer vision checks. The "Extrapolation Distance" field specifies the
maximum distance between holes before a vision check is needed. For example,
setting this to 12mm will allow 1 in 4 holes to be skipped; it will next check the hole
at 16mm, after skipping the holes at 4mm, 8mm, and 12mm. This is a maximum,
and note that ReferenceStripFeeder will perform more checks on the first few holes
at the start of the strip.
```

# Implementation Details

This has been tested on the reference simulated machine. Useful manual tests are:
* 0402 with the default configuration. It picks both components next to each hole for each vision check
* With default configuration, larger components behave the same as previous versions. It will do a vision check before every pick.
* Extrapolation Distance set to 12 allows it to pick several component between each vision check.

This has also been tested on (and developed for) a production run this week using 30 strip feeders of length 180mm. When everything was running smoothly this patch gives a 12% throughput improvement.

(my vision pipeline is sometimes affected by uncleanly punched holes in paper tape, where paper fibres are visible in the hole. So I have secondary benefits from minimising the hole vision checks)

5. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?
yes

7. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.
8. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.
tests pass

# Analysis of the 70% proportional limit

This change performs a vision check if the distance between this hole and the previous is greater than 70% of the span between the two tracked holes. The practical consequence of this is that the first few components on the strip have vision checks controlled by this geometric progression of vision checks. After the first few components, the "70% of the span between the two tracked holes" distance is greater than the user-configured linear limit, and vision is checked for holes at regular intervals.

If the measurement uncertainty on a single hole vision check is normally distributed with standard deviation v, then the worst case uncertainty on the extrapolated position is also normally distributed with standard deviation sqrt(0.7² + 1.7²)v which approximately equals 1.84v.

Vision error is expected to be much smaller than overall pick error, so this is safe.

This 70% number is large enough that this proportional stage will not be very visible to the user. Most of their tape is governed by the linear Extrapolation Distance threshold.  My proposed explanation in the wiki says "note that ReferenceStripFeeder will perform more checks on the first few holes at the start of the strip" and that feels like a sufficient explanation for what they will see.

# UI considerations
The name "Extrapolation Distance" on the button....... Would something less technical be better? I think it is ok. There is a tooltip explaining its purpose in non-technical language, and those words will be searchable on the wiki.


tobydickenson:
This branch was just rebased onto the latest test branch

markmaker:
@tobydickenson , just letting you know I haven't forgotten this. This PR is just a few shoe sizes too large for my spare-time brain at the moment, brain is massively loaded at job-time. 😓

_Mar

tobydickenson:
> @tobydickenson , just letting you know I haven't forgotten this.

I understand. Thanks for the update.

An update from my side: I have been running this in production since sharing and consider it to be functionally robust, and a valuable improvement.

PR #1665, opened 2024-08-29
Add updated GStreamerCamera based on AbstractBroadcastingCamera

ian-arkver:
This is a rebased and updated version of PR #1023 which was merged but later reverted.

It was reverted because at the time gstreamer dependencies broke the population of a dropdown selection option in OpenPnPCaptureCamera. Specifically the resolution dropdown for UVC cameras under Device Settings would set the value and it would be stored to machine.xml, but on restart of OpenPnP the dropdown would show the initial entry again.

Since some interest was shown in GStreamerCamera on the discord server, I retested and discovered that this is no longer the case. The dropdrown populates correctly with the stored value. I'm resubmitting a rebased version of the camera.

Original PR text follows, slightly tweaked...

---

This camera uses a textbox to specify an arbitrary gstreamer launch pipeline.

An appsink is appended to this pipeline and samples are pulled and fed to OpenPnP.

The pipeline should end with an explicit capsfilter element with caps of video/x-raw.

**Description**

Adds GstreamerCamera class, based on ReferenceCamera class which allows use of
an arbitrary gstreamer pipeline as a video source. This can be for example a v4l2src,
nvarguscamerasrc, rpicamsrc, rtsp receiver and decoder, media file reader and playbin, etc., etc.

**Justification**

Other camera sources are available which can read from v4l2 devices, perhaps even using
gstreamer to do so, however the pipelines used are fixed by the libraries. Although many
offer good controls, none allow arbitrary gstreamer sources to be used. This rules out their
use where a specific, non-standard camera source element is needed.

This camera source might also be useful for testing purposes where a pre-recorded video
sequence could be played back.

**Instructions for Use**

GstreamerCamera is added to the list of available camera sources. When added, a wizard is
available with a single text box under "Device Settings". This box is used to specify the gstreamer
launch pipeline, in normal gst-launch-1.0 syntax. The pipeline should be capable of producing
video/x-raw,format=BGRA on little-endian machines and xRGB on big-endian machines.
It should finish with a capsfilter element. An appsink is automatically linked to the end of the
pipeline with these caps.

It is recommended to use gst-launch-1.0 to test the pipeline. Once video is correctly working it can
be copy/pasted into OpenPnP. The pipeline string will be saved in machine.xml.

**Implementation Details**

A new dependency is added to pom.xml to bring in org.freedesktop.gstreamer bindings.
A new GstreamerCamera.java and GstreamerCameraConfigurationWizard.java are added.
The new class is added to the list of available camera types in ReferenceMachine.java.

I believe the new files conform to the coding style.
No spi or model changes are introduced.
Maven tests are not affected as they don't use this camera, however they continue to pass.

The camera has been tested on a Jetson Nano using both a MIPI camera with nvarguscamerasrc
and a USB camera with v4l2src, as well as on a PC using streamed H265 sources.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -241,6 +241,11 @@
             <artifactId>commons-lang3</artifactId>
             <version>3.12.0</version>
         </dependency>
+        <dependency>
+            <groupId>org.freedesktop.gstreamer</groupId>
+            <artifactId>gst1-java-core</artifactId>
+            <version>1.4.0</version>
+        </dependency>
 	</dependencies>
 	<build>
 		<plugins>
--- a/src/main/java/org/openpnp/machine/reference/ReferenceMachine.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceMachine.java
@@ -48,6 +48,7 @@ import org.openpnp.machine.reference.axis.ReferenceControllerAxis;
 import org.openpnp.machine.reference.axis.ReferenceLinearTransformAxis;
 import org.openpnp.machine.reference.axis.ReferenceMappedAxis;
 import org.openpnp.machine.reference.axis.ReferenceVirtualAxis;
+import org.openpnp.machine.reference.camera.GstreamerCamera;
 import org.openpnp.machine.reference.camera.ImageCamera;
 import org.openpnp.machine.reference.camera.MjpgCaptureCamera;
 import org.openpnp.machine.reference.camera.OnvifIPCamera;
@@ -466,6 +467,7 @@ public class ReferenceMachine extends AbstractMachine {
         l.add(SwitcherCamera.class);
         l.add(SimulatedUpCamera.class);
         l.add(MjpgCaptureCamera.class);
+        l.add(GstreamerCamera.class);
         return l;
     }
 
--- a/src/main/java/org/openpnp/machine/reference/camera/GstreamerCamera.java
+++ b/src/main/java/org/openpnp/machine/reference/camera/GstreamerCamera.java
@@ -52,8 +52,8 @@ public class GstreamerCamera extends ReferenceCamera {
         Gst.init();
     }
 
-    @Attribute(name = "pipeline", required = true)
-    private String pipeString;
+    @Attribute(name = "gstPipeline", required = true)
+    private String gstPipeString;
 
     private BufferedImage currentImage;
     private AppSink videosink;
@@ -66,11 +66,12 @@ public class GstreamerCamera extends ReferenceCamera {
 
     @Override
     public synchronized BufferedImage internalCapture() {
-        if (! ensureOpen()) {
+        if (!ensureOpen()) {
             return null;
         }
 
-        // AbstractBroadcastingCamera uses atomic getAndSet on images along with hasNewFrame
+        // AbstractBroadcastingCamera uses atomic getAndSet on images along with
+        // hasNewFrame
         // Should be safe just to return the latest one here.
         newFrame.set(false);
         return currentImage;
@@ -86,20 +87,20 @@ public class GstreamerCamera extends ReferenceCamera {
         close();
         clearCalibrationCache();
 
-        if (pipeString == null) {
+        if (gstPipeString == null) {
             return;
         }
 
         Bin bin = null;
         try {
             // Create ghost src pad as we link the whole bin to the appsink
-            bin = Gst.parseBinFromDescription(pipeString, true);
+            bin = Gst.parseBinFromDescription(gstPipeString, true);
         } catch (Exception e) {
-            Logger.warn("Exception parsing pipeline {}", pipeString);
+            Logger.warn("Exception parsing pipeline {}", gstPipeString);
             return;
         }
         if (bin == null) {
-            Logger.warn("Failed parsing pipeline {}", pipeString);
+            Logger.warn("Failed parsing pipeline {}", gstPipeString);
             return;
         }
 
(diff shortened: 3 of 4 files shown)


markmaker:
Thanks, @ian-arkver,

Frankly, I'm a bit out of my depth here. :sweat_smile: so I'm just probing for an explanation here...

The pipestring seems a bit end-user unfriendly. Is there no other, more "GUIey" way? :thinking: Note, I'm not against such an option per se, for exotic use cases. But there should be a simple GUI letting you select a camera and then a video mode from drop-downs, which then perhaps proposes such a pipestring for you (which the user could still modify, if they know what they are doing).

Like I said, _out of depth_, but a quick _naive_ search seems to suggest that enumeration is possible:

http://gstreamer-devel.230.s1.nabble.com/Enumering-capture-devices-td1475489.html

Is it too complicated in Java? I'm just asking, not blocking. 😇

Also, could you please name it `PipeString` everywhere (you already have some), instead of `Pipeline`, to reduce confusion with the OpenPnP `CVPipeline`s?

_Mark

ian-arkver:
Hi Mark,

There are a couple of GUIs for building gst pipelines, but to be honest I've never used any of them and I'd be surprised if they're very useful. Eg.
https://blogs.igalia.com/scerveau/introducing-gstpipelinestudio-0-3-4/
https://github.com/virinext/pipeviz

The existing OpenCV camera and to an extent OpenPnPCaptureCamera cover all the usual cases where a regular v4l2src element (or V4L2 directly) is used for capture. Both of these already have extensive user "knobs and levers" and these should cover all the standard camera types.

GStreamerCamera is aimed specifically at the "odd corner cases" where the capture element is something other than v4l2src, eg. raspicamsrc or nvarguscamerasrc or some proprietary SDK plugin for a GigE camera and regular V4L2 capture won't work. In those cases the controls are almost certainly non-standard, so it makes sense to construct the pipeline outside OpenPnP first and to do all the camera setup externally. Once capture is working the pipeline is simply copy/pasted in.

I can certainly rename the internal variables and make them consistent. User visible text should use the work "pipeline" to match GStreamer terminology.

Thanks for looking at it!
Ian

ian-arkver:
ps. Yes you could conceivably query an element's properties, names, min/max values etc. and perhaps even dynamically create controls for them, but it's much easier just to spec those values in the pipeline string (imho). Any settable control value can be set in the string.

Also, it wouldn't work for many types of source, for example GigE cameras which have a custom, proprietary method of setting "settings" and storing those directly on the camera in some sort of default profile.

I just don't think it's worth the effort for the few cases where GStreamerCamera would be useful. I realise I'm not selling it well here. ;-)

All the best,
Ian

ian-arkver:
One last thought for today...

The advantage of GStreamerCamera is that it's completely arbitrary what goes in the GStreamer pipeline. It can be *anything* at all, assuming it produces video. It could be a UDP receiver of some remote camera (that's how I use it). It could even be something which pulls stills from a directory of jpegs, or plays a YouTube video. No idea why you'd want to do that (some kind of weird testing?) but it could be done. So the thing is, the Java has *no idea* what's in the pipeline and cannot make any assumptions about it.

markmaker:
> The existing OpenCV camera and to an extent OpenPnPCaptureCamera cover all the usual cases where a regular v4l2src element (or V4L2 directly) is used for capture.

Is that so? I thought GStreamer would (finally!) cover them in a much broader way. AFAIK, OpenCVCamera is considered outdated, even deprecated. OpenPnPCaptureCamera only covers USB cameras (it seems?). 

What about for instance the native Raspi camera?

https://davidmays.io/streaming-raspberry-pi-video-via-rtsp-with-gstreamer-and-video4linux/

Note, I'm just fishing for a slightly more user friendly solution, if one is feasible with a few extra lines of code... :sweat_smile:  

_Mark

ian-arkver:
David's article seems to show the regular v4l2src used to directly produce video/x-raw. He's using Raspbian, but the article is a few years old now. I'm not very familiar with the RPi camera plugins on the various distros, but I believe they were changed recently to be more friendly. Perhaps you don't need a special RPi source any more.

For anything that needs v4l2src only (i.e. plain V4L2 capture), I think I'd be more in favour of fixing OpenPnPCapture to work with that source. If that means adding raw video support as opposed to MJPEG to that, I think that would be more beneficial to many more people than this GStreamer hack. I thought OpenPnPCapture was supposed to handle raw (YUV) streams though already? What's the blocker on using that on an RPi?

I don't have a recent RPi, so can't really test any of this.

Adding some Java to try to parse the pipeline, find the source, evaluate the controls, create dynamic UI elements, read/write machine.xml, etc. etc. is *not* going to be "a few extra lines", sadly.

markmaker:
> For anything that needs v4l2src only (i.e. plain V4L2 capture), I think I'd be more in favour of fixing OpenPnPCapture to work with that source.

Like I said, I'm actually out of my depth here. I think OpenPnPCapture [already uses v4l2](https://github.com/openpnp/openpnp-capture/blob/master/linux/platformcontext.cpp). 

Is this ready for merge?

_Mark

vonnieda:
IMHO go ahead and merge this. If we improve the UI later, great. If not, users of Gstreamer will already be familiar with the concept. If this gets users to be able to use a wider array of cameras, especially industrial ones, that's a big win.

Thanks Ian!

ian-arkver:
Thanks Jason.

Mark, let me review the variable naming and maybe make it more consistent. I'll look at that this evening and push another commit if necessary.

Yes, OpenPnPCapture uses V4L2 natively, which is why I think it's the better place to support normal V4L2 capture cameras. Perhaps we need to see whether it needs improvement for non-UVC cameras.

ian-arkver:
Did some renaming. Also renamed the attrib so it appears in machine.xml as "gst-pipeline" to avoid confusion with the vision pipeline child elements further down.

Maven tests passed (not that we have a test for this camera). Actual test on the machine passed (visual homing, config wizard shows).

Should be good to merge now I think.

markmaker:
Thank you!

ian-arkver:
Thanks Mark!

Hommey99:
Hi Everyone. I want Example Pipeline please thank you

ian-arkver:
Pipeline depends totally on what your camera source produces. I don't have any handy which use an actual video capture source, since I stream my cameras remotely. As it says above, experiment with gst-launch-1.0 first, then use the front part of the pipeline. If you're not familiar with gst-launch syntax, then GStreamerCamera probably isn't your best choice.

Here's an example which receives and uses H265 streamed over UDP...
`udpsrc port=9092 caps="application/x-rtp,media=(string)video,clock-rate=(int)90000,encoding-name=(string)H265" ! rtpjitterbuffer latency=0 ! rtph265depay ! h265parse ! avdec_h265 ! videoconvert ! capsfilter caps=video/x-raw`

PR #1667, opened 2024-09-02
bug fix: Search angle in minAreaRect

tobydickenson:
# Description

This is a bug fix

Openpnp has two implementations inside the minAreaRect vision stage. The first implementation hands over to opencv. The second implementation, in java, supports extra parameters leftEdge, rightEdge, topEdge, bottomEdge, and searchAngle.

The switch between the two implementations only checks the first four parameters. If a non-default `searchAngle` is specified (and the other four parameters have the default true value) then it uses the opencv implementation, and ignores the users specified searchAngle.

This fix checks for a non-default searchAngle as a reason to use the full-featured implementation.

# Justification

Is it useful for a vision pipeline to specify a non-default searchAngle?

I occasionally see misdetection of sot23 components.... sot23 are approximately triangular, and there are three possible orientations for packing a triangle into a rectangle.  I was hoping that specifying searchAngle of 15° would force the bottom vision to pick the bounding box with approximately horizontal orientation. A better bottom camera would help too of course. I have not yet proven whether this patch is a good solution for my sot23 bottom vision, but this vision stage now behaves as expected when given a non-default searchAngle.

# Implementation Details
   a. Manually tested with bottom vision, changing component orientation and reducing searchAngle from the default to prove it finds the expected rectangle.
   b. Leaving the default searchAngle, and proving the default behaviour is unchanged.
3. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? - Yes
5. test pass

### DIFF
--- a/src/main/java/org/openpnp/vision/pipeline/stages/MinAreaRect.java
+++ b/src/main/java/org/openpnp/vision/pipeline/stages/MinAreaRect.java
@@ -241,7 +241,7 @@ public class MinAreaRect extends CvStage {
             return null;
         }
         RotatedRect r;
-        if (leftEdge && rightEdge && bottomEdge && topEdge) {
+        if (leftEdge && rightEdge && bottomEdge && topEdge && searchAngle==45) {
             // All edges, use OpenCv method.
             MatOfPoint2f pointsMat = new MatOfPoint2f(points.toArray(new Point[points.size()]));
             r = Imgproc.minAreaRect(pointsMat);
--- a/src/main/java/org/openpnp/vision/pipeline/stages/MinAreaRect.java
+++ b/src/main/java/org/openpnp/vision/pipeline/stages/MinAreaRect.java
@@ -241,7 +241,7 @@ public class MinAreaRect extends CvStage {
             return null;
         }
         RotatedRect r;
-        if (leftEdge && rightEdge && bottomEdge && topEdge && Math.abs(searchAngle-45)<angularResolution) {
+        if (leftEdge && rightEdge && bottomEdge && topEdge && searchAngle>=45-angularResolution) {
             // All edges, use OpenCv method.
             // This is slightly faster; 30µs per call, compared to 210µs for the java method in the other branch
             MatOfPoint2f pointsMat = new MatOfPoint2f(points.toArray(new Point[points.size()]));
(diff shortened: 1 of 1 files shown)


markmaker:
Hi @tobydickenson,

thank you, good catch! :+1: 

The only thing I'm a bit unsure about is the exact equality condition `searchAngle==45` on `double` values. The value will be converted into text and back (e.g. serialized in `machine.xml`). I guess this usually works out okay, but to be on the safe side, maybe use something like this:

```java
            if (leftEdge && rightEdge && bottomEdge && topEdge && searchAngle >= 44.999) {
            // All edges, use OpenCv method.

```        

This also covers the case where a user enters > 45 degrees, which should still go to the OpenCV method.

See also:
https://stackoverflow.com/a/8081911/15218951
    
_Mark

tobydickenson:
I have pushed something similar to what you suggest, using the existing constant `angularResolution`.

(Equality comparison can be problematic because floating numbers are inexact, but it would have been ok in this case I think because small integers can be represented exactly. This is why a typical `vision-settings.xml` contains `search-angle="45.0"` exactly)

I considered suggesting a different solution of *always* using the new method, and dropping the branch that uses opencv. But opencv is sufficiently faster that I think it is worth keeping the twin implementations. Using System.nanoTime I measure:
opencv method: 30µs per call
java method: 210µs per call

tobydickenson:
> I have pushed something similar to what you suggest, using the existing constant `angularResolution`.

Ah, no. I missed your comment about the >45 case. New patch to follow.....

markmaker:
> opencv method: 30µs per call
> java method: 210µs per call

The difference is heavily dependent on image resolution, if I remember correctly. Better to keep both variants. :grinning:

markmaker:
Thanks, @tobydickenson! :100:

markmaker:
Just FYI:

>  but it would have been ok in this case I think because small integers can be represented exactly

Not true. :wink: 

https://en.wikipedia.org/wiki/Floating-point_arithmetic#Accuracy_problems

Usually, if (and only if) the original number is a decimal number, converting to text with _less_ digits than are theoretically available in decimal representation (i.e. rounding to that last decimal digit), and then converting back from that text, will result in the same binary `double` representation. 

This is true even if one was converted from the Java source code (here) into binary literal in a .class file on a completely different computer/architecture than the one later converting into XML and back on the user's computer. IEEE 754 is pretty nailed down bit-by-bit, and IEEE 754 is likely even guaranteed via the Java spec(?).

_Mark

tobydickenson:
Again, just fyi, try putting some small integers (for example, 45) into this tool: 
https://www.h-schmidt.net/FloatConverter/IEEE754.html

markmaker:
Alright, I stand corrected. Seems that inaccuracy only enters with decimal _fractions_. Good to know. :smile:

PR #1668, opened 2024-09-03
Avoid accidental leading or trailing whitespace in Part IDs etc

tobydickenson:
# Description
This patch removes leading and trailing whitespace in part IDs etc which are created using the gui. It is easy to get extra whitespace when text is copy/paste from a source such as a datasheet, and undesirable to create an ID with leading and trailing whitespace.

# Implementation Details
This patch affects all uses of `JOptionPane.showInputDialog` where the input is used as an ID.
All existing uses already check for null input, which occurs when the cancel button is pressed.
This change uses `String.trim` to remove leading and trailing whitespace, after that null check.
Additionally, if that results in an empty string, it behaves the same as if `showInputDialog` returned null.
1. How did you test the change? - Tested manually, with entering empty strings, strings with some unwanted whitespace, and strings with no whitespace
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. - Tests pass

Any pre-existing IDs are left unchanged. This only affects new IDs as they are created.

This change assumes that noone wants to intentionally create IDs with trailing and leading whitespace, or zero length IDs.

### DIFF
--- a/src/main/java/org/openpnp/gui/BoardPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/BoardPlacementsPanel.java
@@ -484,7 +484,11 @@ public class BoardPlacementsPanel extends JPanel {
             if (id == null) {
                 return;
             }
-            
+            id = id.trim();
+            if (id.isEmpty()) {
+                return;
+            }
+
             // Check if the new placement ID is unique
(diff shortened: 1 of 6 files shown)


markmaker:
Thank you @tobydickenson ! :100:

PR #1671, opened 2024-09-04
Various size check and vision compositing fixes

tobydickenson:
**This is a provisional pull request** shared for review. So far this has only been tested in simulation and I want to test it on the machine.

# Description

This patch contains some bug fixes for vision compositing and part size checking, some defensive changes to prevent such bugs hiding in future, and some unit test extensions.

## Size check preconditions

The part size check logic was not previously validating the measured sizes. While there was not much scope for this to go wrong when its data was provided by minAreaRect,  some other bugs described below in composite vision were causing the measured size to be NaN. NaN size values lead to contradictory error messages from the part size check logic. This patch adds some validation.

## Vision compositing divide by zero

The final stages of vision compositing has various floating point divisions, without checking for a divide by zero. This would result in NaN values in the output.

The divisions relating to the calculation of angle and center point are pretty robust. I have not seen these go wrong, but I have added some defensive checks to raise an exception if the denominator is zero. That is better than allowing a NaN in the return value.

There are also divisions in calculating the scaling for part length and width. Previously the trapezoid vision solution would never collect this scaling information, so always triggered this divide by zero and NaN return value. While this case is fixed, I suspect there may be other valid vision solutions that do not collect scaling information. This patch prevents the divide by zero by assuming unity scale if the denominator is zero.

## Trapezoid vision solution scale

The trapezoid vision solution was previously processing center point, angle, and "width" scale, but not "length" scale. This was therefore triggering the divide by zero and NaN described above. This was the root cause of the bug I have been chasing here.

Width scale was measured in two ways: using the gap between the two corners on the base of the trapezoid, and the gap between the two corners at the top.

The fix for measuring trapezoid "length" scale involves:
* Find the midpoint between the two corners on the base of the trapezoid
* Find the midpoint between the two corners on the top of the trapezoid
* The distance between those two midpoints gives information about the length scale.

## Square vision solution scale

The square vision solution had a size bug. The measured distance is a hypotenuse, so the measured size was previously out by a factor of sqrt(2)

## Unit Test

A unit test for all of the above is achieved by enabling size checking when running the vision compositing unit test. This tests that the composite vision measures sizes correctly, and does not trigger any of those new defensive exceptions.

## Orientation

As discussed at https://groups.google.com/g/openpnp/c/Pd3wYWYiTTU
The part size check compares the longest side vs the longest measurement, and the shortest side vs the shortest measurement. This behaviour is described on https://github.com/openpnp/openpnp/pull/1135 with the rationale "to avoid problems with dimensions being swapped in the package".

The conclusion was to compare both dimensions in their orientation as is

# Justification
* This fixes some cases where vision compositing measured an incorrect size for a part
* This fixes some cases where vision compositing returned a NaN, leading to strange behaviour in other code
* This adds some defensive checks, to avoid strange behaviour if a similar problem arises again

# Implementation Details
1. How did you test the change? - Only in simulation so far (manual tests, and some extensions to the existing unit tests). I still need to test this on a machine with the parts that were previously causing my problems.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -346,11 +346,6 @@ public class ReferenceBottomVision extends AbstractPartAlignment {
             return true;
         }
 
-        // Make sure width is the longest dimension
-        if (partSize.getY() > partSize.getX()) {
-            partSize = new Location(partSize.getUnits(), partSize.getY(), partSize.getX(), 0, 0);
-        }
-
         double pxWidth = VisionUtils.toPixels(partSize.getLengthX(), camera);
         double pxLength = VisionUtils.toPixels(partSize.getLengthY(), camera);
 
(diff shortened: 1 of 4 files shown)


tobydickenson:
The composite vision system is very clever. Thanks for your work on this.

I have just added another commit to this pull request. If there are multiple vision passes, it only needs to check the size after the final pass. This means doing offsets checks before size checks, which I think is ok (or indeed better).

I will do a practical test on the machine tomorrow or early next week. Lets merge it after that.

tobydickenson:
My test on the machine was very successful. Lets merge!

markmaker:
Thanks, @tobydickenson! :1st_place_medal:

Issue #1674, opened 2024-09-11
Reloading the config via a script causes the cameraView to display all camera's twice


mcix:
# Problem Report
## Steps to Reproduce
How can a developer reproduce the issue you are experiencing? Use a step by step format like this:
1. Start OpenPnP
2. Create a (js) script with the contents:
``` config.load(); ```
3. Run the script

## Expected Result
See the same 2 camera's (the config is not changed)

## Actual Result
All camera's are added again to the cameraView

## Notes, Log Files, Screen Captures, Videos, etc. to Show the Issue
<img width="860" alt="Screenshot 2024-09-11 at 13 12 47" src="https://github.com/user-attachments/assets/5de25434-6253-48de-8d7f-4124aa975749">

PR #1675, opened 2024-09-11
Clear the cameraViews to prevent duplicate camera view's

mcix:
# Description
When reloading the config, the camera's appear multiple times in the UI

# Justification
Fixes https://github.com/openpnp/openpnp/issues/1674

# Implementation Details
To reproduce:
1. Start OpenPnP
2. Create a (js) script with the contents:
``` config.load(); ```
3. Run the script

### DIFF
--- a/openpnp.bat
+++ b/openpnp.bat
@@ -2,7 +2,12 @@
 
 set archi=%PROCESSOR_ARCHITECTURE%
 
-if not x%archi:86=%==x%archi% java --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.desktop/java.awt=ALL-UNNAMED --add-opens=java.desktop/java.awt.color=ALL-UNNAMED -jar target\openpnp-gui-0.0.1-alpha-SNAPSHOT.jar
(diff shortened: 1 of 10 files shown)


markmaker:
>  would it be useful to have some sort of restart? 

I can only think of one situation, where I wanted to reload the configuration: After I made a mistake, and misconfigured something (or suspected it), and do not want to save the configuration like that, but rather revert to the former good configuration.

When that happens to me, I currently kill the OpenPnP process using the OS tool (Task Manager, etc.) :smirk: 

So it might be a good idea to add a new `File` menu entry `Exit without saving` or similar. maybe with a confirmation Box to be sure. After doing that, a user can restart OpenPnP and gets everything reverted. I would happily accept such a PR :smile:.

Does this cover your use case? Or what is your actual intention?

Btw. do you know the automatic config backups? A user can always use these:

https://github.com/openpnp/openpnp/wiki/FAQ#how-can-i-go-back-to-an-oldworking-configuration

_Mark

markmaker:
Hi  @mcix,

I see you're still working on that. But you are working directly on your `test` branch. Instead, you should make a git branch, like `feature/restart-app`, push it to your fork, and then create a new PR.

I'll close this one, as it no longer reflects what you are doing. :smile: 

Thanks,
_Mark

mcix:
@markmaker Thanks for your comment and giving some hints on how to do this correctly! Much appreciated, will look into it.

PR #1676, opened 2024-09-15
Fix / Background Calibration: Greyscale vs. HSV value

markmaker:
# Description

In [Nozzle Tip Background Calibration](https://github.com/openpnp/openpnp/wiki/Nozzle-Tip-Background-Calibration), the **Brightness** method worked with a grayscale image, but then used the **HSV Value** channel for masking. The **Value** channel is not the same as grayscale brightness, i.e., it boosts colorful elements, fully saturated colors even reach the maximum value (255). Therefore, in case of background elements being rather colorful,  the pipeline failed to mask them properly.

## Bugfix

This PR fixes this bug by also using the **HSV Value** channel for calibration, so the two always match.

## HSV Value vs. True Brightness Masking

Note: arguably, the **HSV Value** channel is not ideal for masking the background in the **Brightness** mode, i.e. it does not truly indicate "brightness". Vivid colors will show as too bright and trigger a warning. 

Too bright

However, it was concluded that this warning is actually beneficial. Such backgrounds indicate that there is a problem, either the user mistakenly chose the **Brightness** over the **BrightnessAndKeyColor** mode when they used a green Juki nozzle (or similar), or they should fix another underlying problem of having a colorful background in the first place, like missing  [Camera White Balance](https://github.com/openpnp/openpnp/wiki/Camera-White-Balance) or similar. 

Furthermore, an overall change to the possibly more suitable HLS model (where the Luminosity channel truly indicates brightness) was considered too disruptive with regard to existing code, and compatibility with stored pipelines and configurations in the field.

## True Brightness Slider Threshold 

On the other hand, the manual threshold slider now controls the true brightness in the new stock pipeline, and no longer the HSV Value. By combining the HSV model for background elimination, with the true brightness slider threshold, we get the best of both color models. It allows us to combine the advantages of the two models in terms of discriminating background and dark part elements from the bright contacts. Note, if pipelines in the field are left unchanged they will still work as before, and stored threshold slider values will remain correct.

## Testing Grey and Blue-Tinted Nozzle Tips

This PR also added dark and blue-tinted simulation nozzle tips for testing.

Blue tinted nozzle tip

# Justification
User report:
https://groups.google.com/g/openpnp/c/b9vgZJmQZkM/m/0GlZcpCEAwAJ

# Instructions for Use
To enable the new stock pipeline, use the instruction in the Wiki.
https://github.com/openpnp/openpnp/wiki/Computer-Vision#using-new-stock-pipelines

Otherwise the usage remains the same (just a bugfix, and improvement).

# Implementation Details
1. Tested in simulation, using simulated dark grey and blue-tinted nozzle tips.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzleTipCalibration.java
@@ -1020,12 +1020,7 @@ public class ReferenceNozzleTipCalibration extends AbstractModelObject {
             int kernelSize = ((int)getMinimumDetailSize()
                     .divide(camera.getUnitsPerPixel().getLengthX()))|1;
             Imgproc.GaussianBlur(image, image, new Size(kernelSize, kernelSize), 0);
-            if (getBackgroundCalibrationMethod() == BackgroundCalibrationMethod.BrightnessAndKeyColor) {
-                Imgproc.cvtColor(image, image, FluentCv.ColorCode.Bgr2HsvFull.getCode());
-            }
-            else {
-                Imgproc.cvtColor(image, image, FluentCv.ColorCode.Bgr2Gray.getCode());
-            }
+            Imgproc.cvtColor(image, image, FluentCv.ColorCode.Bgr2HsvFull.getCode());
             backgroundImageRows = image.rows();
             backgroundImageCols = image.cols();
             backgroundImageChannels = image.channels();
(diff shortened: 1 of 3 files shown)


tobydickenson:
Hi Mark,

Sorry for the delayed feedback. I tested this today, and everything looks good.

I tested with:
* Color&brightness calibration mode with my blue tinted nozzles.
* Brightness mode with my blue tinted nozzles.
* Brightness mode with a whitebalance correction that makes the nozzles look grey. This whitebalance correction reduces the dynamic range which I need for distinguishing some bright objects, so I cant use it all the time.

With this change I can now use all-stock vision pipelines. I also tested the backwards compatibility case, using these code changes with my previous custom pipelines.

All good. Thank you.

PR #1681, opened 2024-10-04
an extra script event

tobydickenson:
# Description
This adds an extra scripting event related to homing. Openpn already has `Machine.AfterHoming` which gets called when homing is entirely complete. This adds `Machine.AfterDriverHoming`, which gets called between the two phases of the homing process: after homing all the drivers, and before calibration using the homing fiducial.

# Justification
I hope this new scripting event is generally useful, even if my specific use for it is niche.

I have described my requirement on google groups. My gcode homing has not been 100% reliable. I'm working on a fix for that, but until then I would like my openpnp to be able to validate the gcode homing before running its calibration cycle. I can implement that in a script! This PR means I can run the script at the right time.

# Instructions for Use

It is used like any other scripting event.

I will add a corresponding description of this event to https://github.com/openpnp/openpnp/wiki/Scripting

# Implementation Details
1. How did you test the change? Be descriptive -- tested on a lumenpnp, and in the simulated machine.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferenceMachine.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceMachine.java
@@ -532,8 +532,8 @@ public class ReferenceMachine extends AbstractMachine {
         }
 
         getMotionPlanner().home();
+        Configuration.get().getScripting().on("Machine.AfterDriverHoming", null);
         super.home();
-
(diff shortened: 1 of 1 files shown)


markmaker:
Thank you @tobydickenson!

tobydickenson:
scripting wiki has been updated

PR #1682, opened 2024-10-15
Issues solutions guidance

tobydickenson:
# Description
I have recently been redoing my machine setup and calibration using Issues&Solutions, where previously I had followed the opulo-recommended manual method. The end result is very much better with I&S.

However the I&S process was not exactly smooth. I had a few false starts which I was able to correct myself, and a few major misunderstandings which where corrected by the helpful participants on the mailing list. I have made a few changes to the wiki to help the next user following my footsteps; for example noting that there is a maximum diameter for the nozzle-offset-calibration confetti. This patch contains a couple of messages which I think are best delivered through the software.

## 1. Unconventional Z Axis

Mark pointed me to [the convention that working Z is negative](https://github.com/openpnp/openpnp/wiki/Machine-Axes#a-word-about-z-coordinates), which was very helpful. After calibrating Z, this patch adds a warning message with a link to the wiki if a calibration fiducial is at positive Z.

## 2. Tool changer

Changing axis calibration is likely to break any automatic tip changer configuration. This patch adds a warning message if automatic tip changers are enabled before performing any axis calibration.

# Justification

## a. No regressions

* This patch does not change the I&S calibration logic. There is no numerical or process change.

* If a user reads the wiki and follows its guidance meticulously, then they will not see any new messages. This patch adds warnings which are only shown for users who are diverging from the best practice.

## b. Helpfulness

These warning messages would have saved me from a few days of confusion, and having to reprint my tool changer. These are not mistakes that everyone will make, but I feel they are mistakes that any new user could make.

# Instructions for Use

No instructions needed - its just a warning message.

# Implementation Details
    * With positive and negative Z coordinates on both the real machine and simulated machine
    * With automatic toolchangers enabled on 0, 1, then 2 nozzles on the real machine.
3. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? - yes
5. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. - tests pass

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ContactProbeNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ContactProbeNozzle.java
@@ -928,7 +928,7 @@ public class ContactProbeNozzle extends ReferenceNozzle {
                         }
                         if (suggestedCommand != null) {
                             GcodeDriverSolutions.suggestGcodeCommand(gcodeDriver, contactProbeActuator, solutions, 
-                                    GcodeDriver.CommandType.ACTUATE_BOOLEAN_COMMAND, suggestedCommand, false, false);
+                                    GcodeDriver.CommandType.ACTUATE_BOOLEAN_COMMAND, suggestedCommand, false, false, null);
                         }
                         else {
                             String currentCommand = gcodeDriver.getCommand(contactProbeActuator, GcodeDriver.CommandType.ACTUATE_BOOLEAN_COMMAND);
--- a/src/main/java/org/openpnp/machine/reference/solutions/GcodeDriverSolutions.java
+++ b/src/main/java/org/openpnp/machine/reference/solutions/GcodeDriverSolutions.java
@@ -1100,7 +1100,7 @@ public class GcodeDriverSolutions implements Solutions.Subject {
                 public String getExtendedDescription() {
                     String r = "<html>\n";
                     if (rationale!=null && !rationale.isEmpty()) {
-                        r += "<p><b>"+rationale+"</b></p>\n";
+                        r += "<p><strong>"+rationale+"</strong></p>\n";
                     }
                     if (suggestedCommand.isEmpty()) {
                         r += "<p>Delete it.</p>\n";
--- a/src/main/java/org/openpnp/machine/reference/solutions/GcodeDriverSolutions.java
+++ b/src/main/java/org/openpnp/machine/reference/solutions/GcodeDriverSolutions.java
@@ -772,6 +772,7 @@ public class GcodeDriverSolutions implements Solutions.Subject {
         for (CommandType commandType : gcodeDriver.isSpeakingGcode() ?
                 CommandType.values() 
                 : new CommandType[] { CommandType.CONNECT_COMMAND }) {
+            String rationale = "";
             String command = gcodeDriver.getCommand(null, commandType);
             String commandBuilt = null;
             boolean disallowHeadMountables = false;
@@ -797,24 +798,28 @@ public class GcodeDriverSolutions implements Solutions.Subject {
                     else {
                         if (gcodeDriver.getUnits() == LengthUnit.Millimeters) {
                             if (command.contains("G20 ")) {
+                                rationale += "Replace G20 (inches) with G21 (millimeters). ";
                                 commandBuilt = command
                                         .replace("G20 ", "G21 ")
                                         .replace("inches", "millimeters");
                             }
                             else if (! command.contains("G21 "))
-                            { commandBuilt = "G21 ; Set millimeters mode \n" 
-                                    + command;
+                            {
+                                rationale += "Explicitly set millimeters mode. ";
+                                commandBuilt = "G21 ; Set millimeters mode \n" + command;
                             }
                         }
                         else if (gcodeDriver.getUnits() == LengthUnit.Inches) {
                             if (command.contains("G21 ")) {
+                                rationale += "Replace G21 (millimeters) with G20 (inches). ";
                                 commandBuilt = command
                                         .replace("G21 ", "G20 ")
                                         .replace("millimeters", "inches");
                             }
                             else if (! command.contains("G20 "))
-                            { commandBuilt = "G20 ; Set inches mode \n"
-                                    + command;
+                            {
+                                rationale += "Explicitly set inches mode. ";
+                                commandBuilt = "G20 ; Set inches mode \n" + command;
                             }
                         }
                         else {
@@ -1027,7 +1032,7 @@ public class GcodeDriverSolutions implements Solutions.Subject {
                     break;
             }
             suggestGcodeCommand(gcodeDriver, null, solutions, commandType, commandBuilt, commandModified,
-                    disallowHeadMountables);
+                    disallowHeadMountables, rationale);
         }
     }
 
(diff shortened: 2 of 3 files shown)


markmaker:
Hi @tobydickenson,

I very much appreciate the effort to make this easier! :medal_sports:

But instead of a text warning that (by default) will be hidden when the Solution is solved, it should be a separate Solution that pops up on any machine with this disposition. Always, even simply after an update of the OpenPnP version. Inspecting your config permanently, vigilantly, that's how I&S is supposed to work, yes, let's alert all Lumen users!

Also, as a general rule, each Issue or Solutions should address a single, [orthogonal](https://en.wikipedia.org/wiki/Orthogonality_(programming)) _thing_, to make things easy to understand.

Plus I would rather see this immediately after the Safe Z Zone solutions, in the Kinematic Milestone. Comes much earlier, i.e., at the earliest moment when this becomes known. The user should not start off with Z all wrong.

Capture Safe Z

https://github.com/openpnp/openpnp/wiki/Kinematic-Solutions#capture-safe-z

Your Solution could note that Safe Z is (well) in the positive. I guess a slightly positive value could be legal, so it would warn from Safe Z > 2mm or so.

See here for an issue that checks a similar thing at the right moment, your new one could be modeled after this just with different conditions and text, and it should come immediately after this one as an `else` of the second part of the current `if()` condition (which must be split into two nested ones): 

https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/machine/reference/solutions/KinematicSolutions.java#L190-L205

### Bonus :100: 

Furthermore, you could add _one more issue_ that detects when nozzle 1 and nozzle 2 are completely asymmetric in Z, i.e. the Safe Z Zone is not symmetric around zero, which is unexpected for a shared Z axis topology (both cam or negated). This happens, when the homing switch is not at the balance point, and when this is not compensated in the homing code (the homing offset we talked about), e.g. the typical Lumen annoyance. Again this should allow some slight asymmetry but alter when out of whack > 2mm or so.

This one is harder to code, you would have to detect the shared axis. There is code doing this here, you could refactor it into a helper and reuse:

https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/machine/reference/solutions/KinematicSolutions.java#L273-L285

Ideally this solution would propose the right homing offset in the solution text and instruct the user how to apply it on the controller config or homing code (Wiki link?). Obviously the homing offset can be calculated from the midpoint of the Safe Z Zone. 

Be aware of `axis.getDriver().getUnits()` to convert and specify the offset in the right G-code units. Use `Length.convertToUnits()`.

That would really help those user I guess. :sunglasses: :+1: 

_Mark

markmaker:
Hi @tobydickenson 

I wroiting this as a separate comment in order not to mix things up:

You wrote:

>  for example noting that there is a maximum diameter for the nozzle-offset-calibration confetti

I don't see

markmaker:
Sorry, pressed the wrong button! :weary:

tobydickenson:
All agreed. More code to follow.

tobydickenson:
Hi @markmaker 

> See here for an issue that checks a similar thing at the right moment, your new one could be modeled after this just with different conditions and text, and it should come immediately after this one as an `else` of the second part of the current `if()` condition (which must be split into two nested ones):

This is now done, in 55e9f5c8441e0ed0b76538c5d1055f2be5369fd4

> Furthermore, you could add _one more issue_ that detects when nozzle 1 and nozzle 2 are completely asymmetric in Z ....
(the homing offset we talked about), e.g. the typical Lumen annoyance.

The lumen configuration avoids asymmetry using nozzle offsets and a ReferenceMappedAxis. If I understand your suggestion correctly, it would not be caught by this test.

The only possible asymmetry that wont be caught by the first check above is where Z=0 retracts both nozzles, but with an imbalance. Thats not ideal, but not really a problem. I&S secondary nozzle offset calibration will take care of it.

---

There is another new change to improve I&S Gcode solutions usability in 46ea4986163c349de33e0c86af344dd9316cba6d
* It shows the new gcode and current gcode in the solution, to allow the user to review //differences// and assess the impact of the proposed gcode change.
* For some gcode amendments there was an explanation in source code comments. It now puts that same information in the Solution window.

Subject to review, I believe this is all ready to merge.

markmaker:
> The lumen configuration avoids asymmetry using nozzle offsets and a ReferenceMappedAxis. If I understand your suggestion correctly, it would not be caught by this test.

Not true. The Safe Z Zone is ultimately stored on the raw axis, and _it is_ the raw axis that is analyzed, not the mapped nozzle axis. See the use of `HeadSolutions.getRawAxis()`.

This means you will see the asymmetry, and (what is more important) from the midpoint of the axis Safe Zone you can propose the homing offset we talked about earlier. As it is the raw axis and raw coordinates (i.e. the controller's native axis and native controller coordinates).

For Lumen users this would be a much more constructive solution, as it actually not just tells them "Your Z is positive, so it is not recmommended" but also **what they can do about it**. 

Furthermore, the homing offset (or any other means like G92) actually fixes the Z issue at the controller level, where it belongs, IMHO. The user will then see the native Z in both the controller and OpenPnP, and they will not have asymmetric and huge Z head offsets.

___

In your code, I see the issue telling the user to disable the automatic tool changer. What is the rationale behind that?

_Mark

tobydickenson:
OK, I see where you are coming from.

My goal with this PR (and wiki updates) was to intercept the points where my work went in the wrong direction due to a misunderstanding. "Z=0 is retracted" is a useful convention - one that I had not appreciated despite reading the Axes wiki page. My purpose behind the I&S change was to alert a user who was similarly operating under a different Z axis convention. This is a general purpose feature - any novice user on any new machine might have the same misunderstanding, and would benefit from an alert that their setup is unconventional.

Your suggestion is somewhat more ambitious in scope. You are describing a feature that will be //instructional// for lumen users with the stock opulo configuration. But I am unconvinced that the step-by-step I&S model works for this. There are lots of details that would all need to be changed on the machine (connect gcode commands, ReferenceMappedAxis, soft limits, etc) and operating procedures that the user would need to relearn. **what they can do about it** is a big topic which will be beyond my time available for this PR.

----

The advice to disable the automatic tool changer came about because I broke my tool changer!

I made several novice mistakes:
1. Something I did using I&S introduced an incorrect offset to second nozzle Z coordinates. I am not sure how that happened, but PR #1680 maybe?
2. I think I was about to do the "confetti" process, for which I always use the same nozzle tip. When loading that tip, the nozzle collided with the tool changer.

With hindsight, it seems unwise to have the automatic tool changer enabled while adjusting the coordinate system.

markmaker:
> Your suggestion is somewhat more ambitious in scope .... which will be beyond my time available for this PR.

That's okay. Just wanted to set the _technical_ record straight. 

Later, or maybe you can instead just elaborate a bit more in the linked Wiki.

> The advice to disable the automatic tool changer came about because I broke my tool changer! ...  I made several novice mistakes ...   Something I did using I&S introduced an incorrect offset to second nozzle Z coordinates

That' exactly why I wanted the Z issue handled as early as possible in the setup (Kinematic Milestone). With the homing offset handled early and right, _it would almost certainly never happen to any future user!_ 

For those users with an already crippled config, yes, it would make sense to warn about the nozzle tip changer. But again, this must happen in the Kinematic milestone. Better yet, the solution should disable the auto changer for them, when they accept the "asymmetric Safe Z Zone detected" solution we discussed.

So, I still think the comment about disabling the automatic nozzle tip changer is misplaced / not helpful there, thinking of the majority of non-Lumen users. It is solely the moronic Z-homing of Lumen that makes this happen, the asymmetry is the culprit. If not handled earlier, _where it should have been_, it rears its ugly head again in the Z head offsets, _where it shouldn't_. Plus it leaves the user stranded with a disabled automatic nozzle tip changer and no instructions what to do about the potential Z offsets, nor a reminder to re-enable it.

Please remove the extended warning for now. :pray: 

_Mark

tobydickenson:
> Please remove the extended warning for now. 🙏

That is done.

This branch has some back-and-forth development now. Should I rebase to get a clean diff?

markmaker:
Yes please. :cookie:

markmaker:
Oh, there is a trailing <br/> in the VisionSolutions, and left-over import, please restore the original.

tobydickenson:
This branch diff is now clean and tests have been re-run. Ready to merge!

Thanks for your review.

markmaker:
Thanks, @tobydickenson,

Two cosmetic things:

- `if (axisZ.isSafeZoneLowEnabled() && axisZ.isSafeZoneHighEnabled()` is used twice, please wrap as a separate `if()` instead.

- You're using `<b>` for bold text, please use `<strong>` instead. I'm sure you're aware of the idea of "semantics over style" :wink:.

Thanks!

_Mark

tobydickenson:
>     * You're using `<b>` for bold text, please use `\<strong\>` instead. I'm sure you're aware of the idea of "semantics over style" 😉.

Done.

>     * `if (axisZ.isSafeZoneLowEnabled() && axisZ.isSafeZoneHighEnabled()` is used twice, please wrap as a separate `if()` instead.

There is a new local variable for `axisZ.isSafeZoneLowEnabled() && axisZ.isSafeZoneHighEnabled()`

On review I think there was a bigger problem here. The new `if(){}` which generated the new solution would have blocked the "Jog  x over the tallest obstacle on your machine....` solution in the final else. So the if statement nesting has changed too. I'm not 100% confident on this change, so this is a tentative commit for review (again :-).

Toby

markmaker:
You are right. The first issue is a 100% wrong configuration, that you need to undo. However the second could be dismissed by the user, and the the outer `else` should work.

I'll merge it like that, okay?

_Mark

tobydickenson:
> I'll merge it like that, okay?

ok!

markmaker:
Thanks, @tobydickenson!

PR #1685, opened 2024-10-24
a new scripting event for takeBackPart (recycle)

tobydickenson:
# Description
On google groups there has been discussion about a feed count for component stock control purposes. The thread [starts here](https://groups.google.com/g/openpnp/c/W5kmDu2a0a0/m/zs9OIpnxAQAJ?utm_medium=email&utm_source=footer)

Openpnp is pretty close to being able to support a feature like this using scripts. But the "take back" (recycle) feature is a hole which any stock control system might want to monitor. 

This patch adds two script events Nozzle.BeforeTakeBack and Nozzle.AfterTakeBack. There are parallel to Nozzle.BeforePick and Nozzle.AfterPick, and have the same parameters.

# Instructions for Use
I will add documentation for Nozzle.BeforeTakeBack to the wiki when merged. I note there is currently no documentation for [Nozzle.BeforePick](https://github.com/openpnp/openpnp/wiki/Scripting#NozzleBeforePick) - I will add that now

# Implementation Details
1. How did you test the change? - Tested on simulated machine and real machine, with a script which logs component movement events
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. - no changes

### DIFF
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.prefs.Preferences;
 import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collectors;
+import java.util.Map;
 
 import javax.swing.AbstractAction;
 import javax.swing.Action;
@@ -628,9 +629,16 @@ public class FeedersPanel extends JPanel implements WizardContainer {
             nozzle.calibrate();
         }
 
+        Map<String, Object> globals = new HashMap<>();
+        globals.put("nozzle", nozzle);
+        globals.put("feeder", feeder);
+        globals.put("part", feeder.getPart());
+
         // Perform the feed.
         nozzle.moveToSafeZ();
+        Configuration.get().getScripting().on("Feeder.BeforeFeed", globals);
         feeder.feed(nozzle);
+        Configuration.get().getScripting().on("Feeder.AfterFeed", globals);
         return nozzle;
     }
 
--- a/src/main/java/org/openpnp/gui/JogControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/JogControlsPanel.java
@@ -639,11 +639,19 @@ public class JogControlsPanel extends JPanel {
                 if (part == null) {
                     throw new Exception("No Part on the current nozzle!");
                 }
-
+                
                 // go through the feeders
                 for (Feeder feeder : Configuration.get().getMachine().getFeeders()) {
                     if (part.equals(feeder.getPart()) && feeder.isEnabled() && feeder.canTakeBackPart()) {
+                        Map<String, Object> globals = new HashMap<>();
+                        globals.put("nozzle", nozzle);
+                        globals.put("feeder", feeder);
+                        globals.put("part", part);
+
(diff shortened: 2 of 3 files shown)


tobydickenson:
Agreed on the feed vs pick distinction.

But I dont think `Job.Placement.Starting` works. It is triggered for feeding during a job, but not for manual feeding and picking from the Feeders panel.

I considered adding a scripting event for feeding. But initially rejected that since there are multiple callers to `feeder.feed()`. But I now note that Feeders panel has a "feed" button which, in the current implementation, wont trigger any script. Therefore I now conclude that we do need a scripting event for feeding.

So I now propose:
* adding a new method to the feeder base class `.managedFeed()` which triggers scripts at the start and end, and calls `.feed()` in the middle.
* Adding the feeder parameter.
* Change "Nozzle.XxxTakeBack" to "Feeder.XxxTakeBack" as you suggest.
* I also note that "Nozzle.BeforePick" has a part parameter, but "Nozzle.BeforePlace" does not. I suggest adding that for uniformity.

Thank you for the review.




`

markmaker:
Agreed. 

However, a new method `managedFeed()` as a _wrapper-slash-replacement_ feels bad. It is in `org.openpnp.spi` which should not be taken too lightly. Users expect `feed()` to feed, period. 

If it comes to that, I would rather rename all the old `feed()` method implementations to something "internal", and then use an `AbstractFeeder` `feed()` implementation as the wrapper. The new internal method could be declared in  `AbstractFeeder` only. So in the Service Provider Interface everything remains the same. 

_Mark

tobydickenson:
> If it comes to that, I would rather rename all the old `feed()` method implementations to something "internal"

An alternative would be to put these wrapper functions in a new module. I note `utils/VisionUtils.java` which has a wrapper function that generates the "Vision.PartAlignment.Before" script events. Similarly, I could create a new `utils/FeederUtils.java`? That option looks best to me, but I'm happy with either.

tobydickenson:
> Users expect feed() to feed, period

This was a useful architectural insight. Thank you.

> Agreed.

A new implementation is now pushed for review.

I have added code to trigger the Feeder.Before/AfterFeed scripts to **both** ReferencePnpJobProcessor and FeedersPanel (which eliminates the discussion about where the wrapper function should live). Originally I was wary of this duplication, but this is in keeping with other code in FeedersPanel which is documented as being "job-like" and already clones code from the job processor. 
https://github.com/openpnp/openpnp/blob/64883f8abd7a5cbe5575ab14e0d46fba30976fdf/src/main/java/org/openpnp/gui/FeedersPanel.java#L609

tobydickenson:
p.s. #1681 is another PR which adds an extra script event that would be great to have merged too :-)

markmaker:
> Originally I was wary of this duplication, but this is in keeping with other code in FeedersPanel which is documented as being "job-like" and already clones code from the job processor.

Yes, the long term goal for `FeederPanel` would be to create a temporary 1-part Job and then simply execute it, to eliminate all code duplication now, and in  the future. 

But that is likely not be so easy, as the Job is not just passed cleanly as a parameter, but accessed via `Configuration.get()`.

This seems ready to merge, any last words? :grin: 

_Mark

markmaker:
:+1:

tobydickenson:
scripting wiki has been updated

PR #1688, opened 2024-11-04
Refresh placement location after running Job.Placement.BeforeAssembly

tobydickenson:
# Description
`scriptBeforeAssembly()` runs the `Job.Placement.BeforeAssembly` script before placing a part in a job. That is called by `Place.stepImpl()`.

`stepImp()` fetches the placement location from the model using `getPlacementLocation()`. The same value is passed as a parameter to the script, and then to `place()`.

But one valid use case for `Job.Placement.BeforeAssembly` is to tweak the placement location. This patch adds one line to  `stepImpl` to re-fetch the location from the model after running the script.

# Justification
Openpnp determines panel&board fiducial location at the start of a production run. This is used to calculate the board-specific transform. This transformation remains unchanged for the lifetime of the job, and this can lead to placement errors. I am seeing:
 1. Thermal drift I think.... An experiment using a `Job.Placement.Complete` script to monitor the position of my homing fiducial after every placement showed I get a 0.03mm drift during the first 30 minutes of the job.
 2. There is some slop on my 3d printed board holder. It moves a tiny amount if I touch the board by hand during a job.
Those movements are all tiny, but they add up to something significant when placing fine pitch components.

I am now using a `Job.Placement.BeforeAssembly` script to detect fine-pitch components with challenging placement tolerances, then rechecking the board fiducials immediately prior to placement. This is a big hammer to crack a small nut, but is giving me meaningful improvements to placement accuracy:

```
if is_this_a_fine_pitch_part(part):
    locator = machine.getFiducialLocator()
    locator.locatePlacementsHolder(boardLocation)

````

`getPlacementLocation()` is pretty cheap, so there is no problem with the duplicated call for the 99.999% of cases where there is no script.

# Testing

## Baseline Test
* Single step the simulated machine to just after it has completed vision checks on a part
* Offset the board x location by 1mm.
* Single step once more, to perform a placement.
* Observe that the crosshairs showing the placement location are 1mm out

## Without this patch
* Add the `Job.Placement.BeforeAssembly` script above (but without the "if" line, so it runs on every placement).
* Single step the simulated machine to just after it has completed vision checks on a part
* Offset the board x location by 1mm.
* Single step once more, to perform the scripted board fiducial check followed by the placement.
* Observe that the board x location has been changed back, indicating that it has been corrected by the fiducial check.
* Observe that the crosshairs showing the placement location are still 1mm out, indicating that the placement was not corrected.

## With this patch
* Add the `Job.Placement.BeforeAssembly` script above (but without the "if" line, so it runs on every placement).
* Single step the simulated machine to just after it has completed vision checks on a part
* Offset the board x location by 1mm.
* Single step once more, to perform the scripted board fiducial check followed by the placement.
* Observe that the board x location has been changed back.
* Observe that the crosshairs showing the placement location is centred on the footprint, indicating that the placement location was corrected too.

# Implementation Details
4. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? - yes
5. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. - no
6. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. - all test pass

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -943,9 +943,10 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
             final Part part = placement.getPart();
             final BoardLocation boardLocation = plannedPlacement.jobPlacement.getBoardLocation();
 
-            Location placementLocation = getPlacementLocation(plannedPlacement);
+            scriptBeforeAssembly(plannedPlacement);
             
-            scriptBeforeAssembly(plannedPlacement, placementLocation);
+            // Calculate this after running the script, because the script might have fine-tuned the location data
+            Location placementLocation = getPlacementLocation(plannedPlacement);
 
             checkPartOn(nozzle);
             
@@ -1020,13 +1021,14 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
             }
         }
         
-        private void scriptBeforeAssembly(PlannedPlacement plannedPlacement, Location placementLocation) throws JobProcessorException {
+        private void scriptBeforeAssembly(PlannedPlacement plannedPlacement) throws JobProcessorException {
             final Nozzle nozzle = plannedPlacement.nozzle;
             final JobPlacement jobPlacement = plannedPlacement.jobPlacement;
             final Placement placement = jobPlacement.getPlacement();
             final Part part = placement.getPart();
             final BoardLocation boardLocation = plannedPlacement.jobPlacement.getBoardLocation();
             Length partHeight = part.getHeight();
+            Location placementLocation = getPlacementLocation(plannedPlacement);
             Location placementLocationPart = placementLocation.add(new Location(partHeight.getUnits(), 0, 0, partHeight.getValue(), 0));
             try {
                 HashMap<String, Object> params = new HashMap<>();


markmaker:
Hi @tobydickenson 

a valid feature request, it was discussed several times before. Thanks for tackling this. :+1: 

However, honestly, I would prefer this feature directly added to OpenPnP rather than a script. :grin:  So it could have some UI fiducial recalibration checkbox on the package, and some "minimum time" and/or "minimum number of placement between two fiducial recalibrations" setting on the Job Processor to prevent excessive fiducial recalibrations.

I also think this should actually happen first thing in a cycle, as soon as the nozzle-set of placement parts are determined. Not in the middle of placement, zooming around with delicate parts on the nozzle. What if a fiducial detection fails? Do you discard all these parts? 

And the recalibration criteria should be determined according to the most demanding of the whole nozzle-set of parts, not per individual nozzle part.

Finally, some alarm bells are ringing due to pending #1658 and #1654. 

@janm012012 do you see problem with this? Can your solutions cope with these locations being changed? And if so, am I right to assume this better be done up front, at the beginning of the cycle? So pre-rotate can then be final in the subsequent steps?

_Mark

janm012012:
Hi @markmaker!

I don't see any problems. The expected changes in placement location shall be small and hence not matter.

I'm with you, that rechecking board/panel fiducials could be a separate step after Plan() but before OptimizeNozzlesForPick() if it reuses the existing code. (This will likely generates merge conflicts with PR #1654, but they will be easy to resolve.) If this PR adds support for placement local fiducials, it's probably good where it is.

Jan

tobydickenson:
> prefer this feature directly added to OpenPnP rather than a script

Of course! This patch is the just the first step, enabling easy prototyping for the local calibration algorithm.

> some "minimum time" and/or "minimum number of placement between two fiducial recalibrations" setting

Agreed that would be good. It might also be useful to have a "Local Fiducial" type for fiducials that should not be checked as part of the initial board alignment.

> happen first thing in a cycle, as soon as the nozzle-set of placement parts are determined. Not in the middle of placement, zooming around with delicate parts on the nozzle

My thinking is that there are routine causes for interruptions during feeding, picking, and vision. That might trigger human intervention, which might cause the PCB alignment to move. It is better to do a local fiducial check once all the parts are safely loaded on the nozzles, so that there can be no interruption between the fiducial check and placement.

A counterargument is that a part might shift on the nozzle due to all the movement while checking fiducials. But I see that as a reason for minimising the number of fiducials being checked and travel distance, rather than changing the point in the process that checking occurs.

> What if a fiducial detection fails? Do you discard all these parts?

Thats a good question.

I am observing this pre-placement fiducial check to be reliable. Nonetheless, I agree "but what if" is an open question that I dont have a good answer for yet.

tonyluken:
Just some thoughts and questions:

1. Global fiducials (basically what we have now) - I always thought it would be good to have an option to re-perform the fiducial alignment that was performed at the start of the job either periodically, after some number of placements, or just before some selected placement(s) (or some combination of all three, i.e., whichever comes first). This would be exactly the same check and calculations as currently performed at the start of the job.
2. Local fiducials (something new) - these would be somehow associated with a subset of placements. They would not be checked as part of the global fiducial check (item 1 above). They would be checked just before the associated placements are made. I think it is important that the current global affine transform (from item 1) does **NOT** get modified by this check, i.e., it should generate an additional affine transform that applies on top of the current global affine transform and only for those associated placements. In that way, the local fiducial check won't impact placements outside the "local area".
3. Questions:

- How are local fiducials associated with placements? One obvious way would be by distance - Just pick the N closest fiducials (note that the global fiducials should be included as well as the local fiducials when picking the closest set). The placements then would just need a flag to say that they need a local fiducial check. 
- Does the local affine transform expire at some point? Obviously, it needs to expire whenever the global affine transform is updated but I would think it should also expire after some time limit as well. Others?
- If multiple placements are associated with the same set of local fiducials so they could use the same local affine transform, do they all get placed after one local fiducial check or does the local fiducial check happen before each placement? The latter could eat a lot of time unnecessarily. Perhaps the validity of the local affine transform (was it generated from the correct set of fiducials and was it generated recently) could be checked so as to avoid needless fiducial checks. Seems like this could have a lot of impact on job placement sorting order.
- What do you do if the parts loaded on the nozzles need different local fiducials? Also seems like a job placement sorting issue.

Tony

markmaker:
Keeping in mind the accuracy we really need is not microscopic, my [Devil's advocate](https://en.wikipedia.org/wiki/Devil's_advocate) questions would be these:

1. An option to redo the fiducials after a job was paused/interrupted seems the most pressing to me. The user might have had to manually fiddle with something which might have invalidated the fix.
2. Apart from (1), has anybody ever _really_ observed a relevant amount of positional drift that warrants redoing the fiducials _inside_ a job? Do PCB/panels creep in their holder? And if yes, shouldn't users rather do something against the cause rather than the symptom?
3. Regarding local fiducials: Same question. Has anybody ever _really_ observed PCBs that are distorted from the overall fix in a relevant amount? How can that be?
4. I can even less imagine how local deviations could change over time. PCBs made from Swiss cheese, slowly melting :grin:? Maybe with flex-PCBs, okay.
5. If you really need to compensate for local distortions, IMHO it should simply be a part of the global fiducial run. Either a non-linear compensation is needed, perhaps localized piecewise Affine with some blending. But again: how can that be?

_Mark

tobydickenson:
I am replying to some of these questions out of order:

> has anybody ever really observed a relevant amount of positional drift that warrants redoing the fiducials inside a job?
> Do PCB/panels creep in their holder?

* I observed the following 30µm drift during a job. This data was collected by a script which checks the position of my homing fiducial after every placement. X axis position is shown. A thermal effect on belts maybe?

image

* Anecdotally, I believe my 3d printed board holders suffer some creep. Some creepage occurs occurs in the first few minutes after loading the board. I have seen other creep mid-job, but it is hard to say whether that was because the dumb operator touched it. Measuring this would be an interesting experiment - I will do that.

> And if yes, shouldn't users rather do something against the cause rather than the symptom?

I am placing parts with 500µm pitch between pins. A misalignment of maybe 100µm would still lead to acceptable reflow. 30µm creepage doesnt seem unreasonable for a 3d printed board holder, but is still a significant chunck of the position accuracy budget that can be corrected by local fiducial check.

> PCBs made from Swiss cheese, slowly melting 😁? Maybe with flex-PCBs, okay

I partially agree with this sentiment. In my usage, local fiducials are useful because the board moves over time, and of course I am expecting it to have consistent rigid dimensions with no local distortion.

But, thermal expansion coefficient of FR4 is typically 15ppm/°C. On a large board, in the presence of a temperature gradient, the melting cheese analogy is not so silly. I do not have any experience of this being a problem, but I can see how it might happen. So I suggest it is valid to identify two different use cases:

 a. If your boards are small and rigid, then mid-job fiducial checks might be useful because the might board move. This is a time-local transform.
 b. If your boards are large enough then correcting for distortion may be important. This is a spacially-local transform.

To be clear: only the first of those applies to my current pick and place usage.

> One obvious way would be by distance - Just pick the N closest fiducials

I want to do some experiments here. For the time-local use case (i.e. correcting for boards that move over time) I am struggling to justify any solution other than N=all.

> The placements then would just need a flag to say that they need a local fiducial check.

It feels like the footprint carries enough information to determine this automatically for most cases. The footprint shows pad size and pad-gap size. Maybe not universal to all parts though.
 
> Does the local affine transform expire at some point? Obviously, it needs to expire whenever the global affine transform is updated but I would think it should also expire after some time limit as well. Others?

I suggest it makes more sense to handle expiration on the position of individual fiducials, rather than the transform. The transform can easily be recalculated on the fly, immediately prior to each placement.

> If multiple placements are associated with the same set of local fiducials so they could use the same local affine transform, do they all get placed after one local fiducial check or does the local fiducial check happen before each placement? The latter could eat a lot of time unnecessarily. Perhaps the validity of the local affine transform (was it generated from the correct set of fiducials and was it generated recently) could be checked so as to avoid needless fiducial checks. Seems like this could have a lot of impact on job placement sorting order.
> What do you do if the parts loaded on the nozzles need different local fiducials?

Consider the following scenario with three placements:

* The first placement means we need to check all its related fiducials.
* Maybe the second placement uses the same fiducials. The position data for those fiducials has not expired, so there is no need to re-check any positions. The calculated placement transform will be the same as the first.
* Maybe the third placement uses one fiducial which has already been checked, plus one fiducial that has not. We need to check only one fiducial position.

---

All of this is good discussion about a proper local fiducial system, but I note that this PR is just a one-liner change that enables an admittedly crude script-based solution which makes a meaningful difference to my production runs today.

markmaker:
> I observed the following 30µm drift during a job. This data was collected by a script which checks the position of my homing fiducial after every placement. X axis position is shown. A thermal effect on belts maybe?

Interesting. 

Just so you know this should be compensated by bottom vision. If there is creep in the overall motion system, then it will be equally present both in bottom vision, and over the PCB. So the bottom vision adjustment will cancel out the error. This only holds if PCB and bottom camera are close together (as is recommended), or if this creep is mostly an overall "shift", and not so much a "stretching".

_Mark

tobydickenson:
The discussion above has got distracted onto some grand ideas. It would be lovely if those ideas got implemented one day.

Meanwhile, the patch presented in this PR is a super-simple one-line change to recalculate a cached value after running a script, in case the underlying data was modified by that script.

Any objections to merging this in 2.3?

janm012012:
Hi Toby!

I never really saw the need for placement local fiducials and considered it historic. But I agree, with Tony and Mark, that it would be nice to support it in a well styled fashion with UI and expiration and all that fancy stuff. However, as no such PR is - at least to my knowledge - in sight and as I don't see that this PR might cause any harm to users not using the script and after making a final request, you may merge it. And the final request would be to move the position update into the script function and - if possible - link the update to the result of the script. Assuming that in 99% of all cases, no script will run and in 99% of the cases where the script runs, no location update is required, it would be a waist of resources to update it anyhow. Can you make the script return a value that triggers the update? Please also flag your change as "FIXME" mentioning that a placement local fiducial feature will make this modification obsolete. (Other uses reading the code shall be aware, that this is a workaround for a missing feature.)

Jan

tobydickenson:
> But I agree, with Tony and Mark, that it would be nice to support it in a well styled fashion with UI and expiration and all that fancy stuff.

I agree with that too :-)

> Assuming that in 99% of all cases, no script will run....

I have only just appreciated some of the inefficiencies in the scripting subsystem. That 99% common case applies to every scripting event. So lets put this PR on hold (briefly) while I propose some more general scripting improvements.

tobydickenson:
> And the final request would be to move the position update into the script function

This is done. It is much tidier now.

----

> and - if possible - link the update to the result of the script. Assuming that in 99% of all cases, no script will run and in 99% of the cases where the script runs, no location update is required, it would be a waist of resources to update it anyhow.

I dont like an optimisation targeting just this `getPlacementLocation`  recalculation. It saves only 10µs per placement and that is not worth the code complexity. It is not so easy to link the update to the result of the script.

PR #1744 is a better optimisation idea. 10x greater saving and much cleaner code.

----

Finally, I want to retest this on the real machine before merging.

tobydickenson:
This PR has been merged into #1744, for subsequent merging into test.

PR #1689, opened 2024-11-07
Mainframe divider had issue with limits when windows had changed.

tmandys:
# Description
Divider does not work correctly when changing window size. Eg. when switching to multiwindow and back. The divider is stuck at the right border. When changing window size then is pushed left but the limits are limited. Sometimes is stuck in middle because of calculated panel constraints around.

# Justification

# Instructions for Use
Switch to multiwindow and back or play with the window size. Now the divider should be usable.

# Implementation Details
Left and right panel have forced getMinimumX in onresize now . The solution is not too sophisticated. The better solution should consist of constraining all panel hierarchy, e.g. minimum size of jog panel to see all controls. But this solution does not work as expected, still divider issues, so I used mentioned fix.

### DIFF
--- a/src/main/java/org/openpnp/gui/MainFrame.java
+++ b/src/main/java/org/openpnp/gui/MainFrame.java
@@ -137,6 +137,7 @@ public class MainFrame extends JFrame {
     private static final int PREF_MACHINECONTROLS_WINDOW_WIDTH_DEF = 490;
     private static final String PREF_MACHINECONTROLS_WINDOW_HEIGHT = "MachineControlsFrame.windowHeight"; //$NON-NLS-1$
     private static final int PREF_MACHINECONTROLS_WINDOW_HEIGHT_DEF = 340;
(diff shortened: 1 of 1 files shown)


markmaker:
Hi @tmandys 

thanks a lot for addressing this. It has plagued many. :rocket: 

I'm quite clueless in Java Swing, so I can't really review the code.

@Others, any insights?!

I assume you tested it thoroughly? Including collapsing the Machine Controls etc.?

(Next time, please follow the PR template to give us some insight re testing.)

_Mark

tmandys:
Jog collapsion has no effect as it is on top/bottom panel on left panel being divided. I tested it, as I wrote it is not perfect but better than now. I found it when switched back from multiwindow and had no tab panel because divider was stuck on right side.

PR template: I used template provided by PR and I have deleted comments "how to fill it". What was wrong in my description ?

PR #1690, opened 2024-11-07
Save current distance value to preferences and restore when started

tmandys:
# Description
Distance slider position is not stored and when restarting it is hardcoded to finest 0.01 position

# Justification

# Instructions for Use
Set distance and restart sw. The slider should be reverted to previous position

# Implementation Details
Value is saved in preferences (not in XML as it is related to manual jog control)-

### DIFF
--- a/src/main/java/org/openpnp/gui/JogControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/JogControlsPanel.java
@@ -296,11 +296,21 @@ public class JogControlsPanel extends JPanel {
         panelControls.add(sliderIncrements, "18, 3, 1, 10"); //$NON-NLS-1$
         sliderIncrements.setOrientation(SwingConstants.VERTICAL);
         sliderIncrements.setMajorTickSpacing(1);
-        sliderIncrements.setValue(1);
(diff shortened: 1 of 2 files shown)


markmaker:
Hi @tmandys,

please open a discussion in the discussion group. We should ask users whether they want that. After all, there is a certain safety aspect to it. Moving 100mm instead of 0.01 can make the difference of a destroyed machine part, when one is a bit distracted.

_Mark

tobydickenson:
> please open a discussion in the discussion group. 

Discussion was here:

https://groups.google.com/d/msgid/openpnp/8e6a5ae9-bc96-49d9-a19f-f75b48027807n%40googlegroups.com?utm_medium=email&utm_source=footer

tobydickenson:
A final comment before merging: It would be better if the property name was more specific. `setDistance` is not a great name:
 1. There are lots of different distances. This is the jog distance.
 2. It is not really a distance - it is an index of positions of the jog distance slider.

So `setJogSliderIncrement` might have been better. But this is OK and I will merge it now.

tmandys:
Aha, it was meanwhile merged. So I won't fix name

PR #1693, opened 2024-11-07
Jobviewer board selection support to visualize particular boards

tmandys:
# Description
Board placement preview did show all placement for all boards in jobs. But when e.g. PnP boards depanelized from panel in one job then all placement are generated in the same area because boards are placed in machine origin position. So selection what placement should be displayed is welcome

# Justification

# Instructions for Use
Job panel grid row selection is passed to job viewer. So user can precisely select what to inspect.
Note: the board rectangles are displayed at original panel (KiCAD) location as there is no way how to get what is origin of particular subboard. I suggest add new parameter in list ... board origin. In this case the viewer can display correct board borders. Also PnP command "go to board origin" could move head to correct position.

# Implementation Details
JobViewer is extended to support selection argument. Jobpanel icon is enabled/disable by selection state of grid. Works also with panels and subboards.

Note: alternate implementation to display only enabled boards is more affecting as changing job configuration itself by clicking checkboxes. Selection is only visual action.

- How did you test the change?
manually test job panel with/without related jobviewer with boads, panels, child boards

- Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?
yes

- org.openpnp.spi or org.openpnp.model packages
no change

- mvn test
yes

### DIFF
--- a/src/main/java/org/openpnp/gui/BoardPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/BoardPlacementsPanel.java
@@ -656,7 +656,7 @@ public class BoardPlacementsPanel extends JPanel {
         @Override
         public void actionPerformed(ActionEvent arg0) {
             if (boardViewer == null) {
-                boardViewer = new PlacementsHolderLocationViewerDialog(new BoardLocation(board), false);
+                boardViewer = new PlacementsHolderLocationViewerDialog(new BoardLocation(board), false, null);
                 boardViewer.addWindowListener(new WindowAdapter() {
                     @Override
                     public void windowClosing(WindowEvent e) {
--- a/src/main/java/org/openpnp/gui/JobPanel.java
+++ b/src/main/java/org/openpnp/gui/JobPanel.java
@@ -75,6 +75,7 @@ import org.openpnp.events.PlacementsHolderLocationSelectedEvent;
 import org.openpnp.events.JobLoadedEvent;
 import org.openpnp.events.PlacementSelectedEvent;
 import org.openpnp.events.PlacementsHolderLocationChangedEvent;
+import org.openpnp.gui.JobPanel.OpenRecentJobAction;
 import org.openpnp.gui.components.AutoSelectTextTable;
 import org.openpnp.gui.components.ExistingBoardOrPanelDialog;
 import org.openpnp.gui.processes.MultiPlacementBoardLocationProcess;
@@ -174,22 +175,22 @@ public class JobPanel extends JPanel {
                 new ActionGroup(captureToolBoardLocationAction, moveCameraToBoardLocationAction,
                         moveCameraToBoardLocationNextAction, moveToolToBoardLocationAction,
                         twoPointLocateBoardLocationAction, fiducialCheckAction,
-                        setEnabledAction, setCheckFidsAction, setSideAction);
+                        setEnabledAction, setCheckFidsAction, setSideAction, viewerAction);
         singleSelectionActionGroup.setEnabled(false);
         
         multiSelectionActionGroup = new ActionGroup(captureToolBoardLocationAction, setEnabledAction, 
-                setCheckFidsAction, setSideAction);
+                setCheckFidsAction, setSideAction, viewerAction);
         multiSelectionActionGroup.setEnabled(false);
         
         singleTopLevelSelectionActionGroup = new ActionGroup(captureToolBoardLocationAction, removeBoardAction, captureCameraBoardLocationAction,
                 moveCameraToBoardLocationAction,
                 moveCameraToBoardLocationNextAction, moveToolToBoardLocationAction,
                 twoPointLocateBoardLocationAction, fiducialCheckAction,
-                setEnabledAction, setCheckFidsAction, setSideAction);
+                setEnabledAction, setCheckFidsAction, setSideAction, viewerAction);
         singleTopLevelSelectionActionGroup.setEnabled(false);
         
         multiTopLevelSelectionActionGroup = new ActionGroup(captureToolBoardLocationAction, removeBoardAction, setEnabledAction, 
-                setCheckFidsAction, setSideAction);
+                setCheckFidsAction, setSideAction, viewerAction);
         multiTopLevelSelectionActionGroup.setEnabled(false);
         
         jobTableModel = new PlacementsHolderLocationsTableModel(configuration);
@@ -341,6 +342,9 @@ public class JobPanel extends JPanel {
                             }
                         }
                         MainFrame.get().updateMenuState(JobPanel.this);
+                        if (jobViewer != null) {
+                            jobViewer.setPlacementsHolder(job.getRootPanelLocation().getPlacementsHolder(), getSelections());
+                        }
                     }
                 });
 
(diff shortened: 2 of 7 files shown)


tonyluken:
Hi @tmandys,

Rather than hiding all the boards/panels that are not selected in the Job panel, imho it would be better to always display all of them and highlight (possibly with a thicker outline) those that are selected. That way the operator can focus on the selected boards/panels but still see the context of where the selected boards/panels are in relation to the others in the job. What do you think?

tmandys:
@tonyluken, you suggestion is good for situation when you have all panelized PCBs still in panel. I tried to solve also situation when I removed PCBs from panel defined in centroid file. Particular boards are assembled separately, i.e. mounted in origin of machine and "positioned" via fucidials. So parts are in physical position but borders in virtual position with virtual orientation away from parts. Bonus of my solution is also that view should follow GUI when assembling (hopefuly, not yet tested in real PnP) and here should be selected one board.
Works rather surprisingly smoothly and practically when selecting boards in grid with opened viewer at side.

Clicking on borders in viewer could control selection in grid. But I think there is no back reference from viewer to board now.

tonyluken:
@tmandys,

I'm not sure I follow what you are trying to do. It sounds like you have physically broken the panel apart into its separate PCBs so you no longer actually have a panel. Why not just create a single board definition that describes the PCBs and put that into you job?

tmandys:
Because definition of broken panel has reference to original position in board. To achieve what you suggested means:

1. clone copy paste panel board to separate file (AFAIK Kicad cannot generate more boards from one schema so I have them in panel which goes to PCB factory)
2. remove all boards but one
3. move board to position 0
4. save as new file
5. generate centroid files
6. repeat for all boards in panel

tonyluken:
I'm sorry, I think we must have a bit of a language barrier as I still don't understand what you are doing. To me it sounds like you have a single physical PCB on your machine but have a panel definition with multiple PCBs loaded into the job.

- Do you have a physical panel made up of multiple physical PCBs or do you just have the individual physical PCBs (no longer part of a panel)?

- Are the PCBs all the same or are there multiple different designs?

- What does the KiCad output contain - one PCB or a panel of PCBs?

- How did you create the OpenPnP board and panel definitions?

- Can you send the files (*.job.xml, *.panel.xml, and *.board.xml) to me?

tmandys:
@tonyluken 

1. I have 9 different PCB designs for particular project panelized in Kicad to save PCB fabrication costs.
2. I do not need PnP all PCBs in one job as a) the panel is rather large, b) I do not need all PCBs in the same time c) there are plenty of issues in PnP process so I need minimize time till I put PCB to oven
4. export from Kicad 9 .pos file for each subboard (Kicad pcb_schema: just delete temorarily all boards but 1, export .pos,  undo, ...), So I have .pos with coords as was in panel
5. import boards to openpnp (9 boards)
6. create 1 job for 9 boards, I can selectively enable what board want to PnP
7. depanelize physically board 
8. put subboard in machine
9. synchronize fiducials 
10. placement can be displayed at real position, but border not as there is no info about subboard origin, so borders are displayed in virtual position and orientation of original panel

Even your suggestion select board by clicking on border won't work as all boards would be approx. on the same position as they are mounted at machine origin coord

tonyluken:
@tmandys,

Ok, now I think I understand what you are doing. The main problem is that the 9 .pos files you are creating all have their origin at the panel's origin rather than at each of the 9 PCB's origins. Could you not easily fix that in your step 3 by setting the origin for the placements (in KiCad) to the PCB origin just before you export the .pos file? It's easy to do using the "Place origin point for drill files and component placement files" button in KiCad:

PCB Editor

tmandys:
@tonyluken good idea, thanks

tonyluken:
@tmandys,

Are you ok then with closing this PR without merging it?

tmandys:
No! It is general improvement how to select what to display in job viewer. Currently is everything displayed stacked so it is useless for any case with more non-panelized-boards which are to be assembled in one job (or boards non-distributed-to-unique-places in machine which is very unlikely). Our discussion was only about if makes sense to select board by clicking on the border in viewer and why boarded is discontinued from its PnP parts. In theory could be implemented to change selection in grid by clicking in jobviewer. But I think it is not clean solution as there is no reference viewer->grid and would hardcode grid to viewer.
I leads back to question if is not better option display enabled boards in grid only. I chose row selection because it has no influence on xml file as it is only for visualization vs. enable for assembly process. But enable is also logical option. Advantage of selected row is that during assembly is auto displayed board being currently processed and not all (enabled) boards

tonyluken:
You should not have multiple boards in a job that can't possibly be physically on the machine at the same time. Those should be in separate jobs.

tmandys:
It depends on user preferences. Somebody can get (at least me ;-) that maintaining one job is better option than dozens of files for one PnP project. So the pull request extended support also for this options.

_You should not have multiple boards in a job that can't possibly be physically on the machine at the same time._
So what is reason enable checkbox on/off in grid and why are visualized disabled boards/panels ? When used jobviewer during PnP what benefit brings visualization of boards which are not processed ?

As I wrote earlier: alternative option is visualizing only enabled jobs but visualization based on row selection is more general

tonyluken:
@tmandys,

> So what is reason enable checkbox on/off in grid and why are visualized disabled boards/panels ? When used jobviewer during PnP what benefit brings visualization of boards which are not processed ?

The most common scenario for disabling a board is to disable all placements on a particular PCB that is known to be defective. For instance, you receive a panel of multiple PCBs from the fabricator, but because they have electrically tested all the PCBs, they know which of them is defective (have opens or shorts). Showing the disabled board(s) in the job viewer is a good confirmation to the operator that they have disabled the correct board(s) so no parts will be wasted by populating a defective PCB(s).

Another case may be that you have a job with four identical PCBs in it (and room on your machine for all four at the same time), but you only have 17 PCBs total to populate. That means you will run the same job five times, but on the fifth run you will need to disable three of the boards. It is much easier to see which one of the four is still enabled by seeing all four boards in the job viewer (with three of then crosshatched in red) rather than just the one board that is enabled.

Perhaps a compromise would be to keep the existing behavior and add an option to the job viewer to only display selected items. That way it will still work the same for everyone else and you can get what you want as well. I'd suggest adding a third state "Viewing Selected Only" to the button shown below (or change that button to a set of three grouped radio buttons):
Job Viewer

tmandys:
Or what about combobox with 3 items ? Radio button occupies more space and current switching button does not tell what are other options ?

tonyluken:
A combobox would be fine as well.

tmandys:
Done as you suggested:

image
image
image
image
image
image

Is it OK ?

tonyluken:
@tmandys,

Looks pretty good - I like it. But please fix the alignment of the items in the GUI. I think you can just add the following line after you create the combobox:

`cbxViewingOptions.setAlignmentX(Component.LEFT_ALIGNMENT);`

And maybe add a vertical strut of 5 pixels between the Viewing Side button and the combobox to add a little space between them.

Thanks,
Tony

tmandys:
Yup! Thanks.

image

tonyluken (approved):
@markmaker, @tmandys,

This looks good to me. I think it is ready to be merged.

markmaker:
Great work,@tmandys !
Great review, @tonyluken !

@tonyluken, Jason will invite you to get merge rights for OpenPnP. So you can merge it yourself! Thanks again, I really appreciate the help!

_Mark

vonnieda:
Hi @tonyluken,

Mark suggested that we offer you merge rights for OpenPnP, and I agreed wholeheartedly. Thank you for all your contributions to OpenPnP, all your help on the mailing list, your Wiki contributions, and everything else you've done for OpenPnP over the years.

You can now merge pull requests (your own, and others) and commit directly to the repo. I've sent you an invite, but I think it's already set since you were already a Wiki contributor.

Thanks,
Jason

PR #1698, opened 2024-11-15
improve JobProcessorConfigurationWizard translation

QUTAOJUN:
# Description
Extract the Border title：General.
Add the toolTipText for 'Job order' and ‘Max Vision Retries’，following the description in the Wiki.
JobProcessorPanel

# Justification
Perhaps adding some annotated information will enable users to further understand the functionality.
Since the descriptions exist in the Wiki, I'd still recommend adding them to the interface.

# Instructions for Use
Add the toolTipText for 'Job order' and ‘Max Vision Retries’

# Implementation Details
Tested in my Eclipse.
Yes.
N/A.
Tested in my Eclipse.


QUTAOJUN:
I'm a little confused about what is the ‘retry’ in this ‘Max Vision Retries’.
Is it related to Vision? But after Vision failure, no retry.

QUTAOJUN:
For the new function `Nozzle tip loading strategy` in the JobProcessorPanel，I want to rePR.

janm012012:
Hi QUTAOJUN!

Many thanks for your PR! 

> I'm a little confused about what is the ‘retry’ in this ‘Max Vision Retries’. Is it related to Vision? But after Vision failure, no retry.

It seems this is actually an error in the Wiki:

In the pick step https://github.com/openpnp/openpnp/blob/a16a8a86ac436782457d50369de8263e55b255ff/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L708 the parts pickRetryCount is used, which defaults to 0, so no retries. This calls feederPickRetry https://github.com/openpnp/openpnp/blob/a16a8a86ac436782457d50369de8263e55b255ff/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L827 which tries to pick from the same feeder using its pickRetryCount setting. This calls https://github.com/openpnp/openpnp/blob/a16a8a86ac436782457d50369de8263e55b255ff/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L873 to detect if a part has been picked and **that** only support vacuum part on sensing.

'maxVisionRetries' is used in the align step of the job processor: https://github.com/openpnp/openpnp/blob/a16a8a86ac436782457d50369de8263e55b255ff/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L951 and does not support picking a new part if bottom vision fails due to no part on the nozzle.

I've updated the Wiki. Please check it is more clear now.

Jan

janm012012:
Hi @QUTAOJUN!

Are you still interested in getting this PR merged? The JobProcessors configuration page has changed a little, so there currently a merge conflict. Thank You!

Jan

QUTAOJUN:
I'm sorry that I'm late. It's no problem to close it. It's out of time.
Because I found that the recent update frequency is very high, some new
modifications and functions have been added, and more entries must have
been introduced. I like it very much. In this way, I will definitely be
busy for Chinese translation.   :)  thank you very much.

Jan ***@***.***> 于2025年3月12日周三 18:26写道：

PR #1699, opened 2024-11-15
Driver side delaying

janm012012:
# Description
This PR adds support for driver-side delaying using G4 P<> and uses it for static pick and place dwell time.

# Justification
OpenPnP currently uses Thread.sleep() to delay operation which requires full machine coordination including wait for still stand. With driver-side delaying OpenPnP sends the required command to delay to the driver letting the driver time the delay. This provides a better and tighter utilization of the machine while reducing the scheduler induced timing jitter.

# Instructions for Use
Users will have to specify the new DELAY_COMMAND. Issues & Solutions suggests such a command.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **I've run a test job on a physical machine and verified, that pick and place is still precise.**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **yes: the driver interfaces now provides a delay() method with a default implementation using the current Thread.sleep().**

### DIFF
--- a/src/main/java/org/openpnp/machine/neoden4/NeoDen4Driver.java
+++ b/src/main/java/org/openpnp/machine/neoden4/NeoDen4Driver.java
@@ -425,6 +425,15 @@ public class NeoDen4Driver extends AbstractReferenceDriver {
     }
 
     @Override
+    public void delay(int milliseconds) throws Exception {
+        // ToDO: execute delay in hardware. As workaround, use Thread.Sleep()
+        // force full machine synchonization
+        waitForCompletion(null, CompletionType.WaitForStillstand);
+        // time delay using OS
+        Thread.sleep(milliseconds);
+    }
+
+    @Override
     public void setGlobalOffsets(Machine machine, AxesLocation location) throws Exception {
         // Take only this driver's axes.
         AxesLocation newDriverLocation = location.drivenBy(this);
--- a/src/main/java/org/openpnp/machine/neoden4/NeoDen4Driver.java
+++ b/src/main/java/org/openpnp/machine/neoden4/NeoDen4Driver.java
@@ -425,15 +425,6 @@ public class NeoDen4Driver extends AbstractReferenceDriver {
     }
 
     @Override
-    public void delay(int milliseconds) throws Exception {
-        // ToDO: execute delay in hardware. As workaround, use Thread.Sleep()
-        // force full machine synchonization
-        waitForCompletion(null, CompletionType.WaitForStillstand);
-        // time delay using OS
-        Thread.sleep(milliseconds);
-    }
-
-    @Override
     public void setGlobalOffsets(Machine machine, AxesLocation location) throws Exception {
         // Take only this driver's axes.
         AxesLocation newDriverLocation = location.drivenBy(this);
--- a/src/main/java/org/openpnp/machine/reference/ContactProbeNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ContactProbeNozzle.java
@@ -92,7 +92,7 @@ public class ContactProbeNozzle extends ReferenceNozzle {
     private double contactProbeSpeed = 0.05;
 
     @Attribute(required=false)
-    private long sniffleDwellTime = 250;
+    private int sniffleDwellTime = 250;
 
     @Element(required = false)
     private Length contactProbeAdjustZ = new Length(0, LengthUnit.Millimeters);
@@ -440,7 +440,7 @@ public class ContactProbeNozzle extends ReferenceNozzle {
         return sniffleDwellTime;
     }
 
-    public void setSniffleDwellTime(long sniffleDwellTime) {
+    public void setSniffleDwellTime(int sniffleDwellTime) {
         this.sniffleDwellTime = sniffleDwellTime;
     }
 
@@ -543,7 +543,7 @@ public class ContactProbeNozzle extends ReferenceNozzle {
                 for (int i = 0; i < count; i++) {
                     probedLocation = probedLocation.subtract(probeIncrement);
                     moveTo(probedLocation);
-                    Thread.sleep(sniffleDwellTime);
+                    delay((int)sniffleDwellTime);
                     if (! isPartOff()) {
                         // We got contact.
                         probedLocation = probedLocation.add(new Location(contactProbeAdjustZ .getUnits(), 0, 0, contactProbeAdjustZ.getValue(), 0));
--- a/src/main/java/org/openpnp/machine/reference/ContactProbeNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ContactProbeNozzle.java
@@ -543,7 +543,7 @@ public class ContactProbeNozzle extends ReferenceNozzle {
                 for (int i = 0; i < count; i++) {
                     probedLocation = probedLocation.subtract(probeIncrement);
                     moveTo(probedLocation);
-                    delay((int)sniffleDwellTime);
+                    delay(sniffleDwellTime);
                     if (! isPartOff()) {
                         // We got contact.
                         probedLocation = probedLocation.add(new Location(contactProbeAdjustZ .getUnits(), 0, 0, contactProbeAdjustZ.getValue(), 0));
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
@@ -1142,7 +1142,7 @@ public class ReferenceNozzle extends AbstractNozzle implements HeadMountable {
         else {
             // simple method, just dwell
             Logger.trace(getName()+" dwell for pick vacuum "+milliseconds+"ms");
-            Thread.sleep(milliseconds);
+            delay(milliseconds);
         }
     }
 
@@ -1177,7 +1177,7 @@ public class ReferenceNozzle extends AbstractNozzle implements HeadMountable {
         else {
             // simple method, just dwell
             Logger.trace(getName()+" dwell for place vacuum dissipation "+milliseconds+"ms");
-            Thread.sleep(milliseconds);
+            delay(milliseconds);
         }
     }
 
@@ -1236,7 +1236,7 @@ public class ReferenceNozzle extends AbstractNozzle implements HeadMountable {
             else {
                 // simple method, just dwell 
                 Logger.trace(getName()+" dwell for part off probing, open valve "+probingMilliseconds+"ms");
-                Thread.sleep(probingMilliseconds);
+                delay(probingMilliseconds);
                 if (dwellMilliseconds <= 0) {
                     returnedVacuumLevel = readVacuumLevel();
                 }
@@ -1275,7 +1275,7 @@ public class ReferenceNozzle extends AbstractNozzle implements HeadMountable {
             // simple method, just dwell and then read the level
             if (dwellMilliseconds > 0) {
                 Logger.trace(getName()+" dwell for part off probing, closed valve "+dwellMilliseconds+"ms");
-                Thread.sleep(dwellMilliseconds);
+                delay(dwellMilliseconds);
                 returnedVacuumLevel = readVacuumLevel();
             }
             // return the vacuum level, either from before or after valve closed
--- a/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java
@@ -1283,14 +1283,14 @@ public class ReferenceNozzle extends AbstractNozzle implements HeadMountable {
         }
     }
 
-	/**
-	 * Delay for a given time in milliseconds and take the respective vacuum valve into account.
-	 * @param milliseconds
-	 * @throws Exception 
-	 */
-	private void delay(int milliseconds) throws Exception  {
+    /**
+     * Delay for a given time in milliseconds and take the respective vacuum valve into account.
+     * @param milliseconds
+     * @throws Exception 
+     */
+    private void delay(int milliseconds) throws Exception  {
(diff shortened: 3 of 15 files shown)


markmaker:
Thanks @janm012012,

very well designed re the drivers that need to participate! :rocket: 

Also superb with Issues & Solutions support :+1: 

Some thing left to do, though:

You should make sure the planner is sending pending motion before doing the delay. Add this at the top of the method:

```java
        // Plan and execute any queued motion commands. 
        executeMotionPlan(completionType);
```
Otherwise you'll get an inversion, i.e. the delay is sent _before_ the queued motion.

Note, this addition might not be necessary in current scenarios (and in your tests), because delays typically happen after an actuation, which does the coordination for you. But it might not be universally be the case in future code, and/or some actuators might be completely uncoordinated, which should work too. So please add it. It will do nothing if the queue is already empty.

Discussion: 

1. If the drivers of the `hm` axes are not the same as the driver of the vacuum valve actuator, for instance, it is not guaranteed that they would wait for each other. On the other hand, such a user would then have to switch on **After Actuation** coordination, and specifically with `WaitForStillstand` rather than `CommandStillStand` (with your new option). So it would still work okay. Agreed?

3. Theoretically, the delay could also come without any actuation before it. There could be motion by a different driver still pending, when we delay on another driver. So we'd need a coordination similar to `interlockMotionAcrossDrivers()`. Or we'd need a full `WaitForStillstand` inside `delay()`. Alternatively, for such a scenario _the calling code_ could be made responsible to do the `WaitForStillstand` before calling `delay()`. But then this should be in the comments of the `hm.delay()` method. Agreed?

One more thing:

I believe some G-code dialects take _seconds_ rather than _milliseconds_. At least in the past there were some that did so exclusively.

For example 

`G4 S0.01 ; dwell 10 milliseconds`

You should define a _second_ variable substitution with unit seconds (divide by 1000.0).

https://reprap.org/wiki/G-code#G4:_Dwell

_Mark

janm012012:
Hi  @markmaker!

Many thanks for your kind review!

I've added the two changes. They are good to have, IMHO.

Concerning the discussion: 

1.  I thought that the axes assigned to the actuator shall be considered then delaying. Therefore I'd assume, that the vacuum actuator can reference the effected nozzle axes and hence the delay will be applied to their driver(s).
2. At present I only see the use case of executing static pick/place dwell time in the driver. There the relation between actuator and axes can be configured (see previous point) and shall hence be unique. If in the future anyone will request a delay that can not be correlated to a set of axes/drivers, we have to rethink about it. At present I'm aware of machines with a single controller (like my) which is always fine and machines with multiple controller or controllers with extension using one for head related stuff and a second (or the main driver board) for the remaining stuff namely the primary axis X and Y. I'd assume that this scenario can be covered by evaluation the axes of the actuator which shall reference X and Y on the main controller and Z and C on the head controllers as well. (Btw: I don't understand how such a scenario can synchronize XY with Z.) If you feel more safe, we could make multi-controller setups stick with the Thread.sleep() code... Which would be likely a pity for all thus users with a secondary controller for feeders only.

Jan

markmaker:
Thanks, @janm012012 

### Discussion

> 1. I thought that the axes assigned to the actuator shall be considered then delaying. Therefore I'd assume, that the vacuum actuator can reference the effected nozzle axes and hence the delay will be applied to their driver(s).

But you are only executing the `delay()` method on the nozzle (as `hm`), not on the actuator. See your own code.

And I would not assign axes to an actuator just for that purpose. Too much baggage in safe Z moves etc. and confusing/dangerous for UX, especially Z (crashes via vacuum actuator would not be taken kindly). And btw. what Z and C do you assign to the camera light or pump actuators, N1's or N2's? Machine actuator don't even expose axes, but the same need for coordination might be needed.

Like I said, I believe the user is responsible for setting the right machine coordination on the Actuator, if they are not on the same driver and/or don't use a driver at all (`ScriptActuator`, `HttpActuator`).

> 2. At present I only see the use case of executing static pick/place dwell time in the driver. 

Well you could also add some of these :smirk::

https://github.com/search?q=repo%3Aopenpnp%2Fopenpnp+Thread.sleep&type=code

What if we wanted _this one_ do the dwell on the driver?

https://github.com/openpnp/openpnp/blob/03410fe95cd5938699e27c37ddcb878a183106a2/src/main/java/org/openpnp/machine/reference/camera/AbstractSettlingCamera.java#L663-L666

The light actuator that comes before this, might be uncoordinated (recommended), or missing (light always on), or have no driver at all (`ScriptActuator`, `HttpActuator`). That's why even today the code explicitly does this:

https://github.com/openpnp/openpnp/blob/03410fe95cd5938699e27c37ddcb878a183106a2/src/main/java/org/openpnp/machine/reference/camera/AbstractSettlingCamera.java#L656-L657

With the delay delegated to the driver, we would have to do it _after_ the delay (but only in the fixed time method, not with the vision settling!). 

So I believe it is okay to say the caller of `delay()` is responsible to add the `waitForCompletion()`.

Hmm, this case actually makes me think. I guess you need to add 

```java
    motionPending = true;
```

to your `GcodeDriver.delay()`. A dwell is to be treated like a pending motion (mathematically, velocity == 0 is just a special case :sunglasses: ). Otherwise the driver will optimize away a subsequent `waitForCompletion()`, which **would be wrong!** See the code:

 https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java#L883-L884

Next issue:

> There the relation between actuator and axes can be configured (see previous point) and shall hence be unique. 

Afraid, I don't agree, see argument about point 1.

> If in the future anyone will request a delay that can not be correlated to a set of axes/drivers, we have to rethink about it.

IMHO, like in the case of the camera fixed settle time, we need to do the `waitForCompletion()` in the code if we need it.

### G4 Variable Substitution

The "second" (dual meaning :smirk:) variable substitution is fine.

But adding the `{TimeSeconds:S%.3f}` to the `DELAY_COMMAND` defaults/proposal is wrong, they are now both substituted. 

I just meant to support it _as a variable_ that a user can use on exotic/old controllers that only support `S`. But all default/proposed commands should only use the `P`. I don't know what happens if both are specified :fearful: 

_Mark

janm012012:
Hi  @markmaker!

I fixed the default DELAY_COMMAND and replaced some more Thread.sleep with delay() in case its obvious to me that the controller can do better.

### Discussion


> > 1. I thought that the axes assigned to the actuator shall be considered then delaying. Therefore I'd assume, that the vacuum actuator can reference the effected nozzle axes and hence the delay will be applied to their driver(s).
> 
> But you are only executing the `delay()` method on the nozzle (as `hm`), not on the actuator. See your own code.

I still think it's ok as is because the respective nozzle has to be at it's desired location before the delay shall take place and the valve gets actuated.


> So I believe it is okay to say the caller of `delay()` is responsible to add the `waitForCompletion()`.

I agree. The mentioned camera settling is something I'd consider to stay better off in OpenPnP anyhow because OpenPnP will access the camera next.

> Hmm, this case actually makes me think. I guess you need to add
> 
> ```java
>     motionPending = true;
> ```

Agreed and added. I now understand the use case like the camera settling you pointed out.

Jan

markmaker:
Hi Jan,

thanks. 

For starters, a simple one: please undo the one with `dollarWaitTimeMilliseconds`. We specifically want _no serial communication_ during this sleep. Background, see here: 

https://github.com/openpnp/openpnp/wiki/GcodeDriver#dollar-command-wait-time

But for the rest of the issues, I'm afraid, we haven't thought far enough. :frowning_face: 

For the pump actuator case:

```java
pump.delay(getPumpOnWaitMilliseconds());
```

this is actually one of the brittle cases I spoke about. 

I know, in your current code it will work, because you handle machine actuators (with `head == null`) as `null`, which in turn will emit the delay on all drivers (more on that later). You then need to coordinate before on the valve. But this is really hanging on by a very thin thread, it is undocumented and counter-intuitive behavior. Maybe the next guy will think this should be optimized, and issue the delay on the actuator's _own_ driver instead. Or another guy allows head actuators for pumps (as we do for lights), etc. and this will topple this house of cards. Another might copy&paste the code, and (rightfully) think this works for any kind of actuator, but it would not for head mountable actuators with no axes, or where the axis drivers are different from those that need to wait for it.

The first thought would be to say "hey it is not the pump that we want to delay, but the valve opening". So it should be:

```java
valve.delay(getPumpOnWaitMilliseconds());
```

But what if the pump actuator is on a different driver than the valve? Maybe the pump driver is the same driver as motion axes, and motion is still pending. So the valve driver would dwell, but maybe start to do so long before the pump got its command!

Okay so you say "okay, then enable machine coordination before and/or after on the pump and valve actuators, and we replace the pump with the nozzle to enact the delay, so it is motion bound". But this wastes all the G-code queuing optimization you wanted to achieve in the first place. 

```java
nozzle.delay(getPumpOnWaitMilliseconds());
```

The longer I think about this, the longer I'm desperate for a clear solution.

Maybe we need to pass a _whole array_ of `hm`s to delay? So we would pass the hm which actually denotes what happens before (in this case the pump actuation) as the `this` parameter, but then we also pass the nozzle and the valve in an array? And we must also take the actuator's own dedicated drivers into account, in addition to all axis drivers, if they have any.

Is this complexity worth the effort after all? Or do we need to say "when there is more than one driver on the machine, then fall back to `Thread.sleep()`"?

Another issue pops up when really thinking about "all drivers". When you emit the `delay()` on all drivers (which is already the case with `hm == null`) then all drivers need to support `delay()`. But some drivers might be simple actuator or feeder drivers that don't even have G-code, but maybe a binary protocol, let alone the notion of a dwell command. Currently your code would throw an exception with the missing command, i.e. effectively become a show-stopper on such machines :fearful: 

Instead, I propose a `boolean` return that would indicate `false` when the driver does not have the command. Then in the motion planner, if _any_ of the drivers return `false`, you need to fall back to `Thread.sleep()`. Note this way you can also change your current `default` implementation to just return `false`. This would fix another bug, the current implementation would multiply the sleep time by number of such drivers, because each of them sleeps _after_ the other.

No easy solution...

Phew... such "physically involved" PRs are never really easy. Thank you for hanging in there. :+1: 

_Mark

janm012012:
Hi @markmaker!

I've reverted dollarWait and pump to Thread.sleep(). For pump I have changed my mind in the meantime anyhow, because I'd prefer to delay asynchronously (immediately actuate, remember the startup, synchronize with code when pump is actually needed applying the remaining delay as needed. But that's beyond the scope of this PR.)

I've change the type of delay() and moved the Tlread.sleep() fallback into the MotionPlanner. I hope there will be no negative responses from users using a secondary GCode driver for feeders where delay is not supported and not needed.

Jan

janm012012:
Hi @markmaker!

I've just reverted my changes to SampleJobText as they are not needed anymore with the new way of handling the default delay() implementation. Previously the DELAY_COMMAND was required, which made the modification mandatory.

Jan

janm012012:
Hi Mark!

I reviewed all your commends and reworked my code in order to fix them all. It seems the difficult part is to make **all** affected drivers wait before work continues. I've solved that by following your suggestion letting delay accept an array of `HeadMountable`s. If this array is empty, all drivers are instructed to wait. Otherwise each `HeadMountable` is queried for its driver and all driver for all of it's mapped axes. All this drivers are then instructed to wait.

To test the new code, I've setup a simulation machine that uses different drivers for all its axes and for all its actuators. It emits the following GCode when commanded to feed and pick a part.

```
2025-01-31 23:38:21.830 GcodeAsyncDriver DEBUG: [N1VAC:simulated: port 57601] commandQueue offer >> M800 ; on, open solenoid valve, RED led ON on soleonid
2025-01-31 23:38:21.839 GcodeAsyncDriver DEBUG: [N1VAC:simulated: port 57601] commandQueue offer >> G4 P20 ; Delay for given time in [ms]
2025-01-31 23:38:21.840 GcodeAsyncDriver DEBUG: [X:simulated: port 57591] commandQueue offer >> G4 P20 ; Delay for given time in [ms]
2025-01-31 23:38:21.840 GcodeAsyncDriver DEBUG: [Y:simulated: port 57593] commandQueue offer >> G4 P20 ; Delay for given time in [ms]
2025-01-31 23:38:21.840 GcodeAsyncDriver DEBUG: [Z:simulated: port 57595] commandQueue offer >> G4 P20 ; Delay for given time in [ms]
2025-01-31 23:38:21.841 GcodeAsyncDriver DEBUG: [N1:simulated: port 57597] commandQueue offer >> G4 P20 ; Delay for given time in [ms]
2025-01-31 23:38:21.883 GcodeAsyncDriver DEBUG: [Z:simulated: port 57595] commandQueue offer >> M204 S15209 G1 Z-9.0 F37010 ; move to target
```

One can clearly see, that once the vacuum valve has been opened (M800) all axes of the nozzle (X, Y, Z, N1) and the vacuum valve actuator (N1VAC) are requested to delay for the configured pick dwell time.

I made a second test with all axes and actuators on a single controller and a second unused controller and there only the main controller is commanded to wait:

```
2025-01-31 23:51:00.112 GcodeAsyncDriver DEBUG: [GcodeDriver:simulated: port 57634] commandQueue offer >> M800 ; on, open solenoid valve, RED led ON on soleonid
2025-01-31 23:51:00.120 GcodeAsyncDriver DEBUG: [GcodeDriver:simulated: port 57634] commandQueue offer >> G4 P20 ; Delay for given time in [ms]
2025-01-31 23:51:00.145 GcodeAsyncDriver DEBUG: [GcodeDriver:simulated: port 57634] commandQueue offer >> M204 S15209 G1   Z-9.0    F37010 ; move to target
```

If you find the time, please have a look and tell me, if there is anything that still might cause problems. Thank You!

Jan

PR #1700, opened 2024-11-15
Skip auto-focus on up-looking camera calibration

janm012012:
# Description
This PR provides the option to skip auto-focus calibration of the up looking camera via issues & solutions.

# Justification
At present there is no convenient option for force bottom vision to take place at a certain Z level because issues & solutions always uses auto-focus to set the camera Z.

# Instructions for Use
A checkbox has been added to the solution allowing to disable auto-focus. There is no other impact.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **on my physical machine**

### DIFF
--- a/src/main/java/org/openpnp/gui/components/IssuePanel.java
+++ b/src/main/java/org/openpnp/gui/components/IssuePanel.java
@@ -35,6 +35,7 @@ import java.awt.event.MouseListener;
 import javax.swing.ButtonGroup;
 import javax.swing.DefaultComboBoxModel;
 import javax.swing.JButton;
+import javax.swing.JCheckBox;
 import javax.swing.JComboBox;
 import javax.swing.JLabel;
 import javax.swing.JPanel;
@@ -68,6 +69,7 @@ import org.openpnp.model.Solutions;
 import org.openpnp.model.Solutions.Issue;
 import org.openpnp.model.Solutions.Issue.ActionProperty;
 import org.openpnp.model.Solutions.Issue.DoubleProperty;
+import org.openpnp.model.Solutions.Issue.BooleanProperty;
 import org.openpnp.model.Solutions.Issue.IntegerProperty;
 import org.openpnp.model.Solutions.Issue.LengthProperty;
 import org.openpnp.model.Solutions.Issue.MultiLineTextProperty;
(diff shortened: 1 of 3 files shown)


markmaker:
Thank you @janm012012 !

this is much appreciated. :+1: 


I see you implemented a new `BooleanProperty`. So far I implemented such a choice with a two radio buttons choice between `true` and `false`. This way I can separately explain both variants. See an example here:

https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/machine/reference/solutions/GcodeDriverSolutions.java#L210-L225

I think it would be much appreciated by users to get more information about the consequences of their choice, _either way_. For instance, you should definitely **CAUTION users that the Auto-Focus part height detection will not work**, when they choose _not_ to do the Auto-Focus calibration here. We talked about this in the discussion group, and you agreed:

> The automatic part height detection is something I was not aware off. Seems like a need feature. I'm afraid a PR that changes the UI to allow disabling the auto-focus setup of the camera could result in voiding this feature and in turn causing more support trouble. 

Exactly!  :smiley: 

You then asked:

> Would it be possible to somehow disentangle the camera's focal plain Z from the bottom vision Z in the software?

We then [discussed](https://groups.google.com/g/openpnp/c/z0Wl1kdpWXI/m/uK08-q_LBAAJ) a solution. But unless that solution is ready, a big fat "CAUTION" note is required.

Regarding the new `BooleanProperty`, that's perfectly implemented and you should certainly keep it in this PR. It could still become useful in the future, e.g., when it is about (multiple) "extras".

Finally, a small cosmetic issue: you added quotes around the camera name. I see the rationale, but then I think it should be _consistently_ done that way, not only for cameras but also for every other `getName()` in any solution text, for cameras elsewhere, but also for nozzles etc.

So either change them all, or leave it be. :grin: 

Personally, I think it is sufficiently clear (in English) via the capital first letter of such a name.

_Mark

janm012012:
Hi @markmaker!

I've tried changing my checkbox into a choice but failed because I don't yet see how to combine Issue.Choice[] with Issue.CustomProperty[]. I only found that Issue.Choice[] is used standalone but here it has to be combined with IntegerProperty and ActionProperty. Would you please hint me? Thank You!

I've reverted the single quotes I added. Now the issue reads again "Determine the up-looking camera Up Looking position and initial calibration.". Maybe I shall give my camera a better name...

Jan

markmaker:
> I've tried changing my checkbox into a choice but failed ... Would you please hint me? Thank You!

Does the linked example not work as a "how-to"?

https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/machine/reference/solutions/GcodeDriverSolutions.java#L210

Maybe this one is simpler to copy&paste:

https://github.com/openpnp/openpnp/blob/5dd12c73972c74a5f3fa1810fc9167ce4919cec4/src/main/java/org/openpnp/machine/reference/solutions/KinematicSolutions.java#L129


_Mark

janm012012:
Hi @markmaker!

Unfortunately now. If I replace my code as suggested

```
           return new Solutions.Issue.CustomProperty[] {
                   new Solutions.Issue.Choice[] {
                        new Solutions.Issue.Choice(true,  
```

the compiler complains "Type mismatch: cannot convert from Solutions.Issue.Choice[] to Solutions.Issue.CustomProperty".

According to my research, Choice() is not yet used in combination with CustomProperty[], only standalone. But here I think we need Choice() in combination with IntegerProperty() and ActionProperty(). Maybe some sort of ChoiceProperty() might be needed grouping Choice[] int a single element of CustomProperty type...

Jan

markmaker:
Are you sure the two can't be combined?

(can't check at the moment)

_Mark

janm012012:
I tried my very best, but did not found a solution. It seems, that IntegerProperty (and others) inherit CustomProperty which seems to allow them to be combined, while Choice stands for itself. I can change Choice to inherit CustomProperty, however then there is no UI available as that stands by itself as well... It would probably make sense to pull Choice into CustomProperty to provide the same usability, but its beyond my scope if that might cause any negative side effects...

Maybe you can have a look once you have some time for that...

Jan

markmaker:
Looks good! Can I merge?

janm012012:
I just pushed a (last) cosmetic change: added "This is the recommended setting." to he default choice and styled "CAUTION" in red, as Advance Camera Calibration does.

Yes, I think the code is ok, used it a few times today on my physical machine.

Jan

markmaker:
Thank you @janm012012 !

:medal_sports: 

_Mark

PR #1702, opened 2024-11-17
Actuator coordination

janm012012:
# Description
This PR changed the "Before Actuation?" coordination to CommandStillstand as WaitForStillstand is not required.

# Justification
As discussed on the mailing list, CommandStillstand is sufficient for actuator SET operations and provides a better machine utilization and hence a better performance.

# Instructions for Use
this optimization will work immediately. However, in case of miss configurations where an actuator is actuated and read, the addition "Before Read?" coordination is required where before the "Before Acutation?" was sufficient.

This PR will cause merge conflicts with PR #1699 because both change the implementation of ReferenceNozzle for static dwell time handling.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **This code was used for the tests and demonstration discussed on the mailing list using a physical machine and a real job.**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **yes: the coordination handling has been relocated**

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ContactProbeNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ContactProbeNozzle.java
@@ -819,18 +819,18 @@ public class ContactProbeNozzle extends ReferenceNozzle {
                 }
                 else if (getContactProbeActuator() instanceof AbstractActuator) {
                     AbstractActuator contactProbeActuator = (AbstractActuator) getContactProbeActuator();
-                    if (contactProbeActuator.getCoordinatedAfterActuateEnum() != ActuatorCoordinationEnumType.WaitForStillstand) {
+                    if (contactProbeActuator.getCoordinatedAfterActuateEnum() != ActuatorCoordinationEnumType.WaitForUnconditionalCoordination) {
                         solutions.add(new Solutions.Issue(
                                 contactProbeActuator, 
-                                "Contact probe actuator needs machine coordination after actuation.", 
-                                "Enable After Actuation machine coordination.", 
+                                "Contact probe actuator needs unconditional machine coordination after actuation.", 
+                                "Set After Actuation machine coordination to WaitForUnconditionalCoordination.", 
                                 Severity.Error,
                                 "https://github.com/openpnp/openpnp/wiki/Motion-Planner#actuator-machine-coordination") {
 
                             @Override
                             public void setState(Solutions.State state) throws Exception {
                                 if (state == Solutions.State.Solved) {
-                                    contactProbeActuator.setCoordinatedAfterActuateEnum(ActuatorCoordinationEnumType.WaitForStillstand);
+                                    contactProbeActuator.setCoordinatedAfterActuateEnum(ActuatorCoordinationEnumType.WaitForUnconditionalCoordination);
                                 }
                                 super.setState(state);
                             }
--- a/src/main/java/org/openpnp/machine/reference/ContactProbeNozzle.java
+++ b/src/main/java/org/openpnp/machine/reference/ContactProbeNozzle.java
@@ -53,6 +53,7 @@ import org.openpnp.spi.Machine;
 import org.openpnp.spi.Nozzle;
 import org.openpnp.spi.NozzleTip;
 import org.openpnp.spi.base.AbstractActuator;
+import org.openpnp.spi.base.AbstractActuator.ActuatorCoordinationEnumType;
 import org.openpnp.spi.base.AbstractHead;
 import org.openpnp.util.Collect;
 import org.openpnp.util.MovableUtils;
@@ -818,7 +819,7 @@ public class ContactProbeNozzle extends ReferenceNozzle {
                 }
                 else if (getContactProbeActuator() instanceof AbstractActuator) {
                     AbstractActuator contactProbeActuator = (AbstractActuator) getContactProbeActuator();
-                    if (!contactProbeActuator.isCoordinatedAfterActuate()) {
+                    if (contactProbeActuator.getCoordinatedAfterActuateEnum() != ActuatorCoordinationEnumType.WaitForStillstand) {
                         solutions.add(new Solutions.Issue(
                                 contactProbeActuator, 
                                 "Contact probe actuator needs machine coordination after actuation.", 
@@ -828,7 +829,9 @@ public class ContactProbeNozzle extends ReferenceNozzle {
 
                             @Override
                             public void setState(Solutions.State state) throws Exception {
-                                contactProbeActuator.setCoordinatedAfterActuate((state == Solutions.State.Solved));
+                                if (state == Solutions.State.Solved) {
+                                    contactProbeActuator.setCoordinatedAfterActuateEnum(ActuatorCoordinationEnumType.WaitForStillstand);
+                                }
                                 super.setState(state);
                             }
                         });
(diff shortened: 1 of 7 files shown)


janm012012:
Hi @markmaker!

According to the logs, the SimpleJobTest failes with 

> 2024-11-17 23:05:06 [pool-1-thread-1] org.openpnp.machine.reference.camera.ImageCamera.isPartLocation()
ERROR: No good PnP location recognized in target area at location (177.148874, 40.631649, 0.000000, 180.000009 mm). Best  match score 0.3819229304790497 does not meet 0.85, distance 0.31747499999999995
> 2024-11-17 23:05:06 [pool-1-thread-1] org.openpnp.machine.reference.camera.ImageCamera.isPartLocation()
ERROR: No PnP location recognized in target view at location (177.110546, 40.606516, 0.000000, 183.425981 mm)

Is there a coordination issue that is reviled by my changes? Any idea how I can fix that?

Jan

markmaker:
Hi Jan,

I'm not happy with just replacing it. 

Remember, we talked about making it another **option** on the actuator.

I said:
> We could expose that distinction in the Actuator machine coordination options, then OpenPnP would just command it, but not wait for it. So you get the best of both worlds, and (importantly) on all controllers.

At one time you said the following that sounded like a switchable option:
> I've remoddelt the acutator coordination code in OpenPnP as proposed to provide an easy way to switch between WaitForStillstand and CommandStillstand and to apply CommandStillstand only where possible


Why is this necessary? 

Actuators and motion axes can be on _two_ (or more) separate drivers. Or actuators can have no driver at all (Http, Script). Then when you just _command_ stillstand, the motion on the first driver will eventually comply, but that might only happen _seconds later_, but the actuator on the second driver will already be in stillstand, or have no notion of "stillstand" at all, or no driver in the first place, so it will _actuate immediately_.

That's the same effect that you are effectively seeing in the SimpleJobTest (strictly speaking an artifact of a simplification in the simulation machine, i.e., should be moved to the GcodeServer, but that was a bit too hard to implement). Good it  got caught, as this test actuator could have been on a different driver or on an Http or Script actuator.

You also had to add

```java
waitForCompletion(CompletionType.WaitForStillstand)`
```

to the nozzle dwelling. That's another artifact of this not being an option. Calling from code should not be necessary. Instead the user is responsible to set the right machine coordination on the valve actuator.

This will also be true once #1699 is merged. Remember, even #1699 might fall back to `Thread.sleep()`, and we can still have actuators on different drivers or no drivers. So `WaitForStillstand` will still be needed in these cases, and it will be the user's responsibility to select the right option. So the tooltips and the Wiki will be important to explain this properly and warn of the consequences of getting it wrong. The option must be defensively on `WaitForStillstand` by default. The new option can only be switched on, when the actuators and motion axes are all on the same driver, and that driver also supports G4/delay (assuming #1699 merged and configured).

_Mark

janm012012:
Hi @markmaker!

Ok, seems the multi-driver option makes it again a hard nut to crack. 

First of all, this can be considered work in progress and at present only contains the code I used to perform the tests I reported. I'm willing to change everything you'd like to have.

> At one time you said the following that sounded like a switchable option:
> 
> > I've remoddelt the acutator coordination code in OpenPnP as proposed to provide an easy way to switch between WaitForStillstand and CommandStillstand and to apply CommandStillstand only where possible
> 
> Why is this necessary?

Initially I thought is would be more difficult to separate SET operations from others. On SET OpenPnP does neither check nor read the result and hence can just instruct the controller to do its duty without taking the exact time frame into account. For READ operations OpenPnP needs to make sure the controller has executed the requested action before the read will be successful. While making the code changes, it turns out to me, that all instances of "Before Actuation?" can be considered SET and hence can be executed using CommandStillstand. This assumption was still valid after executing my demonstration.

If I understand you correctly, my test case is successful because the controller used to drive axes is the same as the one operating the actuator _and_ it supports waiting for pending motion using M400 before changing the actuator. If axes and actuator would be controlled using different unrelated drivers they can not synchronize themself. In that case OpenPnP needs to provide the synchronization layer by issuing M400 to both, wait for both to respond and only then command the actuator to change. This sounds like a rather exotic case to me...

What's you're suggestion on the UI change? Shall I try to convert the boolean "Before Actuation?" state into an enum with three options in a drop-down or rather add a drop-down or checkbox to enable CommandStillstand? How shall they be named? I fear if the user is to choose here, there might be a lot of support required...

Jan

janm012012:
Hi @markmaker!

I've now change the coordination type from boolean to enum, adjusted the UI and provided an upgrade path for existing configuration.

Jan

janm012012:
I've reverted ReferenceNozzle as we discusses, that executing WaitForCompletion() before Thread.sleep() is the responsibility of the actuator coordination setup. This also removes the merge conflict with PR #1699.

janm012012:
Hi @markmaker!

After discussing the possibility to detect a controller in error state and using this to replace my drag pin interlock the modification of this PR arouses desire for even more find grade coordination control: if the drag pin is going down, CommandStillstand is needed before actuation for correct timing. The same applied for the drag pin going up. If the controller could end up in error state while the drag pin is going up, WaitForStillstand seems to be needed in order for OpenPnP to catch up. Without this coordination OpenPnP would send all commands to bring the placements to the up-looking camera for bottom vision or event further if the placement does required bottom vision. For good efficiency the WaitForStillstand coordination after actuation shall only be applied when the drag pin is going up. This is at present not configurable and would require an option for boolean actuators to specify the coordination per actuated/not actuated state. What do you think? Is it worth implementing?

Jan

markmaker:
Hi @janm012012,

the Halt condition is special in that it kills everything in the queue. In any case _nothing_ is executed after that. Only `M999` can unlock it and the queue will always be empty after that. OpenPnP might send more commands after that, but they are safely ignored. Once the `ERROR_REGEX` is received by OpenPnP it will abort the machine task. This will happen as long as `M999 `is not sent, even if the user tries to start more tasks.

So no `WaitForStillstand` is needed. 

This seemed to work on my machine, for instance after I crashed into the end switches. Not sure about other controllers/firmwares. But safety considerations really, _really_ demand this behavior.

_Mark

janm012012:
Hi @markmaker!

That sound promising! I already changed the controller to go into error state if the drag pin does not retract, but did not yet tested it...

The reason I was wondering is, that with this PR in place, there will (at least could) be much less coordination points in place as we have today. So in an extreme case without bottom vision and my drag feeders, there could be not a single coordination point in the entire job anymore (valves and drag pin only use CommandStillstand). So I was wondering how OpenPnP could keep track if the controller goes into error state and continue with the next open placements? Anyhow, this discussion might be beyond the scope of this PR.

Jan

markmaker:
> there could be not a single coordination point in the entire job anymore

Yep. 

If you don't have vacuum sensing.

> So I was wondering how OpenPnP could keep track if the controller goes into error state 

I just had a look and it is even much _stricter_ than I remembered. :fearful: 

Once it matches `ERROR_REGEX` it sets `GcodeDriver.errorResponse`, and then `GcodeDriver.bailOnError()` kills all attempts to issue new commands. The machine task is shot down by Exception.

In fact, the whole connection is doomed. The Machine need to be disconnected/connected (in the UI this even means disable/enable) in order to recover. What I said earlier about an Actuator sending `M999` is nonsense. This also means you absolutely need to re-home.

I remember why it need to be as drastic: In a multi-driver scenario, you need to make sure _all_ the drivers are on the same page. So not only the one issuing the HALT must be closed, but _all_ of them.

This is probably not what you want to hear. :weary:  There is a cost for letting the controller handle things autonomously. It's a bit like a data base transaction and rollback. All or nothing. This will only make sense, if the probability of a stuck drag pin is very small.

_Mark

janm012012:
Hi Mark!

May I finally ask for your opinion on this last PR as well? Would you support merging it (into the fresh test branch) or do you see any potential issues here?

Jan

PR #1707, opened 2024-11-20
optimize fiducial checking in job using travelling salesman

janm012012:
# Description
This PR enhances the panel and board fiducial check steps of the job processor by routing the list of fiducials to check through a travelling salesman optimization.

# Justification
After seeing the latest videos of Zdenko's machine in action, I notices, that board/panel fiducial check is not optimized yet. So I just added that...

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **on my physical machine**

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -413,14 +413,41 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
     protected class PanelFiducialCheck implements Step {
         public Step step() throws JobProcessorException {
             FiducialLocator locator = Configuration.get().getMachine().getFiducialLocator();
+            Location currentLocation;
             
-            for (PanelLocation panelLocation : job.getPanelLocations()) {
-                if (!panelLocation.isEnabled()) {
-                    continue;
-                }
-                if (!panelLocation.isCheckFiducials()) {
-                    continue;
-                }
+            // collect all panel fiducial locations
+            List<PanelLocation> panelLocations = job.getPanelLocations()
+                    .stream()
+                    .filter(pl -> { return pl.isEnabled(); })
+                    .filter(pl -> { return pl.isCheckFiducials(); })
+                    .collect(Collectors.toList());
+
+            // try to get current location as start location
+            try {
+                currentLocation = head.getDefaultCamera().getLocation();
+            }
+            catch (Exception e) {
+                currentLocation = null;
+            }
+            
+            // optimize panel fiducial using travelling salesman
+            TravellingSalesman<PanelLocation> tsm = new TravellingSalesman<>(
+                    panelLocations, 
+                    new TravellingSalesman.Locator<PanelLocation>() { 
+                        @Override
+                        public Location getLocation(PanelLocation locatable) {
+                            return locatable.getGlobalLocation();
+                        }
+                    }, 
+                    // start from current camera location
+                    currentLocation,
+                    null);
+            
+            // Solve it (using the default heuristics).
+            tsm.solve();
+            
+            // perform fudicual check in optimized order
+            for (PanelLocation panelLocation : tsm.getTravel()) {
                 fireTextStatus("Panel fiducial check on %s", panelLocation);
                 try {
                     locator.locatePlacementsHolder(panelLocation);
@@ -438,18 +465,42 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
         
         public Step step() throws JobProcessorException {
             FiducialLocator locator = Configuration.get().getMachine().getFiducialLocator();
+            Location currentLocation;
+            
+            // collect all panel fiducial locations
+            List<BoardLocation> boardLocations = job.getBoardLocations()
+                    .stream()
+                    .filter(bl -> { return bl.isEnabled(); })
+                    .filter(bl -> { return bl.isCheckFiducials(); })
+                    .filter(bl -> { return !completed.contains(bl); })
+                    .collect(Collectors.toList());
 
-            for (BoardLocation boardLocation : job.getBoardLocations()) {
-                if (!boardLocation.isEnabled()) {
-                    continue;
-                }
-                if (!boardLocation.isCheckFiducials()) {
-                    continue;
-                }
-                if (completed.contains(boardLocation)) {
-                    continue;
-                }
-                
+            // try to get current location as start location
+            try {
+                currentLocation = head.getDefaultCamera().getLocation();
+            }
+            catch (Exception e) {
+                currentLocation = null;
+            }
+            
+            // optimize panel fiducial using travelling salesman
+            TravellingSalesman<BoardLocation> tsm = new TravellingSalesman<>(
+                    boardLocations, 
+                    new TravellingSalesman.Locator<BoardLocation>() { 
+                        @Override
+                        public Location getLocation(BoardLocation locatable) {
+                            return locatable.getGlobalLocation();
+                        }
+                    }, 
+                    // start from current camera location
+                    currentLocation,
+                    null);
+            
+            // Solve it (using the default heuristics).
+            tsm.solve();
+            
+            // perform fudicual check in optimized order
+            for (BoardLocation boardLocation : tsm.getTravel()) {
                 fireTextStatus("Fiducial check for %s", boardLocation);
                 try {
                     locator.locatePlacementsHolder(boardLocation);


tobydickenson:
I have done a preliminary review. I note some comments say "panel fiducial" when it should be "board fiducial".

tonyluken:
I don't think the way you are sorting all the job's panel locations is a simple as you are treating it. Remember panels can now be "nested" (to any level) so you always want to make sure a parent panel is fiducial checked before any of its child panels are checked. The method job.getPanelLocations() returns a list of panels with each parent always occurring before its children so just performing the fiducial checks in that order ensures the parents are always done before their children.  That said, you are free to optimize the order of the children of a panel (just keep in mind that the children may also have children and so on).

Tony

janm012012:
HI @tonyluken!

Many thanks for pointing that out!

If I understand you correctly, the location and transformation of descendant panels depend on the location of their parent and hence the parent has to be checked first?

According to the code, getPanelLocations() is designed to return an "Collections.unmodifiableList". It seems that either panel locations are not part of any automated test sequence or the way I modified them does not trigger the expected "UnsupportedOperationException".

I suppose implementing a fully optimized panel check in full depth will required optimizing per nesting level. Is that correct? If so, is there an easy wait to collect the locations per nesting level? If that's difficult, could we restrict the optimization to the probably most common use case of one root panel with some boards in it? Is it easier to detect if a panel does not contain any descendant panels? (I could probably restrict the optimization using "if (getRootPanelLocation().getPanel().getDescendantPanelLocations().isEnpty()") What do you think?

Jan

tobydickenson:
I have recently been taking a close look at fiducial location precision, as part of PR #1688. On my machine I see:
 * The [precision](https://en.wikipedia.org/wiki/Accuracy_and_precision) of a fiducial position check is between 4µm and 6µm if approached from the same direction. This is remarkably good considering I have 13µm camera resolution.
 * The precision of a fiducial check reduces to 25µm if the fiducial is approached from a random direction and distance. I believe this is residual error after backlash compensation.

In order to preserve maximum precision from the fiducial check, it would be good for the "optimized" fiducial scan order to be consistent. That is, consistent for every board in a panel, and consistent from one run to the next. Is the Traveling Salesman implementation deterministic? I see there is a comment `// make this repeatable by...` but have not reviewed that in detail.

janm012012:
Hi all!

Simulated Annealing is one methods the Travelling Salesman can be used with to vary the order of locations to visit. It is known to work well for a large set of locations with a high probability for a good solution. For small sets other methods might converge faster or always provide the best solution. For us the result matters and for our small set of locations with likely a good significance in different travels the result shall always be the best solution. (And we actually don't care on which way Simulated Snnealing guided the Travelling Salesman towards the result because that's what you gets by seeding the random number generator Simulated Annealing is based on. As seeding always generates the same sequence of "random" numbers, you'll always take the same path to the same potentially imperfect result. That's why seeding is good for debugging but not for production.) However, depending on from where you start, the result might be different and if you see a difference depending on from which direct you approach, then I'm afraid there will be almost always a difference, even without this PR.
At the end the question is how much a placement location jitters with fiducial based corrections applied. I'd assume, that due to the fact that always more then one fiducial is in effect, the result of a single fiducial will not be significant. And finally a physical argument: for a 0.5mm pitch part, a displacement of 0.1 does not matter, IMHO.

Jan

janm012012:
Hi @markmaker!

That's what I'm saying. 25µm is nothing to worry about, IMHO. If Toby wont's to fight it anyhow, backlash calibration with something like OneSidedPosition might help. This would apply to both fiducial check and place action. Anyhow, I don't see any reason to make any changes.

Never the less, Tony's objection requires more attention.

Jan

tonyluken:
> If I understand you correctly, the location and transformation of descendant panels depend on the location of their parent and hence the parent has to be checked first?

Yes, if done in the wrong order, the transformation on all the children will get reset to the identity transform so only the parent's transformation will be used - that's still better than no correction but not as good as doing it in parent then child order.

> I suppose implementing a fully optimized panel check in full depth will required optimizing per nesting level. Is that correct?

Yes, starting at the root level of the job, you would optimize the order of the panels and boards at that level, and then for each of the panels at that level, optimize the order of its children (which are possibly panels whose order of children will need to be optimized and so on). Sounds like a recursive routine that would be first called with the job's root panel and then it calls itself for each of the child panels. Note that this won't be exactly optimum since the last fiducial checked on the last child of the panel won't be the same as the last fiducial of the panel (which was assumed when computing the order of the parent and its siblings). But that said, it should still be pretty good. Also note that if this is implemented, the separate board fiducial check after the panel fiducial check will become unnecessary as they will have all been done.

> If so, is there an easy wait to collect the locations per nesting level?

That is relatively easy. For a given PanelLocation you would call its getPanel().getChildren() method which will return a list of all the panel's direct children. Then iterate through that list, each element will be either instance of PanelLocation or BoardLocation.

Tony

janm012012:
Hi @tonyluken!

Many thank for your hints! I've tried to use them to improve my implementation, unfortunately without success yet: the collection process does not find all boards/panels. On one of my tests with just a board, this is ignored. Would you please have a look why that's missing?

To preserve the nesting order while still allowing a certain degree of optimization, I've extended PlacementsHolderLocations by a nesting level integer, representing the nesting level it was collected with. This level is then evaluated by the travelling salesman to reject all paths that would violate the ordering.

In addition I had to noticed, that PlacementsHolderLocation references all fidicuals of a panel/board and that ReferenceFiducialLocator::locatePlacementsHolder() always assumes that the origin of the panel/board is the next location to go to for its optimization. The first is likely not that easy to break up because of the math that's applied per board/panel while the second is easy by extending the interface. This would allow the job processor to specify the next boards/panels center (center of gravity between all fiducials) as end/next location.

Please have a look. Thank You!

Jan

janm012012:
Hi @tonyluken!

I how may have found a solutions: seems that I did not took isEnabled() and .isCheckFiducials() for panels correctly into account.

Em I assuming correctly, that with this PR Job::getBoardLocations() and alike could be considered deprecated?

Jan

tonyluken:
Hi @janm012012,

I see several issues here:

1. In the collectAllBoardLocations method, the following is not correct:

                // get all descendants of the rootPanel
                List<PlacementsHolderLocation<?>> children = ((Panel)panelLocation.getPlacementsHolder()).getDescendants();

                // loop over all children and collect their descendants
                int nextNestingLevel = nestingLevel +1;
                locations.addAll(children.stream()
                        .map(pl -> { return collectAllBoardLocations(pl, nextNestingLevel); })
                        .flatMap(List::stream)
                        .collect(Collectors.toList()));

The correct way is actually much simpler:

                // get all children of the panel
                List<PlacementsHolderLocation<?>> children = panelLocation.getPanel().getChildren();
                for (PlacementsHolderLocation<?> child : children) {
                   locations.addAll(collectAllBoardLocations(child));
                }


Also, there is no need to collect the children of a panel if it is not enabled as all of its descendants will be disabled.

2. Your override of getTravellingDistance will not force the travelling salesman optimizer to obey the nesting level since getTravellingDistance is never actually called during the optimization processing (in the TravelingSalesman's simulateAnnealing method).

3. Note that the nesting level numbering you've created doesn't really help the problem. As shown in the diagram below, it is valid to jump either up or down in nesting level. The check really needs to make sure that each element in the list is not a descendant of any elements later in the list.
PanelNestingLevels

janm012012:
Hi @tonyluken!

Many thanks for your comments!

I've fixed the collection process as you suggested. It now works in all my tests and the test code again.

Yes, my implementation overriding getTravellingDistance() was wrong. This method is only used at start and for debugging. I would  have to override getSwapDistance(), which looks quite difficult to me at present because it's private and requires access to travelInput, which is also not accessible, to read parent/nesting information to reject swap/twist requests violating the parent-child relation. Before making all this changes, I'd like to ask for your opinion if its worth to make this modification. At present I consider it advantageous to unify board and panel fiducial checking into a single step. Therefor I'd keep this PR even without the optimization.

I also (temporarily) replaced the nesting level by a pointer to the parent to allow for a better parent-child relation checking as I have found instances where the nesting level I had was not good enough.

What ever you think would be a good way to go (incl. closing this PR) all code is available and I'd be willing to continue. (@markmaker: do you have an opinion as well?)

One final question: the original code uses a different behavior for panel and board fiducial checks. For panels all fiducials are checked as a single step while for boards the check is executed one board at a time. At present this PR replicates the one board/panel at a time behavior. Is that ok or do you prefer the all in one step version? (The later requires in little less and simpler code, making it a tiny little preferable of the other, at least to me)

Jan

tonyluken:
I think the simplest thing to do is just collect all the panels and boards at the first level, sort them, and perform the fiducial checks for that level; and then do the collecting, sorting, and fiducial checks for the next level; and repeat until there are no more levels. That will obviously not necessarily give the best overall routing, but it should still be pretty good in the vast majority of the cases (typically, I'd expect most cases to only have one or two nesting levels). The advantage this has is that it preserves the nesting levels and there is no need to modify the travelling salesman algorithm.

> the original code uses a different behavior for panel and board fiducial checks.

I don't believe that is really true. The original code first does all the panels one at a time (and in an order that ensures a panel is fiducial checked before any of its descendant panels). Once all the panels are checked, then the boards are checked one at a time (their order doesn't matter since a board never has any descendants).

janm012012:
Hi @tonyluken!

Many thanks for your quick response!

> I think the simplest thing to do is just collect all the panels and boards at the first level, sort them, and perform the fiducial checks for that level; and then do the collecting, sorting, and fiducial checks for the next level; and repeat until there are no more levels.

Ok, I'll modify my previous version to implement that.

> > the original code uses a different behavior for panel and board fiducial checks.
> 
> I don't believe that is really true. The original code first does all the panels one at a time (and in an order that ensures a panel is fiducial checked before any of its descendant panels). Once all the panels are checked, then the boards are checked one at a time (their order doesn't matter since a board never has any descendants).

Please have a look at the code, thats currently in use:

```
    protected class PanelFiducialCheck implements Step {
        public Step step() throws JobProcessorException {
            FiducialLocator locator = Configuration.get().getMachine().getFiducialLocator();
            
            for (PanelLocation panelLocation : job.getPanelLocations()) {
                if (!panelLocation.isEnabled()) {
                    continue;
                }
                if (!panelLocation.isCheckFiducials()) {
                    continue;
                }
                fireTextStatus("Panel fiducial check on %s", panelLocation);
                try {
                    locator.locatePlacementsHolder(panelLocation);
                }
                catch (Exception e) {
                    throw new JobProcessorException(panelLocation, e);
                }
            }
            return new BoardLocationFiducialCheck();
        }
    }
 ```
and
```  
    protected class BoardLocationFiducialCheck implements Step {
        protected Set<BoardLocation> completed = new HashSet<>();
        
        public Step step() throws JobProcessorException {
            FiducialLocator locator = Configuration.get().getMachine().getFiducialLocator();

            for (BoardLocation boardLocation : job.getBoardLocations()) {
                if (!boardLocation.isEnabled()) {
                    continue;
                }
                if (!boardLocation.isCheckFiducials()) {
                    continue;
                }
                if (completed.contains(boardLocation)) {
                    continue;
                }
                
                fireTextStatus("Fiducial check for %s", boardLocation);
                try {
                    locator.locatePlacementsHolder(boardLocation);
                }
                catch (Exception e) {
                    throw new JobProcessorException(boardLocation, e);
                }
                
                completed.add(boardLocation);
                return this;
            }
            
            return new Plan();
        }
    }
```

PanelFiducialCheck executes the entire list before returning while BoardFiducialCheck returns after each board. The later allows to step each board but requires `completed` to preserve the state of checked boards. I see pros and cons in each but I don't know how you'd like to see/have/prefer it. That's why I ask.

Jan

tonyluken:
Ah! Yes, I missed the `return this;` inside the loop. Yeah, the panel check probably should do the same thing.

markmaker:
@tonyluken, a question:

Is  it really necessary to do the detections hierarchically? 

First, I'm assuming that in _practice_, all fiducials in certain environment are made equal. I don't have much practical experience, so correct me if I'm wrong! :laughing: 

Fiducials have a certain size and a keep-out area around them, that by implication gives the maximum allowable offset from nominal to real location in detection.
 
If that is true, then fiducials are allowed to be offset by the same maximum distances from their nominal locations, _regardless_ of their nesting level. In panels, if all the outermost frame fiducials have the largest allowable offsets, then logic dictates that PCB fiducials inside the panel cannot have larger offsets. In fact it is likely that their offsets are smaller, i.e. there is likely a rotation over the whole rigid panel, that is limited by the outer frame.

As a consequence, one could just walk all fiducials via one _globally_ optimized Travelling Salesman route. Detect and store their _real_ locations in some map. Only then go through the panels/PCBs in the required nested manner, looking up real locations from the map instead of physically detecting them with only _locally_ optimized route.

Or am I missing something?

_Mark

janm012012:
Hi  @tonyluken!

I've now implemented the per-nesting-level optimization including the step wise processing.

Now there are two open questions for me:

1. The optimization uses PlacementsHolderLocation::getGlobalLocation() which to my understanding returns the location the board/panel is located but actually has only little significance with respect to the fiducial locations to visit. Shall I modify that to use the center between all fiducial locations? (This would require modifying spi/FiducialLocator.java) 
2. The ReferenceFiducialLocator locally optimizes the path between all fiducials to visit under the assumption that once the check has finished, the head will move to getGlobalLocation() (endLocation of TravellingSalesman). That's what happens if the fiducial check is initialized by UI but within a job it's either the next board/panel to process or unknown. Shall I change that? (would again require modifying spi/FiducialLocator.java to accept the endLocation as input or without modifying spi I could remove the endLocation requirement)

Jan

janm012012:
Hi @markmaker!

I'm sorry to jump in, even if I'm not Tony.

> Is it really necessary to do the detections hierarchically?

In one of this early reply's Tony explained:

> > If I understand you correctly, the location and transformation of descendant panels depend on the location of their parent and hence the parent has to be checked first?
> 
> Yes, if done in the wrong order, the transformation on all the children will get reset to the identity transform so only the parent's transformation will be used - that's still better than no correction but not as good as doing it in parent then child order.

So the actual reason is that the child's transformation is a function it's parent. I understand it that way, that a panel of boards gets it's fiducials checked first to identify the overall location of the panel itself and then each board gets checks correcting local inconsistencies. If done the other way around no local corrections would be applied.

Side node: I recently tried to change "Max. Distance" for a specific fiducial type, but failed: it seems like a single global parameter at present.

Jan

janm012012:
Ok, I think I've a solution. Here is my proposal:

At present `ReferenceFiducialLocator::locatePlacementsHolder` is called with a single `PlacementsHolderLocation` which is used to extract it's fiducials, and then visit each and in parallel calculate it's expected location. Finally the transformation is derived based on measured and expected locations.

To extend that as Mark proposed, `locatePlacementsHolder` has to accept a list of `PlacementsHolderLocation`. It would collect the fiducials of all of them and visit each taking an optimized route just measuring it's location. As Mark pointed out, the uncorrected expected location shall be close enough in general. Then, in a second loop, the list of `PlacementsHolderLocation` would be iterated again in the original unoptimized order calculating the expected location of the fiducials and based on that and the previously measured locations derive the correction. By executing the loop in the original order of `PlacemetnsHolderLocation` and only calculating the expected locations after all previous transformations have been derived the dependency of the child's locations on it's parent's transformation shall be respected.

This would certainly require some extra layering to keep a 'PlacementsHolderLocation' grouped with it's fiducials and their measured locations. But that would be grunt work and result in a perfectly efficient fiducial check - as intended. However, we would loose the ability to step thought each panel/board separated. FiducialCheck would become a single uninterruptible step in the JobProcessor.

@tonyluken: what do you think, would that work?

Jan

tobydickenson:
> Is it really necessary to do the detections hierarchically?
> First, I'm assuming that in _practice_, all fiducials in certain environment are made equal.

Hi Mark,

It is not uncommon to have different types of fiducial on the same board.
1. A panel fiducial can have a large keep-out area, where there is lots of slack space in the panel margin.
2. A board fiducial might be smaller, with a smaller keep-out, if the board it tight for space.

This situation benefits from hierarchical detection and transform. The panel transform helps with board fiducial detection, because the accurate starting location helps the fiducial locator avoid getting distracted by nearby vias etc, or other round features of the right diameter. Ideally the board fiducials would have a large enough keep-out that they could be detected without the help from the panel transform, but that is not always possible.

janm012012:
Hi Toby!

Yes, that's a valid objection and use case. However, I see two concerns

1. IIRC there is no option (yet) to configure the detection range per fiducial (part). `Max. Distance` is a global parameter of  ReferenceFiducialLocator.
2. To my experience the ReferenceFiducialLocator seems to sort detected features by significance rather then distance. On my HAL version of the [OpenPnP Demo Board v2](https://github.com/openpnp/openpnp/tree/develop/samples/Demo%20Board) it tends to favors the upper cycle of the "8" of "F8" on the (white) silk screen over the fiducial even if perfectly spot on. (Yes, I know that HAL finish is difficult...)

Jan

tobydickenson:
>     1. IIRC there is no option (yet) to configure the detection range per fiducial (part). `Max. Distance` is a global parameter of  ReferenceFiducialLocator.

Yet.

Having that as an option would tidy up some of my pipelines, so I will look at adding that.

tonyluken:
@markmaker,

> Is it really necessary to do the detections hierarchically?

Technically no, assuming the fiducials can be reliably detected. Also, note it's not just a matter of using the measured fiducial positions in the correct order to compute the transforms. As the higher-level transforms are computed, they need to be "unapplied" to the measurements used to compute the transforms for the lower-level boards/panels. If someone wants to expend the effort to do it, I don't see any reason it couldn't be done. However, as @tobydickenson pointed out, doing the detections (and transformation calculations) hierarchically, does have the advantage of making the lower-level fiducial detections easier since they already have a correction applied to them. That means the camera is better centered on them and therefore less likely to mis-detect some other feature. That alone seems to be a good reason to do it in a hierarchical order.

Tony

janm012012:
Hi all!

I think I've found a good solutions that covers all requirements we've discussed: I enhanced the `ReferenceFiducialLocator` to process a list of `PlacementsHolderLocation` in a optimized way by first visiting all fiducial locations and then calculating all affine transformations respecting the input order. This second step also recalculates the expected location taking all parent transformations which might have been updated in a previous step are taken into account. This shall provide the same results as with the current method.

In addition I've added the `fiducialLevel` parameter to the `ReferencePnpJobProcesssor` which specifies how many nesting levels of fiducials to process one by one before processing all remaining fiducials at once. This shall support Toby's use case of wider panel fiducials and very narrow board fiducials while still providing a better overall performance. If set to 0 all fiducials will be checked at once (as Mark suggested). If set to a large number fiducials will be checked nesting level by nesting level.

@tonyluken: please review my changes to the ReferenceFiducialLocator.

Jan

tonyluken:
I haven't spent a lot of time on this, but I did run some quick tests with some surprising results. I used this [Test Panel](https://user-images.githubusercontent.com/50550971/210090575-b79e87be-54bc-4c34-99ae-9af6c4098f24.png) which has some deliberate scaling and shearing errors.

I used the latest test branch to get a baseline, and it took 77 seconds to check and compute the affine transforms for the outer-most panel, the four subpanels, and the 12 boards.

Then with this PR and with fiducial-level at its default value of zero, the same test took about 146 seconds to visit all the fiducials - almost twice as long! And then to top it off, it stopped with an exception, "Fiducial locator results are invalid because: the x shearing = 0.05177 which is outside the expected range of [-0.05000, 0.05000]". I am almost certain the exception is due to a misdetection on one or more of the fiducials, but I can't say which one(s) because there is no instrumentation to say which affine transform is being computed when the exception occurred. It should be relatively easy to add some more info to the exception so that it would be easier for the operator to debug the problem.

Now the question is, why did the optimized route take almost twice as long? Well, the latest test branch performs the fiducial checks in hierarchical order, so only the outermost panel's fiducials have no correction applied when positioning the camera over them. That usually means the camera needs to take two or three shots of the fiducials before it is centered. For all the inner subpanels and boards, there is already a correction being applied so for those fiducials, the camera nearly always falls close enough that it doesn't require more than the first shot.  With this PR, all the fiducial positions are collected first which means most of them require two or three shots with the camera. All that extra moving of the camera for each fiducial is using way more time than is being saved by using an optimized route.

I did try setting fiducial-level to 10 and did get better results, it took about 72 seconds and all checks were successful. Finally, I tried setting fiducial-level to 1 and got even better results, it took about 64 seconds and all checks were successful. I think the default value of fiducial-level probably should be changed to 1.

Tony

janm012012:
Hi Tony!

Many thanks for testing! Sounds more or less as expected. I agree, that the default `fiducialLevel` shall be 1 and that logging and exception has to be extended to indicate the board/panel in question. I've added just added that.

May I assume from your tests, that the code is functional wise ok preserving the fiducial check results from the current test version?

If you think there is anything else that should be changed/enhanced - in respect to this PR - please let me know.

Jan

tonyluken:
Please change the exception info to use placementsHolderLocation.getUniqueId() instead of placementsHolderLocation.getId() so the operator can see the full "path" to the board/panel that is having the problem.

tonyluken:
> Anything else?

Since you are already changing ReferenceFiducialLocator.java, you might as well make the same change to use the getUniqueId() method on the following lines as well: 198, 201, 246, 249, 253, 260, 264, 274, and 466.

Other than that, I think this is good to go.

tobydickenson:
I tested this on a real machine with my most fiducial-heavy panel, running a few test placements. In summary it all looks good.
 * This panel has 4 fiducials were checked first as expected.
 * It has 9 boards x 5 fiducials per board. Each board has two which use the parallax vision method. One of these has some nearby vision distractions so definitely needed to have the panel placements checked first. This all worked as expected.
 * The start-of-job fiducial scan is reduced from 159 (sample of 1) to 152±1 seconds (sample of 6)
 * I checked accuracy: standard deviation of the transformed location of a placement was as expected (sample of 6)
 * This is a board which I place using my #1688 crude script-based local fiducial hack. That method appears to work unchanged with this patch.

janm012012:
@tonyluken: all nine changed. Many thanks for reviewing!

@tobydickenson: many thanks for testing! Sounds it works good for you, although I would have hoped for greater advantages..

Jan

tonyluken (approved):
All looks good!

PR #1708, opened 2024-11-21
Optimize selection of the placement for the N+1's nozzle

janm012012:
# Description
This PR changes the SimplePnpJobPlanner to choose the placement for the N+1's nozzle such, that the distance for the head to travel is minimized.
For this purpose it reuses the pickLocator and other code that is already used for optimization.

# Justification
After all the other optimizations to the JobProcessor, it makes sense to also optimize the N+1's placement. This implementation respects the sorting strategy applied by the JobProcessor by choosing the next compatible placement for the first nozzle and the next compatible nozzle tip for any nozzle. It's just the placement for the N+1's nozzle (with and without nozzle tip change) that is chosen based on the distance to the already planned placements.

# Instructions for Use
The optimization is always active.

# Developers:
Due to the fact, that the SimplePnPJobPlanner is declared static, I had to apply static to the placeLocator as well and I was not able to check for optimizeForMultipleNozzles to disable this optimization. Please advice.
The method calcCenterLocation() had to be relocated to reuse it.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **using simulation and physical machine**

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -105,9 +105,9 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
 
     protected Head head;
 
-    protected Locator pickLocator;
-    protected Locator alignLocator;
-    protected Locator placeLocator;
+    protected static Locator pickLocator;
+    protected static Locator alignLocator;
+    protected static Locator placeLocator;
     
     protected List<JobPlacement> jobPlacements = new ArrayList<>();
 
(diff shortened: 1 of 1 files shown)


markmaker:
Hi @janm012012,

I don't understand. This flies in the face of your #1688. 😵

You can either optimize pick (feeders) colocation, or placement colocation, not both. As feeders are typically spaced much farther apart (all around the machine), the former is more promising (unless you have humongous panels, perhaps).

Even if this was a valid alternative, you would have to do it via the placements "order" like in #1688, i.e. ordering the placements overall, perhaps with a Travelling Salesman, ideally by nozzle compatibility group, not greedily inside the cycle, otherwise the "order" becomes completely meaningless, and that's not what users expect.

Or am I missing something? 🤔

_Mark

janm012012:
Hi @markmaker!

You're correct, I'll have to link it to PnpJobPlanner.Strategy such that a) pick _and_ place locations are taken into account or b) the feeder is kept constant and only the place location is used as selection criteria or c) no further selection beyond just taking the next placement takes place.

Jan

markmaker:
Yes b), but then this would best be added to #1688, right? Like a "sub-criteria".

_Mark

janm012012:
Hi @markmaker!

I've enhanced the PR as suggested. Now the planner strategy is used to keep the feeder of the first part (which would have been chosen without this PR) and the sorting always takes the distance between pick and place location into account.

Concerning PR #1658: the planning/sorting there does not take other nozzles actively into account at all. It's more by luck if an other nozzles can take the same part and hence benefits from the sort-by-feeder strategy. This is the first time, that there is any optimization for the second nozzle. And I think there is a benefit if other nozzles can not take the same part because then a placement with short distances to the others is selected.

Btw: while working on the fiducial PR #1707 I think I found a way to implement a full optimizer. Once that's in place, it will likely supersede this optimization.

Jan

PR #1710, opened 2024-11-30
Shift key will decrease current distance movement by 100

tmandys:
# Description
Jog movement distance is 100 times finer when SHIFT key pressed together with main key or mouse button. So e.g. when distance slider is 100 then movement is 2 step finer, i.e. 1

# Justification
general usability improvement

# Instructions for Use
Check jog movement and press SHIFT key to see difference

# Implementation Details
1. manually tested with PnP machine
2. yes
3. no
4. yes mvn test passed

### DIFF
--- a/src/main/java/org/openpnp/gui/JogControlsPanel.java
+++ b/src/main/java/org/openpnp/gui/JogControlsPanel.java
@@ -145,11 +145,15 @@ public class JogControlsPanel extends JPanel {
     }
 
     public double getJogIncrement() {
+        int val = sliderIncrements.getValue();
+        if (MainFrame.get().getShiftDown()) {
+            val = Math.max(1, val - 2);   // finer movement by 2 levels
+        }
         if (configuration.getSystemUnits() == LengthUnit.Millimeters) {
-            return 0.01 * Math.pow(10, sliderIncrements.getValue() - 1);
+            return 0.01 * Math.pow(10, val - 1);
         }
         else if (configuration.getSystemUnits() == LengthUnit.Inches) {
-            return 0.001 * Math.pow(10, sliderIncrements.getValue() - 1);
+            return 0.001 * Math.pow(10, val - 1);
         }
         else {
             throw new Error(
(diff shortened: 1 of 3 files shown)


tobydickenson:
I generally like a rich keyboard interface, am a regular user through a programmable keypad. I can see why Ctrl+Shift+Arrow makes sense in isolation. But I'm not convinced by this change

1. Having the shift key affect an on-screen mouse button is an unusual interaction.
2. Openpnp already has keyboard control of jog distance, using keystrokes Ctrl-Shift-F1 to F5. It would be good to explain how this change complements the existing feature.

tmandys:
Using a triple keystrokes when I'm focused on setting something in machine is useless. Also camera does not help when position is beyond vision range. I found myself (Shift+)Arrow as very effective. Solution is fully configurable keyboard interface which could solve issue with unlogical hot keys, e.g. for Z-axis and missing ones, e.g. for Special actions. Jog keypad is nice but there is the same issue I would like the fine movement with Shift as well.

- what is collision with Shift+Mouse and how is mouse combined with arrows in one time ?

tobydickenson:
> Using a triple keystrokes when I'm focused on setting something in machine is useless. Also camera does not help when position is beyond vision range. I found myself (Shift+)Arrow as very effective. 

That makes sense. Shift+Arrow works with one hand, when you have the other hand and your eyes on the machine. 🙂

>     * what is collision with Shift+Mouse and how is mouse combined with arrows in one time ?

I meant that it is an unusual user-interface feature.
* Having shift affect the arrow key makes sense, and is very ordinary
* Having the on-screen mouse-clickable arrow button work the same as an arrow key makes sense
* But I cant think of another case where an on-screen mouse-clickable button is affected by the shift key. Anyway its not a problem, just an observation.

> Solution is fully configurable keyboard interface 

Agreed

----

This PR looks good to me :+1:

tobydickenson:
I propose merging this, plus PR #1690 which also relates to the keyboard interface, into the v2.3 test branch. @tmandys do you agree? Any last comments? Once this is merged, please can you update the wiki and announce it on the mailing list.

tmandys:
Sure. Where shall I update wiki, to find the best place where to document or as a change log ?

tobydickenson:
Thank you. Wiki here: https://github.com/openpnp/openpnp/wiki/User-Manual#keyboard-shortcuts

I will update the change log after the next batch of merges.

PR #1711, opened 2024-11-30
Special fuctions to clear tip and part on nozzle

tmandys:
# Description
During PnP process on "non perfect machine" with "non perfect user" many time real state differs from OpenPnP state.
The first case is when OpenPnP think loaded nozzle or part on the tip but in reality is unloaded because of various reason.
Part may be lost, wrongly detected, grabbed by user or so. Nozzle tip may be lost when loading, typically happens when tuning nozzle changer. Fake loaded nozzle is rather very danger issue as OpenPnp will try to unload nozzle and it will likely crash to the station.
Part may be discarded to bin but it will likely move accross machine just to change software state.

So new jog special command were implemented to clear part / nozzle state. Also appears on head/nozzle/tips table in machine config

Note: 

- Clear part stop also vacuum
- no script action is triggered
- I have plenty of issues with real vs. openpnp world when PnP, this is first simple improvement. As they are real time issues it is not easy to reproduce on demand ;-)

# Justification
general usability improvement 

# Instructions for Use
Load nozzle/part and check Clear* command. The OpenPnP will change internal state (and for part stop vacuum).

# Implementation Details
1. manually on machine
2. yes
3. yes, there are new functions exposed because supposed private state manipulation was not possible with current interface. But it is backward compatible, see diff
4. passed


tobydickenson:
Hi tmandys,

I am puzzled by this PR!

> The first case is when OpenPnP think loaded nozzle or part on the tip but in reality is unloaded because of various reason. Part may be lost, wrongly detected, grabbed by user or so. 

How does this feature differ from the existing "discard" feature?

It saves moving the head over to the discard bin. But that is a small overhead, and is a sensible precaution against having parts dropped at random locations on the machine if there is something on the nozzle even when the operator thinks there is not.

> Nozzle tip may be lost when loading, typically happens when tuning nozzle changer. Fake loaded nozzle is rather very danger issue as OpenPnp will try to unload nozzle and it will likely crash to the station. Part may be discarded to bin but it will likely move accross machine just to change software state.

I have been in that situation, but dont see how this would help.

If openpnp and reality disagree about what tip is loaded, there are currently several methods available to fix the discrepancy:
1. A: Manually remove a tip, then B: Manually fit a new tip. This is changing reality to match what openpnp thinks.
2. Disable the automatic changer then trigger a tip change. But don't actually touch any tips when instructed by openpnp. This is changing the openpnp state to match reality.

Those two methods are a normal use of existing general-purpose features, usable by everyone today.

A "clear tip" feature is special-purpose, and it only solves half the problem. It removes the wrong tip from openpnp state, but it doesnt set the *right* tip in openpnp state.

These both seem like misfeatures. What am I missing?

tmandys:
Current tip status is rather bugged in code plus unreliable feeders, argh!. There are some not easily reproducible circumstances when real status does not correspond to code state. I even had state when OpenPnp complained about Discard and only Stop was solution. I suspect it is some unproper exception handling when internal state is not reverted correctly. And Discard itself is a pain when you need just change state to "empty" and wait dozens of times during job to remove non-existing part on tip to discard bin to proceed further. Actually currently I mostly use "Recycle" as it also set "Skip next feed" if possible. But it is not possible when both tips of dual nozzle need to be cleared.

It is definitely not about "dropping part on random location" as I have nothing to drop ;-) When you have reliable feeders you are lucky and do not know what I am complaining about ;-)

Clear nozzle tip I nedded when tuning nozzle station and shuffled with tips. Also when I remove manually tips and forget where they shoud be. OpenPnp state survives restart. So often Pnp wanted to drop tip on loaded position and nozzle crashed.

I plan to implement visual Planner table and status widget to see what PnP is doing and have a chance to adjust state. Also a statistics and timing can be assigned to steps already done. Something like user friendly debugger+profiler in common IDE. This should solve many problems but it will take plenty of mandays. So current clear nozzle/tip is quick workaround.

janm012012:
Hi tmandys, hi Toby!

I'm reluctant concerning this PR too.

As I don't have automatic nozzle tip changes, I have no experience with its possible short comings. With manual nozzle tip changing OpenPnP and reality can be easily brought back to sync. Therefore, I'd say if `Clear tip` is a requested features, why not add it.

Concerning the pick/vision issues: vacuum part on detection has to be configured and as such may fail. IMHO `Clear part` shall not be a feature to work around that. I'm running without vacuum part on detection and hence face the issue that vision part size checking fails. This always requires to verify the machine state and fix the causal issue, which is usually a feeder issue. Finally I usually move the tool to the pick location to pick the part again or perform a feed and pick to pick a new part (both using the appropriate buttons on the feeders toolbar). I never made use for the discard feature. Therefore I don't see a use case for `Clear part` yet. 

Jan

tmandys:
Jan, how you can (re)pick part again when sw status is "tip has loaded part" ? Your case is exactly what I _Clear part_ use for. Change tip status to empty, manually check feeder and repick (often with _Skip next feed_) using toolbar action. -T.

janm012012:
There is no such error and nothing that prevents me from my workflow: if cover tape was not peeled off, I fix that and just move the nozzle to the pick location. In any other case I just executed a feed/pick cycle. Both is easily accessible from the feeders toolbar. Yes, OpenPnP thinks there is something on the nozzle tip, but that does not generate any concerns. Even the rotation is completely find to immediately continue the job.

PR #1712, opened 2024-11-30
Check if nozzle tip is empty, i.e. no part, when feed command is trig…

tmandys:
# Description
Do not allow feed on nozzle tip when part is already loaded on tip. 
Typically: 
When PnP something wrong happens in multi nozzle scenario, e.g. BCAM part not detected on Nozzle1 but Nozzle2 has been 
meanwhile loaded then is no option to retry step in job as next step succeeded. But it is still possible manully put part on tip, which is tricky when orientation is important or part is too small, or trigger manually feed command on Feeder/Part panel. This works (not sure about orientation) but it does blindly feeds part on selected nozzle even is correctly loaded. Usually I forget to double check what nozzle is selected above jog in combo box and it breaks Nozzle2 as well.

# Justification
UI improvement

# Instructions for Use
Load part on nozzle and try to run feed command on Part/Feeder panel

# Implementation Details
1. manual test with machine
2. yes
3. no
4. yes

### DIFF
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -600,7 +600,7 @@ public class FeedersPanel extends JPanel implements WizardContainer {
             UiUtils.submitUiMachineTask(() -> {
                 Feeder feeder = getSelection();
                 // Do the feed and get the nozzle that would be used for the subsequent pick. 
-                Nozzle nozzle = feedFeeder(feeder);
+                Nozzle nozzle = feedFeeder(feeder, true);
             });
         }
     };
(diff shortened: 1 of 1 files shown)


tobydickenson:
The title of the pull request is "Check if nozzle tip is empty when feeding" but a more accurate description of the implementation would be "Check if nozzle tip is empty when picking".

Your implemented behavior allows feeding to occur at any time, but checks that the nozzle tip is empty before picking. That behavior makes sense to me.

Would it not be better to avoid adding the extra boolean parameter to `feedFeeder`? Checking for an empty nozzle at the top of `pickFeeder` would be simpler, and more cohesive.

tmandys:
@tobydickenson Ok, implemented suggested simplification

tobydickenson:
This looks good. I propose to merge this later this week.

I strongly suspect this patch would have saved me from a couple of boards needing rework today!

janm012012:
Hi tmandys, hi Toby!

As I've tried to explain in [PR #1711](https://github.com/openpnp/openpnp/pull/1711#issuecomment-2717459657), I usually pick a new part if bottom vision indicates a part size problem, which actually is a pick problem. If you'd merge a strict don't-pick-if-OpenPnP-thinks-there-is-a-part-on-the-nozzle-tip behavior, this would void my workflow. May I ask you to a) try to find out if it would be possible to fix the root cause that the wrong nozzle is selected and b) consider adding a popup that warns that a pick is requested on a nozzle that is considered to already carry a part. This popup would prevent unintended picks while still supporting my workflow. Thank You!

Jan

tmandys:
Jan, #1711 solves it in third way ;-) It is more general. a) unfortunately seems code is not robust to satisfy always correct status. I would require extensive unit tests to simulate plenty of possible error cases b) yes, seems logic if yes/no/cancel  dialog is raised instead exception. But not sure if user interaction is logical in such place of code. -T.

tobydickenson:
> consider adding a popup that warns that a pick is requested on a nozzle that is considered to already carry a part.

That sounds fine to me. Could this be a three-way choice, with a discard-then-pick option? This is useful to be sure the nozzle is clean.

> But not sure if user interaction is logical in such place of code

I think this is fine, noting that this only applies to the manual pick button on the Part or Feeder tabs.

janm012012:
I think you'd have to throw an exception and process it. You may have a look how manual nozzle tip changing is done today. It swallows unload and load exceptions into one and adds a continuation to automatically resume the job. You can very likely reuse it for your purpose.

tobydickenson:
This was superseded by #1870

PR #1713, opened 2024-11-30
Strip feeder parallax

tobydickenson:
# Description

ReferenceStripFeeder uses vision to determine the location of tape sprocket holes. This is no problem for tapes made of white paper or black plastic, but with transparent tape it can be difficult to see the holes. This patch uses the same parallax technique as is implemented for fiducial vision.

This approach was (briefly) discussed on [google groups](https://groups.google.com/d/msgid/openpnp/c89261c7-4e8e-48d4-8838-885218e313d2%40makr.zone?utm_medium=email&utm_source=footer)

# Justification
This benefits all users of ReferenceStripFeeder with transparent tape.

# Instructions for Use

(instructions inspired by [the equivalent for fiducials)](https://github.com/openpnp/openpnp/wiki/Fiducial-Locator#parallax-operating-principle)

As an option, ReferenceStripFeeder allows for an alternative "Parallax" method for detecting tape sprocket holes. With transparent tape it can be difficult to see the holes. Elaborate lighting systems, such as co-axial (half-way mirror) lights can overcome this problem. If using more conventional ring lights, observing the sprocket holes from a parallax view-point will allow the tape to reflect the bright diffuser of the camera light, and the sprocket hole becomes visible.

If the *Parallax Diameter* field is set to zero, then the sprocket hole is observed from directly above its expected location. This gives the best accuracy and performance but only works with opaque tapes. A non-zero value allows the hole to be observed from the off-center viewpoint. The distance from camera to hole is half the Parallax Diameter. The *Parallax Angle* is the angle of the view-point offset.

I will add the above instructions to the wiki.

# Implementation Notes

## 1. Consistency

I have added the Parallax Diameter and Parallax Angle fields with exactly the same semantics, API, and GUI as for fiducial vision. Consistency is good.

## 2. Branch Baseline

This patch builds on top of my PR #1662 . Both are adding GUI fields to ReferenceStripFeeder wizard, and java swing makes it difficult to create two independent patches. So please consider this a provisional PR until that PR is reviewed. Commits in September are PR #1662. The one commit in November 8aae2a49 implements this parallax feature.

I can separate them if needed, but hope that both patches can be reviewed and approved at the same time.

# Implementation Details
    * In the simulated machine
    * On the real machine with a production run covering various parallax and non-parallax feeders. white tape, black tape, and transparent tape.
    * The parallax technique uses two vision checks, from opposite sides, to cancel out any x/y parallax error due to z error from the off-center viewpoint. My worst case for this is a dome led, where the nozzle tip drops down over the dome to pick at a Z significantly lower than the sprocket holes. I tested this part on the real machine.
4. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? *yes*
5. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. *not changed*
6. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. *tests pass*


tobydickenson:
Some practical observations:
* I have never had much luck with the stock ReferenceStripFeeder vision pipeline.  All my feeders are using a vision pipeline equivalent to the stock fiducial pipeline, using DetectCircularSymmetry. After c078e05 the parallax feature now works with both.
* In google groups I shared some photos with tape on a green carrier. The parallax image shows a white/grey reflection of the diffuser, with the green background visible through the sprocket hole. This continues to work very well. 
* I also have some carriers printed with a grey filament. The equivalent parallax image is a grey circle on a grey background, and this does not work well 😭 The sprocket hole circle was still not reliably detected with either pipeline, with or without parallax.

In conclusion, I continue to find this to be a good feature. But it is not quite so universally applicable as I find the equivalent parallax feature for fiducial vision.

PR #1714, opened 2024-12-01
ReferenceStripFeeder distance calculation from reference holes fix

tmandys:
# Description
ReferenceStripFeeder had bug when calculating distance between reference sprocket holes. The first reference whole is hole following part center. Not sure if it is intended but I updated tooltip text to satisfy current state.
More serious bug is that it did not work if distance between the first and second reference hole was not part pitch. Intended (documented) behaviour is that second hole is any whole in tape continuation. So unless holes defined such way then part #2, #3 were located by hole pitch (4, 8, 12, ...mm), not e,g 16, 32, 48 ... mm in case of 16mm part pitch.

# Justification
bugfix in general feeder

# Instructions for Use
Include a tape in strip feeder with part pitch >8mm and test movement over particular parts

# Implementation Details
1. manual test with machine and feeder
2. yes
3. no
4. yes


tmandys:
Yes, in PnP process it likely works, I used it as well. But does not when manually controlling on Feeder panel. Try change Feed count 1, 2, 3, ... when hole distance is 4mm and part pitch 16mm. Head moves by 4mm, not 16mm.

image

tobydickenson:
Aha, yes. It calculates the right distance if it had already performed a feed and collected some vision data. But the distance was calculated wrong for feedcount>1 if `getIdealLineLocations` was returning data based only on the configured reference locations. It was possibly also wrong if vision was disabled.

Thanks for the extra details. I agree this fix looks good, and the fix for 0402 looks good too. 🙂

tobydickenson:
I propose merging this fix in 2.3 branch. @tmandys do you agree? any last comments?

tmandys:
Just do it ;-)

PR #1715, opened 2024-12-10
remember the global best route in case a swap is accepted due to anne…

janm012012:
# Description
According to the discussions on the [list](https://groups.google.com/g/openpnp/c/HXvHZ2NuCyg/m/f4I2I5TZAwAJ) the travelling salesman implementation requires fixing that in case a worth route is accepted due to the annealing probability, the best route still needs to be considered then returning. This PR intends to fix that.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **using the automated test sequences**

### DIFF
--- a/src/main/java/org/openpnp/util/TravellingSalesman.java
+++ b/src/main/java/org/openpnp/util/TravellingSalesman.java
@@ -84,6 +84,12 @@ public class TravellingSalesman<T> {
     private static final int debugLevel = 0;
 
     /**
+     * Factor to scale current travel distance before coping it over to global
+     * best distance to avoid excessive copies due to rounding effects.
+     */
+    private static final double globalBestDistanceScalingFactor = 1.0 - 1e-5;
+    
+    /**
      * Plain old data TravelLocation for faster processing. Improved solving by a factor of 6 from using
      * OpenPNP Locations directly. These are always in Millimeters, no conversions needed.  
      */
@@ -212,15 +218,15 @@ public class TravellingSalesman<T> {
             System.out.println("Simulated Annealing, size: "+this.travelSize+" temperature: " + startingTemperature + ", max iterations: " + maxIterations + ", cooling rate: " + coolingRate);
         }
         int i = maxIterations;
-        int swaps = 0, twists = 0;
+        int swaps = 0, twists = 0, copies = 0;
         double bestDistance = getTravellingDistance();
         double t = startingTemperature;
         if (debugLevel > 0) {
             System.out.println("Initial distance of travel: " + bestDistance);
         }
         if (this.travelSize > 1) {
-            List<TravelLocation> globalTravel = new ArrayList<>(this.travel);   // global best route
-            double globalBestDistance = bestDistance;                           // cost of global best route
+            List<TravelLocation> globalTravel = new ArrayList<>(this.travel);           // global best route
+            double globalBestDistance = globalBestDistanceScalingFactor * bestDistance; // cost of global best route
 
             // make this repeatable by seeding the random generator
             Random rnd = new java.util.Random(0);
@@ -255,7 +261,7 @@ public class TravellingSalesman<T> {
                         double newDistance = getTravellingDistance();
                         this.swapLocations(a, b, twist);
                         if (Math.abs((newDistance - oldDistance) - swapDistance) > 0.1) {
-                            System.err.println("** Swap distance wrong - newDistance:" + newDistance + ", oldDistance:" + oldDistance +", swapDistance: "+swapDistance + " != "+(newDistance - oldDistance)+", twist: "+twist);
+                            System.err.println("** Swap distance wrong - newDistance: " + newDistance + ", oldDistance: " + oldDistance +", swapDistance: "+swapDistance + " != "+(newDistance - oldDistance)+", twist: "+twist);
                         }
                     }
 
@@ -266,8 +272,9 @@ public class TravellingSalesman<T> {
                         // if the new route is better then the best, remember it
                         if (bestDistance < globalBestDistance) {
                             // remember slightly worth distance do avoid excessive copies due to rounding effects
-                            globalBestDistance = 0.999 * bestDistance;
+                            globalBestDistance = globalBestDistanceScalingFactor * bestDistance;
                             globalTravel = new ArrayList<>(this.travel);
+                            copies++;
                         }
                         swaps++;
                         twists += twist ? 1 : 0;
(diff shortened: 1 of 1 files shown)


janm012012:
Hi @tonyluken!

This is my proposal to fix the travelling salesman. Please review and comment. Thank You!

Jan

tonyluken:
Please revert your changes related to seeding of the random number generator. We always want repeatable answers regardless of the debug level.

Wouldn't it be better to initialize `globalBestDistance` to `bestDistance` right after line 216 and initialize `globalTravel` right after that as well with a **deep** copy (see below) of `this.travel`? Then in the loop you can just do this check: 
```
                           if (bestDistance < globalBestDistance) {
                               globalBestDistance = bestDistance;
                               globalTravel = DEEP COPY OF this.travel (see below)
                           }
```

right after this line:
`                        bestDistance += swapDistance;   // keep bestDistance up-to-date
`

When the loop is done, `globalTravel` will always contain the best route found.

Note, that the current way you are copying will result in `globalTravel` pointing to the same elements as in `this.travel` and on the next loop iteration, the elements pointed to by both lists will get modified by a swap. The proper way to deep copy is to iterate through the elements of `this.travel` and create new instances of each element and add them to the `globalTravel` list. Probably the cleanest way is to add another constructor to the `TravelLocation` class that makes a copy of another instance. It would look something like this:

```
        private  TravelLocation(TravelLocation travelLocation) {
            super();
            this.index = travelLocation.index;
            this.x = travelLocation.x;
            this.y = travelLocation.y;
            this.z = travelLocation.z;
        }
```

then to create a deep copy of `this.travel` you would do something like this:
```
        globalTravel = new ArrayList<>();
        for (TravelLocation travelLocation : this.travel) {
            globalTravel.add(new TravelLocation(travelLocation));
        }
```

janm012012:
Hi @tonyluken!

Many thanks for reviewing! Your suggestion to always remember the best route is good. It increases the readability why only adding two extra copies. I thinks that's worth it.

I reverted the modification I did to the seeding.

Concerning the deep copy I have to disagree: the route is given by the order of `TravelLocation` in the list. The input array `travelInput` is converted once (at line 67) into `TravelLocation` which are per design read-only. When changing the route, `swap()` (line 182) is used to change the order of the list. The elements itself are kept unchanged. I especially made sure that `new ArrayList<>(this.travel)` creates a copy of the list. This new list has the order of the input and references their elements (without copy), which is ok here as the elements are read-only.

Jan

tonyluken:
Oh yeah, you are correct. I though swap was modifying the elements but it's not, so no deep copy required

But wouldn't it still be better to do the initialization outside the loop and simplify the logic inside the loop?

tonyluken:
Just some minor thoughts:

1. I wouldn't modify bestDistance within the debug if statements. I think it would be more appropriate to use a local variable in those places. It would also be good to include the value of globalBestDistance in the debug printouts.
2. When we find a globally better solution, in the assignment `globalBestDistance = bestDistance;` I think it would be good to apply a factor slightly less than one so that we don't spend a lot of time saving solutions that are insignificantly better (due to math truncation or rounding errors). Something like `globalBestDistance = 0.999 * bestDistance;` so that in the next iterations, we will only save a new better solution if it is at least 0.1% better than the previous global best.
3. I think the random number generation could be speed up by a factor of roughly two by using `rnd.nextInt(this.travelSize)` rather than `(int) (rnd.nextDouble() * this.travelSize)`.

janm012012:
Many thanks Tony, I've added all three suggestions. (Although the random number generator is faster by a factor of two, I can't see an overall effect on the TravellingSalesman.)

Side question: do you by chance know why simulated annealing was chosen as solver? It seems to come with a lot of factors that effect its performance and even with small tuning, the reference test suite generates better results and/or takes much more time. The termination criteria (t > 0.1) in combination with the cooling rate seems to be especially effected. (I observed the final distance and the .svg graphics generated.) Googling around it looks as if nowadays branch and bound bases algorithms are more common claiming to provide the perfect solution in much less then brute-force tries...

Jan

tonyluken:
I think you missed line 281, it is still modifying bestDistance. 

> Side question: do you by chance know why simulated annealing was chosen as solver?

No, I don't know why it was chosen. I assume someone found an easy example somewhere that they ported over to OpenPnP.

I do know that there are better TSM solvers out there, but I haven't looked into them to see which can provide better solutions in a similar amount of time.

janm012012:
> I think you missed line 281, it is still modifying bestDistance.

Ups, overlooked that...

> I do know that there are better TSM solvers out there, but I haven't looked into them to see which can provide better solutions in a similar amount of time.

IIRC I noticed some time back that the current solver takes much more tries then brute force if used for just a few cities which is the main use case for multi-nozzle optimization in the job processor. I'll investigate that in a other PR. I hope we can continue the discussion there.

markmaker:
> Side question: do you by chance know why simulated annealing was chosen as solver? It seems to come with a lot of factors that effect its performance and even with small tuning, the reference test suite generates better results and/or takes much more time.

That was me. :sweat_smile: And I admit I implemented it mostly out of academic interest. Years ago I read this huge book about optimization problems. After laboring half-way through the book, with ever more complex and heavy-weight deterministic/memorizing search algorithms being discussed, it suddenly turned to chaotic/random search algorithms. I simply couldn't believe that the same problems discussed earlier in the book could be solved so much easier, and often in less computation time. Simulated annealing particularly fascinated me, and I always wanted to try it out. When the BlindsFeeder needed efficient cover opening, I finally implemented it. For the opening of a few dozen feeder covers it performed okay. Then it got used in more and more other places inside OpenPnP, which might be more demanding.

Clearly, it can be replaced by a more sophisticated implementation and/or using a different algorithm. The interface should be easily adaptable/reusable.

One thing I wish for, would be some real world before/after benchmarking. See how bad it was (or maybe not).

Speaking of benchmarking: for this PR it would be nice to _log_ when the "global best route" was actually better than the conventional result. If I understand or rather "feel" the algorithm right, it should happen very, very rarely. :wink:

_Mark

janm012012:
Hi Mark!

I've added the logging line and did not found a single occurrence in my test cases. 

I'm also counting the number of copies now and found that its rather low. For the 610 cities test case I see "swaps: 726611, twists: 495020, copies: 3263", which is about 1 copy per 370 swaps or twists.

Many thanks for your explanation on the history of the algorithm. I've played around with a new (realistic) metric and found (watching the .svg's of the TravellingSalesmanTest) that the results are not very good (with the default they are ok). By playing around with the various factors I could improve it, but not always. Without seeding the results where not very stable. That's why I started looking around for other solvers. I'll try if any is still reasonably fast, which the current admittedly is...

Jan

tonyluken:
Speaking of counting swaps and twists - shouldn't swaps only be incremented when twist = false?  Maybe move the incrementing of swaps and twists here:

                    // choose the better option
                    if (twistDistance < swapDistance) {
                        twist = true;
                        swapDistance = twistDistance;
                        twists++;
                    }
                    else {
                        swaps++;
                    }

janm012012:
Yes, I think that would be better. Changed.

tonyluken (approved):
This all looks good to me. Thanks!

PR #1716, opened 2024-12-12
Feature/feed options to control movement

tmandys:
# Description
Added new feed options to cope with issues when something does not go smoothly. When e.g. part is lost/wrongly detected or so then re-feed will move to next feeder position even current one might be loadede with part, e.g. manually reloaded with lost part or so (or simply user is testing machine, setting up feeders or so). 

Related to https://groups.google.com/g/openpnp/c/QRtgZP6b5Os 

# Justification
General usability improvement

# Instructions for Use

4 feeders (I currently use) have now option ("Normal/Skip next/disable feed"). "Skip next" won't do feed, just picks part and set option to "Normal" so it will pick manually placed part in feeder. "Disable" will ignore any feed movements and with keep using current part which is handy for testing. 
Recycle is considered as well.

New option for ReferenceAutoFeeder "Recycle supported" supports "Recycle" command on "Special" jog panel also for this feeder.

image

image

image

image

# Implementation Details
1. manually tested on machine
2. coding style: yes
3. model change: no (new xml tags in configuration)
4. mvn test: yes


tobydickenson:
I will be adding this patch to the build I use in production for further testing. Feedback to follow.

tmandys:
I moved options to feeder table as it should be visible immediately. It avoids also copypasting code for each feeder and avoids annoying "Do you want to apply" dialog even when changed from code.

image

tobydickenson:
I like this idea, but have some reservations about the implementation. I hope to find time for a detailed code review this week.

src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java
```diff
@@ -403,7 +410,7 @@ private Location findClosestHole(Camera camera) throws Exception {
      */
     @Override
     public boolean canTakeBackPart() {
-        if (feedCount > 0 ) {
```
tobydickenson, line 406: Spurious whitespace change

tobydickenson:
Hi,

My apologies for the delay in writing up this review. I very much like this feature, so thank you for developing it.

1. Your patch has some spurious changes; FUNDING.yml and some whitespace. It is a huge help if the branch is kept clean of such unwanted changes and merges cleanly with the test branch. I expect there are some new merge conflicts with the `ReferenceStripFeeder` changes I merged last week too (sorry!).

2. I do like your UI for this feature.

3. We dont need the new base class. We cant add a new base class every time we add a new feature! I suggest the content of `FeederWithOptions` should be in `ReferenceFeeder`, plus a new `boolean supportsFeedOptions()` method.

4. Some of your new `feed()` methods are getting quite large, and now lack [cohesion](https://en.wikipedia.org/wiki/Cohesion_(computer_science)) because they are handling both the feed options check plus the normal feed logic. `BambooFeederAutoVision` in particular. The big block indented under `if (getFeedOptions() == FeedOptions.Normal)` could be usefully refactored into its own method. You might find a better name, but I guess `doNormalFeed()` would be ok. 

5. The "but to be sure" checks in `takeBackPart` are duplicated between classes. There was duplication before your patch of course, but lets now rationalise all of these. Move this to a new `FeederUtils.java`? Does it make sense to refactor the "put the part back on the location of the last pick" block into a new method in `FeederUtils` too?

6. You have some new text that should be in the translations file.

7. What is the rationale for making part recycle optional for AutoFeeder?

8. It is not a problem that this does not add support to every feeder type, but it would be good to have other contibuters cover the other feeders before the next release. I can add support for PushPullFeeder.

tmandys:
3. I did not want to affect feeders which I cannot modify to new feature as I cannot test them so I added new class. Not sure if the feature is general for all possible feeders. 

6. That's question when a java file currently has all texts hardcoded in code then shall I refactor it completely and move all translations, follow current style or put there only new text. I use #2 (I believe).

7. Optional recycle is implemented for backward compatability and I can imagine a feeder which physically cannot support "to put part back" operation. Even now I can recycle only the latest part taken from feeder, e.g. cannot recycle both nozzle tip loaded from the same feeder (at least from auto feeder as there is no reverse motion).

tmandys:
Fixed in branch

6. I do not think I can transparently extract "if" code to separate method `doNormalFeed()` as there are more "ifs" and I do not want to copypaste code.

tobydickenson:
Thank you for making those changes. The outline design of this now looks great.

I plan to do a detailed line-by-line review later this week, then merge if nothing else comes up.

janm012012:
Hi tmandys, hi Toby!

This features formally cries out to be part of a generalized [feed/feedcount](https://github.com/openpnp/openpnp/wiki/Programming-ToDo's-and-NiceToHave's#from-httpsgroupsgooglecomdmsgidopenpnp47246f36-80ab-41ec-813e-4a73d653dca740makrzone-on-feedcount-and-remaining-parts-per-feeder) handling as discussed many times in the past. If grouped with feedcount/remaining and a dedicated wizard, it could be easily applied to many feeders.

Jan

tobydickenson:
Hi Jan,

> feedcount

The use case for this patch is complementary to what I understood from that feedcount discussion.

Currently the job processor always operates on a cycle of feed / pick / place. Those three steps are performed unconditionally.  But sometimes that feed step is wasteful. For example after manual handling of an autofeeder you might leave it with a part exposed, and the job processor could use that part by doing a pick without feeding.

I dont think any feedcount-based mechanism could avoid the motor turning on an autofeeder.

> could be easily applied to many feeders

I think that applies to this PR too.

janm012012:
Hi Toby!

I did not intended to derail this PR. I only wonted to remark, that we had discussions in the past, that a feed/feedcount handling might be better off if separated from the feeder code itself. (Yes, the actual feed operation itself would still be a feeder specific implementation.) This refactoring - that was the reason for the old discussion - could then easily be enhanced by providing a remaining counter and/or add an interface to an inventory system. To me this PR looks as if it would also fit well into such a new concept. Please feel free to continue with this PR as is if you think its not yet time for the refactoring.

Jan

src/main/java/org/openpnp/machine/reference/feeder/ReferenceAutoFeeder.java
```diff
@@ -33,6 +33,7 @@
 import org.openpnp.spi.base.AbstractActuator;
 import org.openpnp.util.MovableUtils;
 import org.pmw.tinylog.Logger;
+import org.python.apache.commons.compress.compressors.snappy.FramedSnappyDialect;
```
tobydickenson, line 36: This import is unnecessary?

tobydickenson:
Hi tmandys,

This all looks good. The new feature looks good, and your refactoring has improved the existing code. Thank you for your effort here. I am looking forward to implementing `FeedOptions` in PushPullFeeder.

I noted one spurious import. Please fix that and I will merge right away.

tmandys:
Done. Note: I do not understand why branch diverges and I had to force push to proceed when I use rebase on top of test branch.

tobydickenson:
Thank you tmandys

PR #1718, opened 2024-12-18
The fiducial locator can take multiple passes, with the vision center…

tobydickenson:
# Description

I have been looking at the max-distance limit for the fiducial locator. If it is working correctly then the multi-pass process causes the camera to incrementally step towards the fiducial center, which is necessary to avoid parallax errors. When it works, it works well.

But it can go wrong if the maxDistance property is not set correctly, and one possible outcome is confusing and surprising. The fiducial locator can chase from one visual distraction to the next, and travel outside the expected search area.

The fiducial locator code predates the `DetectCircularSymmetry.center` property override mechanism, and therefore each step of the multi-pass process was using a vision pipeline with a new search area centered on the new camera view point. The original fiducial locator code blocks any false-positive output by calculating the distance between starting and ending point, and it raises an exception if it exceeds the specified maximum.

The code here changed when the parallax fiducial method was added.  This added explicit control of the search area using a `DetectCircularSymmetry.center` property override, but this change did not affect the underlying logic. The search area center point still walked with each step of the multi-pass process, and it can still chase distractions outside of the expected boundary.

This patch simplifies the process by always using the nominal location as the search area center point. (Or, if the parallax method is enabled, the center point is a parallax offset from the nominal location). This is better because:
* Every stage of the multi-pass process is constrained to same search area.
* While each step of the multi-pass process moves the camera location, it does not change the search area. There is a different viewpoint, but looking at the same board features.
* It is no longer possible for the multi-pass process to walk outside the specified maximum distance. Each run of the pipeline can only return a feature which conforms to the maximum distance specification. The camera only moves within the specified maximum distance.
 
# Justification

## Usability Justification

This change makes the system behave in a more predictable manner if the fiducial maximum distance limit is not quite right. The camera does not move beyond the specified maximum distance.

I have not found any scenario where its behavior will be less good.

## Software Engineering Justification

It is more cohesive to have the the maximum distance processed entirely by the pipeline. Previously it was processed by the vision pipeline, but then checked by the fiducial locator. This cohesion becomes more important when there are other ways to specify the fiducial maximum distance (I have another PR to follow)

# Instructions for Use

This is refining the behavior of an existing feature.

# Implementation Details
    * On a real machine
    * Testing fiducial scans on my production boards
    * Testing boundary conditions: maximum distance, misalignment, and nearby distractions.
5. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. -- test pass

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -46,6 +46,7 @@ import org.openpnp.model.Location;
 import org.openpnp.model.PanelLocation;
 import org.openpnp.model.Part;
 import org.openpnp.model.Placement;
+import org.openpnp.model.PlacementsHolderLocation;
 import org.openpnp.spi.Camera;
 import org.openpnp.spi.Feeder;
 import org.openpnp.spi.FiducialLocator;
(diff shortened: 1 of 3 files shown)


tobydickenson:
There are several problems with this PR sorry :-( fixed on #1720

PR #1719, opened 2024-12-18
Fiducial Max Distance control

tobydickenson:
# Description

This patch adds a ParameterNumeric stage to the default fiducial detector pipeline for configuring the detection range. This benefits from the nice slider GUI that shows a preview of the consequence of adjusting the detection range.

For backwards compatibility with every user's existing fiducial pipelines, the maxDistance property of the fiducial locator remains unchanged. The fiducial detector supplies this value as a pipeline property override if it finds that the ParameterNumeric stage is not present.

# Justification

#1707 noted that it is useful to have some fiducial parts using a different detection range, and that there was no convenient way to achieve this. Fiducials at different levels of the panel/board hierarchy can benefit from a different maximum distance.

Another rationale is that different boards might benefit from different settings.

This feature is fully backwards compatible. Existing system behavior is unchanged when software is upgraded. Updating the pipeline provides the new `ParameterNumeric` control.

# Dependencies

Some wiki updates when this patch is merged:

* #1717 A required bug fix in ParameterNumeric
* #1720 Is both a usability improvement, and better code cohesion which makes this patch smaller.

# Instructions for Use

On https://github.com/openpnp/openpnp/wiki/Fiducials

4. Make sure that the fiducial is within a certain distance from where it is expected to be. Go to your fiducial part / Fiducial Vision Settings, and and adjust Max. Distance if needed. If you are using an older version of the vision pipeline this is available as a global setting at Machine Setup / Vision / Fiducial Locator. This prevents from locking to something in the camera view which is recognized as fiducial, but really isn't.

On https://github.com/openpnp/openpnp/wiki/Fiducial-Locator

**Max. Distance** makes sure that the fiducial is within a certain distance from where it is expected to be. This prevents from locking to something in the camera view which is recognized as fiducial, but really isn't.

Also I will update the Configuration screenshot on that page.

# Implementation Details
 * Tested the upgrade scenario in production: new software with old pipelines
 * Tested the full solution in production: new software with new pipelines. This included a panel which requires several different fiducial vision pipelines, which can now all be stock pipelines.


tobydickenson:
I have just noted that the Footprint Fiducial pipeline needs a similar update.

tobydickenson:
NB tests are currently failing due to dependency bug #1717

tobydickenson:
Review:

This feature was originally split over three orthogonal Pull Requests. Those are all now merged into this branch.

Previously `ReferenceFiducialLocator` enforced the limit on the distance between found location and expected location. This PR moves that responsibility into the vision pipeline. As noted above, the stock footprint fiducial pipeline needs to be updated to fulfill this responsibility. Doing so will need extra support in the `MatchTemplate` pipeline stage, which currently has no support for distance or center point. I guess that means the parallax option is currently broken for the footprint fiducial pipeline too.

Once the fidicuial pipeline is fixed, this PR is good to be merged into test.

PR #1722, opened 2024-12-28
change Length class to be immutable

tobydickenson:
# Description

There is lots of code which is written as if the `Length` class is immutable. But, to my great surprise, it is not! This patch removes the Length methods which change its value. It changes a small amount of code which uses those methods to use alternatives which return a new `Length` object.

# Justification

## Justification in code that uses Length

For example, class `ReferenceNozzleTip` has `public Length getMaxPartHeight()` to allow other code to access an attribute.

It also has `public void setMaxPartHeight(Length maxPartHeight)` to allow other code to change that attribute.

It is somewhat surprising that this property could also be changed by code such as `tip.getMaxPartHeight().setValue()`. Thats not how we want this API to be used, and it is not how we want future APIs (exposing a `Length` attribute) to be written. The presence of methods  `Length.setValue` and `Length.setUnits` feels like a bug in the architecture.

## Justification within Length

If `Length` was intended to be a mutable shareable object, then various optimisations inside `Length` are confusing and unsafe. For example, `convertToUnits` has   `if(this.units==units) return this;` which means the return value is sometimes shared, and sometimes not.

## Justification of consistency

The `Location` class has a similar interface, and is used in a similar manner, and it **is** immutable. I find it surprising for these two classes to be different.

# Instructions for Use

There is no end user impact.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. -- unit tests pass
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. -- Yes, and `org.openpnp.model.Length` is one of the lowest level classes!
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. -- unit tests pass.

### DIFF
--- a/src/main/java/org/openpnp/gui/components/ComponentDecorators.java
+++ b/src/main/java/org/openpnp/gui/components/ComponentDecorators.java
@@ -76,12 +76,9 @@ public class ComponentDecorators {
     }
 
     private static void convertLength(JTextField textField, String format) {
-        Length length = Length.parse(textField.getText(), false);
+        Length length = Length.parseWithDefaultUnits(textField.getText(), Configuration.get().getSystemUnits());
         if (length == null) {
             return;
-        }
-        if (length.getUnits() == null) {
-            length = length.changeUnits(Configuration.get().getSystemUnits());
         }
         length = length.convertToUnits(Configuration.get().getSystemUnits());
         textField.setText(String.format(Locale.US, format, length.getValue()));
--- a/src/main/java/org/openpnp/gui/components/ComponentDecorators.java
+++ b/src/main/java/org/openpnp/gui/components/ComponentDecorators.java
@@ -81,7 +81,7 @@ public class ComponentDecorators {
             return;
         }
         if (length.getUnits() == null) {
-            length.setUnits(Configuration.get().getSystemUnits());
+            length = length.changeUnits(Configuration.get().getSystemUnits());
         }
         length = length.convertToUnits(Configuration.get().getSystemUnits());
         textField.setText(String.format(Locale.US, format, length.getValue()));
--- a/src/main/java/org/openpnp/gui/support/LengthConverter.java
+++ b/src/main/java/org/openpnp/gui/support/LengthConverter.java
@@ -44,12 +44,9 @@ public class LengthConverter extends Converter<Length, String> {
 
     @Override
     public Length convertReverse(String s) {
-        Length length = Length.parse(s, false);
+        Length length = Length.parseWithDefaultUnits(s, Configuration.get().getSystemUnits());
         if (length == null) {
             throw new RuntimeException("Unable to parse " + s);
-        }
-        if (length.getUnits() == null) {
-            length = length.changeUnits(Configuration.get().getSystemUnits());
         }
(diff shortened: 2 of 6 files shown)


janm012012:
Hi Toby!

I agree with you, that Length shall be immutable. To force that, I'd suggest to declare value and units final. By doing so, the compiler generates about 13 errors (without setValue(), setUnits() and Length()). which you all fixed.

I see two addition improvements you might wont to consider:

1. LengthUnits could be refactored into its own class allowing it to be inherited to Length and Location (and potentially others) generating cleaner code.
2. In FootprintTableModel::setValueAt LengthCellValue::getLength() returning null is fixed four times using identical code. I'd suggest to refactor that into a single method, something like `Length getLength(LengthUnits defaultUnit)`.

Besides that, the PR looks good to me!

Jan

tobydickenson:
> I agree with you, that Length shall be immutable. To force that, I'd suggest to declare value and units final. By doing so, the compiler generates about 13 errors (without setValue(), setUnits() and Length()). which you all fixed.

Adding `final` causes exceptions during xml persistence. This exception looks like it might be fixable, but I know `final` can cause problems for other persistence frameworks too. So I (sadly) suggest not doing this.

>    LengthUnits could be refactored into its own class allowing it to be inherited to Length and Location (and potentially others) generating cleaner code.

The shared LengthUnits enum seems ok to me. Maybe I'm missing something.

>    In FootprintTableModel::setValueAt LengthCellValue::getLength() returning null is fixed four times using identical code. I'd suggest to refactor that into a single method, something like `Length getLength(LengthUnits defaultUnit)`.

I had changed the `setUnits` mutable method to `changeUnits` which returns a copy.

In practice we see that this was only used by code which tests whether the units are null, and supplies a sensible default unit. So lets change that again to a new Length method `changeUnitsIfUnspecified`.

Furthermore, I note that this was frequently called on the return value of the `parse` method, which an optionally return a length with unspecified units. So lets add a new method `parseWithDefaultUnits` for this operation.

janm012012:
Hi Toby!

I like how to did it! That looks clean and self explaining.

I'm sorry for the suggestion to add `final`. I missed the dependency on the simplexml framework. I was (at least to me) temporarily useful to detect all occasions where Length was not immutable. (Btw: I did the same for Location as you suggested on the list and found much no issues.)

Concerning the refactoring: I was under the impression, that the entire unit handling could be refactored into a new class which then would store the current unit and provide scaling into a new unit. Unfortunately this will very likely not work for the same reason as above: if unit is also accessed by simplexml, the refactoring would break the dependencies. So we can only keep it as is referencing a single LengthUnit and providing individual storage locations, getter methods and scaling.

From my perspective this PR is good as it is. Unfortunately I think I don't have permission to merge, even though the "Merge pull request" button is present and green...

Jan

tobydickenson:
Any objections to merging this into 2.3 test now?

janm012012:
merging? what? ;-)

Jan

PS: Thanks a lot for cleaning up the API!

PR #1724, opened 2025-01-06
Trigger the update to the placement counter on the status bar after c…

tobydickenson:
# Description

This status bar was unreliable.

It was updating when the Job tab is open. But, if the user switches to a different tab, this data was frozen even though placements were progressing.

image

Previously the update to this status was triggered by the *rendering* of the job table 😵‍💫 The table is not rendered when viewing a different tab, so no status bar update was triggered.

This patch moves the trigger to the job processor, after completing a placement.

# Justification

The status bar was misleading when it did not update. This fix helps all users.

# Instructions for Use

No instructions are required; it fixes a feature to work as expected.

# Implementation Details
   * Only tested on a simulated machine so far
   * Various regressions tests: The GUI interactions update this status as expected (for example ticking a placement as complete, or enabled/disabled). The status updates when viewing the job tab during job activity.
   * The status updates when viewing the part tab during job activity too. Previously it would have been frozen in this tab.
5. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. -- all tests pass

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ a complete change list, only those that may directly interest or affect users.
 ## Bug Fixes
 
 * Fix bug causing manual nozzle tip changes to get swallowed if the corresponding placement is set to defer errors. [PR 1741](https://github.com/openpnp/openpnp/pull/1741)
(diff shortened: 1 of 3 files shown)


tobydickenson:
fix has been rebased to latest test branch

PR #1725, opened 2025-01-09
Fix PandaplacerVisionFeeder pick calibrated position calculation

tmandys:
# Description
PandaplacerVisionFeeder wrongly calculates calibrated pick position under some circumstances related to angle rotation.
3 issues:
- typo angleTape vs. part rotation angle
- wrong logic for rounding 3.5, 5.5, 7.5, (9.5), 11.5mm distance
- rounding provides also wrong result for negative distance (because of tape angle)

Table of rounding pick-to-hole distance
relativePickLocation.getY() | Required result| Old algorithm| Fixed
-- | -- | -- | --
0 |   | 0,5 | -0,5
0,4 |   | 0,5 | -0,5
0,7 |   | 0,5 | 1,5
1 |   | 0,5 | 1,5
1,4 |   | 0,5 | 1,5
1,6 |   | 2,5 | 1,5
2 |   | 2,5 | 1,5
2,4 |   | 2,5 | 1,5
2,6 | 3,5 | 2,5 | 3,5
3 | 3,5 | 2,5 | 3,5
3,4 | 3,5 | 2,5 | 3,5
3,5 | 3,5 | 4,5 | 3,5
3,6 | 3,5 | 4,5 | 3,5
4 | 3,5 | 4,5 | 3,5
4,4 | 3,5 | 4,5 | 3,5
4,6 | 5,5 | 4,5 | 5,5
5 | 5,5 | 4,5 | 5,5
5,4 | 5,5 | 4,5 | 5,5
5,6 | 5,5 | 6,5 | 5,5
6 | 5,5 | 6,5 | 5,5
6,4 | 5,5 | 6,5 | 5,5
6,6 | 7,5 | 6,5 | 7,5
7 | 7,5 | 6,5 | 7,5
7,4 | 7,5 | 6,5 | 7,5
7,6 | 7,5 | 8,5 | 7,5
8 | 7,5 | 8,5 | 7,5
9 |   | 8,5 | 9,5
10 | 11,5 | 10,5 | 9,5
11 | 11,5 | 10,5 | 11,5
12 | 11,5 | 12,5 | 11,5

# Justification
General bugfix

# Instructions for Use
Use Pandaplacer and test calibrated pick position to see that now should not pick beyond feeder

# Implementation Details
1. Use Pandaplacer feeder in different rack positions and check different part angles to see that now should work, i.e. calibrated position is intended
2. yes
3. no
4. yes

### DIFF
--- a/src/main/java/org/openpnp/util/FeederVisionHelper.java
+++ b/src/main/java/org/openpnp/util/FeederVisionHelper.java
@@ -672,14 +672,24 @@ public class FeederVisionHelper {
                                 Location pickLocation = partLocation.convertToUnits(LengthUnit.Millimeters);
                                 Location relativePickLocation = pickLocation
                                         .subtract(hole1Location);
-                                // rotate from old angle
-                                relativePickLocation =  relativePickLocation.rotateXy(-pickLocation.getRotation())
+                                // rotate by tape angle, so dX multiply of partPitchMinMm, and dY pick-hole line distance
+                                relativePickLocation =  relativePickLocation.rotateXy(-angleTape)
                                         .derive(null, null, null, 0.0);
                                 // normalize to a nominal local pick location according to EIA 481
+                                // vertical hole to part distances are 3.5, 5.5, 7.5, 11.5mm for 8, 12, 16, 24mm tapes
                                 if (normalizePickLocation) {
+                                    double dY = Math.abs(relativePickLocation.getY());
+                                    if (dY > sprocketHoleToPartGridMm / 2) {
+                                        dY = Math.round((dY - sprocketHoleToPartMinMm)/sprocketHoleToPartGridMm)*sprocketHoleToPartGridMm + sprocketHoleToPartMinMm;
+                                        if (relativePickLocation.getY() < 0) {
+                                            dY = -dY;
+                                        }
+                                    } else {
+                                        dY = relativePickLocation.getY();
+                                    }
                                     relativePickLocation = new Location(LengthUnit.Millimeters,
                                             Math.round(relativePickLocation.getX()/partPitchMinMm)*partPitchMinMm,
-                                            -sprocketHoleToPartMinMm+Math.round((relativePickLocation.getY()+sprocketHoleToPartMinMm)/sprocketHoleToPartGridMm)*sprocketHoleToPartGridMm,
+                                            dY,
                                             0, 0);
                                 }
                                 // calculate the new pick location with the new hole 1 location and tape angle


markmaker:
Note this class does not only affect PandaplacerVisionFeeders, but also ReferencePushPullFeeder. It was refactored out of that a while back and I don't know if there were any bugs introduced then. But it was refactored from the existing code and I'm absolutely sure the existing code had no such "off by one" rounding bug, otherwise this would never have worked. The proposed code also does not make sense, mathematically. This is a linear thing and you don't need "if's" in linear.

I'm less sure about the rotation bug though. It could have existed before (admittedly, I only ever tested with "south" feeders) or been introduced in refactoring. However, the new code cannot be right, as this must be the _old_ angle, not the new one just calculated.

Whoever reviews this, be extra cautious, and demand testing with ReferencePushPullFeeder too.

_Mark

tmandys:
> Note this class does not only affect PandaplacerVisionFeeders, but also ReferencePushPullFeeder. It was refactored out of that a while back and I don't know if there were any bugs introduced then. 

And where is refactored code ? Is it _master, develop, test_ or not merged feature branch ? Maybe I use old code, I work with _test_ branch with applied patches I made and are in pull request queue.

> This is a linear thing and you don't need "if's" in linear.

Function is linear but not symmetrical around zero as should be. See table in original pull request. _If_ is also there  not to normalize distance  when makes no sense. How to adjust when configured distance is 0.5mm and minimal normalized is 3.5mm? Also see e.g. that in original code distance 3.5mm is wrapped to 4.5mm. 
Rotation bug is typo when in one direction is rotated by part orientation and back by hole line orientation. When equal you won't see the bug.

```
relativePickLocation = relativePickLocation.rotateXy(-pickLocation.getRotation())
calibratedPickLocation = calibratedHole1Location.add(relativePickLocation.rotateXy(angleTape))
                                        .derive(null, null, pickLocation.getZ(), angleTape);

```
Also question what to do with 9.5mm which is skipped in row, when comment says it is being normalized to EIA 481.

prusk:
@markmaker Looks like this change is in a very complex area that I left completely untouched when creating the FeederVisionHelper so if the original didn't have any issue, the Utils class doesn't either.
I do remember working hard to understand the logic and as far as I tested a lot the location calculations, I did not find any error in logic.

To prove the point that nothing changed in the calculation logic, see attached the 2 files with the findFeatures function for:
- the latest FeederVisionHelper code: https://github.com/openpnp/openpnp/blob/test/src/main/java/org/openpnp/util/FeederVisionHelper.java#L437
- the PushPullFeeder code before any helper changes: https://github.com/openpnp/openpnp/blob/a7a99223f3d0986dec62a72f84bdd02058ebcd90/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java#L1468

One can see the only differences are due to slight approach with parameters.

[FeederVisionHelper-NewCode.java.txt](https://github.com/user-attachments/files/18708249/FeederVisionHelper-NewCode.java.txt)
[PushPullFeeder-OldCode.java.txt](https://github.com/user-attachments/files/18708252/PushPullFeeder-OldCode.java.txt)

markmaker:
> Function is linear but not symmetrical around zero as should be.

For context: this is supposed to be a "snap to grid" implementation, i.e. a manually captured, imprecise location is turned into a precise one using the sprocket holes mid-point as reference.

**Important note 1**: the "snap to grid" rounding happens offset by `sprocketHoleToPartMinMm`, which is 3.5mm according to EIA 481. 

**Important note 2**: the 2D relative feeder coordinate system is right-handed:

EIA 481

The sprocket holes are at the top of the tape according to EIA 481 definitions. The origin is between the two sprocket holes. X points in the direction of unreeling, Y points up. So the `sprocketHoleToPartMinMm` offset is used with _negative_ sign (letter F in the EIA 481 figure).

This explains the Y rounding expression like this:

`-sprocketHoleToPartMinMm+Math.round((relativePickLocation.getY()+sprocketHoleToPartMinMm)/sprocketHoleToPartGridMm)*sprocketHoleToPartGridMm
`
Regarding symmetry around zero: The original input is a manually jogged and captured location. This means the location is not precise, but it is quite close to the grid point. The Y grid `sprocketHoleToPartGridMm` is 2mm, i.e., a user is well capable to set the pick location with _better than 1mm precision_ (half-way grid) manually, using the magnified camera view. So the asymmetry around zero you name, where the input pick location was exactly between two grid-points, is artificial, it will never realistically happen in the real world. So the rounding-direction at the half-way point between two grid points ([in Java half away from zero](https://en.wikipedia.org/wiki/Rounding#Rounding_half_away_from_zero)) is entirely _irrelevant_.

> See table in original pull request.

See my **Important note 2**. I believe you did not notice that we're entirely in the _negative_ Y. So the table doesn't even contain any of the relevant relative Y. They start at Y-3.5mm for the 8mm tape and then go further down to Y-5.5mm for 12mm and so on in -2mm steps.

This misunderstanding about the right-handed 2D coordinates, and the sprockets being at the top might also explain why you thought the rotation is wrong.

_Mark

tmandys:
@markmaker so do you think dY should be always negative and is not only because of rotation bug? Part rotation in tape has no effect on pickposition evaluation. At least for right side Panda feeders there is bug in _test_ branch.

```
-                                // rotate from old angle
-                                relativePickLocation =  relativePickLocation.rotateXy(-pickLocation.getRotation())
+                                // rotate by tape angle, so dX multiply of partPitchMinMm, and dY pick-hole line distance
+                                relativePickLocation =  relativePickLocation.rotateXy(-angleTape)
```

markmaker:
Note, due to historical nomenclature in OpenPnP feeder classes, the term "pick location" is not very well defined/named. The feeder location does not actually store the _part_ pick location, it is more a _tape_ pick location, and orientation. It stores the pick location X, Y and the orientation (Rotation) of the tape in absolute machine coordinates. You can think of it as a coordinate _transformation_ (translate and rotate) that transforms from feeder local coordinates to machine coordinates. 

It was actually @tonyluken, who unified this thinking. At least on modern feeders (with tapes) it should work that way (there are non-tape feeders like the tray feeder, which have different concepts). 

The "Angle in Tape" of the part is then added on top, to give the final _part_ pick location.

So, I'm not sure about the rotation. There was a simplification, @prusk made when he refactored the code, discussed here:

https://github.com/openpnp/openpnp/pull/1623#issuecomment-2083846754

Maybe the rotation got lost then. Or maybe there was a bug all along. Sorry, I don't have time for deeper investigation.

_Mark

PR #1729, opened 2025-01-24
Do not show table cell editor immediately when click on unselected cell

tmandys:
# Description
When selecting row/cell in a JTable object, Parts, Packages, Jobs, .... then first click on any cell beyond current column opens cell editor or makes checkbox action. It is potentially dangerous because majority of clicks are for row selection. Especially at Job tab to select particular part where unless clicked on a "safe" cell then typically opens combobox list where can be made unintended change. Now the editor opens when cell is selected.

# Justification
General usability improvement

# Instructions for Use
Open e.g. Parts tab, click on a Package cell. Only row/cell is selected. Next click will open combobox list

# Implementation Details
1. How did you test the change? manually in GUI
2. Did you follow the ? yes
3. no
4. yes

### DIFF
--- a/src/main/java/org/openpnp/gui/components/AutoSelectTextTable.java
+++ b/src/main/java/org/openpnp/gui/components/AutoSelectTextTable.java
@@ -22,6 +22,7 @@ import javax.swing.SortOrder;
 import javax.swing.SwingUtilities;
 import javax.swing.table.DefaultTableModel;
 import javax.swing.table.TableColumnModel;
+import javax.swing.table.TableCellEditor;
 import javax.swing.table.TableModel;
 import javax.swing.text.JTextComponent;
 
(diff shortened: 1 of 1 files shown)


janm012012:
Hi @tmandys!

Many thanks for this PR! I think its a need feature. However, I'm not sure everyone will be happy with the modification, especially not if eg. multiple parts on the jobs placements panel need to be disabled. If we need to click twice now, I'm afraid it might cause rejection. I'd suggest to provide a configuration option to disable this behaviour, maybe in the View menu in a new "Option" entry.

In addition, I'm a big fan of an enabled button that needs to be toggled to allow editing (at least for the placements of a job). If switching tabs if may (but must not) go back into disabled state. This would remove the hassle of accidentally changing eg. parts...

Jan

tobydickenson:
@tmandys current implementation means that a **double-click** will toggle the "enabled" checkbox, or cause the placement part selector to drop down. That is easy enough for bulk edits, and solves the real problem of accidental data changes.

I would be happy to merge it to test (for the subsequent release) as it is, on the understanding that we might need to fix and subtle gui issues discovered during testing.

tmandys:
Note: it is not doubleclick but the first click selects cell and when selected then next click opens in-cell editor. So main reason was safely split row selection and editor request (or even checkbox change)

tobydickenson:
I propose merging this into 2.3 test branch. And, when we get to a beta release and put out a wider call for testing, we ask for feedback on the annoyance/safety balance of this UI change. @janm012012 thoughts?

PR #1730, opened 2025-01-24
New job order to allow use original placement list, i.e.one can defin…

tmandys:
# Description
Job processor sorts components by a key for placement. It bring uncertainity what is real order. User can define own placement order and use "sequence" options. Order will consider the same criteria (unplaced, enabled parts, nozzle strategy, ...)

# Justification
General usability improvment

# Instructions for Use
Select job order "sequence", define exact placement order xml definition and execute job

# Implementation Details
1. How did you test the change? manually
2. yes
3. there is new enumeration in xml
4. yes


janm012012:
Hi tmandys!

This feature is now available after merging PR #1658. However, it leaks usability because there is no way to see the current order or provide a custom order. As you already did a few nice updates to the UI, would you mind adding that? IMHO a new column showing the order would be required plus maybe a drag-drop feature or up/down buttons like on the profile actuators wizard to reorder placements.

Otherwise I'd suggest to close this PR.

Thank you again for your contributions!

Jan

tmandys:
Yes, I plan to implement general sortable table grid class. Handy not only for job.

Issue #1735, opened 2025-01-26
The dismiss button does not work for I&S "Unconventional Z Axis" warning


tobydickenson:
The dismiss button has no effect. It should have the effect of dismissing this warning.

I plan to investigate early next week.

tobydickenson:
Reported by jbussmann

https://groups.google.com/g/openpnp/c/edgNiukDUp4/m/1gyheAOUAgAJ

tobydickenson:
fixed by #1736

PR #1736, opened 2025-01-27
this warning should be a PlainIssue

tobydickenson:
# Description

This fixes issue #1735. A bug where the "unconventional Z axis" I&S warning could not be dismissed.

The bug was introduced in my PR #1682. In that PR markmaker suggested I make this warning message its own Solution, and pointed out some code to copy. But I copied too much of the surrounding code.

The fix mostly involves deleting code. Changing the class to `PlainIssue` makes this a warning that can be dismissed, rather than an issue with code that implements its own status changes.

# Justification
I propose including this bug fix in the imminent release. Issue #1735 turns this warning into a hard error, which is problematic for users affected by this situation. We want to warn them that their approach is unconventional, but not block them from continuing.

It would be good for the user who found this bug to confirm he is satisfied using a daily build

# Instructions for Use
n/a.

# Implementation Details
   * tested on my machine with my normal configuration - the warning is not shown
   * I set a limited safe z zone, which is indicative of the unconventional z axis configuration. The warning was shown
   * Pressing the "dismiss" button changes the I&S 

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/solutions/KinematicSolutions.java
+++ b/src/main/java/org/openpnp/machine/reference/solutions/KinematicSolutions.java
@@ -207,21 +207,14 @@ public class KinematicSolutions implements Solutions.Subject {
                                     else {
                                         if (bothSafeZoneHighAndLowEnabled
                                             && hm.getSafeZ().convertToUnits(LengthUnit.Millimeters).getValue() > 2.0) {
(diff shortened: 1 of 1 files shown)


jbussmann:
I cant test with the real machine right now, however using the NullDriver instead, it works for me as expected. Thanks a lot for your efforts.

tobydickenson:
Thank you for testing.

PR #1737, opened 2025-01-29
Create an Issue if no camera device has been selected

janm012012:
# Description
With this PR new issues are created in the Connect milestone if the camera device has not been selected yet.

# Justification
At present the connection to the controller is enforced only allowing to continue without a camera device selected. For symmetry reasons, camera devices shall be selected as well.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **using fresh setup**


janm012012:
Hi @tobydickenson!

Would you please take a look and help me enhance the text. I think the issues shall be present, but my descriptions are not very good yet... Thank You!

Jan

tonyluken:
I tried this. I'd change the issue to "Camera not connected to OpenPnP" and the solution to "On the camera's Device Settings tab select the correct Device and Format. Then click on the Capture FPS Test button. An image from the camera should appear in the camera's view pane."

Both strings should also be externalized and retrieved using the Translations.getString() method.

janm012012:
Many thanks, Tony!

I've applied your suggestions.

While testing them, I thought it would be nice to provide a button that gets the user to the camera's Device Settings tab. Do we have such a facility to reuse?

Jan

tonyluken:
That would be nice, but I don't know if such a facility already exists. I think there would be several other places that would be useful as well.

PR #1732, opened 2025-01-25
Version number 2.1

tobydickenson:
# Description
As discussed on google groups, this changes the version number from 2.0 to 2.1.

It adds this version number to Help/About.

It also changes the copyright message from:
Copyright © 2011 - 2019 Jason von Nieda
to:
Copyright © 2011 - 2025 Jason von Nieda and contributors

This change has been implemented in a low-impact low-risk manner, because this branch is very close to release. For the subsequent release (2.2?) I will tidy up this version number handling. Currently the version number appears in several places in the source code, in translation files, and some source code file names contain `_2_0`. This should all be much neater.

# Justification
The version number 2.0 has been used since 2019. We need to be able to distinguish this release from all that history.

# Instructions for Use
n/a

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted.
    * delete `~/.openpnp`, start openpnp, and check that the welcome dialog box looks right
    * Help / About looks right
3. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? -- no code changes
5. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. -- no changes
6. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. -- tests pass

PR #1739, opened 2025-01-30
Version 2.3

tobydickenson:
**DO NOT MERGE UNTIL AFTER THE NEXT RELEASE**

# Description
This adds a new file `VERSION.txt` containing just the version number; 2.3

openpnp.Main has a new method `getVersionString()`, and all the other places that previously included the version number in string constants now reference that method.

openpnp.Main has always had a method `getVersion` which retrieved the build date and git revision number. This is used, for example, at the start of the log file. This method now also includes the version number.

# Justification

As discussed on PR #1732 

# Instructions for Use
n/a

# Implementation Details
    * Same tests as #1732.
    * It would be good to check that the VERSION.txt gets correctly copied into all the various distribution packages, but the most expedient time to test that is after it is merged.
4. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? -- yes
6. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. -- tests pass


janm012012 (commented):
Hi Toby!

Many thanks for providing this PR! I'm wondering why you choose to store the version number in a separate file. Wouldn't it be sufficient to store it in the code, maybe in the getVersionString()?

Would you please revert the extra blank line that has crept in in AboutDialog.java line 44.

Would it make sense the add the version number to the file name, the installer generates? IIRC the information is generated calling "openpnp --version" (Main.java line 137 - without this PR).

Jan

janm012012 (approved):
Thank You, Toby!

PR #1744, opened 2025-02-05
Most script events do nothing

tobydickenson:
# Description

In PR #1688, Jan noted that in more than 99% of all events, no script will run. This patch introduces 2 optimisations for this common case.

# Justification

## Optimisation 1

Openpnp has to do quite a bit of work to establish that no script will be run. It has to open the `~/.openpnp2/scripts/Events/` directory, iterate over *all* the file entries in the directory, and process the file names.

In this patch, the event name is added to a HashSet if that iteration completes without finding any matching script files. When the same event is triggered a subsequent time, the HashSet allows it to skip the (relatively) expensive filesystem IO work.

Timing with java `System.nanoTime()` shows that this saves approximately 40µs on a system with no event scripts, on each call to `Scripting.on()`. Each saving is tiny, but it adds up. Note there are now lots of script events for micro-operations: Feeder.BeforeFeed, Feeder.AfterFeed, Nozzle.BeforePick, Nozzle.AfterPick, etc. This saves 40µs on each one.

One nice feature of this optimisation is that the implementation impact is limited to the Scripting class. The `feed()` method is 80µs faster (in the common case) with no changes.

The "Clear Scripting Engine Pool" menu handler has been extended to clear this HashMap too. This menu is now needed when adding a script. That will be familiar to script users, who currently need to use this menu after changing a script. I will add this to the wiki after merge.

## Optimisation 2

A client subsystem wishing to trigger a scripting event needs to prepare a HashMap of script parameters.

In most cases this is a very quick operation, and no optimisation is justified. However, if the process of preparing a HashMap is expensive then the caller can use a new method which allows it to skip the whole process if the scripting subsystem already knows that there are no scripts.

This optimisation is used by the caller of Job.Placement.BeforeAssembly, saving 110µs in the common case.

# Instructions for Use

The only user-visible change is noted above; the need to use the "Clear Scripting Engine Pool" menu after adding a script, not just when changing a script.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. -- tested on simulated machine. performance testing using `nanoTime()`

### DIFF
--- a/src/main/java/org/openpnp/scripting/Scripting.java
+++ b/src/main/java/org/openpnp/scripting/Scripting.java
@@ -34,10 +34,12 @@ public class Scripting {
     private final File eventsDirectory;
     private final HashMap<String, String> extensionToEngineNameMap;
     private final GenericKeyedObjectPool<String, ScriptEngine> enginePool;
+    private final HashSet<String> eventsWithoutScripts;
 
     public Scripting(File scriptsDirectory) {
         this.scriptsDirectory = scriptsDirectory;
         extensionToEngineNameMap = new HashMap<>();
+        eventsWithoutScripts = new HashSet<>();
         enginePool = new GenericKeyedObjectPool<>(
                 new ScriptEngineKeyedPooledObjectFactory(this.manager));
         // Allow unlimited engines, but evict all but five per key after a short idle time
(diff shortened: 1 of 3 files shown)


janm012012:
Hi Toby!

I like this extension. It's a good idea to remember if a script has been found earlier or not.

While looking thought the patch, I notices, that you added the method `eventHasScripts()`. I would name it `hasScript()` because to my knowledge no event can have more then one script and `event` is what the argument describes.

You may put your workaround recalculating the pick location into this PR as well. (Please remember to flag it as workaround so we know it can be safely removed once a better solution has been implemented.)

Jan

src/main/java/org/openpnp/scripting/Scripting.java
```diff
@@ -202,24 +204,57 @@ public void execute(File script, Map<String, Object> additionalGlobals) throws E
...
+            return true;
+        }
+    }
+
     public void on(String event, Map<String, Object> globals) throws Exception {
         Logger.trace("Scripting.on " + event);
         if (eventsDirectory == null) {
             return;
         }
+        if (eventsWithoutScripts.contains(event)) {
```
janm012012, line 228: Please use hasScript() here.

janm012012:
There is one more thing I notices: hasNoScript() is what you're collecting and what you know for sure if the contains() returns true. I would rename it like that to cover the third case (don't know yet) as false. (To me `true` should denote proven knowledge only.)

Jan

tobydickenson:
> (To me `true` should denote proven knowledge only.)

Thats better. Thank you

tobydickenson:
PR #1688 has been merged into this PR branch

tobydickenson:
These two PRs have been rebased and squashed to one commit for clean history.

Final testing to follow before I merge.

PR #1694, opened 2024-11-10
Package pad mark to indicate specific pin

tmandys:
# Description
Package layout defines pad position but there is no orientation mark so it is not possible guess what is supposed orienation on nozzle. The feature adds pad mark which is printed on selected pad. 

# Justification
General GUI enhancement

# Instructions for Use
Go to Packages/Footprint, there is new icon to optionally select just one pad, typically #1. See mark in package projection.

See thread https://groups.google.com/g/openpnp/c/YXGE2IURrfk with screenshots.

# Implementation Details
1. Tested manually in GUI
2. the came style as code around
3. new attribute in XML. Backward compatible as it has default value (False).
4. `mvn test` passed

PR #1749, opened 2025-02-11
Through-Hole Parts and Dynamic Safe Z

tobydickenson:
# Justification

Dynamic Safe Z lifts a part just enough to clear the obstacles. This height to which it is lifted is based on part height.

But this goes wrong for parts with through-hole features. For example, mechanical alignment pips, through-hole electrical pins, lenses on down-firing leds, and connectors with features that overhang the side of the board.

This was discussed [on google groups](https://groups.google.com/d/msgid/openpnp/8b21ded0-d044-4cc0-a117-31ff3e6a0498n%40googlegroups.com?utm_medium=email&utm_source=footer)

# Description

This PR adds a "Through-Hole Depth" property to Parts. Default zero for backwards compatibility.

There is an extra column on the Parts tab for this new property. There are tooltip messages for the two height fields.

Parts have a new method `getHeightForSafeZ` which is now used in safe z calculations. This method returns the sum of `getPartHeight()` and `getThroughHoleDepth()`.

# Wiki Update
[This section](https://github.com/openpnp/openpnp/wiki/User-Manual#parts) needs an updated screenshot. Also add:
 * **Height:** The physical height of the part as measured with calipers, or found in a data sheet. This is the distance between the board surface, and the point where the nozzle will pick it up. Part height is important because it determines how high above the board OpenPnP will stop when placing the part. You should only need to measure the part height for a specific part once, and then you can refer to the part in as many boards as you like.
 * **Through-Hole Depth:** The size of any through-hole feature which is sunken below the board surface when placed. This is added to the Part Height when determining how high the nozzle needs to be lifted to avoid obstacles when carrying the part. This should be zero for purely surface-mount SMT parts.

# Testing

On the simulated machine:
1. set the Z for the "Upper Strips 1" 0805 simulated strip feeder to -20
2. set the simulated zN1 axis Safe Zone Low to -10
3. perform a simulated pick. Note the DRO moves to -20 for pick, then -10 for safe Z
4. enable Dynamic Safe Z on simulated Nozzle N1
5. perform another simulated pick. Note the DRO moves to -20 for pick, then -9 for safe Z because the part height is 1mm
6. set the TH Depth for the simulated part to 0.5mm
7. perform another simulated pick. Note the DRO moves to -20 for pick, then -8.5 for safe Z.

This has also been tested in a production run placing connectors with a TH mechanical alignment pip.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. -- as above
6. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. -- **YES** `openpnp.model.Part` has a new property, and a new method.

### DIFF
--- a/src/main/java/org/openpnp/gui/PartsPanel.java
+++ b/src/main/java/org/openpnp/gui/PartsPanel.java
@@ -26,6 +26,7 @@ import java.awt.datatransfer.Clipboard;
 import java.awt.datatransfer.DataFlavor;
 import java.awt.datatransfer.StringSelection;
 import java.awt.event.ActionEvent;
+import java.awt.event.MouseEvent;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.io.StringReader;
@@ -171,7 +172,15 @@ public class PartsPanel extends JPanel implements WizardContainer {
 
         tabbedPane = new JTabbedPane(JTabbedPane.TOP);
 
-        table = new AutoSelectTextTable(tableModel);
+        table = new AutoSelectTextTable(tableModel) {
+            @Override
+            public String getToolTipText(MouseEvent evt) {
+                int column = columnAtPoint(evt.getPoint());
+                if(column==2) { return Translations.getString("PartsTableModel.Column.Height.toolTip"); } //$NON-NLS-1$
+                if(column==3) { return Translations.getString("PartsTableModel.Column.ThroughHoleDepth.toolTip"); } //$NON-NLS-1$
+                return null;
+            }
+        };
         table.setSelectionMode(ListSelectionModel.MULTIPLE_INTERVAL_SELECTION);
         table.setDefaultEditor(org.openpnp.model.Package.class,
                 new DefaultCellEditor(packagesCombo));
--- a/src/main/java/org/openpnp/gui/PartsPanel.java
+++ b/src/main/java/org/openpnp/gui/PartsPanel.java
@@ -177,7 +177,7 @@ public class PartsPanel extends JPanel implements WizardContainer {
             public String getToolTipText(MouseEvent evt) {
                 int column = columnAtPoint(evt.getPoint());
                 if(column==2) { return Translations.getString("PartsTableModel.Column.Height.toolTip"); } //$NON-NLS-1$
-                if(column==3) { return Translations.getString("PartsTableModel.Column.ThroughHoleDepth.toolTip"); } //$NON-NLS-1$
+                if(column==3) { return Translations.getString("PartsTableModel.Column.ThroughBoardDepth.toolTip"); } //$NON-NLS-1$
                 return null;
             }
         };
--- a/src/main/java/org/openpnp/gui/tablemodel/PartsTableModel.java
+++ b/src/main/java/org/openpnp/gui/tablemodel/PartsTableModel.java
@@ -78,7 +78,7 @@ public class PartsTableModel extends AbstractObjectTableModel implements Propert
 
     @Override
     public boolean isCellEditable(int rowIndex, int columnIndex) {
-        return columnIndex >= 1 && columnIndex <= 6;
+        return columnIndex >= 1 && columnIndex <= 7;
     }
 
     @Override
--- a/src/main/java/org/openpnp/gui/tablemodel/PartsTableModel.java
+++ b/src/main/java/org/openpnp/gui/tablemodel/PartsTableModel.java
@@ -40,7 +40,7 @@ public class PartsTableModel extends AbstractObjectTableModel implements Propert
             new String[] {Translations.getString("PartsTableModel.ColumnName.ID"), //$NON-NLS-1$
                     Translations.getString("PartsTableModel.ColumnName.Description"), //$NON-NLS-1$
                     Translations.getString("PartsTableModel.ColumnName.Height"), //$NON-NLS-1$
-                    Translations.getString("PartsTableModel.ColumnName.ThroughHoleDepth"), //$NON-NLS-1$
+                    Translations.getString("PartsTableModel.ColumnName.ThroughBoardDepth"), //$NON-NLS-1$
                     Translations.getString("PartsTableModel.ColumnName.Package"), //$NON-NLS-1$
                     Translations.getString("PartsTableModel.ColumnName.SpeedPercent"), //$NON-NLS-1$
                     Translations.getString("PartsTableModel.ColumnName.BottomVision"), //$NON-NLS-1$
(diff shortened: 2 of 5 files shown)


tobydickenson:
I would appreciate a review of this PR by someone familiar with Safe Z calculations.

janm012012:
Hi Toby!

Many thanks for providing this PR! It's a great feature improvement as there are many new components today that are partly sunken into the PCB.

I agree with your discussion on the list that "through-hole" is actually not the best word. Unfortunately I don't have any better then maybe "sunk depth". It would be cool if a native speaker could provide something better. I even checked a few datasheets and found none were this feature is named. It is good, that you routed everything through the translation layer. That makes it easy  to change the visible part later. However, changing the names of the variables is nearly impossible. That's why I'm a little concerned.

Reading the patch I'm not very happy with the hard coded `columnIndex` values. That suggests that you had to make much more changes then you actually did. Is it because of the UI editor? From reading the code, I'd prefer to convert them into an enum, but I'm not sure it that would be compatible with the UI editor. (PR #1694 has/had the same issue...)

I'd like you to verify one last thing: with this PR merged, we shall be able to safely handle parts, which total hight is given by the sum of Height and Through-hole Depth. For each nozzle tip we specify a Max Part Height. Are there any checks, that compare the actual part height (total part height) with this Max Part Height setting which now might require adjustments too? Checking the code I only found issues in case the remaining Z travel is less then Max Part Height for Dynamic Safe Z.

Besides that, feel free to merge.

Jan

tobydickenson:
> It would be cool if a native speaker could provide something better.

I will ask some colleagues tomorrow.

> hard coded columnIndex values

Agreed.

> I'd prefer to convert them into an enum

It would be nice to have a more complete abstraction than just an enum. There is also an array of column titles, and a separate array of column types, and it would be nice to not have to manually keep those all in sync.

> Are there any checks, that compare the actual part height (total part height) with this Max Part Height setting which now might require adjustments too?

I also have a doubt that something else might need to change, but haven't found anything yet.

janm012012:
> > I'd prefer to convert them into an enum
> 
> It would be nice to have a more complete abstraction than just an enum. There is also an array of column titles, and a separate array of column types, and it would be nice to not have to manually keep those all in sync.

Enums in Java can be extended like that. One can have the enum itself and related data (eg. https://howtodoinjava.com/java/enum/java-enum-string-example/).

Jan

tobydickenson:
An office survey revealed an interesting split. Some colleagues thought "Through Hole Depth" was ok. Some understood that "Through Hole" referred to being soldered on the opposite side, and therefore it specifically excluded mechanical alignment pegs etc. So I agree "Through Hole" could be misleading - for some users at least.

"Negative Height" was suggested. I dont like it,  but maybe.

I do like the text in the first paragraph on the PR, and will add this to the tooltip: "For example, mechanical alignment pips, through-hole electrical pins, lenses on down-firing leds, and connectors with features that overhang the side of the board."

Lets hold this PR for a bit and maybe we get a better idea.

ian-arkver:
What about "Through Board" parts? Implies something similar to through-hole without the solder reflow connotations? Obviously a hole is involved too, but... just a thought.

janm012012:
@tonyluken: do you have a good idea how this feature could be named?

tobydickenson:
> What about "Through Board" parts? Implies something similar to through-hole without the solder wave connotations? Obviously a hole is (usually) involved too, but... just a thought.

This is my preference so far. It is the best option at avoiding the possible misunderstanding of having any connection to the board bottom surface. I have updated the branch with this name.

I propose merging this next week unless there are any objections or better ideas.

janm012012:
Hi Toby!

It's good as it is. As you externalized all strings - which is very nice -, we could still change it later without breaking anything. Please feel free to merge it.

Nice feature, many thanks for adding it!

Just a very small remark for the future: please try to keep translation.properties sorted. If it is opened using eg. the editor that comes with WindowBuilder, it gets automatically sorted alphabetically. On save, you'll then have a difference that is larger then expected. (I know that the file is not properly sorted at present...)

Jan

jbussmann:
I know that this is already merged but how have you made sure that no existing code ignores the new through board depth by circumventing `getHeightForSafeZ()` and directly calling `getHeight()`?

I searched for "part.getHeight(" and found at least two occasions, where this potentially happens: [here](https://github.com/openpnp/openpnp/blob/b8fd4944c584a871ab45c300b9c6c97928923ae8/src/main/java/org/openpnp/gui/MachineControlsPanel.java#L252) inside `MachineControlsPanel.checkJogMotionSafety()` and [here](https://github.com/openpnp/openpnp/blob/b8fd4944c584a871ab45c300b9c6c97928923ae8/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java#L392) inside `ReferenceNozzle.moveToPlacementLocation()`. I think both cases should respect the height including the through board depth.

I also skimmed through all 200+ matches for ".getHeight(" and only found one additional instance of it referring to `Part`, which is `getPart().getHeight()` inside `PnpJopProcessor.JobPlacement.getPartHeight()`. But this doesn't seem to be used anywhere.

tobydickenson:
> I know that this is already merged but how have you made sure that no existing code ignores the new through board depth by circumventing `getHeightForSafeZ()` and directly calling `getHeight()`?

I was expecting I might have missed some....
 
> [here](https://github.com/openpnp/openpnp/blob/b8fd4944c584a871ab45c300b9c6c97928923ae8/src/main/java/org/openpnp/gui/MachineControlsPanel.java#L252) inside `MachineControlsPanel.checkJogMotionSafety()`

I agree this one needs changing too. Thank you for the review. New PR to follow.

> [here](https://github.com/openpnp/openpnp/blob/b8fd4944c584a871ab45c300b9c6c97928923ae8/src/main/java/org/openpnp/machine/reference/ReferenceNozzle.java#L392) inside `ReferenceNozzle.moveToPlacementLocation()`. 

This movement is the actual placing movement, which puts the part onto the solder paste. So I think this is correct as it is.

janm012012:
Hi jbussmann!

I've checked your concerns and agree with Tobys analysis: the first one (checkJogMotionSafety()) is valid (should take the new depth into the board into account) but the second (moveToPlacementLocation()) not (motion uses moveToLocationAtSafeZ() which lifts the nozzle high enough and final target is height about pcb surface).

Many thanks for checking! Do you see any other potential issues?

Jan

tobydickenson:
Thank you everyone for the great review work here.

A fix for the first is #1767

jbussmann:
Ok, this is fine. I had a misunderstanding of what `place()` does and glanced over that `getEffectiveSafeZ()` accounts for dynamic save z.

PR #1758, opened 2025-02-21
Kicad footprint import support for multiline pad definition

tmandys:
# Description
At least Kicad 8 uses "pretty" multiline footprint pad definitions. Previous OpenPnp import used regex requiring all pad stuff on one line. To work now we need join lines till matching right parentheses is found.

```
	(pad "1" smd roundrect
		(at -0.5 1.325)
		(size 0.6 1.55)
		(layers "F.Cu" "F.Paste" "F.Mask")
		(roundrect_rratio 0.25)
		(uuid "b519665e-8b7a-4a3f-a5dc-a119f4a60599")
	)

```

# Justification
General bug fix

# Instructions for Use
Import package pads from Kicad 8 .kicad_mod file

# Implementation Details
1. Import .kicad_mod into package pads 
2. yes
3. no
4. yes

### DIFF
--- a/src/main/java/org/openpnp/gui/importer/KicadModImporter.java
+++ b/src/main/java/org/openpnp/gui/importer/KicadModImporter.java
@@ -170,6 +170,40 @@ public class KicadModImporter {
             String line = reader.readLine();
             while (line != null) {
                 if (line.trim().startsWith("(pad ")) {
+                    int parentheses_cnt = 0;
+                    int pos = 0;
+                    do {
+                        boolean quoted = false;
+                        while (pos < line.length()) {
+                            switch (line.charAt(pos)) {
+                                case '(':
+                                    if (!quoted) {
+                                        parentheses_cnt++;
+                                    }
+                                    break;
+                                case ')':
+                                    if (!quoted) {
+                                        parentheses_cnt--;
+                                    }
+                                    break;
+                                case '"':
+                                    quoted = !quoted;
+                                    break;
+                            }
+                            pos++;
+                        }
+                        if (parentheses_cnt > 0) {
+                            String line2 = reader.readLine();
+                            if (line2 == null) {
+                                break;
+                            }
+                            line2 = line2.trim();
+                            if (line2 != "") {
+                                line += " " + line2;
+                            }
+                        }
+                    } while (parentheses_cnt > 0);
+
                     KicadPad kipad = new KicadPad(line.trim());
                     if (kipad.getType().equals("smd") && kipad.isTopCu()) {
                         Pad pad = new Pad();


tobydickenson:
I leaving this comment after reviewing the PRs that remain open. The code looks good but I am reluctant to merge this without a unit test.

We don't have 100% code coverage by unit tests in openpnp, but file parsers should be easy to test, and benefit greatly from having a test. In this case, it would give us confidence that the changes have not broken compatibility with old format kicad files.

Ideally this feature would already have a unit test, and this PR would simply need to add a few extra test cases. But it doesnt :-( So here are some suggestions on writing the unit test code:

The original code construction has poor testability because the `KicadModImporter()` method includes the logic that benefits from an automated unit test. But it also shows a popup dialog box to select the filename, which get in the way of writing an automated test. I suggest refactoring that method to:
```
public KicadModImporter() throws Exception {
    processImport(selectImportFile());
}

public File selectImportFile()) throws Exception {
    ... dialog box stuff
    return new File(...);
}

public processImport(File()) throws Exception {
    .... import logic
}
```

`src/test/java/ModelUnitsTest.java` is a good example for creating the unit test. Each test operates by creating an instance of the object being tested, calls some methods on the object (for this object it calls `processImport`), then makes some assertions that check the method has done the right thing.

Sample kicad files can go in a new directory under `src/test/resources/samples/`. `src/test/java/EagleMountsmdUlpImporterTest.java` shows how to access those files.

dkonigsberg:
I just started testing this, and it seems to mostly fix the KiCad footprint import feature (which is basically broken in the latest release, when importing current KiCad footprints).

There's one exception, however.  Footprint pads which use the "custom" pad type don't import.
I'm not sure of the correct way of tackling that one, since these would be non-rectangular pads, and I don't think OpenPnP has a way of representing such a thing.  My guess would be that you could make "best guess" pads that need manual tweaking, or something along those lines.  Not entirely sure how important this actually is to get right.

tobydickenson:
> I just started testing this, and it seems to mostly fix

Thank you for the test report.

> There's one exception, however. Footprint pads which use the "custom" pad type don't import.

Please can you attach some example files that demonstrate this.

----

As noted above, the main barrier to this getting merged is the lack of unit tests.

dkonigsberg:
[test-custom-footprints.zip](https://github.com/user-attachments/files/30189932/test-custom-footprints.zip)

Added a zip file with two KiCad footprints from one of my projects showing this issue:

- `SunLike_LED_S1S0.kicad_mod` - LED part where both pads are "custom"
- `Micrel_MSOP-10-1EP_3x3mm_P0.5mm_EP1.73x1.7mm_ThermalVias.kicad_mod` - IC with mostly normal pads, but where the center thermal pad is "custom"

Issue #1747, opened 2025-02-09
Zoom behaviour in camera view


jbussmann:
The zoom functionality in the camera view is implemented in a linear/additive manner roughly like so:
`zoom = zoom + mouseTicks * zoomIncrement` where `zoom` is valid in a range from 1 to 100.
This is very ill suited because an increment of 1 changes the zoom level anywhere from 1% to 100%. It would be much more useful to implement zoom in a logarithmic/multiplicative fashion like so:
`zoom = pow(zoom,  mouseTicks * zoomIncrement)` where `zoomIncrement` obviously needs to be a number larger than 1.
This would result in a much smoother and even zooming experience, especially on cameras with a large FOV, where a substantial part of the available range from 1 to 100 is effectively used.

tonyluken:
I don't see how your proposed equation would work. When zoom = 1, how would it ever change since 1 raised to any non-zero power is still 1? Perhaps you meant `zoom = zoom * pow(zoomIncrement, mouseTicks)`?

jbussmann:
You are of course right, this is how I meant it to be. I was a bit in a hurry and therefore glanced over two other things that might or might not be of importance.

- The code implies that if `renderingQuality == RenderingQuality.BestScale` is `true`, then `zoom` can only be incremented in integer steps. I don't understand why this is but since it initially increases `zoom` by a factor of 2, the proposed method would yield an initially equivalent behaviour if `zoomIncrement = 2`.

- Currently the zoom increments range from 0.001 to 10.0. Is there a reason for such extreme values? They both seem useless to me. Additionally, I don't see the current zoom value displayed anywhere so I argue that those numbers are meaningless to the average user and words like large/course, medium and small/fine would be better suited. As for the specific values of `zoomIncrement` I propose the middle three of the currently five values, simply incremented by 1.0.

jbussmann:
Adressed in #1766.

PR #1766, opened 2025-03-02
Change camera view zoom behaviour from linear to log

jbussmann:
# Description
This pull request changes how the zoom level is calculated, when a `MouseWheelEvent` is received. The calculation is now multiplicative instead of additive as before. The increments are adapted and reduced from 5 to 3 in number. Additionally, the numbers in the the context menu are replaced with text and a tooltip explanation because without any indication of the current zoom level, these numbers are meaningless.

During testing two issues emerged:
- The new increments need to be larger than 1, while previous increments could also be 1 or smaller. Because these settings are stored using the Preferences API, they prevail and cannot easily be reset. If the previous increment was 1, zooming is effectively disabled. If the previous increment was smaller than 1, zooming is inverted. This is likely to confuse users und must therefore be avoided. For this, the value is checked inside setCamera() after being loaded from the preferences and reset to default if it isn't one of the allowed values. The downgrade path doesn't lead to such confusing behaviour and will only result in the current setting not being marked in the popup, as none of the new values are match the previous values.
- When doing a careless right mouse click inside the camera view, dragging is registered and drag jogging starts. At the same time, the popup appears and drag jogging stays active, if the next click happens inside the popup and not inside the camera view. This detaches the drag cursor from the mouse pointer and the former remains hanging in the camera view. This is mitigated by blocking the appearance of the popup, if drag jogging has been started. I would prefer to block right mouse button drag jogging altogether but I'm not sure how to implement this in a way that it works with all platforms and mice.

Inside `createZoomIncMenu()` the increments are hardcoded as magic numbers although the are effectively stored inside `CameraView`. Additionally each number occurs twice and they need to match in order for the logic to work. ~~I wanted to replace them with a HashMap (as in Python dictionary), but failed to achieve it.~~ I got around to implement this using an EnumMap, which additionally allows to easily check if the loaded value is valid or not.

# Justification
Zooming the camera view is a handy feature, especially on cameras with an extreme FOV as the LumenPnP's bottom camera. The previous linear zoom behaviour steadily decreases it's effectiveness the further zoomed in the image is. This change introduces a logarithmic zoom behaviour with a constant effectiveness over the total zoomable range.

Adresses #1747.

# Instructions for Use
As before.

# Implementation Details
Manually by clicking and scrolling around the camera view while observing changes in the current values of `zoom` and `zoomInc`.
Yes
No
Done

### DIFF
--- a/src/main/java/org/openpnp/gui/components/CameraView.java
+++ b/src/main/java/org/openpnp/gui/components/CameraView.java
@@ -307,10 +307,10 @@ public class CameraView extends JComponent implements CameraListener {
 
         // load the zoom increment pref, if any
         zoomIncPerMouseWheelTick = prefs.getDouble(getZoomIncrementPrefKey(), DEFAULT_ZOOM_INCREMENT);
-// reset invalid value from before v2.3
+        // reset invalid value from before v2.3
         if (zoomIncPerMouseWheelTick <= 1) {
             setZoomIncPerMouseWheelTick(DEFAULT_ZOOM_INCREMENT);
-}
+        }
 
         // load sub.pixel rendering prefs, if any.
         try {
@@ -1783,7 +1783,7 @@ public class CameraView extends JComponent implements CameraListener {
         }
         @Override
         public void mouseReleased(MouseEvent e) {
-            if (e.isPopupTrigger()) {
+            if (e.isPopupTrigger() && !isDragJogging()) {
                 new CameraViewPopupMenu(CameraView.this).show(e.getComponent(), e.getX(), e.getY());
                 return;
             }
--- a/src/main/java/org/openpnp/gui/components/CameraView.java
+++ b/src/main/java/org/openpnp/gui/components/CameraView.java
@@ -49,9 +49,11 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
+import java.util.EnumMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.TreeMap;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


tonyluken:
To make the menu more meaningful to the operator, rather than have "Large", "Default", and "Small" options I think I would change the menu to read, "Number of mouse wheel ticks for 2x zoom" and have options of 1, 2, and 4. In other words, it is literally the number of mouse wheel ticks required to double the zoom level. Then zoomInc = pow(2, 1/mouseWheelTicksFor2XZoom) or approximately 2.0000, 1.4142, and 1.1892 respectively for those three options. The first option takes only 7 ticks to exceed 100x zoom, the second takes 14 ticks to exceed 100x zoom, and the last takes 27 ticks to exceed 100x (just fyi, with my mouse, I get about 10 ticks with one swipe of my finger, so those three options seem reasonable).

Speaking of exceeding 100x zoom, I see no real reason to enforce that limit anymore now that we can easily get there and back to 1x zoom relatively quickly.

jbussmann:
Do we care if the menu is meaningful? Does anyone care if the menu is meaningful? My answer is 2x a clear no. I can't think of a program where I know the zoom increment or where I can change the increment. Often it's just scrolling through a list of predefined and roughly geometrically spaced values. But in essence I really don't care. Also I fully expect to be able to sweep across the whole practical zoom range with one swipe (I do around 5 per swipe). This is the secondary reason for this change. Before it was 10, 100, 1k, 10k and 100k ticks to exceed 100x zoom.

I actually wanted to change the menu to just "Zoom" but figured this change might be hard to justify. But most definitely I don't want to make it even longer than it already is. As for the upper limit, I certainly would want to keep it, even more than before. Due to the much more effective behaviour, it's a lot easier to accidently reach a level where the Image seems broken because it's only a few pixels, in the worst case all black. Also note, I reduced the upper limit to 64, a power of 2. This is the simplest way to make sure best scale still uses an integer, even after running against the upper limit. Although this is less then before, it's still far more than I think is useful and I would even want to cap at 32.

As for the values, I think your suggestion of 7, 14 and 27 ticks to exceed x100 is a likely improvement over my implementation of 7, 49 and 463 ticks. I chose these, because they closer resemble the old behaviour and I have no knowledge about the reasoning for the original values. Additionally, I remember situations where mice would take very large scroll steps, potentially by sending multiple ticks at once. Having a large number of ticks might therefore actually be useful to a few individual users.

I will leave it as is until someone else favours your option too, or you make it a requirement for merging.

tonyluken:
Ok, how about a compromise? Changing the menu to read, "Zoom Sensitivity" with options of "High", "Medium", and "Low". And then add a tool tip to each option that describes the settings: "One tick of the mouse wheel changes the zoom by 2x", "Two ticks of the mouse wheel changes the zoom by 2x", and "Four ticks of the mouse wheel changes the zoom by 2x".

I see your point on keeping the max limit. 64x seems like a reasonable limit to me.

jbussmann:
> Changing the menu to read, "Zoom Sensitivity" with options of "High", "Medium", and "Low"

That's actually the idea I was looking for all along.

> And then add a tool tip

Personally, I don't think this to be necessary but added it for the purpose of the compromise.

Additionally, I put all the values inside an EnumMap, which allows for a cleaner upgrade path because the value can easily be checked to be part of the map, when being loaded.

During testing, I noticed that the new values are quite close to each other to the extent that the different sensitivities are almost indistinguishable to me. Which made me wonder if the feature of changeable sensitivity is actually really necessary? It was initially introduced in #984 and changed the zoom increment from 0.01 to 0.001/0.01/0.1/1/10 with the justification to increase zoom sensitivity. This PR changes the increments to equivalents of 0.19/0.41/1. Personally, I am in favour of removing the changeability altogether, provided the behaviour is logarithmic.

Finally, I updated the PR to reflect the changes by the additional commits.

tobydickenson:
Is this ready for merge?

jbussmann:
As far as I am concerned, it is ready for merge.

PR #1771, opened 2025-03-09
Fix Issues & Solutions auto-focus default on uplooking camera

jbussmann:
# Description
Make the focus choice on uplooking camera align with the cameras setting. If auto-focus is configured, enable it by default in the solution, otherwise not. 

# Justification
If users don't want to use auto-focus and disable it in the camera settings, Issues & Solutions should respect that and default to whatever the camera setting is. 

# Implementation Details
Check if changes in camera setting are reflected in Issues & Solutions, run uplooking camera calibrations on the simulated machine with and without auto-focus.
Yes
5. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.  
No
7. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.  
Done

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/solutions/VisionSolutions.java
+++ b/src/main/java/org/openpnp/machine/reference/solutions/VisionSolutions.java
@@ -44,6 +44,7 @@ import org.openpnp.machine.reference.ReferenceMachine;
 import org.openpnp.machine.reference.ReferenceNozzle;
 import org.openpnp.machine.reference.ReferenceNozzleTip;
 import org.openpnp.machine.reference.camera.AbstractSettlingCamera.SettleMethod;
+import org.openpnp.machine.reference.camera.ReferenceCamera.FocusSensingMethod;
 import org.openpnp.machine.reference.camera.AutoFocusProvider;
 import org.openpnp.machine.reference.camera.ReferenceCamera;
 import org.openpnp.machine.reference.camera.SimulatedUpCamera;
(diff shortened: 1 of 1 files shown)


janm012012:
Hi jbussmann!

I just tried to understand the code and failed. Maybe you can help me out. You changed the Issue to read the default setting using getFocusSensingMethod(). That's ok and likely a good idea. However, getFocusSensingMethod() reads fucusSensingMethod which defaults to None. This in tern changes the Issue to default to the unsafe/not recommended "Use nozzle Z location" option with a big red warning. I'm not convinced that this is easy to understand for new users.

Do you by chance have some background about the Focus Sensing Method dropdown you're trying to link the issue to? Why/when was it introduced?

Jan

jbussmann:
Hi Jan,

Thanks you for taking the time to look at this PR. I apologize for making you look at something I have no more intention to merge in the current state but haven't found the time to change it yet. I now updated the code and changed the description. Following some insight into my thinking for you and maybe others to better understand.

The core issue is that one can choose to use auto-focus in Issues & Solutions and one can choose to use auto-focus in machine settings. However these two settings are totally disconnected from each other.

Initially, my thinking was that Issues & Solutions should reflect the current machine settings (which is off by default). This is what I implemented in the first commit. In the discussion [here](https://groups.google.com/g/openpnp/c/DmtywCBtzbw/m/RTgdVeYqAQAJ) it was argued that this might not be the best way for everyone because auto-focus seems useful for others than me. It was also argued that new users should not need to fiddle with the machine tree to use auto-focus. This is why I changed the behaviour in the second commit so that the machine settings now reflect the choice made during calibration with Issues & Solutions. I hope it is now more understandable and it better reflects the recommended workflow.

What remains however is my disagreement over the usefulness of auto-focus and thus my stance that auto-focus should be disabled by default (not only in machine settings but also in Issues & Solutions). This likely originates in my lack of understanding of the inner working of the feature and the subsequent inability to use it in a repeatable fashion.

PR #1764, opened 2025-02-26
sort/reformat translations using the editor of WindowBuilder

janm012012:
# Description
This PR just sorts/reformats translations.properties using WindowBuilders editor to get it into a good shape again.

# Justification
In order to avoid unnecessary large diffs with new PRs and to follow the recommendation at https://github.com/openpnp/openpnp/wiki/Getting-Started-with-Eclipse#translations-in-eclipse translations.properties shall be kept sorted.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **na**
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? **na**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **na**
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. **na**

PR #1772, opened 2025-03-09
add a unit test to confirm that the properties file is in normal form

tobydickenson:
# Description

## New Unit Test

Add a unit test to confirm that the translations properties file is in normalised form. That is, a form which will not lead to spurious changes when opened in WindowBuilder.

If this test fails, the test output includes:
```
java.lang.Exception: 
This file containing translated text is not in normalised form:
./src/main/resources/org/openpnp/translations.properties
You can correct this by copying the corrected file from:
./src/main/resources/org/openpnp/translations.properties-normalised
```

The developer can fix this problem (restore the WindowBuilder-friendly format) by moving `translations.properties-normalised` to `translations.properties`

## Update the file

As noted on #1764 the existing translations properties file was in non-normalised form and needed to be fixed.

This PR includes the fixed file generated by this new unit test having failed on its first run. The diff looks very similar to that on #1764. I also fixed a bungled multi-line string in the original file.

# Status

Does this PR achieve its goal? Can WindowBuilder open and save this file without leaving spurious changes that would lead to a messy git history? I expect this will be ok, but I need to learn how to use WindowBuilder to confirm this. For now this is a provisional PR.

# Implementation Details
    * New unit test pass & fail conditions tested
    * One test is outstanding as described above


tobydickenson:
> Does this PR achieve its goal? Can WindowBuilder open and save this file without leaving spurious changes that would lead to a messy git history? I expect this will be ok, but I need to learn how to use WindowBuilder to confirm this.

This is confirmed. I made a minor change in WindowBuilder and it generated a minimal diff as expected.

This PR is ready for review.

tonyluken:
Thanks for this. I've run into this issue several times in the past. Just FYI, I've also seen WindowBuilder totally scramble the translations.properties file for some unknown reason. It usually works fine for me, but I suspect there is some particular sequence of doing things that can cause it to scramble the file. I never was able to reliably reproduce the problem so I can't say what I did to cause the issue.

janm012012:
Hi Toby!

I like this PR! It certainly supersedes PR #1764 and shall be part of the usual test sequence that is applied to every PR and build.

Looking into the code, I'd like to you to consider two things:

1. Could you indicate the line number where the diff is when generating the error message? This might help to understand the issue rather then just coping over a somehow fixed version (which is btw a nice feature).
2. There is a commended out section at line 85 that I'd rater not merge.

Many thanks for your nice work!

Jan

tobydickenson:
>     1. Could you indicate the line number where the diff is when generating the error message? This might help to understand the issue rather then just coping over a somehow fixed version (which is btw a nice feature).

I'm not sure about that. It is not like a source code error where we expect the IDE to jump to that line for the user to fix it themselves. If they looked at that line, they would not see anything "wrong" per se; it would just be different to the WindowBuilder canonical form.

But it does make sense to have a wiki page explaining this issue, and I will add that link to the error message.

>     2. There is a commended out section at line 85 that I'd rater not merge.

Will do

janm012012:
Many thanks, that looks good to me as it is now.

tonyluken:
@tobydickenson,

Since this change went in, I can't get the test to pass. It first gave me the message about the file not being in normalized form even though I thought it was. In any case, I deleted the existing translations.properties file, renamed the translations.properties-normalized file to translations.properties and reran the maven tests. Then a different maven test failed saying that the translations.properties file didn't have the correct line endings. So, I changed the line endings from CRLFs to LFs and re-ran the tests again. Then I was back to the test saying that the file is not in normalized form. 

I'm running on a windows machine and it appears that this line:

`properties.store(normalisedFileBodyWriter,newComments);` 

is using the system line terminator (CRLF for windows) which of course will cause the comparison to the original file (with LF line endings) to fail. It also appears to add a backslash escape to any # characters that are not already escaped. For instance, this line:

`PackageVisionWizard.PadsPanel.Action.ToggleMark.Description=Toggle mark for the currently selected pad, i.e. the indicate pin #1 or cathode.`

gets converted to:

`PackageVisionWizard.PadsPanel.Action.ToggleMark.Description=Toggle mark for the currently selected pad, i.e. the indicate pin \#1 or cathode.`

which, of course, also causes the test to fail.

I can fix the line ending problem by adding another matcher to convert all the CRLFs to LFs:

`
      normalisedFileBody = Pattern.compile("\r\n",Pattern.MULTILINE).matcher(normalisedFileBody).replaceAll("\n");
       normalisedFileBody = Pattern.compile("^#.*\n",Pattern.MULTILINE).matcher(normalisedFileBody).replaceAll("");
`

Probably need to add one for Mac line endings as well.

But I don't know what to do about the extra backslash on # characters. Do you have any ideas?

tonyluken:
@tobydickenson,

Ok, I figured out that the missing backslash escape was due to my error in merging. I had modified PackageVisionPanel to be PackageVisionWizard on that line and missed the fact that the backslash had been added with your PR. In any case, I am opening a new PR to fix the line ending problem.

PR #1773, opened 2025-03-12
enable the save&reset buttons only when a gui control has been update…

tobydickenson:
I am boldly sharing this small PR when I dont fully understand the original code!!! Review is needed please.

# Description

I am sure everyone is familiar with the "Feeder X changed.  Apply changes?" which pops up when changing to a different feeder by clicking on a different row in the feeder table. It also pops up when changing rows in other tables, if you have "linked tables" enabled. It pops up far too often, even when **there are no changes to be saved**.

This message is shown if [wizard.isDirty](https://github.com/openpnp/openpnp/blob/8eac82aeb97f79387f347133dc0142941519570d/src/main/java/org/openpnp/gui/FeedersPanel.java#L344) returns true, which checks if [the apply button is enabled](https://github.com/openpnp/openpnp/blob/8eac82aeb97f79387f347133dc0142941519570d/src/main/java/org/openpnp/gui/support/AbstractConfigurationWizard.java#L197l). 

That button gets enabled in the listener method changed by this PR. This is a callback in interface [AbstractBindingListener](beansbinding/src/org/jdesktop/beansbinding/AbstractBindingListener.java), called by [AutoBinding](beansbinding/src/org/jdesktop/beansbinding/AutoBinding.java), used by [AbstractConfigurationWizard](src/main/java/org/openpnp/gui/support/AbstractConfigurationWizard.java).

The Binding is used for bidirectional synchronisation of properies between two objects.
a. It synchronises from the model object (Feeder in this case) to the gui control directly. For example if the job is running in the background then the feeder's feedCount property is updated in the model as it performs a feed, and this gets updated in the gui immediately.
b. It synchronises in the reverse direction. This is where my understanding is vague. It synchronises *from* the gui control but I am not sure how the update to the model object is delayed until the Apply button is pressed.

Previously the Apply button was enabled in the `synced()` callback, which is called following synchronisation in either direction. This is conceptually not quite right, and the source of the original problem. It should enable the Apply button in scenario b but not scenario a. This PR changes it to use the `targetChanged` callback; we enable the Apply button if the gui form control has changed.

This change also affects the data flow in error cases. That is, if the synchronisation fails, for example if a number field has been edited with some text that can not be parsed as a number. `targetChanged` is called earlier, before that exception is raised, whereas `synced()` was only called if the synchronisation was successful. I do not think this difference matters; the `syncFailed` callback will sweep up all the difference.

# Implementation Details
     * A normal property editing cycle in a feeder wizard, manually pressing apply
     * A normal property editing cycle in a feeder wizard, changing feeder table rows and using the popup message
     * Changing feeder table rows while a job is running
     * Changing feeder table rows while a job is running, with an unsaved property edit
     * Editing a numeric feeder property: Enter "1"; the apply and reset buttons becomes enabled. Press the "a" key so the field contains "1a"; the field goes red because it is unparseable, the apply button is disabled, and the reset button remains enabled
     * Perform that "1a" process while a job is running
     * Perform that "1a" process while a job is running, then change feeder table rows
     * Repeat all of the above with a different type of wizard; I used the ReferenceControllerAxis wizard.
4. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. -- no, but openpnp.gui.support is heavily reused and I have not tested every use.

### DIFF
--- a/src/main/java/org/openpnp/gui/support/ApplyResetBindingListener.java
+++ b/src/main/java/org/openpnp/gui/support/ApplyResetBindingListener.java
@@ -24,6 +24,7 @@ import javax.swing.Action;
 import org.jdesktop.beansbinding.AbstractBindingListener;
 import org.jdesktop.beansbinding.Binding;
 import org.jdesktop.beansbinding.Binding.SyncFailure;
+import org.jdesktop.beansbinding.PropertyStateEvent;
 
 public class ApplyResetBindingListener extends AbstractBindingListener {
     private final Action saveAction;
@@ -40,7 +41,7 @@ public class ApplyResetBindingListener extends AbstractBindingListener {
     }
 
     @Override
-    public void synced(Binding binding) {
+    public void targetChanged(Binding binding, PropertyStateEvent event) {
         saveAction.setEnabled(true);
         resetAction.setEnabled(true);
     }


janm012012:
Hi Toby!

This PR has an amazingly small diff while promising a major enhancement. I agree with your, that apply/reset shall only be effected if changes have been applied on the UI and not if the model has changed and the UI is just updated to reflect them.

Besides feedcount, there are other use case where the "<feeder> changed" shall be suppress and will be by this PR: all types of vision operations like pick location calibration or part assignment updats. I think the user shall not be bothered to confirm them (they shall - and they very likely are already - noted in the logs).

Maybe we can requested @vonnieda as the original author to review this.

Jan

vonnieda:
Thanks for the ping. I'll have a look.

vonnieda:
I think this change is good and correct in that it does what it says it should. But, I think that there are probably some hairy wizards that may behave oddly with this change. 

I would suggest additional testing in OpenPnpCaptureCamera's wizards, especially with regards to hardware properties, and in the ReferenceStripFeeder with regard to feed count. Those are places that have always been problematic with the Apply/Reset stuff. 

Additionally, I'll reiterate my support of removing the Apply/Reset system altogether if someone wants to take that on. Ideally it would come with Undo support, but even without, I think users would generally prefer their changes to just be applied immediately in almost all cases.

Jason

janm012012:
Hi Jason!

Many thanks for your review!

> Additionally, I'll reiterate my support of removing the Apply/Reset system altogether if someone wants to take that on. Ideally it would come with Undo support, but even without, I think users would generally prefer their changes to just be applied immediately in almost all cases.

Yes, I'm with you. It has always been a source of frustration that changes got lost because Apply was not pressed in time. Maybe - at some later stage - a lock button would be needed to prevent accidental medications.

Jan

PR #1775, opened 2025-03-16
Fix PartsComboBoxModel and board import problem.

tonyluken:
# Description
Fixes the PartsComboBoxModel so that the current selection is not disturbed when the configuration parts list is updated. Also removes the associated propertyChangeListener from the configuration parts list when the comboBox is no longer needed - this was in effect a memory leak.

Fixes a problem with using the File->Import menu option that could potentially import into the wrong board if tables were unlinked.

# Justification
Bug fixes.

# Instructions for Use

# Implementation Details
1. How did you test the change? **Tested with scenarios that were previously causing feeders to erroneously have their part selection changed. Now they no longer change when the parts list is changed.**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **No changes**.
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. **All tests ran and passed.**

### DIFF
--- a/src/main/java/org/openpnp/gui/BoardPlacementsPanel.java
+++ b/src/main/java/org/openpnp/gui/BoardPlacementsPanel.java
@@ -602,6 +602,8 @@ public class BoardPlacementsPanel extends JPanel {
                 
                 importedBoard.dispose();
                 
+                tableModel.fireTableDataChanged();
+                
                 configuration.getBus()
                     .post(new DefinitionStructureChangedEvent(board, "placements", BoardPlacementsPanel.this)); //$NON-NLS-1$
             }
--- a/src/main/java/org/openpnp/gui/BoardsPanel.java
+++ b/src/main/java/org/openpnp/gui/BoardsPanel.java
@@ -273,7 +273,7 @@ public class BoardsPanel extends JPanel {
         });
     }
 
-    private void selectBoard(Board board) {
+    public void selectBoard(Board board) {
         if (board == null) {
             boardsTable.getSelectionModel().clearSelection();
             return;
--- a/src/main/java/org/openpnp/gui/FeedersPanel.java
+++ b/src/main/java/org/openpnp/gui/FeedersPanel.java
@@ -288,6 +288,11 @@ public class FeedersPanel extends JPanel implements WizardContainer {
 
                     Feeder feeder = getSelection();
 
+                    for (Component comp : configurationPanel.getComponents()) {
+                        if (comp instanceof AbstractConfigurationWizard) {
+                            ((AbstractConfigurationWizard) comp).dispose();
+                        }
+                    }
                     configurationPanel.removeAll();
                     if (feeder != null) {
                         priorFeederId = feeder.getId();
--- a/src/main/java/org/openpnp/gui/MainFrame.java
+++ b/src/main/java/org/openpnp/gui/MainFrame.java
@@ -91,6 +91,7 @@ import org.openpnp.gui.support.MessageBoxes;
 import org.openpnp.gui.support.OSXAdapter;
 import org.openpnp.gui.support.PropertySheetWizardAdapter;
 import org.openpnp.gui.support.RotationCellValue;
+import org.openpnp.model.Board;
 import org.openpnp.model.BoardLocation;
 import org.openpnp.model.Configuration;
 import org.openpnp.model.Configuration.TablesLinked;
@@ -914,6 +915,7 @@ public class MainFrame extends JFrame {
 
                 @Override
                 public void actionPerformed(ActionEvent e) {
+                    boardsPanel.selectBoard((Board) jobPanel.getSelection().getPlacementsHolder().getDefinition());
                     boardsPanel.getBoardPlacementsPanel().importBoard(boardImporter.getClass());
                 }
             });
--- a/src/main/java/org/openpnp/gui/support/AbstractConfigurationWizard.java
+++ b/src/main/java/org/openpnp/gui/support/AbstractConfigurationWizard.java
@@ -122,6 +122,14 @@ public abstract class AbstractConfigurationWizard extends JPanel implements Wiza
         applyAction.setEnabled(false);
         resetAction.setEnabled(false);
     }
+    
+    /**
+     * Override this method if the wizard needs to do any cleanup like removing property change 
+     * listeners that may have been added during the wizard's construction
+     */
+    public void dispose() {
+        
+    }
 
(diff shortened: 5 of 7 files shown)


janm012012:
Hi Tony!

Great, that you found the issue and cool that you already solved it. I just verified, that the workflow reported on the list and the one I noticed are now fixed. However, I still see the lag between importing placements and switching back to the feeders tab, I mentioned on the list. Do you by change have an idea where this might come from? (I tried profiling using visualvm a few weeks back, however I still did not manage to get function level profiling to work. I just did not found out how to get across "Performing instrumentation, please wait...". Do you have any suggestion on that?)

Initially I though there might be other places where the same might happen (with packages or vision) but I did not found any. They all use the new inheritance model that works without drop down.

The code looks good to me. You may merge it when ready. 

Jan

tonyluken:
Hi @janm012012,

I don't really see much of a delay on my computer. I added some instrumentation to the BoardPlacementsPanel.importBoard(Class<? extends BoardImporter>) method and saw that the "Pitot Rev B mount top panel.csv" file was read in 17ms, adding the parts to the board took 42ms, and the rest of the time was insignificant. I can switch back to the Feeders tab about fast as I can move the mouse so I'm not sure where the delay is coming from that you are seeing. My test scenario is to start OpenPnP, go to the Feeders tab, select Lever-22, select the Job tab, add a new Board to the job (giving it a new file name), the use the File->Import Placements->Reference CSV, select the csv file, and click Import. I also tried it with tables linked and unlinked and it didn't appear to make much difference. Are you doing something differently?

janm012012:
Hi Tony!

Many thanks for trying to understand my issue! I've setup a debug session within Eclipse and prepared it with the files Greg send us. I then switch to feeders tab and select Lever-22. If I now select a different (Lever) feeder using eg. the curer keys, it takes about 1..2 seconds until the UI has been updated. If I keep Lever-22 selected, switch over to the boards tab, create a new board and import the top data, this all works with no noticeable lag. If I now switch immediately back to the feeders tab, it takes about 15 seconds until the UI is up to date. (While preparing this response, I once switched to the feeder tab a few minutes after importing the data and then the UI was up to date almost instantaneously.) Does the import queue something that takes a significant amount of time to be processed?

I would say that my PC is not super fast but also not incredible old/slow, somewhat in the middle, a some years old i7-4790 with 16G. If you have a faster PC, you shall still see about half of the delays, I guess.

According to VisualVM, there is significant CPU load (on one core) and if I understand the Sampler -> CPU -> thread CPU time view correctly in "AWT-EventQueue-0". If switching to the feeder tab after the import it's the same thread that has maximum CPU usage.

Jan

tonyluken:
Hi @janm012012,

I am using Greg's files as well.

>Does the import queue something that takes a significant amount of time to be processed?

Not that I know about other that it does fire off some property change events when it's done that causes all the GUI displays to be updated as appropriate. Before my change to remove the parts property change listener when switching feeders, it was possible to keep building up more and more hidden parts comboboxes (actually I think since the combobox was still linked via the listener, the entire feeder wizard was still linked and therefore could never be garbage collected) which, when the events were fired, would all go through the exercise of updating their lists even though they could never be seen by anyone - so that could bog things down eventually (more and more CPU and memory usage). I wouldn't be surprised if there are more property change listeners with similar issues, but they are just hard to find since everything works ok even if they should have been removed. I find them to be one of the more difficult things to keep track of and remove when they are no longer needed. 

My pc has an Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz with 16GB ram running Windows 10 Pro. I compared my CPU vs yours
[here](https://cpu.userbenchmark.com/Compare/Intel-Core-i7-8700K-vs-Intel-Core-i7-4790K/3937vs2384) and it shows that I have 6 cores 12 threads vs your 4 cores 8 threads - don't know if that could be the difference.

I am also running OpenPnP in debug mode via Eclipse. I tried selecting different feeders using the curser keys and notice no delay. If I press and hold the curser key, I can induce noticeable a lag in the GUI and it takes several seconds to recover once I release the key. But I don't think that is what you are talking about.

I have never tried setting up and running a profiler, so I am of no help there.

janm012012:
Hi Tony!

> My pc has an Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz with 16GB ram running Windows 10 Pro. I compared my CPU vs yours [here](https://cpu.userbenchmark.com/Compare/Intel-Core-i7-8700K-vs-Intel-Core-i7-4790K/3937vs2384) and it shows that I have 6 cores 12 threads vs your 4 cores 8 threads - don't know if that could be the difference.

OK, that's likely not a big difference.

If you don't see my issue, then it's likely a local (my) problem only. Hence I'd suggest to merge your fix to solve the issue Greg reported.

Just for interest: does the propertyChange feature has something to group updates to prevent excessive use like updating the parts dropdown you fixes only once even if many parts are added or removed maybe only if the respective UI gets in view?

Jan

tonyluken:
Ok, I went ahead and merged the changes.

>Just for interest: does the propertyChange feature has something to group updates

I really don't know. It's all wrapped up inside Java Beans which I don't know about very much.

tonyluken:
Looks like ScriptingTest is failing for some reason during the build process under Java 19. AFAIK, I didn't change anything related to scripting and don't know anything about the test.

janm012012:
> Looks like ScriptingTest is failing for some reason during the build process under Java 19. AFAIK, I didn't change anything related to scripting and don't know anything about the test.

I've seen build and deploy issues as well while merging a PR a few days before. If you continue to see issues, I'd suggest to contact Jason and ask him to have a look.

tonyluken:
@janm012012,

I was able to duplicate the delay in the GUI updating you were seeing. If I turn on table linking, exit and restart OpenPnP, then I start getting the delay (about 5 seconds). Once I get the delay, I continue to see it regardless of how the table linking is set. To recover back to minimal delay, I need to set table linking off, exit and restart OpenPnP. Once it's back to minimal delay, then it continues to operate with minimal delay regardless of how I have table linking set. I'll try to figure out what is causing it.

janm012012:
Hi Tony!

Many thanks for persistence! I tried your findings but can't actually get it to sync with my experience: for me switching between (reference push pull) feeders feels laggy. It feels more laggy with linked table selection and less without. And I don't notice a difference if OpenPnP is restarted. I'm sorry, that I can't configure...

Jan

PR #1779, opened 2025-03-18
Job error handling

tobydickenson:
# Justification

Each placement currently has an Error Handling property. Either 'Alert' or 'Defer'.
There are some placements where I definitely want "Alert". I dont want it dumping lots of expensive parts in the discard pile.
There are some placements where I want "Defer". If the number of placements for one part exceeds my feeder capacity then I want the job to continue when the feeder is inevitably exhausted.
But for most placements the desired error handling policy depends on whether or not **I am near the machine** and want to respond to errors as they occur.
So I suggest it would be nice to have a separate global error handling property in the job configuration, and have a prominent gui button which the operator can toggle when they are near the machine.

# Description

This feature has been discussed on google groups twice, [thread 1](https://groups.google.com/d/msgid/openpnp/CAH35ure%3DJW1cQyRueP7qsoqQx%3D4NWmYLKBSXNXGGNQ3DEOQubQ%40mail.gmail.com?utm_medium=email&utm_source=footer) [thread 2](https://groups.google.com/d/msgid/openpnp/706e24e9-b2b3-44d6-8ef6-d5a2c0d176de%40googlemail.com). 

Mark suggested that this fitted into a new class which I would call "Part Handling and Placement Settings", which would include properties like Pick Retry Count and Speed %, and could be assigned at Placement, Part, or Package level. That architecture makes some sense, but:
1. It is a big project that I dont have time for right now.
2. It is mostly orthogonal to this change; setting the error handling policy at job level with a big gui button. 

# Instructions for Use

Instructions will go [here](https://github.com/openpnp/openpnp/wiki/User-Manual#handling-errors)

When a job is running and something goes wrong, OpenPnP will try to recover from the problem if it can. In many cases, though, OpenPnP will stop and ask you how to proceed. You will see this message box if the job is set to alert on errors, or if a placement-specific error handling policy has been changed to "Alert". Otherwise the machine will continue and errors will be reported at the end of the job.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. -- tested in production runs


tobydickenson:
This has become a valuable  part of our production workflow.

I propose merging later this week.

PR #1780, opened 2025-03-18
non-portable solutions

janm012012:
# Description
This PR contains a proposal to extend Issues & Solutions with non-portable solutions. Solutions can now be marked as non-portable. If a solutions is marked as such, its hash value (that identifies it internally) is appended by some local data (currently the mainboard's serial number) which makes it considered unsolved if evaluated on a different computer (mainboard).

# Justification
In the past a lot of support issues where raised because a machine.xml was shared and the user was not aware of which issues he still has to solve and which are still ok as they have been solved by the provide. With this PR some Issues will automatically open up if a machine.xml is shared.

# Instructions for Use
Existing setup will be handled as if the machine.xml has just been copied over. To prevent this, the flag "disableNonPortableSolutions" may be set to true which considers all solutions a portable again as is the current state.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **using simulation and physical machine making sure the solutions I currently would expect to open up are open**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **yes: Solutions() has been extended**

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/solutions/CalibrationSolutions.java
+++ b/src/main/java/org/openpnp/machine/reference/solutions/CalibrationSolutions.java
@@ -206,7 +206,8 @@ public class CalibrationSolutions implements Solutions.Subject {
                                 "Calibrate backlash compensation for axis "+axis.getName()+".", 
                                 "Automatically calibrates the backlash compensation for "+axis.getName()+" using the primary calibration fiducial.", 
                                 Solutions.Severity.Fundamental,
-                                "https://github.com/openpnp/openpnp/wiki/Calibration-Solutions#calibrating-backlash-compensation") {
+                                "https://github.com/openpnp/openpnp/wiki/Calibration-Solutions#calibrating-backlash-compensation",
+                                true) {
 
                             {
                                 tolerance = oldAcceptableTolerance;
@@ -318,7 +319,8 @@ public class CalibrationSolutions implements Solutions.Subject {
                         "Advanced camera "+camera.getName()+" calibration.", 
                         "Automatically calibrates the camera "+camera.getName()+" using the primary and secondary calibration fiducials.", 
                         Solutions.Severity.Suggestion,
-                        "https://github.com/openpnp/openpnp/wiki/Calibration-Solutions#advanced-camera-calibration") {
+                        "https://github.com/openpnp/openpnp/wiki/Calibration-Solutions#advanced-camera-calibration",
+                        true) {
 
                     @Override 
                     public void activate() throws Exception {
@@ -376,7 +378,8 @@ public class CalibrationSolutions implements Solutions.Subject {
                     "Calibrate precise camera ↔ nozzle "+nozzle.getName()+" offsets.", 
                     "Use a test object to perform the precision camera ↔ nozzle "+nozzle.getName()+" offsets calibration.", 
                     Solutions.Severity.Fundamental,
-                    "https://github.com/openpnp/openpnp/wiki/Calibration-Solutions#calibrating-precision-camera-to-nozzle-offsets") {
+                    "https://github.com/openpnp/openpnp/wiki/Calibration-Solutions#calibrating-precision-camera-to-nozzle-offsets",
+                    true) {
 
                 private Location oldNozzleOffsets = null;
 
@@ -476,7 +479,8 @@ public class CalibrationSolutions implements Solutions.Subject {
                     "Advanced camera "+camera.getName()+" calibration.", 
                     "Automatically calibrates the camera "+camera.getName()+" using the nozzle "+defaultNozzle.getName()+".", 
                     Solutions.Severity.Suggestion,
-                    "https://github.com/openpnp/openpnp/wiki/Calibration-Solutions#advanced-camera-calibration") {
+                    "https://github.com/openpnp/openpnp/wiki/Calibration-Solutions#advanced-camera-calibration",
+                    true) {
 
                 @Override 
                 public void activate() throws Exception {
--- a/src/main/java/org/openpnp/machine/reference/solutions/VisionSolutions.java
+++ b/src/main/java/org/openpnp/machine/reference/solutions/VisionSolutions.java
@@ -290,8 +290,8 @@ public class VisionSolutions implements Solutions.Subject {
 
     public class VisionFeatureIssue extends Solutions.Issue {
 
-        public VisionFeatureIssue(Subject subject, ReferenceCamera camera, Length featureDiameterIfKnown, String issue, String solution, Severity severity, String uri) {
-            super(subject, issue, solution, severity, uri);
+        public VisionFeatureIssue(Subject subject, ReferenceCamera camera, Length featureDiameterIfKnown, String issue, String solution, Severity severity, String uri, boolean isNonPortable) {
+            super(subject, issue, solution, severity, uri, isNonPortable);
             this.camera = camera;
             featureDiameter = 20;
             if (camera.getUnitsPerPixelPrimary().getX() != 0
@@ -300,6 +300,9 @@ public class VisionSolutions implements Solutions.Subject {
                 // Existing diameter setting.
                 featureDiameter = (int) Math.round(featureDiameterIfKnown.divide(camera.getUnitsPerPixel().getLengthX()));
             }
+        }
+        public VisionFeatureIssue(Subject subject, ReferenceCamera camera, Length featureDiameterIfKnown, String issue, String solution, Severity severity, String uri) {
+            this(subject, camera, featureDiameterIfKnown, issue, solution, severity, uri, false);
         }
 
         private ReferenceCamera camera; 
@@ -708,7 +711,8 @@ public class VisionSolutions implements Solutions.Subject {
                     "Determine the up-looking camera "+camera.getName()+" position and initial calibration.", 
                     "Move the nozzle "+defaultNozzle.getName()+" over the up-looking camera "+camera.getName()+" and capture the position.", 
                     Solutions.Severity.Fundamental,
-                    "https://github.com/openpnp/openpnp/wiki/Vision-Solutions#up-looking-camera-offsets") {
+                    "https://github.com/openpnp/openpnp/wiki/Vision-Solutions#up-looking-camera-offsets",
+                    true) {
 
                 @Override 
                 public void activate() throws Exception {
--- a/src/main/java/org/openpnp/model/Solutions.java
+++ b/src/main/java/org/openpnp/model/Solutions.java
@@ -24,6 +24,9 @@ package org.openpnp.model;
 import java.awt.Color;
 import java.beans.PropertyChangeListener;
 import java.beans.PropertyChangeSupport;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
@@ -68,10 +71,61 @@ public class Solutions extends AbstractTableModel {
     @Attribute(required = false)
     private boolean showIndicator = true;
 
+    /**
+     * The flag disableNonPortableSolutions considers solutions marked as non-
+     * portable as portable.
+     * Non-portable solutions are intended to make sure they are raised even
+     * if a machine.xml was shared. This may happen if a manufactures sells
+     * identical machines to multiple users. Sharing a machine.xml makes it
+     * easy to start but otherwise hides the requirement that some issues
+     * and calibration steps have to be solved anyhow.
+     * The disableNonPortableSolutions flag is considered as workaround for
+     * machines that have been setup already to prevent raising solutions that
+     * have already been solved locally.
+     */
+    @Attribute(required = false)
+    private boolean disableNonPortableSolutions = false;
+
     private boolean showSolved;
 
     private boolean showDismissed;
 
+    /**
+     * Provide local unique data that is populated with the Mainboard's
+     * serial number.
+     */
+    private static String localUnique = null;
+    
+    public Solutions() {
+        // create a local Unique string using the mainboards serial number
+        // code taken from https://stackoverflow.com/questions/35698321/how-to-get-motherboard-serial-number-on-gui-in-java
+        if (!disableNonPortableSolutions) {
+            try
+            {
+                String result = null;
+                Process p = Runtime.getRuntime().exec("wmic baseboard get serialnumber");
+                BufferedReader input
+                        = new BufferedReader(new InputStreamReader(p.getInputStream()));
+                String line;
+                while ((line = input.readLine()) != null)
+                {
+                    result += line;
+                }
+                if (result.equalsIgnoreCase(" ")) {
+                    System.out.println("Result is empty");
+                } else
+                {
+                    localUnique = result;
+                    Logger.trace("Mainboard serial number read, localUnique set to " + localUnique);
+                }
+                input.close();
+            } catch (IOException ex)
+            {
+                Logger.error("Reading mainboard serial number failed: " + ex);
+            }
+        }
+    }
+    
     public enum Milestone implements Subject, Named {
         Welcome(Translations.getString("Solutions.Milestone.Welcome.name"), //$NON-NLS-1$
                 "welcome",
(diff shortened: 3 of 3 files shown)


tobydickenson:
This is currently using a PC serial number. But the PC is the one component we can change without affecting I&S solutions.

My lumenpnp M115 command response contains a UUID. I assume that is is a machine id, which would be more suitable. I will ask on the lumenpnp discord where that UUID comes from.

tobydickenson:
Confirmed that [lumenpnp Marlin is using the stm32 device id](https://github.com/sphawes/Marlin/blob/f80c299ec48b672088166ec29591f163dbf9a12d/Marlin/src/gcode/host/M115.cpp#L81)

An extract from my machine.xml (with some hex digits changed to x):
```
<detected-firmware><![CDATA[FIRMWARE_NAME:Marlin bugfix-2.1.x (Mar 15 2024 14:51:01) SOURCE_CODE_URL:github.com/MarlinFirmware/Marlin PROTOCOL_VERSION:1.0 MACHINE_TYPE:LumenPnP REV04 EXTRUDER_COUNT:0 AXIS_COUNT:6 UUID:00xxxxxxxxxxxxxxxxxxxxxx]]></detected-firmware>
```

janm012012:
Hi Toby!
Many thanks for your suggestion. However, I have concerns with it:

- I'm not aware, that the serial number/UUID is available on other controllers. Maybe Duet/RRF has it, but my Smoothie V1 does not have it.
- If we link issues to this serial number, we have to start with a situation in which it is not available yet. M115 is part of the connect solution. Yes, we could update the local unique part once it has been read. But before, we have to handle that somehow different.
- How would be know, that the firmware shall be read again to get the new serial number? Once you connected to your machine, this issue has been solved. If we'd declare it non-portable, we do not have information that signals that it might have changed.

It seems, that the UUID is likely a good idea if paired with other local unique data. One could use eg. the mainboard serial number for the connect issue and the controllers serial number for backlash and other calibration. How would be code that? Turn isNonPortable into an enum that can be set to Portable, NonPortablePC and NonPortableController?

The other question that we need to discuss is what Issues shall be raised if the controller/machine has changed? Currently I've "Determine the up-looking camera Up Looking position and initial calibration.", "Calibrate backlash compensation for axis [XY]", "Calibrate precise camera ↔ nozzle <x> offsets." and "Advanced camera Up Looking calibration." declared non-portable. Advanced camera calibration for down-looking is also declared non-portable but not visible due to isOverridingOldTransformsAndDistortionCorrectionSettings() returning true. I don't yet understand if that's ok or needs different handling.

Jan

alexanderwiller:
I have a somewhat-adjacent comment that concerns this multi-machine topic:

With more than one machine and having slot feeders with a unique ID, the information which part is loaded in a given feeder needs to be synchronized across all machines. This is because a part can be assigned to a feeder on one machine (or even external to the machine, with some maintenance tool!), then the feeder can be loaded into any machine.

Once we get there with our setup, we will be looking at implementing a database feature, which when enabled will save all slot feeder config not in the `machine.xml`, but in said database. This feature enables a critical aspect of multi-machine capability for us.

**This was just the intro to the actual point:**

The `machine.xml` contains quite some _machine state_ that frequently changes - as opposed to _machine config_, which requires more deliberate action to change. This makes version control of machine config difficult, as ephermal machine state blows up the size of the diff, making it hard to interpret or to maintain a "clean repo" state.

Separation of state and config could be a desirable characteristic in version-controlled and multi-machine scenarios. The case of the I&S system portability is one of those challenges in multi-machine scenarios, as is the slot feeder case.

I wonder if this general view of "state vs. config" is something on the radar in the context of "configuration portability" and just wanted to bring attention to it. Hope this is a suitable place, if not, I can create a separate issue for it.

janm012012:
Hi Alexander!

Many thanks for your comment. I'm with you that machine.xml shall be split into related concerns. However IMHO that's at present almost impossible or would result in a nasty solution because the framework that's used to read xml files can not handle excessive elements. It just generates an exception. I've seen some places where transformations inside the xml have been done, so it's not impossible. It would just require someone to try it... There is also light at the end of the tunnel that we might get a replacement for the xml framework. It's in a quite good state, so it will hopefully turn into a PR soon. This might make the splitting easier.

If you'd like to the splitting, I'll do my best to assist.

Jan

tobydickenson:
If M115 is being used for this purpose then it would be reasonable to refetch that data when the machine is enabled.

Currently I&S only works when the machine is enabled and homed. But I have occasionally thought it would be better if this was not the case. If I&S worked on cached machine state then we could write regression tests that check what issues are reported for a variety of machine states, and this might have caught #1774.

Jan, what issues would be reopened when the PC mainboard changes? I had assumed that I should to be able to transfer my machine.xml (etc) to a new PC and carry on without interruption?

I'm not sure what number could be used on linux. You can maybe extract the mainboard serial number with `dmidecode`, but that requires root permission. Some systems (raspberry pi for example) do not have that number.

alexanderwiller:
@janm012012

Thanks for the feedback. For now, I'll put this on our internal todo list. We want to become more active on the mailing list and I expect to bring this up again in a more appropriate context. Will watch the PRs for the XML framework progress.

**On the matter of this PR:**

How about creating a file like `.machine-uuid`, that is generated when OpenPnP is launched for the first time? Would remove all the complexity of determining a "suitably unique" ID some other way, and the naming implies it should not be duplicated to any other machine.

janm012012:
Hi Alexander!

> How about creating a file like `.machine-uuid`, that is generated when OpenPnP is launched for the first time?

That's a very good idea. I like it because it provides solutions for the two major issues this PR tries to solve:

1. changing PC is easy because one just copies the UUID along the other settings.
2. sharing a machine.xml is safe because the UUID is a separated file and not shared.

I'll try that!

Jan

jbussmann:
I already shared my thoughts on this topic [here](https://groups.google.com/g/openpnp/c/_QEXHaW3Y6s/m/V5nL6j6PAgAJ) as I thought about this as well.

I came to the conclusion that there is no way to reliably identify a majority of unique machines. While the STM MCUs have a UUID and Marlin exposes this, this is not a general practise as far as I know. And even if, how are controller replacements handled?

alexanderwiller:
I think this introduces a lot of complexity while not addressing the key matter:

The pairing of (machine.xml, physical machine) makes a machine unique. In this "uniqueness definition", exchanging the controller to a replacement one (thus potentially having a new serial number) or exchanging the PC (once again, potentially giving a new serial number) does not affect the uniqueness of the machine.

I'm fine with calling the file `machine-uuid` without hiding it, and adding a comment line explaining what it is about. In addition, there should be instructions on "how to use a config with multiple machines" in the wiki.

I am strongly against having a "save as template" button or similar mechanisms, that feels like putting the cart before the horse and completely breaks any future gitops workflow, which requires that the config can just be transferred as-is, without any GUI interaction.

**In addition to the `machine-uuid`, I have no objections against saving the PC serial number in this file too and prompting the user in case of a mismatch, asking them about there intentions and linking them to the wiki page about how machine identification works.** I think it's not either-or, but at the core should be a UUID that is generated if none was generated yet.

alexanderwiller:
Sorry if I was a bit harsh about the "save as template" button, I think the concept has merit and I followed up on the general problem of configuration portability with a bigger write-up in it's own issue: https://github.com/openpnp/openpnp/issues/1782

janm012012:
Hi @jbussmann!

> I already shared my thoughts on this topic [here](https://groups.google.com/g/openpnp/c/_QEXHaW3Y6s/m/V5nL6j6PAgAJ) as I thought about this as well.
> 
Yes, I followed the discussion and it drove me to prepare this PR as a forum to discussion the details.

> I came to the conclusion that there is no way to reliably distinguish a majority of unique machines. While the STM MCUs have a UUID and Marlin exposes this, this is not a general practise as far as I know. And even if, how are controller replacements handled?
> 
Yes, that's the major issue against the obvious solution and the reason why discussions are required.

> My proposal is to track at least the controller firmware, the PC hardware, the OS and maybe also the OpenPnP build number. If any of these change, the user is asked to choose between this being a backup from this exact machine or a template from another machine. In case of the latter, all non portable configs are cleared from machine.xml. I think this is a more comprehensive approach to the underlying issue.
> 
"Tracking" controller firmware or OpenPnP build versions is likely nothing that would be accepted by the majority of users. The same applies to brute force and forced PC identification.

> > How about creating a file like .machine-uuid, that is generated when OpenPnP is launched for the first time?
> 
> I don't think this is of much use because people just copy the whole config folder.

But isn't that exactly what users expected to do if they e.g. move their machine controls to a different PC or back it up?

Going one step back, I've to repeat my findings about how Issues & Solutions works:

- Issues are created on the fly based on the machine configuration and state and other things like active Milestone. E.g. if the machine is switched off you see other issues then if switched on. This makes it hard to collect _all_ issues.
- Issues are collected first and then compared against a list and an overwrite button ("Include Dismissed?") before been presented to the user.
- Issues are identified internally using a hash calculated from visible text elements.  This hash is stored in lists of solved and dismissed issues. Due to the hash at least I can't say which of the Issues shall be reopened.

Now we can go back to what this PR is trying to archive: allow support to share a machine.xml with preconfigured data in that reliably detects the shared nature and presents its user only thus issues that can not be shared. In addition this PR shall consider existing setups as entirely solved.

To me the hash as identification criteria and the inability to get a list of all issues makes it difficult to "just" remove non-portable issues form the solved/dismissed lists. I therefore modified the hash generation to apply uniqueness to thus issues, that are marked as non-portable. This may blows up the solved/dismissed list with issues, that will never appear because if they are solved/dismissed on one computer I can not identify them on an other (with a different uniqueness). I'd consider this acceptable.

So the remaining question is how to generate uniqueness and how to use it. And here Alexanders proposal looks like the best idea so far to me: with uniqueness stored in a separate file, one can safe/restore/transfer it and one can decide not to share it. I think what is considered uniqueness is more or less irrelevant as long as collisions are unlikely. So even a time stamp would be sufficient for this purpose. If enriched with a MAC address the uniqueness is almost perfect. And finally if the uniqueness in the file is manually reset to null, OpenPnP could consider this as what we currently have, all issues are considered solved if solved once using any computer and any machine.

Em I missing anything?

Jan

janm012012:
Hi!

I've now changed the implementation to a) use an enum to specify the non-portability in a more verbose manner and b) take uniqueness from a file which gets initialized to a random UUID if it does not exist. If the file exists, but is empty, the uuid is considered null falling back to the state we currently have where all issues/solutions are considered portable.

I'd say that this how might provide a way to allow machine.xml files to be safely shared. Some solutions will then popup automatically (due to the changed uuid) while most will be still considered solved. Please have a look if there are any other solutions that shall be considered non-portable as well.

It seems, that most other files are read using methods that Configuration provides. I'm unsure if reading the uuid file shall be moved there too.

What do you think?

Jan

janm012012:
Hi all!

I've now moved the uuid file handling to Configuration.java where all the other .xml files are handled as well. This provides a cleaner solution. I'd say this is ready for testing how. We might need to rearrange the list of issues to get them into a reasonable order, but that's something for an other PR. What do you think?

Jan

tobydickenson:
This looks good to me

tobydickenson:
Hi Jan,

Thinking about this overnight, and I have changed my mind.....

Consider all the current users who have completed I&S then upgrade to this version. Does this change mean that their backlash compensation (for example) gets marked as incomplete? I deduce this will happen because the pre-uuid I&S solution was complete, but the new post-uuid I&S solution is not. That is an undesirable speed bump for all existing users.

Who does benefit from this feature?

l think it is a mistake to think that regular lumen users will benefit from this.  Opulo have been consistent in their policy that [backlash does not need to be calibrated on each machine](https://docs.opulo.io/misc/backlash-compensation/). Their approach is to provide a super-simple setup that gets their users placing their first board with the minimum setup time. That approach got me through my first few boards. It is not the approach I choose today, but I cant argue it is wrong for their customers.

Consider the situation where you, or I, or any other user familiar with I&S is setting up a clone of an existing machine.  Would you use this feature? Would you change the uuid? Probably not because it would be no harder to just reopen and redo the I&S solutions that you know would be affected.

jbussmann:
Unfortunately I lack the time to fully review the PR but I would like to add some thought none the less, as I drafted them already a week or so ago. Because in my opinion, this implementation has several shortcomings.
- There is a reasonable chance that users won't know or understand that this mechanism exists oder how it works. They might just copy this UUID file, either because they figure out that it works better this way, because the are unaware about what it does or maybe even because the creator of the config instructs them to do so (I can easily imagine several motivations for that).
- As I understand, this only affects the state of Issues & Solutions but not the configurations themselves? We know there are users who don't use Issues & Solutions at all and I can see how one can just ignore it if one is provided with a working configuration. In that case, this implementation is ineffective, unless not only the solutions but also the non-portable configuration is effectively reset.
- As I pointed out earlier, there is no way to make this 100% reliable and waterproof. Therefore, and because I'm convinced it is overall the much better approach, the user should be educated about the issue through the GUI. Everyone with the intention to ignore non-portable config can still do it but those doing it without intention are at least provided with the information to proceed with an intentional choice.
- Adding to the above, I fully agree with Mark on problem 1, as he outlined it in [this](https://groups.google.com/g/openpnp/c/UpUNJFcruu0/m/tdVFAzyCCQAJ) post. While I really appreciate Jan's efforts, I think our efforts would be much better invested in problem-preventing education of the user-base.

> That approach got me through my first few boards. It is not the approach I choose today, but I cant argue it is wrong for their customers.

Same here, as only about a month into using my LumenPnP, I started to seriously think about Issues & Solutions. Before that I would have just ignored everything this PR is about (unless it effectively wipes parts of the config file).

> Their approach is to provide a super-simple setup that gets their users placing their first board with the minimum setup time.

I thought this too at that time. Meanwhile I'm working on a suggestion for an Issues & Solutions based config procedure and although I thought this to be impossible, I'm convinced this is sooo much quicker than all the manual steps their procedure provides.

jbussmann:
Hi Jan,

First off, I really advocate for such a portability feature, I simply disagree with the specific implementation. At the same time, I lack the time to implement the idea I have in mind myself so I'm fully prepared to accept that in the end, another solution might be merged and my opposition to it is ignored. 

> - I'm sorry, I don't know how to fight bad guidance

Here I also agree with Mark's suggestions: "Make a autoritative setup video from scratch. Such a video should be doable via community effort." As mentioned before, I have an idea of such a (LumenPnP specific) thing in mind. I'm looking forward to sharing it and receiving feedback, as soon as it somewhat starts taking shape.

> - What would you think if we combine this PR with some "Tip-of-the-day"

While initially starting to read your comment, another not totally different idea came to my mind. How about if a "change in config" is detected e.g. by a UUID mismatch, the user is presented with a dialog with tick-boxes to choose what should be reset or not.

Apart form that I'm still firmly convinced that for triggering this, a multitude of indicators should be tracked. Like motherboard ID, OS type/version, OpenPnP version, controller firmware/version. Because I'm convinced that a high number of false positives can be tolerated to minimize false negatives. I also remain firmly convinced that not only the solved solutions but also the configuration itself should be reset, so it makes no difference whether one uses Issues & Solutions or not.

janm012012:
Hi jbussmann!

Many thanks for your commend! I appreciate your participation in this discussion!

> > * I'm sorry, I don't know how to fight bad guidance
> 
> Here I also agree with Mark's suggestions: "Make a autoritative setup video from scratch. Such a video should be doable via community effort." As mentioned before, I have an idea of such a (LumenPnP specific) thing in mind. I'm looking forward to sharing it and receiving feedback, as soon as it somewhat starts taking shape.
> 
Yes, I'll with you!  We shall try to set that up. Unfortunately I've not experience at all in this, so I can't help...

> > * What would you think if we combine this PR with some "Tip-of-the-day"
> 
> While initially starting to read your comment, another not totally different idea came to my mind. How about if a "change in config" is detected e.g. by a UUID mismatch, the user is presented with a dialog with tick-boxes to choose what should be reset or not.
> 
> Apart form that I'm still firmly convinced that for triggering this, a multitude of indicators should be tracked. Like motherboard ID, OS type/version, OpenPnP version, controller firmware/version. Because I'm convinced that a high number of false positives can be tolerated to minimize false negatives. I also remain firmly convinced that not only the solved solutions but also the configuration itself should be reset, so it makes no difference whether one uses Issues & Solutions or not.

There is a fundamental "problem" (implementation decision) that makes this all almost impossible - at least to my current knowledge: 

- issues/solutions are generated on the fly based on the current milestone and based on the current configuration state (eg you don't see advanced driver issues if no driver has been configured/detected). So you almost never get a list of all solutions.
- Solved/Dismissed is handled using a map of hash values that are derived from the issues text elements. Due to the nature of hash values, you can not deduce from a hash value to the issue. So you can only go forward: is an issues popes up, you can search the maps if it has been solved or dismissed but not the other way around.

All this leads to the problem that you can not reset solved or dismissed issues unless you know their hash value.

Jan

kbralten:
> * There is a reasonable chance that users won't know or understand that this mechanism exists oder how it works. They might just copy this UUID file, either because they figure out that it works better this way, because the are unaware about what it does or maybe even because the creator of the config instructs them to do so (I can easily imagine several motivations for that).

I'd rephrase this problem as solving "things that aren't right when someone gives you the configuration from another machine". If it's descoped to that, it feels much simpler to create a mechanism for "giving someone a configuration" than "detecting a configuration might not be correct in an innumerable set of scenarios". I'd suggest a nice "export configuration" menu/script that either strips out machine-specific state or alternately strips everything that isn't considered portable. If we count on the small minority of people shipping around configuration bundles to use the tool it's simple. Alternately we could empower everyone with an "import configuration" tool that strips it while importing.

PR #1789, opened 2025-03-27
An empty delay command should be treated the same as a missing delay command

tobydickenson:
# Description
PR #1699 added driver-side-delaying. This includes a new configurable gcode command, and fallback logic for the upgrade scenario when that gcode has not been configured.

This PR extends the fallback logic if the DELAY_COMMAND an empty string. Previously the fallback logic was only enabled if DELAY_COMMAND was null.

# Justification
I&S guides the user to set some appropriate gcode DELAY_COMMAND. The I&S solution shows that the gcode is initially blank, and shows the new recommended gcode.

After performing this upgrade, a curious user might want to revert back to java-side-delaying. Maybe to perform a comparison, or maybe because the effect was not quite what he expected. However this does not work. This fallback logic is triggered only if the gcode variable is null, but the gui gcode editor can only set it to an empty string.

# Implementation Details
    * I was that "curious user".
    * Tested on real machine: After the first upgrade to 2.3, I&S upgrade path, and the gcode editing "downgrade" path.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java
@@ -685,7 +685,7 @@ public class GcodeDriver extends AbstractReferenceDriver implements Named {
     @Override
     public boolean delay(int milliseconds) throws Exception {
             String command = getCommand(null, CommandType.DELAY_COMMAND);
-        if (command == null) {
+        if (command == null || command.isEmpty()) {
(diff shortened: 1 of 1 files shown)


janm012012:
Hi Toby!

This is a good change. If that's what happens if the respective field is cleared, that it's a good catch. Unfortunately I never tried that. Only upgrading when the command has never been seen yet or with the command as proposed by I&S.

Please feel free to merge it when ever you're ready.

Jan

PR #1793, opened 2025-04-08
Fix more GUI memory leaks

tonyluken:
# Description
Most wizards created under the Parts, Packages, Vision, Feeders, and Machine Setup tabs are leaking memory due to the lack of cleanup of property bindings when the wizards are removed (just before a new set of wizards are created). This happens whenever a new item is selected in any of the tables (or machine tree) on all of the above-mentioned tabs. To fix this, a "dispose()" method was added to WrappedBinding and AbstractConfigurationWizard to perform the cleanup of bindings. Also, several of the "wizard" tabs under the Parts, Packages, Vision, Feeders, and Machine Setup tabs did not extend AbstractConfigurationWizard but instead JPanel so they were changed to extend AbstractConfigurationWizard. While performing this conversion, the names of the new wizards (which used to be JPanels) were changed to end with Wizard and moved to their proper packages.

Several wizards were directly using Beans Bindings so those were converted to use the binding methods provided by AbstractConfigurationWizard so that the new dispose() method could be used to clean up the bindings when the wizards were no longer needed.

Fixed the ReferencePushPullFeederWizard to show the clone template status in a JScrollPane so that the entire status is visible to the operator.

Fixed the bug where imports would fail if using the File->Import Placements... menu while on the Boards tab.

# Justification
Bug fixes.

# Instructions for Use
No special instructions needed. Most of these changes should be invisible to the operator.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **Started OpenPnP, looked through all the GUI tabs and verified all still appear to be working. However, GUIs are notoriously difficult to test completely so this PR would be worthy of additional testing by others.**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **Only one minor change was made to Part.java to change the setPickRetryCount method to include the old value of the pick retry count in the property change event that it generates.**
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. **All maven tests were run and passed.**


src/main/java/org/openpnp/gui/support/PartsComboBoxModel.java
```diff
@@ -64,38 +66,28 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+        //Compute the indices to remove and those to add to modify the current list
+        List<int[]> indicesToRemove = new ArrayList<>();
+        List<int[]> indicesToAdd = new ArrayList<>();
+        Collect.computeInPlaceUpdateIndices(partsList, parts, indicesToRemove, indicesToAdd);
+        
+        //Remove any parts from the existing combo box that are not in the new list of parts. The
+        //current combo box selection will not change as long as it is not one of the parts being
+        //removed.
+        for (int[] range : indicesToRemove) {
+            for (int idx=range[0]; idx>=range[1]; idx--) { //Reverse order indexing!
```
janm012012, line 78: Does this eventually use the wrong end/start indices?

src/main/java/org/openpnp/gui/support/PartsComboBoxModel.java
```diff
@@ -64,38 +66,28 @@ public void propertyChange(PropertyChangeEvent evt) {
...
-                continue;
-            }
-            else if (cmp > 0) {
-                while (idx < this.getSize() && comparator.compare(part, (Part) getElementAt(idx)) > 0) {
-                    idx++;
-                }
-                if (idx == this.getSize() || comparator.compare(part, (Part) getElementAt(idx)) < 0) {
-                    insertElementAt(part, idx);
-                }
+        for (int[] range : indicesToAdd) {
```
janm012012, line 84: Are this ranges still valid after the remove action?

src/main/java/org/openpnp/gui/tablemodel/PartsTableModel.java
```diff
@@ -169,20 +170,71 @@ public Object getValueAt(int row, int col) {
...
+                    fireTableCellUpdated(rowIdx, 7);
+                    break;
+                case "placementCount":
+                    fireTableCellUpdated(rowIdx, 8);
+                    break;
+                case "assignedFeeders":
+                    fireTableCellUpdated(rowIdx, 9);
+                    break;
+                default:
+                    //ok - property is not visible in the table so no need to update the table
```
janm012012, line 207: How would we detect if a new property has been added that is not handled here yet? I'd consider it safe to handle all properties and generate a warning if any new/unknown property is spotted.

src/main/java/org/openpnp/gui/tablemodel/PartsTableModel.java
```diff
@@ -169,20 +170,71 @@ public Object getValueAt(int row, int col) {
     @Override
     public void propertyChange(PropertyChangeEvent arg0) {
         if (arg0.getSource() instanceof Part) {
-            // Only single part data changed, but sort order might change, so still need fireTableDataChanged().
-            fireTableDataChanged();
+            //Only single part data changed, so notify the table as to which cell to update
+            int rowIdx = indexOf(arg0.getSource());
+            switch (arg0.getPropertyName()) {
+                case "id":
+                    fireTableCellUpdated(rowIdx, 0);
```
janm012012, line 177: I flagged that in an other PR as well: would be possible to convert this numbers/digits into an enum? It looks as if this '0' indicates a column in a table that might change (get out of sync) if the table is redesigned. Using an enum defined at a single place would be safe. We might be even able to combine propertyName with the value into the enum.

src/main/java/org/openpnp/gui/tablemodel/PartsTableModel.java
```diff
@@ -169,20 +170,71 @@ public Object getValueAt(int row, int col) {
...
+            
+            //Compute the indices of those parts to remove and those to add
+            List<int[]> indicesToRemove = new ArrayList<>();
+            List<int[]> indicesToAdd = new ArrayList<>();
+            Collect.computeInPlaceUpdateIndices(parts, newParts, indicesToRemove, indicesToAdd);
+
+            //Remove the unneeded parts in reverse order so as to not disturb indices of those 
+            //parts that are yet to be removed
+            for (int[] idxRange : indicesToRemove) {
+                for (int idx=idxRange[0]; idx>=idxRange[1]; idx--) {
```
janm012012, line 222: Same as above: is the start/end correct for reverse removal?

src/main/java/org/openpnp/gui/tablemodel/PartsTableModel.java
```diff
@@ -169,20 +170,71 @@ public Object getValueAt(int row, int col) {
...
+                fireTableRowsDeleted(idxRange[1], idxRange[0]);
             }
-            parts = new ArrayList<>(Configuration.get().getParts());
-            fireTableDataChanged();
-            for (Part part : parts) {
-                part.addPropertyChangeListener(this);
+            
+            //Insert any needed parts into the table
+            for (int[] idxRange : indicesToAdd) {
+                for (int idx=idxRange[0]; idx<=idxRange[1]; idx++) {
```
janm012012, line 232: Save as above: is the start/end correct? it uses the same order [0] as start and [1] as end, so unless the ordering is different for indicesToAdd compared to IndicesToRemove, one is wrong. Please try to use an enum for the indices instead of '0' and '1' because then the code is more verbose or create new variables which indicate Start/End.

src/main/java/org/openpnp/gui/wizards/PackageNozzleTipsWizard.java
```diff
@@ -28,27 +28,29 @@
...
         this.pkg = pkg;
         createUi();
     }
     private void createUi() {
-        setLayout(new BorderLayout(0, 0));
-        scrollPane = new JScrollPane();
-        add(scrollPane);
+//        setLayout(new BorderLayout(0, 0));
+//        scrollPane = new JScrollPane();
+//        add(scrollPane);
```
janm012012, line 49: If this code is no used anymore, I'd remove it. Same applies for line 51/52.

src/main/java/org/openpnp/gui/wizards/PackageVisionWizard.java
```diff
@@ -250,17 +249,8 @@ public void valueChanged(ListSelectionEvent e) {
         tablePanel.add(tableScrollPane);
 
         showReticle();
-        initDataBindings();
+//        initDataBindings();
```
janm012012, line 252: The method does not exist anymore. I'd suggest to remove the comment.

src/main/java/org/openpnp/gui/wizards/PartSettingsWizard.java
```diff
@@ -69,20 +64,19 @@ private void createUi() {
...
         
         lblNewLabel = new JLabel(Translations.getString(
-                "PartSettingsPanel.pickConditionsPanel.pickRetryCountLabel.text")); //$NON-NLS-1$
+                "PartSettingsWizard.pickConditionsPanel.pickRetryCountLabel.text")); //$NON-NLS-1$
         pickConditionsPanel.add(lblNewLabel, "2, 2, right, default");
         
-        pickRetryCount = new JTextField();
-        pickConditionsPanel.add(pickRetryCount, "4, 2, left, default");
-        pickRetryCount.setColumns(10);
+        txtFieldPickRetryCount = new JTextField();
```
janm012012, line 70: If this variable is of type "JTextField", I'd suggest to name it something like "textField" and not "txtField".

src/main/java/org/openpnp/util/Collect.java
```diff
@@ -143,4 +144,122 @@ public static <T> List<T> extractSublist(List<T> list, int ... indices) {
...
+     * Each entry is a two element array of indices [a, b] where a >= b. All elements in the range a
+     * down to b (inclusive) should be removed from the listToModify. The list of index ranges is 
+     * sorted highest to lowest so that the removals don't change the indices of elements that are 
+     * yet to be removed.
+     * @param indicesToAdd - a list of index ranges that are of the elements in targetList that are 
+     * to be added to listToModify. Each entry is a two element array of indices [a, b] where 
+     * a <= b. All elements in the range a to b (inclusive) need to be copied from the targetList 
+     * and inserted into the listToModify. The list of index ranges is sorted lowest to highest so 
+     * that the insertion point in listToModify is the same as the index of the item in targetList.
+     */
```
janm012012, line 190: That's very nice documentation and almost completely invalidates my concerns about the remove/insert ranges I added! There is just a single note I'm missing: if indicesToAdd is designed to be applied _after_ indicesToRemove, it shall be noted as such.
Maybe you can add a generic method that applies the add/remove operations, which you currently do manually...

janm012012:
Hi Tony!

Thank you for tackling this difficult topic. It sounds like we have some sloppy programming merged. To prevent this in the future, do you think it would be possible to add tests that detect and flag this? Lets say flag anything that does not extend AbstractConfigurationWizard and/or WrapperBinding or flag code that reference JPanel?

Looking into your diff, I noticed, that a few files just import org.openpnp.gui.support.MultisortTableHeaderCellRenderer. That looks like an unused import. However, when compiling the code without this import (tried for BoardPlacementsPanel.java) it does not work anymore generating a error message that MultisortTableHeaderCellRenderer() could not be resolved. Looking into the old code, MultisortTableHeaderCellRenderer() is used but there is no related import. How does that work?

In FeedersPanel.java an empty line has to be slipped in. Unless you insist, I'd revert that.

I put a few more comments to individual lines in the diff. Other then that I did not noticed anything of relevance.

Overall this is a very impressive PR and a huge improvement in cleanup. I'd love to see it getting merged quickly.

I've tested the changes in my simulation machine (with ReferencePushPullFeeders) and did found any issues.

Jan

src/main/java/org/openpnp/gui/support/PartsComboBoxModel.java
tonyluken: Nope, starts removing from the end so no indices change during the removal process.

src/main/java/org/openpnp/gui/support/PartsComboBoxModel.java
tonyluken: Yes, it is designed so that the removal needs to happen first (and in reverse order) and then the insertions occur starting at the beginning and working towards the end.

src/main/java/org/openpnp/gui/tablemodel/PartsTableModel.java
tonyluken: That is up to the software designer. If the table is changed to display more properties, it is the table model that needs to be changed. There is nothing to currently "handle" with properties that are not displayed in the table.

src/main/java/org/openpnp/gui/tablemodel/PartsTableModel.java
tonyluken: As I stated, if the table is redesigned, then it is the table model that gets changed. It might make sense at that point to change to an enum but since it's all in one file, I don't really see the need - at least not for this PR.

src/main/java/org/openpnp/gui/tablemodel/PartsTableModel.java
tonyluken: Yes.

src/main/java/org/openpnp/gui/tablemodel/PartsTableModel.java
tonyluken: Yes, the ordering is different for the two. You really need to read the javadocs for the  Collect.computeInPlaceUpdateIndices(...) method. I'm not sure what the enum would be. Start and End, or Hi and Low? When removing, we need to start with the highest index and go down to the lowest index. When inserting. we need to start with the lowest index and go up to the highest.

src/main/java/org/openpnp/gui/wizards/PackageNozzleTipsWizard.java
tonyluken: I'll take care of that.

src/main/java/org/openpnp/gui/wizards/PackageVisionWizard.java
tonyluken: I'll take care of that.

src/main/java/org/openpnp/gui/wizards/PartSettingsWizard.java
tonyluken: I'll take care of that (although I'm not the one who wrote that code).

src/main/java/org/openpnp/model/Part.java
```diff
@@ -22,6 +22,7 @@
 import org.openpnp.ConfigurationListener;
 import org.openpnp.machine.reference.vision.AbstractPartSettingsHolder;
 import org.openpnp.spi.Feeder;
+import org.pmw.tinylog.Logger;
```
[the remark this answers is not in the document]
tonyluken: Agreed.

src/main/java/org/openpnp/util/Collect.java
tonyluken: I will update the documentation to make it clear that the removals need to take place before the insertions are made. I also thought about adding a static method to do the removals and insertions, but it's not that simple since there is often more work to do than just that. For instance, in the PartTableModel, there are listeners to be removed and/or added as well as table events to be fired.

tonyluken:
>To prevent this in the future, do you think it would be possible to add tests that detect and flag this?

Well, just about anything is possible given enough time and programming ability, but I just don't think it's worth the effort. I think we just need a bit more diligence when reviewing PRs.

>Looking into your diff, I noticed, that a few files just import org.openpnp.gui.support.MultisortTableHeaderCellRenderer

Yeah, that is because I moved it to a different package so now those files need it as an import whereas before it was in the same package as those files (so it didn't need to be imported).

>In FeedersPanel.java an empty line has to be slipped in

I'll take it out.

tonyluken:
@janm012012,

Did you by chance notice if this resolves the "sluggishness" you've been seeing when scrolling through feeders on the Feeders tab?

janm012012:
Hi Tony!

Many thanks for merging this PR! It's a great improvement!

> Did you by chance notice if this resolves the "sluggishness" you've been seeing when scrolling through feeders on the Feeders tab?

I just did: running Gregs machine insight the debugger the new version (with this PR applied) feels the same as a version before this PR then switching between ReferencePushPullFeeders using the courser keys. I can't notice a difference in the switching time between feeders and if I hit the button really fast, it takes about the same time until the last feeder is visible. Also the fan suggests, that  the computer is doing heavy work. Maybe the sluggishness is more a problem of the ReferencePushPullFeeder implementation in the context of Gregs machine with a lot of them.

Unfortunately I have more bad news: while switching between feeders, I noticed, that the memory consumption is slowly going up. I started with about 1.1g and now the meter shows 1.5g. (I'm on Win11 with JDK11, if that matters)

I'm sorry for not having better news for you! If there is anything I can help you with to track that down, please let me know.

Jan

PR #1794, opened 2025-04-14
Fix Neoden 4 Linux support

jsphpl:
# Description
The [PR adding Neoden4 camera support for Linux](https://github.com/openpnp/openpnp/pull/1604) contains a mistake that prevents the shared library from being loaded, due to the addition of the `lib` prefix.

I also noticed that the RS-232 communication with the Neoden 4 driver board is very sensitive to timing and I could only get it to work reliably by sending out the entire buffer in one go. Hence the other change I'm trying to sneak in with this PR.

If there is any reason not to make `ReferenceDriverCommunications::writeBytes()` public, please let me know.

# Justification
Improve support for Neoden 4 under Linux.

## Implementation Details
(copied from the commit message)

Previously, write() was called in a loop over all bytes. This seems unnecessary because that method will in turn call writeBytes() with a length-1 array for each byte.

Now we're exporting writeBytes() as a public method so the machine drivers can use it directly.

Neoden4Driver was adapted to use writeBytes(). Without this change, serial communication would be too slow on the original hardware and the machine controller would not recognize command sequences.

### DIFF
--- a/src/main/java/org/openpnp/machine/neoden4/NeoDen4Driver.java
+++ b/src/main/java/org/openpnp/machine/neoden4/NeoDen4Driver.java
@@ -3,6 +3,7 @@ package org.openpnp.machine.neoden4;
 import java.awt.geom.Point2D;
 import java.io.IOException;
 import java.util.concurrent.TimeoutException;
+import java.util.Arrays;
 
 import org.openpnp.gui.support.PropertySheetWizardAdapter;
 // import org.openpnp.logging.CalibrationLogger;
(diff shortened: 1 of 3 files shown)


tobydickenson:
Is this still needed any ready to merge?

(I apologise for the slow review; I am just now starting to work through the PR backlog)

apr-mcnetic:
Yes, this is still needed. I've been running both commits on a NeoDen 4 in production jobs, and they work reliably:

- Without the `neodencam` library name fix, the camera library can't be loaded on Linux.
- Without sending each command in one `writeBytes()` call, the controller sometimes doesn't recognize command sequences.

I tested it together with #1982, which fixes a separate job hang in `NeoDen4Driver` (#1959). The two PRs are independent and don't conflict.

PR #1795, opened 2025-04-14
Fixed NormalisationTest to convert different OS line endings to LFs before doing the comparison

tonyluken:
# Description
Fixed the NormalisationTest (added in [PR#1772](https://github.com/openpnp/openpnp/pull/1772)) to convert different OS line endings to LFs before doing the comparison. Otherwise, the test would always fail if run on a machine that doesn't use LFs for line endings.

Also added output to show the lines where the first difference was detected in the event the test fails.

# Justification
Bug fix

# Instructions for Use
N/A

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **Ran the test on a Windows machine to verify that it now passes whereas previously it would fail. Also tested with a translations.properties file that was known to not be normalized and verified that the test failed. Then overwrote the contents of the translations.properties file with the contents from the translations.properties-nomalised file and verified that the test passed again.**
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? **Yes**.
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **No changes.**
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. **All maven tests ran and passed.**

### DIFF
--- a/src/test/java/LocalisationTest.java
+++ b/src/test/java/LocalisationTest.java
@@ -1,7 +1,6 @@
 import org.junit.jupiter.api.Test;
 import java.util.*;
 import java.io.*;
-import java.net.URL;
 import java.util.regex.Pattern;
 
 // Any developer using WindowBuilder will cause the translations properties file to be normalised.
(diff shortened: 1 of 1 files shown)


tobydickenson:
I think I see what is happening here:

1. openpnp git is keeping all text files with LF (unix-like) line endings thanks to a639839a276c6adb603d182a20adb4fa45cce210 from Tony 4 years ago.
2. My test code uses `BufferedReader.readLine` to read the property file which handles any line endings
3. The lines returned from `BufferedReader.readLine` are concatenated with `\n` - building a string in memory that has LF line endings.
4. The test code regenerates a file in memory using `Properties.store` which uses `BufferedWriter.newLine` [here](https://github.com/openjdk/jdk/blob/cec48ed270d3bdf704c389a091b42a32c2ed6440/src/java.base/share/classes/java/util/Properties.java#L945) to write a platform-specific line ending
5. Tony's fix transforms the platform-specific line ending to LF.
6. The test then performs its string comparison on two strings that both have LF line endings. If it needs to write a normalised file, that also has LF line endings which matches git's expectation.

So I agree that is a good fix for this test. Good catch. Thank you.

But maybe there is a hole when running WindowBuilder on windows? I guess it would write a file with CRLF endings, then git would report that every line has changed? Anyway thats not a new problem, and we can address it should it occur.

Please merge.

tonyluken:
>But maybe there is a hole when running WindowBuilder on windows?

Maybe, but if so, the maven test that checks for non LF line endings will at least catch it.

PR #1797, opened 2025-04-18
Fix error when dismissing secondary fiducial position solution

jbussmann:
# Description
Dismissing the secondary calibration fiducial position solution throws an error. This is due to the secondary units per pixel being initialized to `null` to indicate a missing 3D calibration. The solution calls `.derive()` on this to restore the old value, which causes a null pointer exception. This fix restores the old value without calling it's methods.

# Justification
Dismissing solutions should not cause errors.

# Instructions for Use
Same as before.

# Implementation Details
Tested on the simulated machine if the solution can be accepted, dismissed and reopened without errors.
Yes
No
Done

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/solutions/VisionSolutions.java
+++ b/src/main/java/org/openpnp/machine/reference/solutions/VisionSolutions.java
@@ -595,8 +595,7 @@ public class VisionSolutions implements Solutions.Subject {
                         head.setCalibrationSecondaryFiducialLocation(head.getCalibrationSecondaryFiducialLocation()
                                 .derive(oldSecondaryFiducialLocation, true, true, false, false));
                         head.setCalibrationSecondaryFiducialDiameter(oldSecondaryFiducialDiameter);
-                        camera.setUnitsPerPixelSecondary(camera.getUnitsPerPixelSecondary()
-                                .derive(oldUnitsPerPixelSecondary, true, true, false, false));
+                        camera.setUnitsPerPixelSecondary(oldUnitsPerPixelSecondary);
                         // Persist this unsolved state.
                         solutions.setSolutionsIssueSolved(this, false);
                         super.setState(state);


tobydickenson:
I agree this looks right, but it is undoing an apparently intentional change from commit c06dcce77ecac6ea407e6688348040f3dfc70a35 back in 2021. Maybe @markmaker can comment.

jbussmann:
Thanks Toby for pointing this out and Mark for explaining your thinking behind this change. Much appreciated!

I implemented the suggested conditional and edited the description accordingly.

> as Z could have been calibrated in the meantime

A clarifying question, because I still feel unsure in the understanding of how the whole Issues & Solutions is built: by meantime you mean the time since `Find Issues & Solutions` was clicked the last time?

markmaker:
yesOn 7 May 2025, at 10:27, jbussmann ***@***.***> wrote:﻿
jbussmann left a comment (openpnp/openpnp#1797)
Thanks Toby for pointing this out and Mark for explaining your thinking behind this change. Much appreciated!
I implemented the suggested conditional and edited the description accordingly.

as Z could have been calibrated in the meantime

A clarifying question, because I still feel unsure in the understanding of how the whole Issues & Solutions is built: by meantime you mean the time since Find Issues & Solutions was clicked the last time?

markmaker:
Hi @jbussmann,

Sorry for thee delay. 

I just looked at https://github.com/openpnp/openpnp/compare/6f2da80995181d0a73347298c51f8101b5b52541..f00f154e38dcccb021fe2e120c32c297c38f11ba and it is still not entirely correct.

If the Z was set in the meantime, it will still get lost if `oldUnitsPerPixelSecondary` happened to be `null`, which is likely on a fresh machine.

The right code would be something like this:

```java
                        // only restore X, Y, preserve Z which might have been calibrated in the meantime
                        if (oldUnitsPerPixelSecondary == null) {
                            camera.setUnitsPerPixelSecondary(camera.getUnitsPerPixelSecondary()
                                    .derive(Location.origin, true, true, false, false));
                        } else {
                            camera.setUnitsPerPixelSecondary(camera.getUnitsPerPixelSecondary()
                                    .derive(oldUnitsPerPixelSecondary, true, true, false, false));
                        }
```

Taking `Location.origin` (all zero coordinates) restores X, Y to zero, but preserves Z.

Even better: fix it in `Location.derive()`, i.e. check for `null` there and use `Location.origin` there.

_Mark

PR #1799, opened 2025-04-29
Job Optimisation nozzle order

tobydickenson:
# Description

It is a bad outcome for a multi-nozzle machine to get stuck using only a single nozzle, while another nozzle is left idle. This occurs if the job reaches a point where the idle nozzle can not find a tip which can handle any of the outstanding parts.

This PR changes the "Nozzle Tips" Job Ordering method with a heuristic to reduce the likelihood of this occuring.

The "Nozzle Tips" method operates by grouping parts according to the nozzle tip, then optimising the order of feeders then placements within each group.

The description in [the wiki](https://github.com/openpnp/openpnp/wiki/Job-Processing#configuration) is that it uses "most efficient order of nozzle tips". A code comment says "keeping all possible tips sorted by amount of placements they can handle will likely provide a good starting point in selecting the next best one". Maybe I am missing something, but I do not have an intuitive understanding of why that is a desirable order to run the job.

This PR uses a different sort order. For each tip, it counts the number of placements that can also be handled by an alternative tip, then this count is the primary sort key. A high count indicates that the nozzle is handling parts which can be placed by more than one nozzle. On a multi-nozzle machine it is desirable to schedule these at the end of the job.

# Concerns

I changed the "Nozzle Tips" policy sort order, but I am open to the suggestion that adding a new policy to the `JobOrderHint` enum would be better:

1. Am I overlooking some advantage to the original "Nozzle Tips" sort order? On a single nozzle machine maybe?
2. Are there users who would be better served by the original sort order?
3. Are there users who have got used to the original sort order and would not appreciate it changing?

# Justification
I have been testing this on my lumenpnp which has 2 nozzles, and 7 tips which can be used on any nozzle.

Having tested the original "Nozzle Tips" on several different jobs, more than half of those got stuck finishing the job with a single nozzle. Typically this occurs when there is a module or connector which needs a larger tip, and only one tip is suitable. Sorting by number of placements means that this tip is scheduled last, because there is a small number of placements using this tip, and the job reaches a point where only this tip is active.

I have tested two of those problem jobs with this PR. It schedules the tip to place those inflexible parts first. The job finishes using the tips with the greatest flexibility, with both nozzles remaining busy until the end of the job.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. -- tested on two real jobs. 
6. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. -- no changes

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -977,7 +977,7 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
             
             // sort lists per nozzleTips by their size such that the nozzle tip that can handle
             // the most jobPlacments is first. As second sorting criteria the nozzleTip's name is used.
-            // For mulit-nozzle machines with different nozzleTips for each nozzle that are compatible
+            // For multi-nozzle machines with different nozzleTips for each nozzle that are compatible
             // with the same jobPlacments this always results in groups of identical amounts of
             // jobPlacments. Taking the name into account makes the sorting unique again.
             perNozzleTipJobPlacements.sort(Comparator.comparing(JobPlacementNozzleTip::size).reversed()
@@ -992,7 +992,9 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
             // the other nozzles. Keeping all possible tips sorted by amount of placements they can
             // handle will likely provide a good starting point in selecting the next best one.
             List<NozzleTip> plannedNozzleTips = perNozzleTipJobPlacements.stream().map(j -> j.getNozzleTip()).collect(Collectors.toList());
-            
+
+            HashMap <NozzleTip,Integer> perNozzlePlacementOptions = new HashMap<NozzleTip,Integer>();
+
             // Remove duplicate placements keeping only the first occurrence
             // Skip this step if there is only one nozzle tip left.
             if (perNozzleTipJobPlacements.size() > 1) {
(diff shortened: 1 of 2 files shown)


janm012012:
Hi Toby!

I'd say this is a chicken and egg problem. If you force the JobProcessor to use different nozzles and visit different feeders in one go, you might end up slower then the current route.

As you might have seen on the list, it seems not uncommon to provide multiple nozzle tips of the same size and use them in parallel. You'll then get very good performance because the distance between feed locations is small. From my practical experience a good setup requires hand tuning per job. The job I run yesterday worked well, if I double one tip but not the other. The other job performed better with the second tip also doubled. (both jobs contain lots of 0402 where doubling is a big win. They contain a few other components requiring 3 other nozzle tips.) Running the job with "Nozzle tip loading strategy" set to Minimize results in worth performance because a lot of 0402s are placed while the other nozzle was still used for larger parts. The big problem was the large distance between feed locations.

I'd say the only really good improvement to the sorting strategy is to run a TSM over the entire job and simulate which order and when to change nozzle tips results in best performance. This would also provide a unified solution that works well for all jobs. I'm thinking about this for some time now but still can't get my head around how to handle the path between picks and places... If you have ideas, I'd be happy to discuss them!

In general, I don't see a reason not to merge this as a new sorting criteria. (The idea we currently have is based on Marks thoughts and has been discussed for some time on the list. I'm confident, that it provides good (or even perfect) results for single nozzle machines.)

Jan

tobydickenson:
In your job with N 0402 placements, do you never see this scenario?
* The Job starts with both tiny tips loaded.
* After N/2 cycles, the first nozzle has placed N/2 0402s, and the second nozzle has done the same. All the 0402s have been placed.
* What remains are all the larger parts. The first nozzle loads a larger tip, and the second nozzle is unemployed.

Maybe you have more than one larger tip (maybe of different sizes) that can come into play at this point. But at some point one tip runs out of possible placements, leaving the other tip working alone.

I am using "StartAsPlanned" strategy.

janm012012:
Toby, I'm with you and I see the scenario you depicted. I'm just not convinced, that your solutions provides overall better results then the one we have. However, I can imagine, that your modification provides better results in some situations. That's why I think we shall keep the solution we have and you may merge (and I'd be happy to assist if you like to) your solution as a separate option.

My last job was actually a double sided panel, so every "problem" is multiplied by 20 (PCBs per panel).

One side consists of 10 0402, a SOT32 and a crystal. This requires three different nozzle tips. With two for 0402 the job runs until dry and then switches over to a tip for SOT23 and one for the crystal. That keeps both nozzles busy but takes quite long because the crystal is large and feeded from fare way. With "StartAsPlanned" and "Minimize" the job runs nice and smooth. With two tips for SOT23 there is a advantage in feeding the SOT23, but a penalty for a third nozzle tip change and the final pass just using a single nozzle for the crystal. (Ok, I could double that as well.) If this configuration is run with "Minimize", the tip change/feeding gets adverse.

The second side contains 16 0402, four 0603, one 0805 and a QFP, This sums up to 4 different nozzle tips. It runs best with two tips for 0402 and two for 0603. The 0805 and QFP can be done in one path. With just a single tip for 0603, the other nozzle runs dry...

I don't see how your sorting would result in a better performance as the difference in quantities per tip are to large. (Side node: you might have noticed that as well, but it seems hardly possible to increase the performance using a second nozzle by more then about 50% compared to a single nozzle.)

As I said earlier, your sorting seems as good as the one we have on a general manor. Its just that the simple hysteric (Grady) can't do better. So, just add your sorting as an other option... (seems the sorting of perNozzleTipJobPlacements is the only function change. That shall be easily made optional by an argument.)

Jan

tobydickenson:
Proposed new text for [the wiki](https://github.com/openpnp/openpnp/wiki/Job-Processing)

* _Nozzle Tips_ first calculates the most efficient order of nozzle tips using the number of placements each nozzle tip can handle and its name and then optimizes each of these groups using Pick and Place Locations.
* _Nozzle Tips (Inflexible Tips First)_ schedules the special-purpose nozzle tips first, the multi-purpose tips last,  and then optimizes each of these groups using Pick and Place Locations. On a machine with multiple nozzles, this can help keep all the nozzles busy though to the end of the job.

janm012012:
Hi Toby!

> The description in [the wiki](https://github.com/openpnp/openpnp/wiki/Job-Processing#configuration) is that it uses "most efficient order of nozzle tips". A code comment says "keeping all possible tips sorted by amount of placements they can handle will likely provide a good starting point in selecting the next best one". Maybe I am missing something, but I do not have an intuitive understanding of why that is a desirable order to run the job.

The point is, that if a placement can be processed using different nozzle tips, it is desirable to use the tip that can handle as many other placements as well. This selects the tip that can handle the most other placements rather then the first tip in the list of compatible tips. If you have placements that can be processed by tips 1 and 2 and other placements that required 2 and 3, the sorting will suggest to use tip nr 2 only. Whereas the (old) default would have chosen tip 1 because it the first in the list of the first placements.

The code looks good now. You may merge it.

Another remark: there is an other filter feature in the Planner open for multi nozzle machines: enforce, that no two parts may collide on any nozzle tips. This can - and shall - be handled by taking tips Max. part diameter into account. This would choose the next tip by making sure, that it will not intersect with any other nozzle tip. Would that effect your new sorting strategy?

Jan

tobydickenson:
Hi Jan,

I will merge after rebasing

> Another remark: there is an other filter feature in the Planner open for multi nozzle machines: enforce, that no two parts may collide on any nozzle tips. This can - and shall - be handled by taking tips Max. part diameter into account. This would choose the next tip by making sure, that it will not intersect with any other nozzle tip. Would that effect your new sorting strategy?

Nice feature :-) That will work as expected.

PR #1800, opened 2025-04-30
calculate travel time using backlash calibration data

tobydickenson:
**This is a Proof Of Concept PR, shared for feedback and ideas. I do not plan to merge this in its current form**

# Description

This loop in the job processor optimises the combination of placements which will be handled simultaneously on a multi-nozzle machine. It optimises based on the total **travel distance**: from one pick location to the next, plus the distance from one placement to the next.

https://github.com/openpnp/openpnp/blob/535678fe6a554a8bf4ab2b0bb3a8f78565f64a3a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L2534-L2555

Optimising based on travel distance is ok, but it ignores:
* Differences between axes. One axis might be stronger and faster
* Acceleration profiles

Anecdotally, I had noticed that the job processor was sometimes picking pairs of placements that were less optimal than they could be. My lumenpnp machine is distinctly limited by acceleration, so it is not a surprise that optimising based on travel distance was picking some poor combinations.

This PR changes that loop to optimise **travel time**, using some of the data collected during backlash calibration to convert movement vectors into travel time. Specifically it uses the data which is plotted on this chart (image copied from the [wiki](https://github.com/openpnp/openpnp/wiki/Calibration-Solutions#backlash-calibration-on-the-axis--diagnostics) ):

image

I have used this on a few jobs, and it looks to be an improvement. A higher proportion of the moves between pick locations and between placement locations are desirable, efficient moves. I do not yet have a quantitative analysis.

# Disadvantages

1. The standard backlash compensation does not record quite enough data. It records timing data for a range of movements up to 10mm. For my testing I increased this to 100mm. That significantly increases the duration of the backlash compensation process, so this is definitely not an approach that I can generally recommend.
https://github.com/openpnp/openpnp/blob/535678fe6a554a8bf4ab2b0bb3a8f78565f64a3a/src/main/java/org/openpnp/machine/reference/solutions/CalibrationSolutions.java#L82-L82
2. The backlash data is specific to each axis. It provides data about movements parallel to an axis, but we have no information about diagonal movements.
3. The backlash data is stored in a `SimpleGraph` object; ideal for plotting as a graph, but not a great format for further calculations.

# Outstanding Questions

A. Does this general approach sound useful?
B. Feedback would be appreciated on the new `MotionUtils` api
C. I think it makes sense to create a new I&S solution to collect motion timing data over a range of steps sizes, at 0° 90° and 45° movements. Maybe 22.5° too. Does that approach make sense?

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -68,6 +68,7 @@ import org.openpnp.util.TravellingSalesman;
 import org.openpnp.util.UiUtils;
 import org.openpnp.util.Utils2D;
 import org.openpnp.util.VisionUtils;
+import org.openpnp.util.MotionUtils;
 import org.pmw.tinylog.Logger;
 import org.simpleframework.xml.Attribute;
 import org.simpleframework.xml.Element;
@@ -2540,15 +2541,15 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
                 Location averagePickLocation  = calcCenterLocation(plannedPlacements, pickLocator);
                 Location averagePlaceLocation = calcCenterLocation(plannedPlacements, placeLocator);
                 
-                // find the placement with the shortest distance to averagePlickLocation and averagePlaceLocation
-                double bestDistance = Double.MAX_VALUE;
+                // find the placement with the lowest cost to averagePickLocation and averagePlaceLocation.
+                double bestCost = Double.MAX_VALUE;
                 for (JobPlacement p : compatibleJobPlacements) {
-                    double distance = pickLocator.getLocation(p, nozzle).getLinearDistanceTo(averagePickLocation) 
-                                    + placeLocator.getLocation(p, nozzle).getLinearDistanceTo(averagePlaceLocation);
+                    double cost = MotionUtils.getMotionCost(pickLocator.getLocation(p, nozzle).subtract(averagePickLocation))
+                                + MotionUtils.getMotionCost(placeLocator.getLocation(p, nozzle).subtract(averagePlaceLocation));
 
-                    // if this placement is closes with respect to its pick and place 
-                    if (bestDistance > distance) {
-                        bestDistance = distance;
+                    // if this placement is closest with respect to its pick and place
+                    if (bestCost > cost) {
+                        bestCost = cost;
                         bestPlacement = p;
                     }
                 }
--- a/src/main/java/org/openpnp/util/MotionUtils.java
+++ b/src/main/java/org/openpnp/util/MotionUtils.java
@@ -1,0 +1,89 @@
+package org.openpnp.util;
+
+import java.awt.geom.Point2D;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.Collections;
+import org.pmw.tinylog.Logger;
+import org.openpnp.model.Configuration;
+import org.openpnp.model.Location;
+import org.openpnp.machine.reference.ReferenceMachine;
+import org.openpnp.spi.base.AbstractAxis;
+import org.openpnp.machine.reference.axis.ReferenceControllerAxis;
+import org.openpnp.util.SimpleGraph;
+import org.openpnp.util.SimpleGraph.DataScale;
+import org.openpnp.util.SimpleGraph.DataRow;
+
+public class MotionUtils {
+
+    // Given a x/y movement vector, return a measurement of the cost of performing that movement.
+    // This return value is suitable for use in optimising a movement plan.
+    //
+    // It might be either:
+    // 1. A time value, if the x and y axes have performed backlash calibration
+    //    which records a profile of single-axis movement times over a range of distances.
+    // 2. The linear distance, as a fallback if there is no way to estimate movement time.
+    //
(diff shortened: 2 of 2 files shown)


tonyluken:
Wouldn't the travel time of a diagonal move just be the longer of whichever axes are moving? For instance, if both X and Y are moving and X takes 250ms and Y takes 175ms, then the combined move would complete in 250ms. I can't think if any reason for it to be slower than that and it certainly can't be any faster than that.

tobydickenson:
> I can't think if any reason for it to be slower than that

I was expecting my response to this to be "yes in theory but in practice there are some other factors that matter".

But, having taken some measurements on the machine, I do indeed see that diagonal movements are as you suggest. That simplifies things greatly, thank you for the suggestion.

----

MoveTiming.py
```
import math,time
from org.openpnp.model import LengthUnit, Location
from org.openpnp.util.UiUtils import submitUiMachineTask
from org.openpnp.spi.MotionPlanner import CompletionType
from java.lang.System import nanoTime

def main():
    submitUiMachineTask(Timing().do)

class Timing():
    def do(self):
        # Move to the start location
        self.hm = machine.getDefaultHead().getDefaultCamera();
        self.position = self.zero_location = machine.getDefaultHead().getCalibrationPrimaryFiducialLocation();
        self.hm.moveTo(self.zero_location)
        machine.getMotionPlanner().waitForCompletion(None,CompletionType.WaitForStillstand)
        #
        # Test loop
        distance = 1
        while distance<160:
            # Measure single-axis movements
            tx = self.move(distance,0)
            ty = self.move(0,distance)
            #
            # Measure a diagonal movement which is mostly in x. We expect this to take
            # the same duration as a movement which is entirely in x
            txh = self.move(distance,distance*0.4)
            # Same for y
            tyh = self.move(distance*0.4,distance)
            # Measure a diagonal movement which is the same distance in x and y. We expect
            # this to take the same duration as the maximum of the single-axis movements
            txy = self.move(distance,distance)
            #
            # Log some numbers. These are are:
            # * distance
            # * 2 columns of durations in milliseconds for x and y moves of that distance
            # * 3 columns of percentages comparing the timings as above. We expect these
            #   to be close to 100%, less than 100% if the diagonal moves are faster, or
            #   more than 100% if the diagonal moves are slower.
            print '%5.1f, %6.1f, %6.1f, %6.1f%%, %6.1f%%, %6.1f%%' % (distance,tx*1e-6,ty*1e-6,
                    100.0*txh/tx, 100.0*tyh/ty, 100.0*txy/max(tx,ty))
            distance *= math.sqrt(2)
        #
        # Move back to the start
        self.hm.moveTo(self.zero_location)
        machine.getMotionPlanner().waitForCompletion(None,CompletionType.WaitForStillstand)

    def move(self,vx,vy):
        # Move a vector (vx,vy) and return the duration in nanoseconds.
        # If the current position is left of home then we move right,
        # and if we are right of home then we move left. This assumes
        # that positive and negative movements are equivalent, and means
        # that all movement will be centred on the home location.
        if self.position.getX()>self.zero_location.getX():
            vx =-vx
        if self.position.getY()>self.zero_location.getY():
            vy =-vy
        vector = Location(LengthUnit.Millimeters, vx, vy, 0, 0)
        self.position = self.position.add(vector)

        start = nanoTime()
        self.hm.moveTo(self.position)
        machine.getMotionPlanner().waitForCompletion(None,CompletionType.WaitForStillstand)
        end = nanoTime()
        return end-start

main()

```

tobydickenson:
I originally intended this branch as a proof of concept, but I think it is now heading towards something that can be merged to test.

I previously noted some disadvantages:

> The standard backlash compensation does not record quite enough data. It records timing data for a range of movements up to 10mm. For my testing I increased this to 100mm. That significantly increases the duration of the backlash compensation process, so this is definitely not an approach that I can generally recommend.

The code in commit 84dbb650301212f28ae93ed800a5aeac7c3f55dc adds some extra movement timing steps to backlash compensation. This runs very quickly and is low impact so I think this is a reasonable addition to the standard backlash process.

> The backlash data is specific to each axis. It provides data about movements parallel to an axis, but we have no information about diagonal movements.

As noted by Tony this is not a problem. The timing of diagonal movements can be predicted very well from the timing of axial movements.

----

Proposed addition to the CHANGES file:
Pick and place routing is now optimised for travel time, rather than travel distance, using the timing data recorded during backlash compensation. To see the full benefit you will need to rerun the backlash compensation, particularly on machines that are more limited by acceleration than maximum speed.

janm012012:
Hi Toby!
I've a PR on this in the pipeline as well. It uses a 3-point motion profile to take the actual axis parameters into account. I needs some more reviewing before I can publish it...

Jan

janm012012:
Hi Toby!

I've now published my implementation as PR #1813. If you'd please take a look to see if it would suit your needs.

I remember that Mark suggest to use the backlash data for travel cost estimation but never really liked the idea because

- the required interpolation is costly,
- the interpolation uses measurement data and hence needs error handling,
- the required data is only available if backlash calibration has been done,
- the interpolate goes beyond the data and
- cost estimation based on acceleration and feedrate is rather simple.

However, I agree with you, that we need to enhance the current implementation. So I'm open to what solution we finally merge.

Jan

tobydickenson:
#1813 does the same thing but better. This PR is closed.

Issue #1804, opened 2025-05-08
job processor exception


tobydickenson:
The following exception was observed during a job. I have not yet investigated.

One strip feeder had just disabled itself due to running out of sprocket holes. I suspect that is relevant.

```
2025-05-08 11:21:56.773 AbstractMachine TRACE: Exception caught, executing pending motion: java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 0
        at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
        at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
        at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)
        at java.base/java.util.Objects.checkIndex(Objects.java:374)
        at java.base/java.util.ArrayList.get(ArrayList.java:459)
        at org.openpnp.machine.reference.ReferencePnpJobProcessor$Plan.updateFeederIndex(ReferencePnpJobProcessor.java:816)
        at org.openpnp.machine.reference.ReferencePnpJobProcessor$Plan.planJobPlacementsByPickLocation(ReferencePnpJobProcessor.java:735)
        at org.openpnp.machine.reference.ReferencePnpJobProcessor$Plan.planJobPlacementsByPickPlaceLocation(ReferencePnpJobProcessor.java:845)
        at org.openpnp.machine.reference.ReferencePnpJobProcessor$Plan.planJobPlacementsByNozzleTips(ReferencePnpJobProcessor.java:1038)
        at org.openpnp.machine.reference.ReferencePnpJobProcessor$Plan.planJobPlacements(ReferencePnpJobProcessor.java:680)
        at org.openpnp.machine.reference.ReferencePnpJobProcessor$Plan.step(ReferencePnpJobProcessor.java:570)
        at org.openpnp.machine.reference.ReferencePnpJobProcessor.next(ReferencePnpJobProcessor.java:169)
        at org.openpnp.gui.JobPanel.lambda$jobRun$2(JobPanel.java:950)
        at org.openpnp.spi.base.AbstractMachine$1.call(AbstractMachine.java:575)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[...]

PR #1812, opened 2025-05-12
a possible fix for issue #1804. feeders empty list

tobydickenson:
# Description
This is a proposed fix for issue #1804. The exception stack trace recorded in that issue occurs when `updateFeederIndex` returns its finishing location, using the location of the last feeder visited by the TSM, but that list is empty. This patch detects this case, skips the TSM operation, and simply returns the `startLocation`.

This change conclusively avoids the exception recorded in issue #1804. It has been tested for 2 days of production usage with no further problems observed.

An outstanding task before merging is to prove that this wont leave some other problems. Previously the return value from `updateFeederIndex` could never have been `null`, because it returns a feeder location and feeders always have a location. However sometimes the provided `startLocation` is `null`. I have not fully proven that this can not cause a problem. This is a provisional PR for now.

# Implementation Details
    * tests pass
    * code review confirms that the reported exception is no longer possible
    * test in production shows no further problems
5. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. -- tests pass.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -780,6 +780,12 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
                 catch (Exception e) {
                 }
             }
+
+            if (feeders.size()==0) {
+                // We were unable to find any feeders, so there is no need for TSM.
+                // We end where we started.
+                return startLocation;
+            }
             
             // route pick locations of all feeders through travelling salesman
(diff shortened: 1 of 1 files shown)


janm012012 (approved):

janm012012:
Hi Toby!

I don't see a reason not to merge this.
Out of curiosity: how can you run a job without feeders? Is the check, that a feeder is available for each part not covering this?

Jan

tobydickenson:
There is a check at the start of a job that every part has an enabled feeder.

But a ReferenceStripFeeder will disable itself when it reaches the end of the strip, so this can occur in the middle of a job too.

PR #1813, opened 2025-05-12
Better metric for motion optimization

janm012012:
# Description
This PR adds a new metric for travel cost estimation based on axes acceleration and feed rate parameters named TravelCost. This new metric is then used by the traveling salesman and in the job processor for n+1 placement selection. For efficient use, the traveling salesman class has got a new constructor that allows to specify the effected head mountable. This head mountable is then used for travel cost estimation by the traveling salesman algorithm. The implementation defaults to the default camera on the default head and is therefore automatically available everywhere where the traveling salesman is used. The new TravelCost class is also available separately for use eg. within sorting loops.

# Justification
At present travel cost is not actually taken into account. All optimization is based on geometrical distance, which provides a high significance for optimization, but does not reflect how controller today work. They either operate all axes independently or delay axes such that all reach the target at once. With the new metric this behavior is now taken into account and different axes parameter are respected as well.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. **using simulation machine**
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. **yes: Length extended with a copy constructor**

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -64,6 +64,7 @@ import org.openpnp.spi.PnpJobProcessor.JobPlacement.Status;
 import org.openpnp.spi.base.AbstractJobProcessor;
 import org.openpnp.spi.base.AbstractPnpJobProcessor;
 import org.openpnp.util.MovableUtils;
+import org.openpnp.util.TravelCost;
 import org.openpnp.util.TravellingSalesman;
 import org.openpnp.util.UiUtils;
 import org.openpnp.util.Utils2D;
@@ -2540,21 +2541,32 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
                 Location averagePickLocation  = calcCenterLocation(plannedPlacements, pickLocator);
                 Location averagePlaceLocation = calcCenterLocation(plannedPlacements, placeLocator);
                 
-                // find the placement with the shortest distance to averagePlickLocation and averagePlaceLocation
-                double bestDistance = Double.MAX_VALUE;
-                for (JobPlacement p : compatibleJobPlacements) {
-                    double distance = pickLocator.getLocation(p, nozzle).getLinearDistanceTo(averagePickLocation) 
-                                    + placeLocator.getLocation(p, nozzle).getLinearDistanceTo(averagePlaceLocation);
-
-                    // if this placement is closes with respect to its pick and place 
-                    if (bestDistance > distance) {
-                        bestDistance = distance;
-                        bestPlacement = p;
+                TravelCost travelCost = null;
+                try {
+                    travelCost = new TravelCost(nozzle);
+                }
+                catch (Exception e) {
+                    // ignore exception and continue without optimization
+                    Logger.trace("TravelCost() failed, skipping second placement optimiation");
+                    travelCost = null;
+                }
+                if (travelCost != null) {
+                    // find the placement with the least cost for motion to averagePlickLocation and averagePlaceLocation
+                    double leastCost = Double.MAX_VALUE;
+                    for (JobPlacement p : compatibleJobPlacements) {
+                        double cost = travelCost.getCost(pickLocator.getLocation(p, nozzle), averagePickLocation) 
+                                    + travelCost.getCost(placeLocator.getLocation(p, nozzle), averagePlaceLocation);
+    
+                        // if this placement is closes with respect to its pick and place 
+                        if (leastCost > cost) {
+                            leastCost = cost;
+                            bestPlacement = p;
+                        }
                     }
                 }
             }
-            else {
-                // no further optimization possible or requested, just choose the first of the list
+            // if bestPlacement is still null, use the first
+            if (bestPlacement == null) {
                 bestPlacement = compatibleJobPlacements.get(0);
             }
             
--- a/src/main/java/org/openpnp/machine/reference/solutions/VisionSolutions.java
+++ b/src/main/java/org/openpnp/machine/reference/solutions/VisionSolutions.java
@@ -1198,7 +1198,7 @@ public class VisionSolutions implements Solutions.Subject {
 
             // Perform zero knowledge calibration motion pattern.
             double displacementAbsMm = zeroKnowledgeDisplacementMm;
-            Location unitsPerPixel = new Location(null);
+            Location unitsPerPixel = new Location();
             Length featureDiameter = null;
             for (int pass = 0; pass < 3; pass++) {
                 double displacementMm = displacementAbsMm;
--- a/src/main/java/org/openpnp/model/Location.java
+++ b/src/main/java/org/openpnp/model/Location.java
@@ -44,12 +44,8 @@ public class Location {
     @Attribute(required = false)
     private double rotation;
 
-    /**
-     * Only used by XML serialization.
-     */
-    @SuppressWarnings("unused")
-    private Location() {
-        this(null);
+    public Location() {
+        this((LengthUnit)null);
     }
 
     public Location(LengthUnit units) {
@@ -62,6 +58,10 @@ public class Location {
         this.y = y;
         this.z = z;
         this.rotation = rotation;
+    }
+
+    public Location(Location l) {
+        this(l.getUnits(), l.getX(), l.getY(), l.getZ(), l.getRotation());
     }
 
     static final public Location origin = new Location(LengthUnit.Millimeters);
@@ -87,10 +87,21 @@ public class Location {
     }
 
     public Location convertToUnits(LengthUnit units) {
-        Location location =
-                new Location(units, new Length(x, this.units).convertToUnits(units).getValue(),
-                        new Length(y, this.units).convertToUnits(units).getValue(),
-                        new Length(z, this.units).convertToUnits(units).getValue(), rotation);
+        Location location;
+
+        // If the units do not change, take the shortcut to return this.
+        // This is safe because all members are guaranteed to not change
+        // after creation. This has been verified defining them final.
+        // Unfortunately this compiler assisted test can not stay in place
+        // because the serializer requires elements to be writeable.
+        if (this.units == units) {
+            location = this;
+        }
+        else {
+            location = new Location(units, new Length(x, this.units).convertToUnits(units).getValue(),
+                            new Length(y, this.units).convertToUnits(units).getValue(),
+                            new Length(z, this.units).convertToUnits(units).getValue(), rotation);
+        }
         return location;
     }
 
--- a/src/main/java/org/openpnp/util/TravelCost.java
+++ b/src/main/java/org/openpnp/util/TravelCost.java
@@ -1,0 +1,196 @@
+/*
+ * Copyright (C) 2025 <janm012012@googlemail.com>
+ * 
+ * This file is part of OpenPnP.
+ * 
+ * OpenPnP is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ * 
+ * OpenPnP is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public License along with OpenPnP. If not, see
+ * <http://www.gnu.org/licenses/>.
+ * 
+ * For more information about OpenPnP visit http://openpnp.org
+ */
+
+package org.openpnp.util;
+
+import org.openpnp.machine.reference.axis.ReferenceControllerAxis;
+import org.openpnp.machine.reference.solutions.HeadSolutions;
+import org.openpnp.model.Configuration;
+import org.openpnp.model.LengthUnit;
+import org.openpnp.model.Location;
+import org.openpnp.spi.CoordinateAxis;
+import org.openpnp.spi.HeadMountable;
+import org.openpnp.spi.Machine;
+import org.pmw.tinylog.Logger;
+
+/**
+ * Methods to estimate the cost of motion between two locations.
+ * 
+ * The basic principle is to estimate the cost using three point motion with
+ * two constant acceleration segments and one optional feedrate limited segment.
+ * The estimation assumes, that motion starts and ends at standstill.
+ */
+public class TravelCost {
+    
+    /**
+     * @param movable head mountable's axis to use for cost estimation, if null
+     * the default camera on the default head is used
+     * @param units the units to calculate the in. static data is prepared in this
+     * units and Length input is convert to this units for the estimation
+     * @throws if no default camera is define or if the movable has no mapped XY axis
+     */
+    public TravelCost(HeadMountable movable, LengthUnit units) throws Exception {
+        super();
+
+        // use default units if not specified
+        if (units == null) {
+            units = defaultUnits;
+        }
+        this.units = units;
+        
+        // collect axis parameter
+        Machine machine = Configuration.get().getMachine();
+        // if no movable is specified, use the default camera of the default head
+        if (movable == null) {
+            movable = machine.getDefaultHead().getDefaultCamera();
+        }
+        CoordinateAxis rawAxisX = HeadSolutions.getRawAxis(machine, movable.getAxisX());
+        CoordinateAxis rawAxisY = HeadSolutions.getRawAxis(machine, movable.getAxisY());
+        if (! (rawAxisX instanceof ReferenceControllerAxis
+                && rawAxisY instanceof ReferenceControllerAxis)) {
+            throw new Exception("HeadMountable "+ movable + "has no XY axis assigned");
+        }
+        ReferenceControllerAxis referenceControllerAxisX = (ReferenceControllerAxis)rawAxisX;
+        ReferenceControllerAxis referenceControllerAxisY = (ReferenceControllerAxis)rawAxisY;
+
+        this.xAxis = new AxisParameter(referenceControllerAxisX, units);
+        this.yAxis = new AxisParameter(referenceControllerAxisY, units);
+        
+        // try to collect information about Z axis
+        CoordinateAxis rawAxisZ = HeadSolutions.getRawAxis(machine, movable.getAxisZ());
+        if (!(rawAxisZ instanceof ReferenceControllerAxis)) {
+            Logger.trace("Travel cost estimation for headmountable " + movable + " not avaiable on Z axis");
+            this.zAxis = null;
+        }
+        else {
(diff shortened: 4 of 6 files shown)


tobydickenson:
## Implementation Strategy

We have several implementation strategies proposed, and I can see advantages to them all.

1. #1813 theoretical model with two constant acceleration segments and one optional feedrate limited segment.
2. Practical model interpolating measured timing data.
   a. Data collected in an extended backlash compensation (implemented on #1800)
   b. Data collected in a new I&S stage (no implementation yet)

I agree with your comments on #1800 that it would be good to avoid measurements.... if the theoretical model is close enough to reality that it gives a good movement plan. My next chance to test this on the machine will be in the week starting 26th and I am excited to try it out. I dont have a good objective comparison, but subjectively it was pretty easy to see that the machine was making more sensible multi-nozzle choices with #1800.

Some possible reasons that the theoretical model might not work so well on my machine:
* Use of ModeratedConstantAcceleration to restrain short movements to minimise machine vibration. This uses the Jerk parameter to calculate a reduced acceleration for short movement segments.
* But Marlin isnt processing that acceleration in the manner expected by ModeratedConstantAcceleration. It has the right general effect (short moves are smooth, long moves are fast) but the kinematic quantitative details are not as expected.
* Marlin is doing is own jerk control too.
* Marlin has a wierd kinematic bug where long movements are sometimes super-slow. My workaround for that involves splitting long movements, and that creates some performance discontinuities. My first threshold for this is 60mm, so a 59.9999999mm movement is 10% faster than an 60.00000001mm movement

All of those details are visible in the measurements, but hidden from the theoretical optimisation. Hopefully that wont matter.

Am I right in thinking you have a single nozzle machine? So your focus is on TSM optimisation? Do you have any measurements on TSM speed improvement?

## Travelling Salesman stability

In #1800 I was focused on optimising the pair of placements handled simultaneously by both nozzles. But I did look at the TSM too.

Having added that `MotionUtils` class to the TravellingSalesman I found that it behaved poorly.  It often seemed like the TSM had not converged on a good solution. Maybe I had a bug, or maybe the TSM solver has problems planning routes in a non-euclidean search space. I will check that code again.

## Code Review

1.

I think it is a mistake to put the `HeadMountable` into the TravellingSalesman constructor. It makes the TSM significantly less general.

Would it be better if `TravelCost` was an abstract class, TravellingSalesman had a `setTravelCost` method, and clients called  `tsm.setTravelCost(TravelCostThreePointMotion(moveable,units))` ?

2.

The old comment "Plain old data TravelLocation for faster processing. Improved solving by a factor of 6 from using OpenPNP Locations directly" is no longer valid. What is happening here?

janm012012:
Hi Toby!

Many thanks for your comment and review!

> I agree with your comments on #1800 that it would be good to avoid measurements.... if the theoretical model is close enough to reality that it gives a good movement plan. My next chance to test this on the machine will be in the week starting 26th and I am excited to try it out. I dont have a good objective comparison, but subjectively it was pretty easy to see that the machine was making more sensible multi-nozzle choices with #1800.
> 
I used it today in production and did not found any issues. I don't know if the results are better compare to the current implementation, but it feels better to provide a more physical metric.

> Some possible reasons that the theoretical model might not work so well on my machine:
> 
>     * Use of ModeratedConstantAcceleration to restrain short movements to minimise machine vibration. This uses the Jerk parameter to calculate a reduced acceleration for short movement segments.
> 
What Jerk values are you using? Any value not 0 and not close to infinity will make the machine slower by requesting a lower acceleration/feedrate then allowed for the axis based on the previous motion. Yes, with this jerk enabled, the metric will predict a faster motion time, but the physical relation is still preserved. 

>     * But Marlin isnt processing that acceleration in the manner expected by ModeratedConstantAcceleration. It has the right general effect (short moves are smooth, long moves are fast) but the kinematic quantitative details are not as expected.
> 
>     * Marlin is doing is own jerk control too.
> 
>     * Marlin has a wierd kinematic bug where long movements are sometimes super-slow. My workaround for that involves splitting long movements, and that creates some performance discontinuities. My first threshold for this is 60mm, so a 59.9999999mm movement is 10% faster than an 60.00000001mm movement
> 
> 
> All of those details are visible in the measurements, but hidden from the theoretical optimization. Hopefully that wont matter.
> 
I don't see how that effect the optimization. The actual time is not relevant, just the change in cost between different options. At present constant cost is expressed by a cycle, which is not physical. Our machines run each axis at their parameters, which forms a rectangular for constant cost. So running on a diagonal has the same cost as running one axis only.

> Am I right in thinking you have a single nozzle machine? So your focus is on TSM optimization? Do you have any measurements on TSM speed improvement?

I'm sorry, no. I've a CHM-T36VA, which is a dual nozzle machine and I quite early discovered, that most developers focused on signal nozzle machines. That's why I started contributing multi nozzle optimization. Today the JobProcessor uses TSM a lot for that.

> ## Travelling Salesman stability
> 
> In #1800 I was focused on optimising the pair of placements handled simultaneously by both nozzles. But I did look at the TSM too.
> 
> Having added that `MotionUtils` class to the TravellingSalesman I found that it behaved poorly. It often seemed like the TSM had not converged on a good solution. Maybe I had a bug, or maybe the TSM solver has problems planning routes in a non-euclidean search space. I will check that code again.

The TSM test code produces graphics you my use to evaluate its performance. Today the solver uses simulated annealing. I once asked why and go the answer, that is was available at the time this was implemented. The code itself is likely ok, but the parameters its operated with, are tricky. I'd say the results are better compared to not using the TSM but fare from perfect. For small problems (like optimizing the order of picks for multiple nozzles) it makes fare to many evaluations and for large problems (placements to process) it does not take enough care to find a good solution. Using the graphics the test generates, you can see the effect. (I read that a branch and bound solver would be more appropriate today and tried it - in an other branch - but the code is not fast enough yet because I don't have a good bounding function... Maybe I'll add it as solver for small problems. It then will not take more evaluations then possible routes and will always return the perfect route.)

> ## Code Review
> 
> I think it is a mistake to put the `HeadMountable` into the TravellingSalesman constructor. It makes the TSM significantly less general.
> 
The default constructor still does not need a head mountable. If a head mountable is supplied (like moving the head for pick/place optimization) its axes are known and can be taken into account for the evaluation. There is no need to use this constructor and existing code does not require any changes.

> Would it be better if `TravelCost` was an abstract class, TravellingSalesman had a `setTravelCost` method, and clients called `tsm.setTravelCost(TravelCostThreePointMotion(moveable,units))` ?
> 
This would mean, that code has to be modified to take advantage of the new metric. The current implementation defaults to the default down-looking camera which makes it useful for all use cases of the TSM, like checking blinds feeders or fiducials.

> The old comment "Plain old data TravelLocation for faster processing. Improved solving by a factor of 6 from using OpenPNP Locations directly" is no longer valid. What is happening here?

Thats bad bashing. Using Location suffered from a performance penalty because the effected locations are converted to the same units before the requested math is applied. And the unit-conversion always created new objects, even if the units don't change. I fixed that (making sure that the Location class is immutable - like we discussed and I've learned from you) Without creating new objects due to unit conversion, the code runs almost as fast as the current hand optimized version. (I actually have never seen the factor 6, the comment claims. It was more a factor 3 or so.) Anyhow, I think the TSM code is cleaner if it makes use of the Location class and it was easier for me to replace the metric if TravelLocation is based on Location.

tobydickenson:
> What Jerk values are you using? Any value not 0 and not close to infinity will make the machine slower by requesting a lower acceleration/feedrate then allowed for the axis based on the previous motion. Yes, with this jerk enabled, the metric will predict a faster motion time, but the physical relation is still preserved.

X
feedrate 950.0
acceleration 5000.0
jerk 400000.0

Y
feedrate 800.0
acceleration 2500.0
jerk 200000.0

I find the moderation of small movements to be an important factor in placement accuracy. The moderation reduces machine vibrations. The lumenpnp is not the most rigid of machines.

> I don't see how that effect the optimization. The actual time is not relevant, just the change in cost between different options. At present constant cost is expressed by a cycle, which is not physical.

Those factors affect the **actual** cost of each movement. The quality of the optimisation relies on having a "cost function" in the software which is an accurate model of the actual cost.

There are some advantages on the other side of course! The measurement-driven approach in #1800 will have an imperfect "cost function" due to noisy measurement data, the linear interpolation between data points, extrapolation for long movements, etc. I am excited to run some tests.

> The TSM test code produces graphics you my use to evaluate its performance.

Thank you for that reference. I will check it out.

> > ## Code Review
> > I think it is a mistake to put the `HeadMountable` into the TravellingSalesman constructor. It makes the TSM significantly less general.
> 
> The default constructor still does not need a head mountable. If a head mountable is supplied (like moving the head for pick/place optimization) its axes are known and can be taken into account for the evaluation. There is no need to use this constructor and existing code does not require any changes.
> 
> > Would it be better if `TravelCost` was an abstract class, TravellingSalesman had a `setTravelCost` method, and clients called `tsm.setTravelCost(TravelCostThreePointMotion(moveable,units))` ?
> 
> This would mean, that code has to be modified to take advantage of the new metric. The current implementation defaults to the default down-looking camera which makes it useful for all use cases of the TSM, like checking blinds feeders or fiducials.

I overlooked that default constructor. Yes, all agreed.

> Thats bad bashing. Using Location suffered from a performance penalty because the effected locations are converted to the same units before the requested math is applied. And the unit-conversion always created new objects, even if the units don't change. I fixed that

Got it! Yes thats a big improvement.

janm012012:
Hi Toby!

> I find the moderation of small movements to be an important factor in placement accuracy. The moderation reduces machine vibrations. The lumenpnp is not the most rigid of machines.
> 
Your values are reasonable and probably most machines will suffer from limited rigidity, so we shall definitely take that into account.

> > I don't see how that effect the optimization. The actual time is not relevant, just the change in cost between different options. At present constant cost is expressed by a cycle, which is not physical.
> 
> Those factors affect the **actual** cost of each movement. The quality of the optimization relies on having a "cost function" in the software which is an accurate model of the actual cost.
> 
For the optimization the **difference** in cost is what counts, not the absolute value. The TSM only checks if a route is faster then the other but never cares how fast it actually is. As long as the metric reproduces that correctly, I don't see a reason why jerk needs to be taken into account. (I did the same with jerk a few years back, but the math gets incredible lengthy because you need to solve a third order equation. I'd avoid that here.) Do you see any situation where the metric would fail for any jerk setting? 

Jan

tobydickenson:
> Do you see any situation where the metric would fail for any jerk setting?

All those items on the bulleted list above will make some difference to the actual travel time. Hopefully it wont matter. Hopefully the (feedrate+acceleration) algorithm is a sufficiently good approximation, and hopefully the generated TSM route and placement order will be satisfactory. We can prove that by testing.

janm012012:
> We can prove that by testing.

I'm looking forward for your test results.

tobydickenson:
Hi Jan,

Thank you for the pointer to the svg output from the TSM test. This has been super helpful.

With your PR 6c7b7b446afc22e29a953d23774f1fa408553b34  I am seeing the following bad TSM solution to the test. The test passes, but its not right:
travelling-salesman13737305920500516589

I can make it do something good with the following patch. This simply applies a linear 10x scaling to the travel cost which should not make a difference to the solution (right?):
travelling-salesman4778932488860243150

```
diff --git a/src/main/java/org/openpnp/util/TravellingSalesman.java b/src/main/java/org/openpnp/util/TravellingSalesman.java
index 49c7f1b5b3..a434a2597a 100644
--- a/src/main/java/org/openpnp/util/TravellingSalesman.java
+++ b/src/main/java/org/openpnp/util/TravellingSalesman.java
@@ -152,7 +152,8 @@ public class TravellingSalesman<T> {
             return 0.0;
         }
         if (this.travelCost != null) {
-            return travelCost.getXyzCost(la, lb);
+            double r = travelCost.getXyzCost(la, lb) * 10;
+            return r;
         }
         else {
             return la.getXyzDistanceTo(lb);
```

I am out of time for debugging tonight. Nothing really stands out as looking wrong in `TravellingSalesman.java`.

janm012012:
Hi Toby!

I found the same issue. The stability/results heavily depend on the parameters, the solver is started with. It seems, that the temperature and the absolute scale of the metric are related via https://github.com/openpnp/openpnp/blob/e468422e544562b1e7b6cb7b3dc88df6de4f9b11/src/main/java/org/openpnp/util/TravellingSalesman.java#L268. This makes them susceptible to the change of the metric and to your change of the scale of the metric. According to Wikipedia there is no good general advice how to choose the parameters `startingTemperatur` and `coolingRate`. Wikipedia also mentions, that a Restart feature is commonly used to prevent the solver from getting stuck in a local minimum. We don't have that yet...

According to https://github.com/openpnp/openpnp/blob/e468422e544562b1e7b6cb7b3dc88df6de4f9b11/src/main/java/org/openpnp/util/TravellingSalesman.java#L317 the solver is already driven by the initial travel distance, which depends on the metric and shall compensate for the change in the scale of the metric. I'm surprised, that this does not work as expected.

An other aspect: the test (in this PR) uses the `BasicJobTest/machine.xml` which does not define any axes. I don't know what parameters default axes get if not defined by config file. I'd suggest that you debug using more realistic parameters.

Finally, the test setup uses rows of random locations in X/Y direction. The new metric has a rectangular characteristic, which makes it less sensitive to additional x/y offsets compared the to current geometrical distance. This makes it probably less easy to find the perfect path because it makes not difference when jumping a row if you go a little up or down as well.

Final question: how much shorter is the path if you scale the metric?

Jan

tobydickenson:
> An other aspect: the test (in this PR) uses the `BasicJobTest/machine.xml` which does not define any axes. I don't know what parameters default axes get if not defined by config file. I'd suggest that you debug using more realistic parameters.

ReferenceControllerAxis supplies the defaults. From memory, 250 mm/s feedrate and the other parameters are small but not zero

tobydickenson:
Hi Jan,

This branch has:
 * A fix for the TSM 'temperature' loop termination condition. This fixes the broken solutions in the pictures above.
 * An improvements to test. It now runs multiple times with different acceleration profiles.

https://github.com/tobydickenson/openpnp/commits/tsm-test-fix/

This changes the `travelCost` attribute to be non-final, which maybe you wont like but it makes the test simpler.

janm012012:
Hi Toby!

Your fix to replace the hardcoded _0.1_ with a meaningful _endTemperature_ based on the startTemperatur is nice and works very well! It provides stable good (not perfect) results (covered by the seeding of the random number generator).  According to my tests even with _endTemperature = startingTemperature/1000_ the results are almost the same for ~75% of the cost (iterations). (I've counted the number of iterations and compared them for the 610 cities test case with seeding switched off. With _endTemperature = startingTemperature/10000_ it takes about 5.6e6 iterations to reach ~195sec. while with _endTemperature = startingTemperature/1000_ you get the same result in 4.2e6 iterations.)

How do we get your fix and the metric together?

In an other round I'd like to check the behavior for small problems with 2..4 cities as we currently very frequently use for pick, align and place optimization on multi-nozzle-machines and for common fiducial checks. IIRC the solver takes more iterations then a brute force solver would require and does not guarantee a perfect result. I'll probably plug my new solver in for this small problems.

Jan

PS: I just noticed that the TravelingSalesmanTest is actually named TavelingSalesmanTest (without the 'r'). Do you think it would be worth fixing this typo?

tobydickenson:
Thank you for testing this. My `endTemperature` coefficient was not chosen with any great care beyond giving approximately the same number of iterations in the unit test scenario.

> How do we get your fix and the metric together?

If you have both branches locally you can merge one into the other. Maybe I misunderstand your question.

> In an other round I'd like to check the behavior for small problems with 2..4 ....  more iterations then a brute force solver would require and does not guarantee a perfect result. I'll probably plug my new solver in for this small problems.

If definitely makes sense to get a guaranteed perfect solution for small problems.

> PS: I just noticed that the TravelingSalesmanTest is actually named TavelingSalesmanTest (without the 'r'). Do you think it would be worth fixing this typo?

Changing the file name breaks `git log <filename>`. That shouldnt stop us doing meaningful refactoring of course, but I dont think fixing a typo is worth it.

tobydickenson:
> I'm looking forward for your test results.

I have some good news.

A numerical analysis of the measurement data recorded by #1800 shows that the timings recorded by backlash correction are noisier than I had expected. The factors I mentioned earlier are real, but, apart from Jerk, they are barely larger than that noise.

Separately, I have tried a production run with this PR. I previously noted the subjective assessment that #1800 was causing the machine to make sensible coordinated multi-nozzle moves, and the same was true of this PR. So I am pleased to announce that #1800 is dead.

janm012012:
Hi Toby!

> If you have both branches locally you can merge one into the other. Maybe I misunderstand your question.

I did not know yet, that I can pull a branch from an arbitrary URL... That made merging your branch easy.

I tested it and I'm not yet convinced with the modification you did to the test. The reason why I added the cost metric is, that travel distance is more or less meaningless if you're interested in getting the job done quickly and with physical parameter. Therefor I don't understand - yet - why you choose to validate the test results against the linear distance. What every you configure the axis, the distance will very likely not be minimal. You can see the effect by comparing the A2, B2 and C2 results, A2 and C2 are at the minimum of the allowed range you specified and B2 is at the max and about 50% longer then A2 and C2.  I don't see how that classifies good and bad TSM results. Why don't you like the old code which compares the result of the TSM against an upper limit?

Jan

janm012012:
> I have some good news.

Many thanks for testing! Good to know that it also works for you!

tobydickenson:
>  Why don't you like the old code which ....

I agree I didnt quite achieve what I was aiming for here.

1. This test is covering lots of code, and there have been several recent changes in different areas. For testing the TSM loop termination change it seemed better to use a metric which is independent of the TravelCost.

2.The test simulates three different machines with different accelerations. Each machine produces solutions which run in very different times because they have different accelerations, but the overall travel distances are all approximately the same. Checking distance,  even with a wide tolerance, is a good way to confirm that the TSM has not converged on a birds-nest solution like the first image in this thread. Agreed that the 50% spread is stretching this.

3. The test runs with different problem sizes, across different machine profiles, and I wanted to avoid providing separate pass/fail limits for all combinations. I agree there was a loss from _removing_ your original test limits, so maybe this can't be avoided.

> ... compares the result of the TSM against an upper limit?

but lets use a lower limit too!

janm012012:
Hi Toby!

I thinks its a good idea to extend the test against different machines/axes configurations. I also agree that using a different metric for verification is a good idea. And I also agree, that we shall add a lower bound as the geometrical metric that's used for verification generates an arbitrary number. This number is very likely go up if the optimization gets worth, but might also go down.

However, I'd like to see tight limits to flag even small degradation from the current state. With the current implementation, we only get a single error (in B2) if we relax the endTemperatur definition because it crosses the upper bound. All other tests (B0 and B1 don't run, so I don't see them) show significant increase in (geometrical) distance but still pass because they are at the lower limit with default conditions. I'd really like to see tighter bounds here that flag more errors to make all tests useful.

Jan

tobydickenson:
All agreed

tobydickenson:
That looks good to me.

We had another production run today on different boards and everything worked as expected, so to me this looks ready to merge.

janm012012:
Many thanks for testing again and reviewing! I'm merge it now.

PR #1842, opened 2025-05-30
How do I make sure part X is placed last?

tobydickenson:
# Description

This PR satisfies several common problems relating to placement sequencing:
 a. How do I make sure part X is placed first?
 b. How do I make sure part Y is placed last?
 c. How do I make sure part J is placed before part K?
 d. How do I make sure that nozzle tip Z is not left until the end, because that leaves my multi-nozzle machine limping along on only a single nozzle. (This was my problem solved by #1799, but the feature in this PR would have satisfied that requirement too.)

I generalize all of those scenarios into three related requirements:
1. User-controlled **strict** sequencing. For example if the board has a short 0603 passive next to a tall aluminium can capacitor, then it is critical that the short part is placed before the tall part. If the tall part was placed first, then the nozzle would collide with it when placing the short part. This sequencing must account for deferred placement errors too: If there is a deferred error when placing the short part then the tall part can not even be considered for placement.
2. User-controlled **weak** sequencing. The user might want to express a hint about sequencing to help the JobPlanner, without the disadvantages implied by strict sequencing.
3. Parts without sequencing. Most parts can be placed in **any** order. The Job Processor and Job Planner make use of this to optimise the nozzle tip sequence and movement path.  We need to give this optimiser as much flexibility as possible.

# Instructions for Use - New Wiki section

## Rank

The **Rank** provides flexible control over placement sequencing. The rank is an integer, and placements are generally processed in rank order from low to high. The default configuration is that all placements are at the same rank which indicates to openpnp that the placements can be handled in any order, and the [Job Planner](https://github.com/openpnp/openpnp/wiki/Job-Processing#job-planner) makes use of this to calculate an optimised tip loading strategy and an optimised path through all of the placements.

A rank difference less than 10 indicates a _order preference_: The Job planner will generally load tips for the smallest rank first, and will generally process the placements in the smallest ranks while also following its global optimisation strategy.

A rank difference 10 or greater indicates a _strict order_. If there is an outstanding placements of rank R then it will block all placements of rank R+10 and greater.

For example the rank feature can be used in a workflow where an expensive IC is only placed on the board after all the low-cost passive parts have already been placed successfully. This can be achieved with the following configuration:

|Rank|Usage|
|----|-----|
|0|All the placements for the passive parts are at rank 0.|
|10|The placement of the expensive ICs are at rank 10. These placements will only be considered after everything else at rank 0 has been successfully placed. |

The absolute value of the rank property does not matter. That configuration using rank 0 and 10 would work the same with ranks 100 and 110.

This more sophisticated example supports a board which has some tall aluminium can capacitors, and lots of other short parts. Some of the shorter parts are located close to the tall can, and it would be impossible to place them after the can without causing a nozzle collision. However there are no specific sequence requirements for all the other parts on the board.

|Rank|Usage|
|----|-----|
|100|The placements which are overshadowed by the tall can are at rank 100. This is a hint to the Job Planner to place them early in the job. |
|105|The majority of placements on the board are at rank 105 because they can be placed at any time. The Job Planner optimiser can choose to place these at the same time as those at rank 100 and at rank 110, because the difference between rank values is less than 10.|
|110|The tall can is at rank 110. This placement will only be considered after the overshadowing parts at rank 100 have been successfully placed. |

# Justification

This is a frequently requested feature.

# Outstanding Work and Concerns

1. Do we like the name "Rank"? Openpnp has already used lots of other words relating to sequence order and I struggled to find something distinctive and searchable.
2. The code has lots of `Logger.info` that need to be `Logger.debug` before it is merged.
3. This works in the simulated machine, but I have not yet tested this on a real machine.
4. This needs a unit test for the strict sequencing characteristic.

# Rejected Ideas

A goal of this implementation is to avoid over-constraining the placement sequence. That is, to avoid unnecessarily reducing the pool of placements processed by the Job Planner. An alternative design based on a **placement dependency tree** would be better still. But I think that makes the user interface more complicated. I hope this design is a good balance between usability and Job Planner performance.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. -- this is still outstanding
6. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. -- **yes** adding the Rank property.
7. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. -- tests pass.


janm012012:
Hi Toby!

This is brilliant work! Its the same idea I considered easy to implement, except that you already extended it by weak sequencing.

Here are a few points I noticed why reading the code:

- Please update the PR description mentioning that it does change files in both module and spi.
- Please add a tool tip that explains the difference between weak and strict sequencing. This might otherwise lead to new support requests claiming that sequencing does not work.
- You're using rank=0 at two locations (in the initializer and when sorting/mapping). I'd recommend to use a variable like defaultRank to clarify its meaning and allow it to be safely modified in the future.
- You mentioned, that weak sequencing shall aid in choosing the "right" nozzle tip. I don't see how you interfaced that. The tip that's chosen next, is selected at https://github.com/openpnp/openpnp/blob/e218f2649b6a8050eb3688d08a3d45712e17f6bb/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L2684 as the first from the list of nozzle tips, which is either build up as part of the placement sorting/optimization step or as list of all nozzle tips.
- You mentioned, that sorting would be stable. Does that mean, that the original order is preserved even if the list is sorted by rank? The original list has been optimized before using the user selected strategy (which now also involves then rank). I think we shall not distort the order. However, it's likely a good idea to chose the first nozzle tip from the list required to process the first placement with the lowest rank.

Jan

tobydickenson:
>     Please update the PR description mentioning that it does change files in both module and spi.

Oops. Thank you

>     Please add a tool tip...

Done.

This change exposed another bug: This tooltip code obviously needed to call `convertColumnIndexToModel` because the _Board_ placements table is built as a clone of the _Job_  placements table but with some columns deleted, and therefore some column index numbers differ. The need was obvious in this case, but all table views need the same treatment because the user can reorder columns. I will fix this for some other table tooltips in a separate PR.

>     You're using rank=0 at two locations...

Done.

>     You mentioned, that sorting would be stable. Does that mean, that the original order is preserved even if the list is sorted by rank?

Yes, the order of items with the same rank is unchanged.

> The tip that's chosen next, is selected at...

Yes, planWithNozzleTipChange was ignoring rank, and I agree that is a mistake. Thank you.

It should be selecting a tip to process the first part on the lowest rank. And if that part allows multiple tips, it should use the Job Processors nozzle preference order.

That means we can do the rank stable-sort very much earlier, and in the same function as the rank+10 filter which is nicely cohesive.

I have pushed that code, but untested.

I have been looking at this logic much closer than previously and this has raised some questions about the exiting design:

a. On a multi-nozzle machine, it will always be the first nozzle which takes this branch
https://github.com/openpnp/openpnp/blob/396cdcbc3f10fbba88b084607c7a91a96dbfa598/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L2595-L2598
and the second nozzle which selects a part based on minimal travelCost. I wonder whether it would be a useful optimisation to consider the alternative combination

b. This logic filters placements if the strategy is  other than _Minimize_. That gives a difference between _Minimize_ and _StartAsPlanned_, but the [wiki](https://github.com/openpnp/openpnp/wiki/Job-Processing#configuration) documents that they are the same except for the start of a job. Is this a code bug or documentation error?
https://github.com/openpnp/openpnp/blob/396cdcbc3f10fbba88b084607c7a91a96dbfa598/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java#L2540-L2542

janm012012:
Hi Toby!

Many thanks for updating the code. I've put some more commend into the changeset. Overall this looks very good, especially with your new tooltip. This will make it a very useful feature almost immediately.

I'm confident, that the strict sequencing this PR adds will work straight away as it splits the job internally into multiple jobs which are each optimized and processed individually.

However, I don't see that this PR provides the weak sequencing it claims. If I understand the documentation correctly, weak sequencing shall favor placements with lower rank. You're trying to enforce this by sorting the placements with respect to rank. In the planner the placement selection depends on the configuration. We currently have three options here:

1. **Minimize** continues until no placement is available for the loaded nozzle tip.
2. **Strict** exactly follow the placement sorting, even if that requires additional nozzle tip changes.
3. **StartAsPlanned** uses **Strict** on first pass and **Minimze** on all other passes. This enforces, that the first placement in the list is always placed first (even if that requires a nozzle tip change) and all other placements are placed minimizing nozzle tip changes.

(I hope, that clarifies the questions you had. IIRC I reused this strategies to enable/disable the optimization for the n+1 nozzle.)

Depending on the planner strategy a placement compatible with a nozzle tip will be favored over its rank and n+1 nozzle optimization might also favor a placement by best cost/distance. Did you checked, that the weak sequencing is at least somehow preserved by this?

Jan

tobydickenson:
Thank you again for the review.

I am currently working on some changes that started out as a unit test for Rank, but this has developed into a more general unit test for the Job Processor & Job Planner optimiser. I am getting some interesting results that (so far) are different to what we both might expect for the baseline behaviour, even without the complications of Rank. I will share code & observations as soon as I am confident the results are reliable.

You might be right about the weak sequencing (that is one reason why I started on this unit test). It is easy to define the strict ordering semantics, but the weak sequencing is trickier.

> Depending on the planner strategy a placement compatible with a nozzle tip will be favored over its rank.

Yes, 'Minimize' has a policy of avoiding nozzle swaps until forced. The intention is that this policy still applies for rank differences less than 10. The Minimize policy is stronger than the sequencing implied by "weak sequencing".

There are several factors that provide some sequencing within the limits of that policy:
1. When forced to swap tips, it always picks a tip that matches a placement on the lowest rank.
2. At least one part from each planning cycle will be from the lowest rank. The Job Planner handles the first nozzle with the branch commented "no further optimization possible or requested, just choose the most preferred placement on the list". This will always select a placement from the lowest rank (filtered for tip changes). But the second (and third!) nozzle uses travelCost optimisation with no rank bias.

tobydickenson:
Hi Jan,

The new push above implements a unit test for the Job Planner.

I have been debating how to present this change in a PR. The unit test has pinpointed some curiosities in the Job Planner behavior. That is, some of the values in assertions are not as we would want them to be. I have fixes for those issues, but including them in this PR makes it far too large. So I am presenting this unit test as-is, and will follow up with Job Planner fixes in a later PR.

## Unit test operation

The unit test uses a familiar simulated machine with the following changes:

1. It is configured for speed. Fiducial checking and bottom vision are disabled, and axis motion limits are set high.
2. It is configured with two nozzles. The nozzles start the job with no tip loaded.
3. The feeders are AutoFeeder, again for speed and for easy analysis because the pick location is unchanged during the job.
4. The Job Planner has been extended to publish its operation on eventbus, and the unit test records these events.
5. That planning data is recorded in a CSV file in a temporary directory for detailed manual analysis.
6. The test makes assertions on the following parameters:
   1. The number tip changes
   2. Nozzle utilisation. That is, the ratio of number of placements and number of pick&place cycles. This value is 2.0 if both nozzles are fully used in every pick&place cycle, or a smaller number if some nozzles go unused in some cycles.
    3. The number of pick&place cycles. This is just a different way to express the previous number, but I found it useful to have both
    4. The number of part changes. This is a small number if the planner empties one feeder before moving on the next, and a larger number if it flits between different feeders
    5. The average 'travelCost' during the planning stages. This is the key measure of performance.

The test performs various full jobs using different Job Planner settings. Some are testing existing JobPlanner options, and some are various scenarios using the new Rank feature.

'testRankWeak' is a test for the scenario where you previously commented "I don't see that this PR provides the weak sequencing it claims". A description of this test is here:
https://github.com/tobydickenson/openpnp/blob/85861c2cec1464e3b65e53ce380353e03fc34345/src/test/java/JobProcessorTest.java#L178-L192
The results on this test show that the first group is placed right at the start of the job, and the second group is placed half way through. Given what I was attempting to achieve that test is a pass, but I agree it does not quite match up to my original description of the expected behaviour. The promised "weak ordering" is being overruled by other Job Planner optimisation priorities. I will think about how to update that documentation.

janm012012:
Hi Toby!

The tests you added are impressive! Just a side note: for me - as a manual-nozzle-tip-user - the time to change a single tip is significant and changing multiple tips takes only marginally longer the changing a single tip. Therefore I prefer to change multiple tips at once.

What is the outcome of your tests with respect to the default settings? Are thous the most (or usually most) efficient ones or would you recommend other defaults?

As far as I can see, weak sequence shall be fine for single-nozzle-machines and work as you intended/described. For multi-nozzle-machines the selection process for the N+1 nozzle seems to void that sometimes. (Btw: this is in sync with my previous observations: with 2 nozzles the gain is only about 70..80% at best.) As weak sequencing is something, that I don't see a big need for, you may optimize it further or merge it as is. My only concern is the sorting you added. This shall not overwrite the intended ordering my more then the rank dependency. You said and documented that it would not. I'd love to see a (dedicated) test that verifies and enforces that.

Jan

tobydickenson:
> The tests you added are impressive! Just a side note: for me - as a manual-nozzle-tip-user - the time to change a single tip is significant and changing multiple tips takes only marginally longer the changing a single tip. Therefore I prefer to change multiple tips at once.

That would be a useful metric to include as an assertion. And it looks like it would be a straightforward optimisation tweak to hold back `planWithNozzleTipChange` for a few cycles if it means multiple changes could be done at the same time.
 
> What is the outcome of your tests with respect to the default settings? Are thous the most (or usually most) efficient ones or would you recommend other defaults?

For this simulated job, the tests show that `JobOrderHint.BoardPart` is best, which is far from the expected result. This simulated job highlights some curiosities that I have glimpsed while running real jobs, and I will follow up with some more PRs with fixes. With all those in place, yes the current defaults are looking good.

> As weak sequencing is something, that I don't see a big need for

I agree there is no *user*  requirement for weak sequencing.

My rationale for including that feature relates to the 'testRankWeak' test. We want both rank -5 and +5 placements to be handled efficiently using travelCost optimisation to mix them in with rank 0. We want to avoid the bad outcome where optimiser schedules the rank -5 parts very late in the job, forcing it to then place the rank 5 parts inefficiently because all the good travelCost pairings have been taken. 

The current implementation achieves that, but I agree this characteristic does not need to be exposed to the user. I agree it makes more sense for the user documentation to say that _rank difference < 10_ is "Non-sequencing rank difference"

> My only concern is the sorting you added. This shall not overwrite the intended ordering my more then the rank dependency. You said and documented that it would not. I'd love to see a (dedicated) test that verifies and enforces that.

Agreed that would be a good test.

> you may optimize it further or merge it as is.

I think we should hold off merging this and other new features into the 'test' branch until after the release of version 2.4.

Should we create a 'test-after-2.4' branch so that we can merge and close these PRs without interfering with the 2.4 beta?

janm012012:
Hi Toby!

> That would be a useful metric to include as an assertion. And it looks like it would be a straightforward optimisation tweak to hold back `planWithNozzleTipChange` for a few cycles if it means multiple changes could be done at the same time.

I'll study your new PR #1852 how you did that. To me this looks not trivial at present, because I have seen a lot of jobs, that
need one tip for about double the amount of placements compared to other nozzle tips (for me 0402 vs. 0805/PLCC/QFN...).

> For this simulated job, the tests show that `JobOrderHint.BoardPart` is best, which is far from the expected result. This simulated job highlights some curiosities that I have glimpsed while running real jobs, and I will follow up with some more PRs with fixes. With all those in place, yes the current defaults are looking good.

I guess I shall read your other PR #1857 before asking if this still holds true if you swap the IDs of two arbitrary placements. It would not make sense to sort for C1, C2, C3... and has to be bad look with the nozzle tip sorting... I'll read your PR...

> I think we should hold off merging this and other new features into the 'test' branch until after the release of version 2.4.

It's your PR and your work. I think its very valuable and deserves merging. If you're ok to hold it back, do so. I don't see very urgent needs yet.

> Should we create a 'test-after-2.4' branch so that we can merge and close these PRs without interfering with the 2.4 beta?

No, I don't think we need that yet. There is likely some merge conflicts that might arise from multiple PRs affecting the same files, but thus are usually easy to solve - later. And even if we would group them into a new branch, developers would have to create their PRs against this new branch to take advantage from it... If it's ok for you, that you work is pending for some weeks, I'd just wait and merge it when the test branch is ready to accept it.

Jan

PR #1843, opened 2025-06-01
photonfeeder: parallelize positioning the nozzle above the feeder and feeding

Jorropo:
# Description

While the feeder is moving the tape, position the nozzle above the pick location to save time.

# Justification

On a random build I happen to be running as I am typing theses words, a Pick Pick Vision Vision Place Place cycle is 14s, ~2s of that is the nozzles moving over to the feeders, ~5s is the feeders feeding,
by masking the time for the nozzles to move to the feeders we should increase CPH by ~14%.

# Instructions for Use

Configure a photon feeder.
Make sure the `Move while feeding?` option is set for it (default to `true`).
Start a job using this feeder.
Observe the nozzle moving above the feeder *while* the feeder is feeding.

# Implementation Details

1. How did you test the change? yes
4. Be sure to run `mvn test` before submitting the Pull Request. yes


Jorropo:
I am experimenting with this, I think the worst that could happen is that the heads accelerating or decelerating could dislodge a component from the tape as it is moving, but this is already a problem given we uncover the tape completely before moving the head.

Jorropo:
It works, rough timing give a 5~15% CPH improvement

First video (just this behavior without #1844)

https://github.com/user-attachments/assets/d5fcdc05-4890-4156-ad03-6e5bef48f449

Second video, current state of this PR:

https://github.com/user-attachments/assets/9181d73b-7048-470f-81e5-879c4d0e2742

Jorropo:
I have reworked this:
- It is now stacked on top of #1844 (since #1844 is much simpler)
- It now copies the implementation of the Reference Feeder and the Bamboo Feeder
- Made the behavior configurable (toggle on / off)

I have tested my changes on my lumen

tobydickenson:
There has been some review discussion on the lumenpnp discord. A summary:
* localisation test needs to be fixed (translations not in order in the properties file)
* property access method names are wrong, `isMoveWhileFeeding` should be `getMoveWhileFeeding` and `setMoveWhileFeed` should be `setMoveWhileFeeding`.
* The photon unit tests are failing because moveWhileFeeding is enabled by default. The tests try to feed, which of course now tries to move. This moves the head, which is accessed from the nozzle. But the unit test nozzle is not connected to a head object. For the purposes of preserving the existing unit tests, it seems expedient to disable moveWhileFeeding for these tests.

Jorropo:
I've applied the fixes from the message above.

Jorropo:
I havn't yet ran a job, but I've confirmed the machine boot, detect feeders and the config work after
> - property access method names are wrong, `isMoveWhileFeeding` should be `getMoveWhileFeeding` and `setMoveWhileFeed` should be `setMoveWhileFeeding`

tobydickenson:
This all looks good in code review. Thank you for the contribution.

We are currently in the middle of the 2.3 beta release, and there are other pending pull requests which I have suggested need to wait for 2.5. However I am tempted to merge this into 2.3 now. This PR is very low risk:
 * The changes are limited to photon feeder
 * The change follows the same implementation strategy as the equivalent feature in ReferenceAutoFeeder.
 * For photon users the change is disabled by default with a configuration checkbox.

Given that it is merged late in the release cycle, would it be prudent to label this as an *experimental* feature in the release notes? I'm sure it will be fine, and the good news will spread quickly in the photon feeder user community.

We should add an explanation to the [photon feeder wiki](https://github.com/openpnp/openpnp/wiki/Photon-Feeder). I suggest the following. @Jorropo feedback on the following paragraph please:

----

The **Move While Feeding** option (New in version 2.4) allows the nozzle to move to above the pick location while the feeder motor is running to feed the part. This option improves throughput. For this feature to work it is necessary to enable the **Allow continuous motion** option on the [advanced motion planner](https://github.com/openpnp/openpnp/wiki/Motion-Planner#motion-planner).

Jorropo:
> New in version 2.4

New in version 2.3 ?

> For this feature to work it is necessary to enable the Allow continuous motion option on the [advanced motion planner](https://github.com/openpnp/openpnp/wiki/Motion-Planner#motion-planner).

I would remove this, I havn't tried it but I don't see how it would not work even without continuous motion.
The GCODE command to move the tape isn't a classical synchronous command, it's a command that does an RS485 transaction.
As far as marlin knows the command is finished (thus it would start moving the head) at the end of the RS485 transaction, which completes before the feeder starts feeding.

---

Labeling it experimental sounds good, however I don't know about making it false by default.
Afait because it's the first time the feature is added it's our chance to choose a default flag.
Because if we do:
- 2.3 exists experimental default false
- 2.4 default true

Someone updating to 2.3 first would append to the config as false
And then if they update to 2.4 openpnp wouldn't enable it for them since it can't differentiate between someone who disabled it intentionally or someone who left it on defaults.

tobydickenson:
Is it easy for you to test what happens with Allow continuous motion disabled? It would be good to know

Jorropo:
Sure, 5 minutes

Jorropo:
I've just tested it, it works perfectly fine, it's worst than with continuous motion but everything is worst when you turn continuous motion off so ...

tobydickenson:
> New in version 2.4

oops, that should have been "in version 2.3". Apologies for the confusion.

PR #1844, opened 2025-06-01
photonfeeder: poll the feeder at 20hz rather than waiting the expected feed time

Jorropo:
# Description

Poll the feeder at 20hz while we are waiting for it to finish feeding.

# Justification

The feeder's firmware seems to be very conservative about it's timing estimates and there is marginal costs to polling instead.
This increase CPH by ~15 more %.

# Instructions for Use

None, just profit.

# Implementation Details

I've tested it, and it works.

### DIFF
--- a/src/main/java/org/openpnp/machine/photon/PhotonFeeder.java
+++ b/src/main/java/org/openpnp/machine/photon/PhotonFeeder.java
@@ -24,6 +24,9 @@ import org.simpleframework.xml.Attribute;
 import org.simpleframework.xml.Element;
 
 import javax.swing.*;
+
+import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.regex.Matcher;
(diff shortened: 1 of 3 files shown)


Jorropo:
https://github.com/user-attachments/assets/9181d73b-7048-470f-81e5-879c4d0e2742

tobydickenson:
>  We could reorder the two PRs but I don't think it matters.

If you did that I could merge it now. This one is obviously right (great catch! thank you!) whereas #1843 may need some discussion.

Jorropo:
Sure fair will do

Jorropo:
@tobydickenson I've repushed this PR, should be an easy merge.
I've tested it and it works as expected.

tobydickenson:
please note the unit test failure

Jorropo:
@tobydickenson please how can I run a single test ? I've tried `mvn test -Dtest="feedWillCheckTheStatusUpToThreeTimesBeforeFailing"` but it doesn't run the test.

tobydickenson:
I'm not at a desk to confirm, but I think this should do it:
```
mvn test -Dtest=org.openpnp.machine.photon.PhotonFeederTest
```

Jorropo:
It does thanks.

tobydickenson:
Hi @Jorropo  any news on fixing this unit test glitch? Any help needed? This is a huge improvement that would be great to have in the next stable release, but it really needs to be merged in the next week before the beta release freeze.
Toby

Jorropo:
@tobydickenson after getting around to fixing the tests here what were changed:
- Fixed a bug where we should always try to read from the feeder three times regardless of expected feed time logic. Could in theory trigger if the reported expected feed time is <= 50ms, probably impossible condition IRL but good hardening none the less.
- Modified implementation detail test `feedWillCheckTheStatusUpToThreeTimesBeforeFailing` into `feedWillCheckTheStatusAtLeastThreeTimesBeforeFailing`.

`mvn test` passed on my machine :sparkles: thanks for getting this in v2.3

PR #1854, opened 2025-06-11
photonfeeder: poll using a monotonic clock

Jorropo:
Avoid a bug if polling the feeder while DST hits.

Using untyped time feels wrong, in Go `Instant`'s equivalent (`time.Time`) is
an hybrid monotonic clock & wall clock and the previous code would just-work™,
but after searching for 5 minutes typesafety in this 2 line case is not worth solving.

I didn't tried running both pieces of code through DST or TZ config change to see what would happen but this feels better.

This PR is stacked on top of #1844


Jorropo:
Sending as a second PR because `System.nanoTime` feels awfully primitive but tbh idk what would be modern / type safe way to achieve this.

PR #1857, opened 2025-06-11
Job Planner fixes

### LINKED ISSUE #1851: Job planner optimization feature request: ensure picks are from different feeders on each nozzle
This was mentioned on lumenpnp discord.

Photon feeders are slow to feed, and at some point I think we will see patches which trigger the feed to start earlier in the JobProcessor cycle. On a multi-nozzle machine it would be good to get multiple feeders starting to move. That would only work if the nozzles are all picking from a different feeder, which requires a JobPlanner change.

That fits in with my changes on PR #1842 and I hope can be implemented at the same time (the JobPlanner change - not the feeder change)

tobydickenson:
Hi Jan,

My apologies that this is quite a large PR.

The Rank feature unit tests have highlighted some bugs in the planner. The overall result of the planner is still quite good, which is a consequence of its robust design. I believe these are all curiosities that I have seen on the real machine, but had previously been unable to pinpoint. The 'averagePlanningCost' metric is showing that Board:Part job order outperforms 'NozzleTips' on this simulated test job, proving that not everything is working as it should. 

I suggest the best way to review is one commit at a time. Each commit is cohesive and fully explained by the commit message, with an overview and more details below. NB commits before 10 June you have seen before on #1842.

As before, there are lots of Logger.info that all need to be Logger.debug before this is merged, and in one case I note some old code in comments that needs to deleted.

## Feeder Focus

Commit 1824363e084ddc7fdd6cc7245b040cdfdc44696f

We might expect that Strategy.StartAsPlanned and Strategy.Minimize would be identical on a machine which starts with no tips loaded, but that recently has not been true.

In commit 71e301fdbf144e13972f691371402e56fe1936d7 in January you added the optimization option for "only consider placements that use the same feeder as the first placement". This feature is enabled for Strategy.StartAsPlanned but not Strategy.Minimize. I like this feature, but it doesnt make sense to enable it with the 'Strategy' since that is explicitly the *tip loading* strategy option. I suggest it is better as an orthogonal option, and I have therefore added a new enum for 'FeederStrategy' with initially two options: AnyFeeder and FeederFocus.

There is a new unit test to confirm that FeederFocus works through the feeders one at a time. In general this option is not a complete guarantee that when the job leaves a feeder it will never return, but in most cases it does achieve this, and for this test job it does.

The unit tests for Minimize and StartAsPlanned now perform identically, which is expected because the test job starts with no tips loaded.

## DifferentFeeder

Commit 4e1afe7a2027678c842a6fb260e6cc7cee22e114

This is a new feature. It is a new third option for the FeederStrategy enum. FeederStrategy.DifferentFeeder  implements a feature requested on the lumenpnp discord. closes #1851. This option ensures that each nozzle picks from a different feeder.

I have included this new feature in this PR because it presents an interesting test case. This is another example where it can't exactly follow the job plan, but we still want the overall algorithm to plan a job that runs efficiently. The baseline job planner does an ok job, but the following bug fixes make FeederStrategy.DifferentFeeder only fractionally less efficient than FeederStrategy.AnyFeeder.

image

[This change will need a corresponding wiki update]

## Job Order start points

Commit e6921665644f36d0e274019d4d50327f995978d5

This fixes a puzzling characteristic I have seen a few times on the real machine, and it occured right away when implementing the unit tests.

This change relates to how it creates the job order using TSM sorting. The job order is a funny thing: It plans a route through the whole job, but that never actually gets used as a "route" in the conventional sense. It uses the first few placements of the route for pick & place, then discards the remainder of the planned route. Then it recalculates a whole new route from scratch on the next planning cycle. The expectation is that the route is largely stable from one cycle to the next, but often it is not.

There are two problem scenarios:
1. Previously when planJobPlacementsByPickLocation ran the TSM to generate the planned route it was not specifying a start point. Although the TSM seems pretty stable when given identical input, this input data set is constantly changing because it is reducing in size as parts are placed. The planned route can therefore randomly switch front-to-back. Below is an extract of the log from 'testNozzleTips' prior to the fix for this problem. This log extract shows the TSM-selected route through the feeders. This is the output from planJobPlacementsByNozzleTips. Note how f0201-1k starts out at the end of the route then randomly swaps to the start at point x.
2. Previously planJobPlacementsByNozzleTips ran a TSM for each nozzle, where the end of one path was used as the start point for the next path. A common pattern is that the planned route for the first nozzle tip is left-to-right through the feeder bank. Then the planned route for the second nozzle tip is likely to be in the reverse direction, right-to-left. And the third nozzle tip goes back again, left-to-right, etc. This causes the planned route to be sensitive to the order in which each nozzle is included. Point y in the log below is where the route [f0805-10k, f0805-1k] swaps to being planned first, and therefore the route involving f0201-1k reverses direction again.

I am working from an intuitive assumption that the job will be most efficient if the planned job order remains consistent wherever possible during the job. I have therefore changed the pick location TSM to always specify a start position, using a policy designed to promote this type of consistency.
a. For the first nozzle, it uses the location of the first feeder in the planned route for the first nozzle in the previous run.
b. For subsequent nozzles N it uses the location of the first feeder in the route for nozzle N-1.

```
INFO: feeders [f0603-10k, f0402-10k, f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0603-10k, f0402-10k, f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0603-10k, f0402-10k, f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0603-10k, f0402-10k, f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0603-10k, f0402-10k, f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0603-10k, f0402-10k, f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0603-10k, f0402-10k, f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0603-10k, f0402-10k, f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0603-10k, f0402-10k, f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0402-10k, f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0402-10k, f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0201-10k, f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0201-1k, f0402-1k, f0603-1k]           <------- x
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0201-1k, f0402-1k, f0603-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0201-1k, f0402-1k, f0603-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0201-1k, f0402-1k, f0603-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0201-1k, f0402-1k, f0603-1k]
INFO: feeders [f0805-1k, f0805-10k]
INFO: feeders [f0805-10k, f0805-1k]                     <-------- y
INFO: feeders [f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0603-1k, f0402-1k, f0201-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0402-1k, f0201-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0402-1k, f0201-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0402-1k, f0201-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0402-1k, f0201-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0201-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0805-10k, f0805-1k]
INFO: feeders [f0805-1k]
INFO: feeders [f0805-1k]
INFO: feeders [f0805-1k]
INFO: feeders [f0805-1k]
INFO: feeders [f0805-1k]
INFO: feeders [f0805-1k]
INFO: feeders [f0805-1k]
INFO: feeders [f0805-1k]
INFO: feeders [f0805-1k]
INFO: feeders [f0805-1k]
INFO: feeders [f0805-1k]
INFO: feeders [f0805-1k]
```

Fixing this bug I definitely see the test jobs progress more systematically through the feeders, rather than randomly jumping around. Ironically this makes some 'averagePlanningCost' scores get worse. It appears that the random jumping from this bug was a mitigating factor for the bug described below.

## Nozzle planning order

Commit 8fa3f8a88aaeef854dc7ff72a95c4d04336e0bb0 then 3ef1291e49393ace61e213791cf5158a5acb0b9a

Planner.plan processes the nozzles in a fixed order:
1. The **first nozzle** is processed first. It receives the next placement from the job order
2. The **second nozzle's** placement is selected second. It is selected using travelCost optimisation, using the first placement as an anchor point.

This leads to an asymmetry between the two nozzles. The first nozzle is following the job order route, while the second nozzle makes optimized placements around it. 'testNozzleTips' has a very good 'averagePlanningCost' score 1.13 if the first nozzle is on the left of the head and the second nozzle on the right. But swap those positions around in the machine configuration and the 'averagePlanningCost' is a very poor 1.67. It gets stuck in a pattern where the nozzle on the left is doing the travelCost search but it only has bad choices because all the remaining placements are off to the right.

I have reduced this asymmetry with a new optimisation that checks whether there is a lower travelCost outcome available by swapping the planning roles of the two nozzles. That is:
1. The **second nozzle** is processed first. It receives the next placement from the job order
2. The **first nozzle's** placement is selected second. It is selected using travelCost optimisation, using the first placement as an anchor point.

If there are more than 2 nozzles then the 3rd and subsequent are processed as normal.

This makes the 'averagePlanningCost' for 'testNozzleTips' uniformly 1.16 for both machine configurations. 
With all of these changes we can finally beat the Board:Part planning order!

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. -- this has been primarily driven by unit test. This is a provisional PR pending testing on a real machine.
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? -- yes (NB there are some pending changes described above)
4. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. -- 
5. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.


tobydickenson:
The unit test has always written an svg file showing the placement plan. It shows movement around the placement area, but that is not very helpful because it ignores movement around the feeder area and therefore I had not previously shared any. But the diagram is useful with a new simpler testAllOnePart test. This is a test where all the placements have been changed to the same part, eliminating the complications due to movement between feeder positions.

The svg file contains symbols like this which represent a pick/place cycle:
symbol
* The red circle is the placement location for the left nozzle, and the black circle is the right nozzle
* The number indicates that this was the 7th pick and place cycle of the job. The number is next to the placement(s) that were in the job order at this point. placements without a number were selected by travelCost optimisation.
* The grey horizontal lines show the offset of the nozzle from head center
* The black line joining the two grey lines shows the head movement between each placement. We want this to be short.

### testAllOnePart After Fixes
testAllOnePart
* All the head movement lines are relatively short. This indicates that the travelCost optimisation is working.
* The numbers cycle through the job from bottom-left to top-right indicating that the placement TSM is working.
* The placements are divided sensibly into zones of red circles (left nozzle) and black (right nozzle)  

### Before Fixes
testAllOnePart
* It is actually not too bad! There are lots of short head movement lines. travelCost optimisation is working
* It includes this very silly move. The placement on the left was picked up by the right nozzle (black circle). Not having the clean separation into zones of red circles and black circles suggests this mistake is made quite frequently.
backwards
* The numbers follow no pattern. The TSM is working, but not achieving anything beneficial.  
* There are several very long movements. With the placement TSM ineffective, the last few placements at the end of the job were at opposite sides of the board and this was the best move available.

janm012012:
Hi Toby!

I've read your description once and I'm very impressed! I'll have to read it again before I can appropriate comments.

If you did all this impressive optimization, will it still work for machines with more then two nozzles and how fare are we actually away from a full optimizer (TSM over the entire job)?

Jan

tobydickenson:
Hi Jan, thank you for the comments.

> If you did all this impressive optimization, will it still work for machines with more then two nozzles

Increasing the number of nozzles makes it a harder optimisation problem. It will still work, but machines with more than two nozzles might benefit from some further work. I do have a list of optimisation ideas. Let me write that up and add it here.

> and how fare are we actually away from a full optimizer (TSM over the entire job)?

There are some limited parts of the job planner that can be described as an "optimiser" - notably the travelCost process - but the higher level strategic planning mostly operates on heuristics. An advantage of this approach is that it makes the overall plan robust to unforeseeable changes. For example:
* When a part is rejected by vision and it needs to be repicked.
* If some parts come from a tray feeder and the pick location changes as parts are used.

We currently have several parameters which tune these heuristics. For example, job order. Sometimes "Nozzle Tips" is fastest. Sometimes "Nozzle Tip (inflexible tips first)" is fastest. In some weird scenarios maybe one of the "legacy" job orders might be fastest. Maybe a more achievable whole job optimisation process would be to auto-tune these parameters, using a new simulated-annealing optimiser which tweaks job planner parameters. The new unit test here provides a way to objectively measure the performance of a job.... Just an idea.

tobydickenson:
We had a first production run with this today.

This is a dense board with lots of parts, so it should be easy for the optimiser to find pick and place movements that are all highly coordinated between the two nozzles. And indeed this is what we observed during most of the job. Overall throughput is up 5%.

There was one exception: We had only one part on a strip feeder, on the opposite side of the machine from the other feeders, so the planner sensibly used both nozzles to pick from this feeder in each cycle. But some of the placements from that feeder were like the "very silly move" shown above. I will investigate.

janm012012:
Hi Toby!

Interesting and good to hear, that your fixes provide better results.

> There was one exception: We had only one part on a strip feeder, on the opposite side of the machine from the other feeders, so the planner sensibly used both nozzles to pick from this feeder in each cycle. But some of the placements from that feeder were like the "very silly move" shown above. I will investigate.

This sounds like artifacts of the Grady style optimizer for the second nozzle. If always the best placement is chosen, it is likely that one ends with one or some very bad combinations and misses the opportunity to take the second best in favor to avoid them and get better overall performance. That's what I meant by "full optimizer": optimize the entire job including all nozzles and do not use the Grady optimizer to overwrite it. If we could get the second nozzle into such an optimizer, we would likely find better combinations without "very silly move"s.

Jan

tobydickenson:
It wasnt quite what I thought. "very silly" moves were those that could be improved with a trivial local optimisation of simply swapping the placements between left and right nozzles. Those are an embarrassing failure. This, as you guessed, is a lack of global view.

I dont know the term "Grady"; do you have a reference for that please?

image
This test is a 6x5 grid of placements. It could cover 6x4 of that with near perfect efficiency in a tiling pattern. But after getting half way (the yellow shaded area) it fumbles the tiling pattern and selects the two placements highlighted in purple. After that it is stuck with making less than optimal moves to fill around the edges.


Incidentally, it appears I have staged the wrong commit for writing the svg images. I need to find the right code in some older branches.

janm012012:
Hi Toby!

I'm sorry, I must have mixed "Grady" with something else. I can't find a reference myself, but I remember that there is a good article on Wikipedia. I meant to say an optimizer that always takes the currently/locally best and does not look ahead.

The pattern you observed might be caused by the cost metric: one unit of motion cost forms a rectangular. Therefore the cost to stay in the same row is the same as for going a row up or down. Due to the xxx style optimizer for the second nozzle not looking ahead there is no indication that jumping a row would result in worth overall performance. I guess this could be solved either by a better (global) optimizer or by allowing the user to hand optimize the job (would require a GUI update to support sorting and an option to job-wise overwrite the processor/planner configuration). To me the first fits better into the general use case of prototyping and small runs.

Jan

tobydickenson:
Maybe [Greedy](https://en.wikipedia.org/wiki/Greedy_algorithm)?

> allowing the user to hand optimize the job

Could this be done using rank?

> I guess this could be solved either by a better (global) optimizer

Agreed that would be good. Do you have an algorithm in mind?


I do think there is life in the current algorithm structure. The current algorithm uses a Greedy travelCost optimisation on one nozzle, after the heuristics have selected the placement on the other nozzle. In practice, for many jobs, the NozzleTips heuristics are pretty effective at mitigating the worst outcomes due to the greedy algorithm not looking ahead. The job works systematically through the feeders, and systematically across the board.

Currently NozzleTips uses a TSM for feeder route planning and placement route planning. This goes wrong in the example above because the TSM plans a route that starts in the middle of this grid of placements. This is the opposite of what we want it to do. So I speculate that the greedy+heuristics model could be improved by swapping the feeder and placement TSMs with a cluster/outlier detection algorithm.

And I agreed a global optimization could outclass that.

tobydickenson:
3 days ago I did a force push from 24a2ebb9a505eec85540895c593087e1a2da2356 to 867dc691fa97406fe959ade17219d2dd86b15e96
The purpose of the force push was a rebase onto the rank branch. There was an extra commit to that branch to remove the "weak ordering" phrase from the tooltip. But the rebase excluded a few commits from the tip of this branch. 

3d4519d807c81f89bd6271273cec140094ed0924 looks right.

janm012012:
Hi Toby!

Here are now some thoughts about your implementation:

> Feeder Focus

I did not wonted to add a new parameter and had the impression that it might be worth selectively restricting the optimization to the same feeder. Adding a new configuration option can not be (easily) undone (at present) due to the XML serializer. Depending on the outcome of your tests, I would not mind forcing one configuration.

> DifferentFeeder

I see the point from what you wrote to the related issue but do not see the use case for the average users. Therefore this option might be causing more questions/issues. You also wrote that it's a workaround for a missing feature. If you'd like to provide it, please consider turning it into a hidden option. 

> Job Order start points

There is in deed a problem with repeated calculations of the same. In an earlier PR I introduced a concept to store the result of the planning and reuse it as long as there is no GUI interaction. Unfortunately it was not merged assuming that multiple calculations are cheap and usually generate the same result and resetting the calculation by the GUI might not work reliably. Maybe we shall reconsider this option...

Concerning the start point: on the first pass, we might use the current location. For the second an onward feeders one might evaluate if a nozzle tip change is required and if so, use the changer location as start. Probably not easy, because its part of the optimization to choose a reasonable set of nozzle tips.

Concerning TSM stability: the TSM currently uses simulated annealing with a seeded random number generator. Due to this combination, the results (for identical problems) are always the same and not guaranteed to be the best. I planning to add a brute-force solver for small problems to enhance that for the most common class of problems (fiducial check, nozzle order optimization). 

> Nozzle planning order

That's a good catch!

In https://github.com/openpnp/openpnp/commit/3ef1291e49393ace61e213791cf5158a5acb0b9a you're adjusting the expected averagePlanningCost() in both directions. Why? Doesn't this optimization provide at least the results as before?

If I understand your implementation correctly, you're "just" running the optimization back and forth, which is fine for a double nozzle machine. How about about a machine with more then two nozzles? Couldn't you add a recursive function that cycles though all combinations?

How concerning all the other question I did not yet responded to:

You mentioned, that you had many loggings set to info which you'd change to debug before merging. May I ask why (not add/keep them at debug)?

Yes "Greedy" is what I meant. Thanks for solving that puzzle and pointing me into the correct direction.

Concerning the rebase/push: I usually force push until I'm happy with the appearance. We merge what is linked to this PR so if you change your mind express it even if it requires force pushing. That's what I did in the past and what I will continue doing until someone explains why that's not a good idea.

> > allowing the user to hand optimize the job
> 
> Could this be done using rank?
> 

Yes, I think one could have very many ranks until all order requirements are satisfied. However, then we do not need the option "Unsorted" anymore... (what did I wrote about parameters that can not be removed, damn...)

> > I guess this could be solved either by a better (global) optimizer
> 
> Agreed that would be good. Do you have an algorithm in mind?

I think a TSM that uses all placements as input and calculates the cost to place them shall do. Probably disabling some optimization (like choosing the best placement for the second nozzle) but leaving others in place (like optimizing the pick, bottom vision and place order).

Jan

tobydickenson:
Hi Jan,

> > Feeder Focus
> 
> I did not wonted to add a new parameter and had the impression that it might be worth selectively restricting the optimization to the same feeder. Adding a new configuration option can not be (easily) undone (at present) due to the XML serializer. Depending on the outcome of your tests, I would not mind forcing one configuration.

My tests show that selectively restricting the optimization to the same feeder only hurts performance. If we only care about performance I would be happy to remove this option.

In the beginning, I anticipated that there may be some usability benefits from "Feeder Focus". I have the bad habit of refilling one feeder while the machine is busy picking from another. Obviously that only works if I have confidence that the machine wont suddenly pick from the feeder I am handling.

My current feeling is that the "feeder TSM" achieves the same thing. It is effective at causing the job to start at one end of the feeder bank, and work gradually and predictably through to the other end. This was not the case prior to 1b08875363bb6d558bf92ed272cb65b5f2a19ec3. So yes, I think I am fully happy to remove the Feeder Focus option.

> > DifferentFeeder
> 
> I see the point from what you wrote to the related issue but do not see the use case for the average users. Therefore this option might be causing more questions/issues. You also wrote that it's a workaround for a missing feature. If you'd like to provide it, please consider turning it into a hidden option.

ok

> > Job Order start points
> 
> There is in deed a problem with repeated calculations of the same. In an earlier PR I introduced a concept to store the result of the planning and reuse it as long as there is no GUI interaction.

I like this aspect of the current implementation because it has very robust behavior that automatically adapts to other mid-job changes such as:
 * A strip feeder pick location gradually creeps along the strip
 * I can use up the tail end of an old tape on a strip feeder, and it automatically switches to a new reel on an auto feeder when the last part has been picked from the tape. The switch is handled in a fairly optimal manner, given that I have not counted the parts and the planner has no way of anticipating the switch.

> Concerning the start point: on the first pass, we might use the current location. For the second an onward feeders one might evaluate if a nozzle tip change is required and if so, use the changer location as start. Probably not easy, because its part of the optimization to choose a reasonable set of nozzle tips.

> > Nozzle planning order
> 
> That's a good catch!
> 
> In [3ef1291](https://github.com/openpnp/openpnp/commit/3ef1291e49393ace61e213791cf5158a5acb0b9a) you're adjusting the expected averagePlanningCost() in both directions. Why? Doesn't this optimization provide at least the results as before?

Suppose we have 4 placements remaining, numbered 1 to 4 in the following locations. The numbers indicate planner job order. We have two nozzles L and R in the following arrangement. All 4 placements are the same part, so feeder pick location is irrelevant.
```
4        2
     3          1                      L         R
```

The original implementation would:
* Use the "no further optimization possible or requested, just choose the most preferred placement on the list" branch for the first nozzle L. It chooses the job order preferred placement 1.
* Use the travelCost greedy optimisation for nozzle R, where it has to choose between placements 2, 3 or 4. All of these are options are the "very silly" case because it has already decided to use the left nozzle to handle the rightmost placement. Anyway, it chooses placement 2 because that is least bad.

The new loop also runs the nozzles in the opposite order:
* Use the "no further optimization possible or requested, just choose the most preferred placement on the list" branch for nozzle R, placement 1.
* The travelCost greedy optimisation for nozzle L chooses placement 3, which has near zero travel cost. Hurrah

I hope that makes sense

The high-level outcome is: if the "placement TSM" job order means that the job is generally working left to right across the panel, you generally want the left nozzle to be given the next part on the left perimeter of the pool of outstanding placements, and allow the right nozzle to select a placement from the geometric interior of that pool, one that conveniently matches up with the nozzle separation.

> If I understand your implementation correctly, you're "just" running the optimization back and forth, which is fine for a double nozzle machine. How about about a machine with more then two nozzles? Couldn't you add a recursive function that cycles though all combinations?

Yes, it could, but that is O(N!) to try *all* combinations. This would not be desirable.

I expect it is not necessary to try all the combinations, but the right heuristic probably needs to come from someone with a better understanding of multi-nozzle machines.
 
> How concerning all the other question I did not yet responded to:
> 
> You mentioned, that you had many loggings set to info which you'd change to debug before merging. May I ask why (not add/keep them at debug)?

I find having lots of visible (info level) log messages scrolling in my terminal is a good way to gain an understanding of what this code is doing while running a test. But no one else needs that much detail in their logs while running their machine.
 
> Yes "Greedy" is what I meant. Thanks for solving that puzzle and pointing me into the correct direction.
> 
> Concerning the rebase/push: I usually force push until I'm happy with the appearance. We merge what is linked to this PR so if you change your mind express it even if it requires force pushing. That's what I did in the past and what I will continue doing until someone explains why that's not a good idea.
> 
> > > allowing the user to hand optimize the job
> > 
> > 
> > Could this be done using rank?
> 
> Yes, I think one could have very many ranks until all order requirements are satisfied. However, then we do not need the option "Unsorted" anymore... (what did I wrote about parameters that can not be removed, damn...)

All agreed

 
> > > I guess this could be solved either by a better (global) optimizer
> > 
> > 
> > Agreed that would be good. Do you have an algorithm in mind?
> 
> I think a TSM that uses all placements as input and calculates the cost to place them shall do. Probably disabling some optimization (like choosing the best placement for the second nozzle) but leaving others in place (like optimizing the pick, bottom vision and place order).

The current unit tests have proven useful. Before we consider further algorithm improvements I think we need some new unit tests containing some more sophisticated boards. It might be good to capture some anonymized copies of users real jobs to assess performance.

janm012012:
Hi Toby!

> > > Feeder Focus
> > 
> > I did not wonted to add a new parameter and had the impression that it might be worth selectively restricting the optimization to the same feeder. Adding a new configuration option can not be (easily) undone (at present) due to the XML serializer. Depending on the outcome of your tests, I would not mind forcing one configuration.
> 
> My tests show that selectively restricting the optimization to the same feeder only hurts performance. If we only care about performance I would be happy to remove this option.
> 
I just look up the discussions on PR #1708 which added this feature and did not found any good reason why it shall stay in place. It was the consents at that time, that sorting feeders is a good idea assuming that feeders are further apart then place location. This still holds true IMHO. Optimizing the placement for the second nozzle was never covered before and if you now prove that leaving the feeder open is advantageous, please remove the option.

> > > Job Order start points
> > 
> > There is in deed a problem with repeated calculations of the same. In an earlier PR I introduced a concept to store the result of the planning and reuse it as long as there is no GUI interaction.
> 
> I like this aspect of the current implementation because it has very robust behavior that automatically adapts to other mid-job changes such as:
> 
>     * A strip feeder pick location gradually creeps along the strip
> 
>     * I can use up the tail end of an old tape on a strip feeder, and it automatically switches to a new reel on an auto feeder when the last part has been picked from the tape. The switch is handled in a fairly optimal manner, given that I have not counted the parts and the planner has no way of anticipating the switch.
> 
Both valid concerns, so lets keep it as is. You said, that you fixed the implementation to stabilize the results across the job, did you?

> > Concerning the start point: on the first pass, we might use the current location. For the second an onward feeders one might evaluate if a nozzle tip change is required and if so, use the changer location as start. Probably not easy, because its part of the optimization to choose a reasonable set of nozzle tips.
> 
If you stabilized the ordering, there is already a good gain. We may leave this optimization as an option for the future. For (my) typical jobs I would say there are about 5 nozzle tips involved which means, that just this about 5 cycles would benefit. Not sure, if it is worth the effort (assuming there is no guarantee that it will get (significantly) better).

> > > Nozzle planning order
> > 
> > That's a good catch!
> > In [3ef1291](https://github.com/openpnp/openpnp/commit/3ef1291e49393ace61e213791cf5158a5acb0b9a) you're adjusting the expected averagePlanningCost() in both directions. Why? Doesn't this optimization provide at least the results as before?
> 
[...]
> The new loop also runs the nozzles in the opposite order:
> 
I think I do understand what you did and don't see any issues in the implementation itself. Its just the modifications to the unit test that comes with it: I see a lot of modifications of the `assertEquals(1.73, averagePlanningCost(), 0.01)` type. For most of them them the number is smaller now assuming that this indicates that the total cost for the job is reduced. This is what I expect from your fix. However, for two occasions (https://github.com/openpnp/openpnp/commit/3ef1291e49393ace61e213791cf5158a5acb0b9a#diff-e53e576a4b5f834279e8fb6245805021ff6454a1033733695f0e9bf6880a50c1L64 and https://github.com/openpnp/openpnp/commit/3ef1291e49393ace61e213791cf5158a5acb0b9a#diff-e53e576a4b5f834279e8fb6245805021ff6454a1033733695f0e9bf6880a50c1L81) the number increases, which would in turn look like the performance got worth. That's what I don't understand.

There is one more think, that might be not optimal: if one nozzle always gets assigned the first placement in the list, this (fix) placement might be at an undesirable location which does not provide options for a good optimization. This probably only happens if the two nozzle tips are not interchangeable.

> Yes, it could, but that is O(N!) to try _all_ combinations. This would not be desirable.
> 
> I expect it is not necessary to try all the combinations, but the right heuristic probably needs to come from someone with a better understanding of multi-nozzle machines.
> 
I would say that O(N!) is ok for small problems like 2..8 nozzles. (A brute force solver for the TSM has the same expense and I'm planning to use it for N ~8..10.) You would have to turn your solutions into a recursive function which stores it's results in a list that can be sorted then. (I did the same for the feeder/nozzle tip optimization). We may leave this for a future optimization.

> > How concerning all the other question I did not yet responded to:
> > You mentioned, that you had many loggings set to info which you'd change to debug before merging. May I ask why (not add/keep them at debug)?
> 
> I find having lots of visible (info level) log messages scrolling in my terminal is a good way to gain an understanding of what this code is doing while running a test. But no one else needs that much detail in their logs while running their machine.
> 
You may debug as you like. I'd just keep the focus on debugging user issues later on. It has been proven that logging expert/developer related information at debug level is useful as it allows to collect enough information to debug a user problem and keeps the log small for all day use (with the default logging level at info). You may (temporarily) change the logging level to debug in the unit test. (If you try, you'll likely see a lot of "subordinate motion queue drained" messages due to the fact, that your test board and panel seems to have an invalid Z of 0.)

> > > > I guess this could be solved either by a better (global) optimizer
> > > 
> > > 
> > > Agreed that would be good. Do you have an algorithm in mind?
> > 
> > 
> > I think a TSM that uses all placements as input and calculates the cost to place them shall do. Probably disabling some optimization (like choosing the best placement for the second nozzle) but leaving others in place (like optimizing the pick, bottom vision and place order).
> 
> The current unit tests have proven useful. Before we consider further algorithm improvements I think we need some new unit tests containing some more sophisticated boards. It might be good to capture some anonymized copies of users real jobs to assess performance.

That's a good idea. How could I verify that using one of my boards? (I have a simulation machine setup but no vision for fiducials, so need to run without fiducial check.)

Jan

tobydickenson:
> For most of them them the number is smaller now assuming that this indicates that the total cost for the job is reduced. This is what I expect from your fix. However, for two occasions ([3ef1291#diff-e53e576a4b5f834279e8fb6245805021ff6454a1033733695f0e9bf6880a50c1L64](https://github.com/openpnp/openpnp/commit/3ef1291e49393ace61e213791cf5158a5acb0b9a#diff-e53e576a4b5f834279e8fb6245805021ff6454a1033733695f0e9bf6880a50c1L64) and [3ef1291#diff-e53e576a4b5f834279e8fb6245805021ff6454a1033733695f0e9bf6880a50c1L81](https://github.com/openpnp/openpnp/commit/3ef1291e49393ace61e213791cf5158a5acb0b9a#diff-e53e576a4b5f834279e8fb6245805021ff6454a1033733695f0e9bf6880a50c1L81)) the number increases, which would in turn look like the performance got worth. That's what I don't understand.

hehe yes. Those two tests already had a good score. The "placement TSM" was indeed working left to right through the job in those jobs. Where the original code always selected the job-order placement on the left nozzle and the travelCost-optimised placement on the right nozzle, it was accidentally optimal for those two tests. With the smarter algorithm added, those two suffer fractionally due to the greedy algorithm lack of planning ahead.

> There is one more think, that might be not optimal: if one nozzle always gets assigned the first placement in the list, this (fix) placement might be at an undesirable location which does not provide options for a good optimization. This probably only happens if the two nozzle tips are not interchangeable.

My view on that scenario is that it is a failure of having the tips on the wrong nozzles. Or having the wrong pair of nozzles loaded in combination.

The job planner currently makes a plan for "tip order". That order indirectly determines which nozzle handles each tip, but we never explicitly optimise the tip:nozzle mapping.

For a global optimiser to find a good solution to this type of puzzle I fear it will need a super-slow annealing rate.

> You may debug as you like. I'd just keep the focus on debugging user issues later on. It has been proven that logging expert/developer related information at debug level is useful as it allows to collect enough information to debug a user problem and keeps the log small for all day use (with the default logging level at info). 

I'm not certain I understand this comment. 3d4519d807c81f89bd6271273cec140094ed0924 leaves all the useful log messages at level debug. Thats right?

> You may (temporarily) change the logging level to debug in the unit test. (If you try, you'll likely see a lot of "subordinate motion queue drained" messages due to the fact, that your test board and panel seems to have an invalid Z of 0.)

Is there a way to configure the logging system to show all the ReferenceJobProcessor debug messages, without also showing debug messages from other classes?

> That's a good idea. How could I verify that using one of my boards? (I have a simulation machine setup but no vision for fiducials, so need to run without fiducial check.)

I will add to my to-do list a tool to "anonymise" a board.xml file. Adding small random offsets to x/y, set rotations to zero, and change all the part and package names to meaningless ids.

janm012012:
Hi Toby!

> hehe yes. Those two tests already had a good score. [...] With the smarter algorithm added, those two suffer fractionally due to the greedy algorithm lack of planning ahead.

You mean, there is a combination in the original code that was almost perfect that got replaced by a better one by the new code? And by choosing this locally better pair the overall performance got worth?

> > There is one more think, that might be not optimal: if one nozzle always gets assigned the first placement in the list, this (fix) placement might be at an undesirable location which does not provide options for a good optimization. This probably only happens if the two nozzle tips are not interchangeable.
> 
> My view on that scenario is that it is a failure of having the tips on the wrong nozzles. Or having the wrong pair of nozzles loaded in combination.

I don't agree with that because you can either double up all nozzle tips which obviously would avoid this scenario or you usually  always end up with two different nozzle tips. Then you will be affected. In the first case you'd have to accept more nozzle tip changes but benefit from picking from the same feeder. In the second case I don't see anything we can do to optimize that. Anyhow, your optimization has proven to provide significantly better results compared to the existing implementation. 

> The job planner currently makes a plan for "tip order". That order indirectly determines which nozzle handles each tip, but we never explicitly optimise the tip:nozzle mapping.

Yes, agreed, All the optimization/sorting in the JobProcessor does not take into account how many nozzles a machine has and hence the JobPlanner will always distort this optimization while choosing a placement for the other nozzle(s). With all nozzle tips doubled up the distortion will be negligible but for other scenarios I don't know yet how to optimize them except using a global strategy.

> > You may debug as you like. I'd just keep the focus on debugging user issues later on. It has been proven that logging expert/developer related information at debug level is useful as it allows to collect enough information to debug a user problem and keeps the log small for all day use (with the default logging level at info).
> 
> I'm not certain I understand this comment. [3d4519d](https://github.com/openpnp/openpnp/commit/3d4519d807c81f89bd6271273cec140094ed0924) leaves all the useful log messages at level debug. Thats right?
> 
Ok, that was not very clear.... I just wonted to mention that it seems dangerous to me to change the code for debugging and before merging changing it again. I prefer to have expert logging at trace/debug and temporarily lower the logging level as required. Then the logging levels can stay as they are and I can get the same information in case a user suffers from any issue just by lowering their logging level.

> > You may (temporarily) change the logging level to debug in the unit test. (If you try, you'll likely see a lot of "subordinate motion queue drained" messages due to the fact, that your test board and panel seems to have an invalid Z of 0.)
> 
> Is there a way to configure the logging system to show all the ReferenceJobProcessor debug messages, without also showing debug messages from other classes?
> 
Not to my knowledge. Reading the logger code, it seams possible to change the logging level on a per class or package base. One would need to create a modified configuration for the logger and set it.

Jan

tobydickenson:
> > hehe yes. Those two tests already had a good score. [...] With the smarter algorithm added, those two suffer fractionally due to the greedy algorithm lack of planning ahead.
> 
> You mean, there is a combination in the original code that was almost perfect that got replaced by a better one by the new code? And by choosing this locally better pair the overall performance got worth?

Note that the "placement TSM" was seriously broken at this point, with fixes in the following commit 61c3ebf5502bffbda95829edaaf547a36f125626 so this test was only ever getting a good score by accident.

> Ok, that was not very clear.... I just wonted to mention that it seems dangerous to me to change the code for debugging and before merging changing it again. I prefer to have expert logging at trace/debug and temporarily lower the logging level as required. Then the logging levels can stay as they are and I can get the same information in case a user suffers from any issue just by lowering their logging level.

Understood, and agreed.

tobydickenson:
> you now prove that leaving the feeder open is advantageous, please remove the option

This has been done.

janm012012:
Hi Toby!

> Note that the "placement TSM" was seriously broken at this point, with fixes in the following commit [61c3ebf](https://github.com/openpnp/openpnp/commit/61c3ebf5502bffbda95829edaaf547a36f125626) so this test was only ever getting a good score by accident.

I'm some issues or problems to understand with this:

- using the current head location as start location shall provide the best result as that  is where the head currently is and will start off when requested to process the next placement. 
- In https://github.com/openpnp/openpnp/commit/61c3ebf5502bffbda95829edaaf547a36f125626#diff-8678aed78661414dfe0dc320f036c63f5475c211c30bec9d867e31206cfdd37eR1088 you're not taking a placements location and not the location of the head. As all optimization use locations relative to the head, this introduces some error. I admit, that this error is small suppose the nozzle to head offset is small compared to the distance between head and next (feeder) location.
- The change of https://github.com/openpnp/openpnp/commit/61c3ebf5502bffbda95829edaaf547a36f125626#diff-8678aed78661414dfe0dc320f036c63f5475c211c30bec9d867e31206cfdd37eR918 is likely not relevant. The optimization only takes placement locations into account. So the transformation shall only add a constant offset. However, I'm not entirely sure if the transformation is indeed required to handle board rotations. Maybe you can verify that board rotated by 90° or 180° is still optimized well. If the transformation is needed, please add a comment so that it will be be removed in the future with the argument is started with...
- The test board/panel you added have z=0 everywhere, which is unphysical and leads to warning from eg. the subordinate motion queuing subsystem. Please correct that.

Please remember:

> As before, there are lots of Logger.info that all need to be Logger.debug before this is merged, and in one case I note some old code in comments that needs to deleted.

I'm looking forward to August for this PR to get merged. Please consider adding a note to the changelog to encourage more people to stay with us on the test branch.

Jan

tobydickenson:
>     * using the current head location as start location shall provide the best result as that  is where the
> head currently is and will start off when requested to process the next placement.

The "placement job order planner" currently uses a TSM, but I think that is not the ideal algorithm for the job. The current implementation is giving us ok results at the moment so I am not in a hurry to replace the TSM for this purpose, but it is definitely a mistake to think that the placement TSM is beneficial for the same reason as, for example, the fiducial TSM.

Differences include:
* When the head moves to a placement, its starting point is always the bottom camera location, or another placement. Current head position during the planning cycle is irrelevant
* The head never moves from placement to placement along its *whole* route. A TSM operates by planning a route through all of the "cities", but that is nonsense for this application.

Lets look at where the output of the placement TSM is used:

On a single-nozzle machine there are not many optimisation opportunities. The nozzle has to move feeder -> vision -> placement and repeat that cycle for each placement. The order in which each placement is handled does not make a huge difference to throughput. So lets ignore the single-nozzle case and look at dual nozzles.

On a dual nozzle-machine, each cycle operates by filling one nozzle from the first placement in the job order, and filling the second nozzle using travelCost optimisation.

In my conceptual model of how this algorithm is now working, the purpose of the "placement job order planner" is as a heuristic to complement the travelCost optimisation stage. We need this heuristic because travelCost is greedy and does not plan ahead.

The strategy is to start work at the outside of the panel, and work inwards. The advantage of this strategy is that it maintains a pool of pending placements that are all sufficiently close to each other that travelCost optimisation will always have some good choices. We implement this strategy by having the "placement job order planner" selecting placements that are on the perimeter of the pool of outstanding placements, allowing travelCost to set up the other nozzle with a placement from the geometric interior of that pool.

We get an ok result using the TSM with the previous pick location as a starting point. At some point I will look at replacing the "placement job order planner" TSM with a dedicated cluster/outlier detection algorithm.
 
>     * In [61c3ebf#diff-8678aed78661414dfe0dc320f036c63f5475c211c30bec9d867e31206cfdd37eR1088](https://github.com/openpnp/openpnp/commit/61c3ebf5502bffbda95829edaaf547a36f125626#diff-8678aed78661414dfe0dc320f036c63f5475c211c30bec9d867e31206cfdd37eR1088)
> you're not taking a placements location and not the location of the head. As all optimization use
> locations relative to the head, this introduces some error. I admit, that this error is small suppose
> the nozzle to head offset is small compared to the distance between head and next (feeder) location.

I will review this in detail.

>     * The change of [61c3ebf#diff-8678aed78661414dfe0dc320f036c63f5475c211c30bec9d867e31206cfdd37eR918](https://github.com/openpnp/openpnp/commit/61c3ebf5502bffbda95829edaaf547a36f125626#diff-8678aed78661414dfe0dc320f036c63f5475c211c30bec9d867e31206cfdd37eR918)
> is likely not relevant. The optimization only takes placement locations into account. So the transformation shall
> only add a constant offset. However, I'm not entirely sure if the transformation is indeed required to handle
> board rotations. 

Without that transformation, the planner only sees placement coordinates as the offset from the board zero. So the TSM sees zero distance between R1 on board 1 and R1 on board 2, because both R1 placements have the same board offset coordinates.

> Maybe you can verify that board rotated by 90° or 180° is still optimized well.

The job used in all these new tests has boards at various different rotations.
 
>     * The test board/panel you added have z=0 everywhere, which is unphysical and leads to warning from eg. the subordinate motion queuing subsystem. Please correct that.

This is the same in the standard simulated board used in all the other unit tests, so I will do a separate PR.

> 
> 
> Please remember:
> 
> > As before, there are lots of Logger.info that all need to be Logger.debug before this is merged, and in one case I note some old code in comments that needs to deleted.
> 
> I'm looking forward to August for this PR to get merged. Please consider adding a note to the changelog to encourage more people to stay with us on the test branch.
> 
> Jan

janm012012:
Hi Toby!

> The strategy is to start work at the outside of the panel, and work inwards.

I understand your intention, however, this might lead to the same issue you're fighting just at the end of the job when most placements have been processed...

> Without that transformation, the planner only sees placement coordinates as the offset from the board zero. So the TSM sees zero distance between R1 on board 1 and R1 on board 2, because both R1 placements have the same board offset coordinates.

Oh, that's  good catch! So panels of identical boards are almost unsorted?

> This is the same in the standard simulated board used in all the other unit tests, so I will do a separate PR.

You're right. I noted that some time ago but don't see a problem that requires fixing it immediately. I mentioned it here to not merge more boards with the same issue. Especially if you're looking at multiple-nozzle machines, you might be affected as the subordinate motion system behaviors differently compared to physical machines.

Jan

tobydickenson:
> * The test board/panel you added have z=0 everywhere, which is unphysical and leads to warning from eg. the subordinate motion queuing subsystem

I have changed all the placements to z=-1. I agree this is better.

However I still see warnings from the subordinate motion queuing subsystem. I am not sure I fully understand what is happening, but I think this is because the test is not using an asynchronous driver so motion is not processed from queues, but the subordinate motion system needs queues.

janm012012:
Hi Toby!

> However I still see warnings from the subordinate motion queuing subsystem.

I just verified the required modifications on the `SampteJobTest`, which loads `pnp-test.job.xml`. If I change `z="0.0"` to `z="-1.0"` in the `<location` properties, the `WARNING: Subordinate Motion queue not empty: [...], drained.` are gone.

> I am not sure I fully understand what is happening, but I think this is because the test is not using an asynchronous driver so motion is not processed from queues, but the subordinate motion system needs queues.

All motion is commanded by sending a virtual `Location` together with the respective `HeadMountable` to the `MotionPlanner`. Location is limited to four dimensions. The motion planner first converts the virtual into a physical location by mapping the head mountables axis and applying offsets and transformations. The result is an `AxisLocation` that can carry any amount of (physical) axis. This information is used by the subordinate motion queuing system to queue motion requests that are flags as `subordinate[X|Y|Z|Rotation]`. For any other motion not flagged as subordinate, the queue and the set of axis to move are inspected. If the entire move from start to end and all head mountables of all moving axis/heads are in safe Z, the subordinate motion segment is merged. For a physical machine with the usual U type motion this is satisfied for the long moves between pick and bottom vision, between bottom vision and place and between place and pick location.

IIRC the problem with the `pnp-test.job.xml` is the second nozzle: if the board is location at z=0.0, placing a part requires the nozzle to go higher (z > 0) by the part height. Due to the shared Z configuration used for the test the other nozzle goes down at the same time. Now the other nozzle is not in Safe-Z anymore which violates the requirements to merge subordinate motion.

Jan

tobydickenson:
Hi Jan, thank you for the comments. That explanation is useful.

Maybe I'm missing something, but I think NullMotionPlanner is dumber than you describe. NullMotionPlanner.moveTo calls two functions, AbstractMotionPlanner.moveTo and machine.waitForCompletion(WaitForStillstand). AbstractMotionPlanner.moveTo can record a subordinate motion, but it will immediately get discarded by waitForCompletion calling drain() and causing the "Subordinate Motion queue not empty" log message. This all happens inside the one call to NullMotionPlanner.moveTo.

In the original SampleJobTest I see that changing the board Z does make a difference to whether the "Subordinate Motion queue not empty" log message is seen. I dont fully understand that yet, but I think that must be due to a difference in whether a subordinate motion is generated, rather than a difference in it being merged into other movements. Because with NullMotionPlanner it never has a chance to be merged.

janm012012:
Hi Toby!

I've just setup the debugger for SampeTestJob to break at the warning and see that the first occurrence happens when the vacuum valve gets opened (for pick). The second occurrence happens from insight the image camera (for bottom vision).

Breaking at AbstractMotionPlanner.moveTo() reveals that the ReferenceAdvanceMotionPlanner is used.

Jan

tobydickenson:
Thank you for unlocking the puzzle.  `src/test/resources/config/SampleJobTest/machine.xml` contains the ReferenceAdvancedMotionPlanner. But `src/test/resources/config/JobProcessorTest/machine.xml` does not, and therefore uses the default NullMotionPlanner when deserialised from xml.

PR #1856, opened 2025-06-11
gcodeasyncdriver: bubble up errors through machine coordination

Jorropo:
# Description

Previous to this patch, a timeout during M400 would softlock openpnp.
Fixes #1848

# Justification

This bug stole hours of my life I will never get back just because I didn't knew I had wrong configuration that made M400 timeout.

# Instructions for Use

None, this is a bug fix.

# Implementation Details
1. How did you test the change?
  On my lumen right after tool changing it goes for bottom vision or Z focus, and the combined tool change plus move always timeout the M400 sent by the bottom camera or vacuum actuator, previous to this patch openpnp softlocks, now it raises an error along the line of « M400 timeout » and continues to be usable.
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)?
  yes.
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.
  I did not.
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted.
  This is running.

PR #1858, opened 2025-06-11
gcodeasyncdriver: change condition variable for confirmationComplete to the formally correct format

Jorropo:
# Description

The previously implemented condition variable is very confusing for no benefit.
The formally correct way to implement this is to lock the loop preheader and header while unlocking in the body.
Change to the formally correct so I don't have to guess if the volatile usage is correct or not.

This also fixes a data race on errorResponse (and arguably confirmationComplete but it's volatile and not everyone would agree).

# Justification

The previous patch was very hard to read for me because it does not follow well known systematically proven concurrency patterns.
It asks the reader to pounder the meaning of volatile in this context and what kind of evil were the previous programmer up to.

Avoiding dataraces and volatile by implementing a formal condition variable has no measurable cost given both critical sections are free of IO, have`< O(N)` behavior, it is impossible to wait on the mutex for very long (unless your OS is starving so bad it doesn't schedule your threads, but that also a problem with the previous code) while making the code easier to read and free of once in a blue moon impossible to debug data races.

# Instructions for Use
None, this is not a feature.

# Implementation Details
  `mvn test`, will check on the machine tomorrow, I have tested an extremely similar patch on my machine and nothings happens.
  `mvn test` passes

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/GcodeAsyncDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/GcodeAsyncDriver.java
@@ -134,7 +134,8 @@ public class GcodeAsyncDriver extends GcodeDriver {
     protected LinkedBlockingQueue<CommandLine> commandQueue;
 
     private boolean waitedForCommands;
-    private volatile boolean confirmationComplete;
+
+    private boolean confirmationComplete;
 
     public boolean isConfirmationFlowControl() {
         return confirmationFlowControl;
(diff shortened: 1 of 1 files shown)


Jorropo:
This PR stacked over #1856.

Do not merge, there is a potential deadlock because `.offer` blocks if the queue is full, but then we would block while holding the lock that the run could try acquiring.
I am not sure the conditions are reachable since we would need multiple machine coordinations in flight, but we can fix this while simplifying the code by using the boolean flag to signal we want to be awken rather than sending an empty item on the queue so I'll do that.

PR #1868, opened 2025-07-02
WaitForStillstand before Thread.sleep if controler side delay is unsuported

### LINKED ISSUE #1867: if using `GcodeAsyncDriver` without `DELAY_COMMAND` delays are masked by long queued commands
New in 2.3 if `DELAY_COMMAND` is not configured and you are using GcodeAsyncDriver it will fallback to java `Thread.sleep` dwell.

However let's assume there already are some commands sent to the controller, let's continue to say it have 1000ms of travel moves in queue.

Then openpnp tries to dwell, so it call `Thread.sleep(200);` however it does not wait for coordination, so the 200ms sleep runs in parallel of the `1000ms` of travel moves; result you don't dwell (200 < 1000).

---

This can be worked around by configuring `DELAY_COMMAND`, with marlin configure `DELAY_COMMAND` to `G4 P{TimeMS:%d}`.

Jorropo:
# Description
With GcodeAsyncDriver we may have let's say 1000ms of moves already in queue; if we Thread.sleep without waiting for the machine to be stillstanding the Thread.sleep is masked by the machine moving as they run in parallel.

Fixes #1867

# Justification
Me and some other lumenpnp user had issues where dwell times not being respected after upgrading to 2.3; this caused the machine to become extremely unreliable.
This is clearly a bug because the code already tries to fallback to `Thread.sleep` over let's say throwing an exception, but it does not do it properly.

# Instructions for Use
None, it just works, when it used to require you to change your config to GcodeDriver or configure `DELAY_COMMAND` after upgrading to 2.3.

# Implementation Details
  On my lumen if I have `DELAY_COMMAND` empty all the vacuum related tests would fail because dwell times were not respected and the pumps take a significant amount of time to spool up.
  With this patch applied even tho I have removed `DELAY_COMMAND` from my config, it works reliably and I can observe it dwell.
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. ***passed***

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/AbstractMotionPlanner.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/AbstractMotionPlanner.java
@@ -292,12 +292,18 @@ public abstract class AbstractMotionPlanner extends AbstractModelObject implemen
             
             // now execute the delay on all effected drivers
             for (Driver driver : drivers) {
-                delayNotExecuted |= driver.delay(milliseconds);
+                if (!driver.delay(milliseconds)) {
+                    driver.waitForCompletion(null, CompletionType.WaitForStillstand); // see issue #1867
+                    delayNotExecuted = true;
+                }
             }
         }
         else {
             for (Driver driver : machine.getDrivers()) {
-                delayNotExecuted |= driver.delay(milliseconds);
+                if (!driver.delay(milliseconds)) {
+                    driver.waitForCompletion(null, CompletionType.WaitForStillstand); // see issue #1867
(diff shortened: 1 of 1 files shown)


Jorropo:
cc (*blame* 0deb6dd58b6d5f5e6f3e09734a8b33edc60124ae) @janm012012 could you please take a look ?

janm012012:
Hi @Jorropo!

Please let's first discuss if this is an issue before changing the code.

Jan

tobydickenson:
this is superseded by #1869

PR #1869, opened 2025-07-07
Delay fix

tobydickenson:
# Description

I believe this is a fix for #1867. A different fix for this issue was proposed in #1868.

I think the bug is very simple and silly. @janm012012 review please.

GcodeDriver returns **FALSE** if we need a java sleep

https://github.com/openpnp/openpnp/blob/c56e1dd5661a5028a1f3011172dcb14c7278fc38/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java#L686-L690

but the motion planner expects drivers to return **TRUE** if we need a java sleep

https://github.com/openpnp/openpnp/blob/c56e1dd5661a5028a1f3011172dcb14c7278fc38/src/main/java/org/openpnp/machine/reference/driver/AbstractMotionPlanner.java#L293-L296

# Testing & Risks

I have reproduced #1867 by removing the delay gcode from my configuration and seen the delay getting skipped during vacuum pick. I have confirmed that this PR restores the expected behaviour.

I have confirmed with a brief test that my machine still works with this change and the delay gcode restored. This definitely needs wider testing. If we agree this is the right fix then we should notify beta testers to update.

But I dont really understand why the #1868 fix worked. What am I missing?

# Instructions for Use

no changes 

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. - described above
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. - **YES** to fix a comment

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/AbstractMotionPlanner.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/AbstractMotionPlanner.java
@@ -289,22 +289,29 @@ public abstract class AbstractMotionPlanner extends AbstractModelObject implemen
                     }
                 }
             }
-            
-            // now execute the delay on all effected drivers
-            for (Driver driver : drivers) {
-                delayNotExecuted |= driver.delay(milliseconds);
-            }
+            delayAllDrivers(drivers,milliseconds);
         }
         else {
-            for (Driver driver : machine.getDrivers()) {
-                delayNotExecuted |= driver.delay(milliseconds);
-            }
+            delayAllDrivers(machine.getDrivers(),milliseconds);
         }
-        
-        // if any driver was not able to execute the delay, fabllback using Thread.sleep()
-        if (delayNotExecuted) {
+    }
+
+    // Execute a delay on all drivers in the list.
+    // If any driver can not execute the delay then we fallback to using Thread.sleep()
+    private void delayAllDrivers(final List<Driver> drivers, int milliseconds) throws Exception
+    {
+        boolean delayExecutedInAllDrivers = true;   // set to false if any driver requested to delay does not support delaying
+
+        for (Driver driver : drivers) {
+            // delay() returns false if the delay was not executed
+            delayExecutedInAllDrivers &= driver.delay(milliseconds);
+        }
+
+        if(!delayExecutedInAllDrivers) {
             // time delay using OS
+            Logger.trace("delay sleep before");
             Thread.sleep(milliseconds);
+            Logger.trace("delay sleep after");
         }
     }
 
--- a/src/main/java/org/openpnp/machine/reference/driver/AbstractMotionPlanner.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/AbstractMotionPlanner.java
@@ -260,7 +260,6 @@ public abstract class AbstractMotionPlanner extends AbstractModelObject implemen
 
     @Override
     public void delay(int milliseconds, HeadMountable... hms) throws Exception {
-        boolean delayNotExecuted = false;   // set to true if any driver requested to delay does not support delaying
         // Plan and execute any queued motion commands. 
         executeMotionPlan(CompletionType.CommandStillstand);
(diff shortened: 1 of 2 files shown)


tobydickenson:
> But I dont really understand why the #1868 fix worked. What am I missing?

Aha. [That PR](https://github.com/openpnp/openpnp/pull/1868/files) also inverted the boolean return value from `delay()`. The `waitForCompletion` call in that PR was a decoy.

janm012012:
Hi Toby!

Many thanks for clearing the fog! There is indeed a bug in the `delayNotExecuted` handling which would also explain the observations in issue https://github.com/openpnp/openpnp/issues/1867. However, I do not like your fix because it introduces more issues: you're changing the commend for the default delay() implementation (in src/main/java/org/openpnp/spi/Driver.java) but not it's return value. Now the default signals that it did execute the delay, which is wrong. Why not fix the problem where it is at `delayNotExecuted` like

```
                            drivers.add(driver);
                        }
                    }
                }
            }
            
            // now execute the delay on all effected drivers
            delayNotExecuted = delayAllDrivers(drivers, milliseconds);
        }
        else {
            // execute the delay on all drivers of the machine
            delayNotExecuted = delayAllDrivers(machine.getDrivers(), milliseconds);
        }
        
        // if any driver was not able to execute the delay, fabllback using Thread.sleep()
        if (delayNotExecuted) {
            // time delay using OS
            Thread.sleep(milliseconds);
        }
    }

    // execute a delay on all drivers in the list. return true if any driver was not able to do so
    private boolean delayAllDrivers(final List<Driver> drivers, int milliseconds) throws Exception
    {
        boolean delayNotExecuted = false;   // set to true if any driver requested to delay does not support delaying
        
        for (Driver driver : machine.getDrivers()) {
            // delay() return false if the delay was not executed
            delayNotExecuted |= !driver.delay(milliseconds);
        }
        
        return delayNotExecuted;
    }
```

This shall fix the problem, add the relevant documentation that delay() returns true on success and collects false results. In addition it groups the two for loops into a more logical function.

What  do you think?

Jan

tobydickenson:
Yes, Mostly all agreed. 

It's better to keep the two boolean variables having the same sense, and avoid the double negation. So I prefer
delayExecutedInAllDrivers &=

Code to follow in a few hours

tobydickenson:
Hi Jan,

> What do you think?
> 
> Jan

All these are actioned. Its not quite the same as you wrote, but I think you will like it. Much tidier than my original.

I am not near the machine to test this right now, but will do so tomorrow.

janm012012:
Hi Toby!

That looks even better. Two things:

1. please remove `delayNotExecuted`, which is unused now.
2. is is really needed to log twice per Thread.sleep()? I'd prefer to see one of three lines in the final log: "delay delegated to driver", "delay skipped for zero duration" and "delay via Thread.speed()". Then the log would be conclusive indicating if, how and where the delay(s) is/where executed.

Jan

tobydickenson:
>     2. is is really needed to log twice per Thread.sleep()?

That is a habit from multithreaded code, where there might be lots of log lines from other threads in between those two. But agreed it is not needed here.

I will do a final test with real hardware today then merge.

janm012012:
> I will do a final test with real hardware today then merge.

Agreed, good luck!

tobydickenson:
This has been merged, but the deployment action has failed for other reasons.

vonnieda:
> This has been merged, but the deployment action has failed for other reasons.

Looking into it.

vonnieda:
Should be resolved now!

tobydickenson:
Thanks. Is there anything I could have done to fix that?

vonnieda:
The Apple notarization of the macOS installer was failing because they had updated the license agreement. The owner of the developer account (me, in this case) has to sign the agreement by logging in and agreeing once in a while. I did that.

So, no, not really much you could do in this case except maybe edit the installer script to disable the macOS build temporarily.

Jason

PR #1893, opened 2025-09-02
photonfeeder: while doing moveWhileFeeding move to safeZ rather than 0

Jorropo:
# Description
While doing moveWhileFeeding move to safeZ rather than 0.

# Justification
This fixes an interesting case where my nozzles go above safeZ while doing a moveWhileFeeding action.

# Instructions for Use
None just profit (it makes an other option `moveWhileFeeding` slightly faster).

# Implementation Details
1. How did you test the change? **TODO**
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. **TODO**


Jorropo:
It doesn't change the behavior, something looks wrong.

Issue #1891, opened 2025-08-29
Design Review request: Feeder Management


tobydickenson:
Hi all,

This is a request for a design review. I am considering implementing some features that relate to how feeders are selected for a part, and placements retried during a job.

Firstly, this will address some problem behaviors relating to job errors. If the job error handling policy is set to "alert" then the user sees an error message box right away. These problem behaviors occur if the error handling policy is set to "defer". Problem behaviors are:
1. Suppose there are 100 placements of a part. Size checking is enabled, but the configured size is slightly wrong. All 100 placements will be attempted, each one resulting in a vision size check error, and 100 parts are systematically dumped into the discard pile.
2. Suppose there are 100 placements of a part. 99 of them are placed correctly, but one suffered a sideways pick error. This is detected at vision and the part discarded. The placement left as a "pending error" to be shown at the end of the job, without any retries.Those problem scenarios are related. The common factor is that the job processor currently has exactly one attempt at each placement. It would be better to:
1. Disable a feeder after several consecutive errors.
2. Have some retries of each placement within the job processor, provided the feeder is not disabled.

I also want to support the following new use cases:
1. Suppose we have few parts left on the tail of a previous tape; this is too short for an auto feeder so it will be loaded on a strip feeder. We want to use those up first, then automatically switch to using the new reel on an auto feeder.
2. We have two auto feeders set up with the same part. If one feeder is working well but the other has a problem, it should automatically use only the good feeder to continue making progress in the job.
3. We have two auto feeders set up with the same part. If both feeders are working well, the job planner optimiser should automatically consider both options and select the nearest

Implementation plan:
1. Some new configuration state is needed; by default to disable a feeder after 3 consecutive failures, or 2 failures out of 5. Store this in the machine object? or a new persistent object inside the machine?
2. Nozzle currently has an attribute which tracks which Part is loaded on the nozzle. It gains a new attribute to track which feeder the part came from.
3. Feeders receive a notification (TBC from Nozzle? or JobProcessor?) when a part has been successfully placed, or when a part is discarded due to a problem. It tracks those statistics and disables itself when thresholds are exceeded.
4. Feeder gains a new "Priority" attribute; low / normal / high.
5. There are several implementations of a findFeeder method; these need to be rationalised.
6. findFeeder() operates by finding all the enabled feeders, filtering by priority, then allowing the job planner optimiser to select if there is more than one option.
[...]

PR #1898, opened 2025-09-09
Feeder fleet

tobydickenson:
This is a preliminary Pull Request. This implements the plan described on #1891 but this is not yet tested in practice on a machine.

# Description

A breakdown of each commit:

* `Nozzle` already tracks which `Part` it is carrying, and there is a corresponding new api for tracking which `Feeder` provided the part. This had more impact than I was expecting because there are many calls to `Nozzle.pick` which all now need to provide a Feeder parameter. (This new API has standalone usefulness, so I will do a separate PR #1899 for just this commit).
* Openpnp already had several layers of retries. This adds tooltips for all the original retry count gui fields.
* In Defer error reporting mode, each `JobPlacement` can be retried in full. The job processor exception handler that previously called `JobPlacement.setError` can now call `JobPlacement.setStatus(Pending)` instead. `JobPlacement` counts the number of times it has has been `Processing`, and the job processor checks that number against the new per-placement retry limit. By default this is set to 5 retries.
* An old bug is fixed. The jobprocessor disables a feeder if it sees `Feeder.feed` throwing exceptions. But there was no notification of this change to refresh `FeederPanel`. I dont think we can do this using the normal binding mechanism because the notification happens in the wrong thread, so I added the `ReferenceFeeder.updateView` method. Is there a better way to do this?
* `ReferenceFeeder` now keeps track of whether the job processor has a successful outcome with its parts. It stores this in a `LinkedList<Boolean>` which acts as a short ring buffer of failure flags. The job processor notifies `ReferenceFeeder` at the same time as driving `JobPlacement` state transitions. `ReferenceFeeder` disables itself if limits are exceeded, and the fault record is cleared when the feeder is re-enabled by the operator. The default configuration (also stored in the job processor) is to disable if there are 3 faults in 6 placements. `FeederPanel` has a new column showing the fault status. This fault status is blank if there are no faults, otherwise it shows an "X" for each fault and "-" for each success. For example, "X-----" indicates it has just had its first fault. This gui display is functional, but I'm not sure I like it yet.
* `FeederUtils` holds the new implementation of `findFeeders()` which returns `List<Feeder>` and `findFeeder` which makes a selection from the list. The selection uses `getLinearDistance` rather than `TravelCost` for compatibility with some old unit tests which set up a machine without the kinematic configuration required by `TravelCost`. The usage of this API in the job processor has been made functional and safe: For job planning options which use a "feeder tsm" it selects the feeder nearest the head of the previous tsm run. For job planning options without, it selects the feeder nearest the head. It will always select the same feeder for prerotation calculations as picking. I expect there are a further optimisation opportunities here: Maybe a greedy search through the list of feeders; maybe changing the "feeder tsm" to include both. This optimisation is an outstanding task.
* There is a fix for an old unit test, which broke because of a typo in its machine.xml configuration of the location of one feeder. Previously that feeder was unused, but it is now because findFeeder selects the nearest.
* `Feeder` gains a priority enum, with values `High` (use these parts up first) `Normal` and `Low` (backup parts). This is the second new column on `FeederPanel`. `findFeeder` checks this to filter for the highest priority class available.
* There are two new units tests. These operate by simulating **intermittent job processing errors, and confirming that the job finishes with 100% placements complete anyway.** This validates the overall goal of this development.

# What has NOT changed
* There are no changes to the multitude of different feeder implementations, and I expect that this will work with them all.

# Justification
Rationale discussed on #1891

# Known issues
* `ReferenceFeeder` needs to update `FeederPanel` from the machine thread, when its status is changed by the job processor. Is there a better way of doing this?
* Feeder faults are displayed using the X legend. Is there a nicer way to display this?

# Outstanding Tasks
* Testing on a real production job
* Wiki documentation:
     * Job processor page
     * A new page describing the overall retry structure.
     * New script event for a job fault.
* A review of out-by-one issues. Does "3 retries" mean 3 attempts, or 4 attempts? Check this in code and documentation.
* Job planner could optimise the choice of feeder, if there is more than one available.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. -- there are new unit tests. testing in a production context is an outstanding task.
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. -- yes!!
    * `Feeder` and `AbstractFeeder` have the new `Priority` enum and accessor methods.
    * `Nozzle` and `AbstractNozzle` have changes that allow it to keep track of the feeder which provided the part it is carrying.
    * `PnpJobProcessor` has two changes: It count changes into `Processing` status, and `JobPlacement` can keep track of the feeder (so that we pick from the same feeder that was used for prerotation calculations).
    * `AbstractPnpJobProcessor` implemenation of `findFeeder` now delegates to the new `FeederUtils`


tobydickenson:
This recent force-push was a rebase after merging #1899

tobydickenson:
Edit: moved to #1901

tobydickenson:
> ## Known issues
> ReferenceFeeder needs to update FeederPanel from the machine thread, when its status is changed by the job processor. Is there a better way of doing this?

On review, the [new updateView method in ReferenceFeeder](https://github.com/openpnp/openpnp/pull/1898/files#diff-eb3641f0b57f44adb90fa5e1e74921b1689009978a499fc2f7066695a209f55f) is not so bad.

> Feeder faults are displayed using the X legend. Is there a nicer way to display this?

Some colleagues have expressed that they like this display, so I will proceed with this as-is in this PR. But I am open to implementing a better display if there are suggestions.

> ## Outstanding tasks
> * Testing on a real production job

This was successfully tested today

> Wiki documentation:  A new page describing the overall retry structure.

[Done](https://github.com/openpnp/openpnp/wiki/Retries).   

> Wiki documentation:  Job processor page,  New script event for a job fault.

This is still outstanding

> A review of out-by-one issues. Does "3 retries" mean 3 attempts, or 4 attempts? Check this in code and documentation.

Still outstanding

tobydickenson:
> > A review of out-by-one issues. Does "3 retries" mean 3 attempts, or 4 attempts? Check this in code and documentation.

There were a few problem here, some old and some recent from me. These are all fixed in 2407f7ffb9.

1. PR 958 introduced the "Max Vision Retries" field with an out-by-one problem. It controls the number of **attempts**, not the number of **retries**. The default 3 attempts is one try then two subsequent retries. There are backwards compatibility reasons for not changing the attribute name or behavior of this field, but I have corrected the gui label. The tooltip was already correct.

2. I previously noted that there were two different fields called "Pick Retry Count". This changes one to "Feed & Pick Retry Count". Again this is just the gui label and tooltip.

3. It was possible for the loop performing vision retries to run zero times. I have fixed this to run a minimum of once.

tobydickenson:
> Wiki documentation: Job processor page, New script event for a job fault

wiki documentation is now all complete.

This PR is all now complete. I plan to merge this next week after some final testing.

PR #1903, opened 2025-09-27
photonfeeder: don't move the head while feeding if the machine is unhomed

Jorropo:
# Description
The feeder will not move the nozzle according to move while feeding if the machine is unhomed.

# Justification
I had to grab a part from a feeder for manual rework.
I've turned on openpnp, scan for feeders, press the feed (but not feed and pick) on the right feeder.
Then the head slammed into the rails.

If we don't know where the head is don't try to move it.
It's not part of feed's contract to move the nozzle, we do it for performance optimization since it's allowed, but anyone consuming this has better code to handle this situation.

# Instructions for Use
None

# Implementation Details
1. How did you test the change? I didn't


tobydickenson:
Looks good. Ready to merge?

Jorropo:
LGTM

tobydickenson:
I have merged this, but note that this may just be one example from a family of similar bugs. ReferenceAutoFeeder has code very similar to that fixed here. Other feeders that use vision will be moving the head too - do they need a similar fix or does the vision subsystem check this already? If this is a common flaw in lots of feeders then maybe it would be better fixed elsewhere. I will review.

Jorropo:
I have no idea, as you point out there are feeders that cannot work if the machine is unhomed.
Maybe the feeder UI should pass a null nozzle if the machine is unhomed ?

tobydickenson:
Do you have "allow unhomed motion" enabled in your driver configuration?

It is that option which prevents ReferenceAutoFeeder moving in the same situation.

Jorropo:
Yes, I use it to jog the Z back to a safe level before homing.

PR #1905, opened 2025-09-30
fix the longer-than-typical dwell time on the last placement of a job

tobydickenson:
# Description
This change adds waitForCompletion after the final machine activities in a job, and before doing all the other processes that do not involve moving the machine.

# Justification
I sometimes see what appears to be an extra-long dwell time on the final placement of the job, or a delay before homing the machine at the end of a job. The async driver has received those movements but there is a delay somewhere. It makes sense to ensure make sure that all machine processes are complete at this point.

# Instructions for Use
no changes

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. -- tested on a real machine


tobydickenson:
(I thought I had created a PR for this already, but I dont see it now. Maybe I forgot to press the final "create PR" button)

PR #1913, opened 2025-11-01
photonfeeder: add Feed 1mm button

Jorropo:
# Description
For photon feeders, add `Feed 1mm` button right next to the `Feed` button.

# Justification
While using the photon feeder I found myself doing this on a regular basis to fine tune the tape position in the pick window:
- change part pitch to 1mm
- apply
- feed
- change part pitch to what it was
- apply

Because the physical buttons on the photon feeder are 2mm feed and just inconveniently located for the half on the other side of the machine.

# Instructions for Use
Go to a photon feeder.
Go to global config, search all slots.
Go back to a photon feeder.
Press `Feed 1mm` button.

# Implementation Details
1. How did you test the change? I tested it on my machine, pressed the button it feeds 1mm.
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? y
3. ~~If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing.~~
4. Be sure to run `mvn test` before submitting the Pull Request. I love CI :heart:

Issue #1896, opened 2025-09-08
use backtransformed bottom vision results for a closed loop feeder pick position offset control


Jorropo:
Notes from a discussion I had with @tobydickenson in the lumenpnp discord.

The photon feeders are not very reliable, they drift overtime. This cause issues where I manually need to use top vision to recenter the pick location onto the actual tape holes.

Let's say you have a 4mm pitch tape, on average the feeder might feed 3.97mm, overtime error add up until the pick window shifted so far the nozzle can't pick reliably or at all.
I was quick to think this a hardware or feeder firmware issue however @tobydickenson pointed out I am expecting all tapes to be precisely 4mm pitch (or whatever) which might not be true.

The photon feeder uses DC brushed actuators, it implements closed loop control using a magnetic encoder (~8µm tape movement resolution) on the drive shaft.
That means it is blind to tape related errors.
Afaik it does not compensate for quantization error so assuming a quantum / 2 average error (~4µm) after ~250 feeds the feeder would be 1mm shifted.

All of this would be solved if the tape position was part of the control loop, however getting an accurate reading of the tape position in the feeder looks complicated to say the least.

### Thus here is an idea where we can use the bottom vision step to control the pick location and feed distance

Let's say the pick location is offset from the part.
That means we will pick parts off-center.
Which will be detected by bottom vision. This will be used to place the part correctly.
But we can run this offset through the inverse of the pick → bottom vision transform and apply this as a pick offset.

Assuming the tape relative position is identical (which is an acceptable assumption within a short amount of parts).

That means we will pick the next part on center. Closing the loop.

I see two issues that will to be solved:
1. the data is noisy, tiny parts tumble when picked, ... I am not sure how to solve this without trying it, I guess the average of the last 10 picked parts and excluding parts with the most extreme offsets might be good enough.
2. for a feeder constantly over feeding or under feeding the pick offset might grow so large as to try to pick outside of the feeder's pick window
   this can only happen in the Y plane relative to the feeder (/ feeders are mounted on the front or back) which can be compensated by tweaking how far we feed the tape, for example the photon feeder protocol accepts 100µm resolution which is orders of magnitude bigger than the pick window size.

tobydickenson:
This is a good idea. I am not aware of any industrial machines that do this, but have seen plenty that would have benefited from it. FWIW the idea probably would have been patentable.

[...]

PR #1914, opened 2025-11-03
photonfeeder: Press Play Walk Away

Jorropo:
# TODO for Ready-For-Merge:
- [ ] ~~split into 3 commits (spi, pick location nudging, tape nudging)~~
- [ ] ~~remove accidental formater diff lines~~ reordering imports is not the best but it's fine
- [ ] ~~extract feature to be usable by other feeders~~ gonna be done in 2.7
- [ ] add some configuration knobs ?
- [ ] add tests
- [x] merge #1913 (stacked PR)
- [x] rename `pickCorrectionOffset` to `pickCorrectionIntegral` to reflect we are implementing an `I`-controler
- [x] **known bugs**:
  - [x] setting the slot offset does not the reset the correction offset
  - [x] having multiple nozzles pick from the same feeder is metastable
  - [x] using test alignment does a null deref in the vision utils code trying to undo rotation 

# Description

Implements #1896 for photon feeder (generic implementation will be implemented in 2.7)

Implements Backtransformed bottom vision → pick location closed loop control.
After doing a bottom vision the VisionUtils code will callback into the feeder it were picked from.

PhotonFeeder uses this to implement two things:
1. Nudge the pick location to aim for perfectly picked components on the first try.
2. Nudge the feed pitch to remove long term drift in pick location nudging.

# Justification
I have ran a couple of before and after jobs.
Feeders are literally >20x more reliable.
On average an operator used to have to manually fix the the pick location alignment every 30 pick.
With this PR I can run a 650 placement job without any manual intervention.

# Instructions for Use
None just enjoy >20x more reliable feeders.

# Implementation Details
  I've ran multiple jobs with it, also created tests jobs making sure the coordinate transformations are bullet proof.
  probably not, maybe
  I did, I am not 100% sure on the final signature form (mainly thinking if we should pass `x` & `y` components individually since `z` & `rotation` are always 0 by definition of the coordinate frame used)
4. Be sure to run `mvn test` before submitting the Pull Request. **TODO** meaningless to do before I add my own tests.

### DIFF
--- a/src/main/java/org/openpnp/machine/photon/PhotonFeeder.java
+++ b/src/main/java/org/openpnp/machine/photon/PhotonFeeder.java
@@ -61,6 +61,7 @@ public class PhotonFeeder extends ReferenceFeeder {
 
     final private double correctionLimit = 5.0; // millimeters
     private Location pickCorrectionOffset = new Location(LengthUnit.Millimeters);
+    private int visionsSinceLastFeed = 0;
 
     private static PhotonBusInterface photonBus;
 
@@ -351,6 +352,8 @@ public class PhotonFeeder extends ReferenceFeeder {
         case Disable:
             return;
         }
+
+        visionsSinceLastFeed = 0;
 
         // To solve long term drift. Nudge the part pitch sent to the feeder if the correction offset is big enough.
         Location nudgeOffset = new Location(LengthUnit.Millimeters);
@@ -665,9 +668,14 @@ public class PhotonFeeder extends ReferenceFeeder {
             Logger.error("{}: Could not get pick location without error for bottom vision callback: {}", getSlotAddress(), e);
             return;
         }
+
         // Add half of the error to the running total to avoid over-correction
         Location before = pickCorrectionOffset;
-        pickCorrectionOffset = pickCorrectionOffset.add(offset.rotateXy(pickLocation.getRotation()).multiply(0.5));
+
+        double factor = 1 / (double)(2 << visionsSinceLastFeed); // Start at 0.5 then have each new vision half.
+        visionsSinceLastFeed++;
+
+        pickCorrectionOffset = pickCorrectionOffset.add(offset.rotateXy(pickLocation.getRotation()).multiply(factor));
         Logger.debug("{}: bottom vision reports pick error of: {}; old pick correction: {} new {}", getSlotAddress(), offset, before, pickCorrectionOffset);
 
         double distance = pickCorrectionOffset.convertToUnits(LengthUnit.Millimeters).getLinearDistanceTo(0, 0);
--- a/src/main/java/org/openpnp/machine/photon/PhotonFeeder.java
+++ b/src/main/java/org/openpnp/machine/photon/PhotonFeeder.java
@@ -62,9 +62,9 @@ public class PhotonFeeder extends ReferenceFeeder {
     final private double correctionLimit = 5.0; // millimeters
     final private double correctionGain = 0.5; // portion of error to correct each bottom vision
 
-    private Location pickCorrectionOffset = new Location(LengthUnit.Millimeters);
+    private Location pickCorrectionIntegral = new Location(LengthUnit.Millimeters);
 
-    private Location pickCorrectionOffsetAccumulatorSinceLastFeed = new Location(LengthUnit.Millimeters);
+    private Location pickCorrectionAveragingAccumulatorSinceLastFeed = new Location(LengthUnit.Millimeters);
     private int visionsSinceLastFeed = 0;
 
     private static PhotonBusInterface photonBus;
@@ -103,7 +103,7 @@ public class PhotonFeeder extends ReferenceFeeder {
     }
 
     private Location getPickCorrectionOffset() throws Exception {
-        return pickCorrectionOffset;
+        return pickCorrectionIntegral;
     }
 
     private Location getPickLocationWithoutError() throws Exception {
@@ -364,7 +364,7 @@ public class PhotonFeeder extends ReferenceFeeder {
         int partPitchNudgeTicks = 0;
         final double feedTickMm = 0.1;
 
-        double yPlaneErrorMm = pickCorrectionOffset.getLengthY().convertToUnits(LengthUnit.Millimeters).getValue();
+        double yPlaneErrorMm = pickCorrectionIntegral.getLengthY().convertToUnits(LengthUnit.Millimeters).getValue();
 
         if (yPlaneErrorMm <= -feedTickMm || feedTickMm <= yPlaneErrorMm) {
             // far enough to nudge the part pitch
@@ -377,13 +377,13 @@ public class PhotonFeeder extends ReferenceFeeder {
             Logger.debug("{}: Nudging tape by {} ticks", getSlotAddress(), partPitchNudgeTicks);
         }
 
-        pickCorrectionOffset = pickCorrectionOffset.add(nudgeOffset);
+        pickCorrectionIntegral = pickCorrectionIntegral.add(nudgeOffset);
 
         try {
             feed(nozzle, getPartPitch() * 10 + partPitchNudgeTicks);
         } catch (Exception e) {
             // Didn't feed, revert correction offset.
-            pickCorrectionOffset = pickCorrectionOffset.subtract(nudgeOffset);
+            pickCorrectionIntegral = pickCorrectionIntegral.subtract(nudgeOffset);
             throw e;
         }
     }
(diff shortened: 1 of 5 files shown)


tobydickenson:
Reviewing this is still on my to-do list. I just wanted to note that I have not forgotten this PR

tobydickenson:
Hi, sorry again for the delayed review. Life suddenly got busy.

This looks really good. I am excited to implement support for this in `PushPullFeeder` to replace of my script-based solution.


* The maths in `VisionUtils` is considerably more simple than what I ended up with. nice work. I havent reviewed the correctness of the maths, but I'm sure you would have noticed if it was not right. 

* The `bottomVisionResultCallback` signature looks fine to me (although not necessarily its name; see below). Passing a `Location` is definitely the right thing to do. I also note that passing the correction in the machine frame of reference is the best that can be done; it would be nice to use the feeder's frame of reference, but sadly there is no feeder API to determine this.

* Everything would be so much simpler if parts were fed/aligned/placed in a strict rotation, but of course this is not so simple. Your code involving `visionsSinceLastFeed` handles the case of "pick nozzle a, pick b, align a, align b" but I'm not sure it works for "Pick a, align a, align a, align a" or "Pick a, pick/align/discard b, pick/align/discard b, pick/align/discard b, align a". fwiw my script-based solution implemented logic for only performing a correction if the part was picked after the preceeding correction was applied, but that throws away some information that your method makes good use of. I'm not sure what api changes to suggest here. Let me think about this some more.

* There is an openpnp convention that homing the machine resets all transient state - not just the stepper controller home. Please add ConfigurationLister code [like this](https://github.com/openpnp/openpnp/blob/b1cb8c37c29e5ddaa7b6192e0b859d740f20b401/src/main/java/org/openpnp/machine/reference/feeder/ReferenceStripFeeder.java#L178) to zero the photon correction offset.

* I think there is some documentation needed. I suggest a new wiki page explaining the underlying principal of operation. I think this is important because this is a new invention; even users familiar with other pick and place machines will not be familiar with this concept. Please let me know if you want me to draft something.

* The feature needs a name. Ideally something that works as both a title for that wiki page, and the `bottomVisionResultCallback` method name. I tentatively propose "Bottom Vision Feeder Correction", but up to you.

* Do you think the photon feeder benefits from a gui option to enable/disable this feature? This can be on by default. This will be useful for whoever fixes the photon firmware unreliable feed distance.

Jorropo:
@tobydickenson 

> Everything would be so much simpler if parts were fed/aligned/placed in a strict rotation, but of course this is not so simple. Your code involving visionsSinceLastFeed handles the case of "pick nozzle a, pick b, align a, align b" but I'm not sure it works for "Pick a, align a, align a, align a" or "Pick a, pick/align/discard b, pick/align/discard b, pick/align/discard b, align a". fwiw my script-based solution implemented logic for only performing a correction if the part was picked after the preceeding correction was applied, but that throws away some information that your method makes good use of. I'm not sure what api changes to suggest here. Let me think about this some more.

My code make sure the correction are never *unstable* in the control theory sense of the term.
This line of code:
```java
double factor = 1 / (double)(2 << visionsSinceLastFeed); // Start at 0.5 then have each new vision half.
```
Make sure the gain of the sum of all corrections is strictly `< 1`.
I believe you can have nonsensical correction in cases you mentioned:
> "Pick a, align a, align a, align a" or "Pick a, pick/align/discard b, pick/align/discard b, pick/align/discard b, align a"

But thanks to limiting the gain it wouldn't be terrible.

This problem goes away if we move it to the JobPlanner execution code, I started like this.
But it felt hacky.

> There is an openpnp convention that homing the machine resets all transient state ...

Ok, Will do.

> I think there is some documentation needed. I suggest a new wiki page explaining the underlying principal of operation. I think this is important because this is a new invention; even users familiar with other pick and place machines will not be familiar with this concept. Please let me know if you want me to draft something.

I'll write something.

> The feature needs a name. Ideally something that works as both a title for that wiki page, and the bottomVisionResultCallback method name. I tentatively propose "Bottom Vision Feeder Correction", but up to you.

« Bottom Vision Feeder Correction » is pretty simple but Ideally the name would convey that the correction apply to the next correction from place correction.
From discussion I have most people expect to use top vision to measure the component location, then use the measurement to pick the component.
What about « **Deferred Vision Pick Correction** » ? or « **Deferred Bottom Vision Pick Correction** » ?

> Do you think the photon feeder benefits from a gui option to enable/disable this feature? This can be on by default. This will be useful for whoever fixes the photon firmware unreliable feed distance.

Yes we also need:
- a button to reset the correction (I wanted it to be so good you wouldn't but Analogy has convinced me otherwise)
- text display to show the live correction (so people can check by themselves it is doing something)
- a selector to configure the feeding plane rather than hardcoding the slot ID, there are "wide body" mods that increase the X axis so both nozzle can reach the full bed, and so you can mount more feeders.

---

Lastly I wanted to extract the I-controler feature so other feeders could use it.
Either as a class extending `ReferenceFeeder` or as a class providing the logic and state but feeders would instantiate as a private member.

Solomute:
This concept works great! Has basically eliminated drift-related picking issues for me.

In order to be ready to become an official feature:
- Needs to be more configurable/visible in general. The feeder configuration window should have a section devoted to this feature, with a checkbox to turn it on/off, a display of the current correction, a button to zero it out, and a way to bake the current correction into the feeder's "official" pick location.
- The feed amount correction is currently hardcoded to assume that feeder slots >25 feed in the -Y direction and feeder slots<=25 feed in the +Y direction. Probably add some configuration in the feeder configuration window. A checkbox to turn feed amount correction on/off, and a dropdown for feed direction, +/- X and +/- Y (some machines in the future may have a feeder rail running down the Y axis, feeding along the X axis)
- Generalizing it so that it can be easily added to other feeder types (i.e. tray feeders) would be awesome. Actually, implementing this for tray feeders would be really nice because when I am using a strip holder to feed parts, and I need to refill it, I can't always line up the new strip of parts to quite the same spot. If I can just get it ballpark, and have openpnp notice how far off and correct, that would be swell.
- Two possible solutions to the "multiple consecutive picks from the same feeder" problem: First, only do a correction the first time a particular feeder is seen in a cycle. Second, the correction should be an average of all of the times a feeder is picked from during a cycle, instead of a sum. You cannot rely on just keeping correction gain low enough to not cause instability with 2 nozzles because... What if someone builds a 4 nozzle machine? Or an 8 nozzle machine? If your solution is to cut your correction gain to 1/8 on an 8 nozzle machine, you may end up with your correction gain not being high enough to converge in a timely manner, or to track high frequency drift.

Bug report:
When the job is paused, using the yellow right-arrow button to step through the job does not work when vision is the current step

Name for feature?
Closed loop pick position correction

tobydickenson:
What is the schedule for getting this ready for merge?

Ideally I would like to see this merged early January, so the enthusiastic testers can try it out before the beta release and feature freeze at the start of February.

Jorropo:
I plan to work on it this month. I want it to be included in the next release.

tobydickenson:
@Jorropo This is a schedule reminder.... Ideally this would be merged a few weeks before the end of January to allow some use by the enthusiastic test branch users before the freeze and beta release.

Jorropo:
Picked a name « Deferred Closed Loop Pick Position » based on @Solomute's idea but with `Deferred` added to signify it works backward and dropped `Correction` for brevity.

n0tmad84:
Usually I wouldn't comment on a MR. I'm using this branch for some weeks now and made 2XX pcbs on my lumenPnP 4.0 and lumen feeders. This is by far the most useful improvement for using these feeder. In combination with "move while feeding" I can achieve 750CPH with very good reliability now. Except for a nullpointer exception when test aligning I have not encounter any issues.

**Thanks a lot!**

syagger:
Hi @Jorropo — thanks for this work, it looks very promising.

Just built the press-play-walk-away branch (commit be22c51) to try it out. Ran into three issues getting it to run; sharing in case they're useful. Another user on the Opulo Discord confirmed hitting the same ones.

1. Checkstyle failure

PhotonFeeder.java:689: 'if' construct must use '{}'s. [NeedBraces]

Worked around with -Dcheckstyle.skip=true, but it blocks a plain mvn package.

2. Existing configs fail to load — PhotonFeederSlots.java

Error while reading machine.xml
(Constructor not matched for class org.openpnp.machine.photon.PhotonFeederSlots$Slot)

The diff against upstream/test shows the private no-arg constructor was replaced by the new field:

diff
-        private Slot() {}
+        private PhotonFeeder feeder;

SimpleXML needs the no-arg constructor to deserialize. Adding it back alongside the new field fixes loading. I think this affects anyone with existing Photon feeders in their config.

3. NPE in PhotonFeeder.setSlotAddress()

Cannot invoke "PhotonFeederSlots$Slot.getFeeder()" because "oldSlot" is null

Line 502 dereferences oldSlot without a null check, while line 501 does check getSlot() != null. Fixed locally with:

java
if (oldSlot != null && oldSlot.getFeeder() == this) {

With those three changes the branch builds and runs fine on my LumenPnP v4.1. About to start testing the actual correction behaviour on 0402 parts — will report back.

dkonigsberg:
For anyone who is interested, I've decided to maintain my own fork of this PR here:
https://github.com/dkonigsberg/openpnp/tree/press-play-walk-away

Its basically a copy of the version @Jorropo has here:
https://github.com/Jorropo/openpnp/tree/press-play-walk-away

With a few of the commits edited to resolve various merge/build/test/run issues, including those mentioned by @syagger above.  I've noted these changes in the commit messages of those edited commits.

I did this because I want something that applies cleanly to the upstream [openpnp/test](https://github.com/openpnp/openpnp/tree/test) branch, without stomping on any other changes, to make it easier for us to continue to use this feature until such time that its either unnecessary or complete enough to actually get merged into openpnp/test officially.

Jorropo:
Hey I have done a lot of refactoring and rebased onto current `test`, I've renamed the feature to `Pick Offset Correction` to avoid a needlessly long name.

I have hoisted the current correction algorithm to it's own class that feeders can leverage with minimal boiler plate.

There are two algorithms, one that supports 0 DOFs feeders, and one that extends on the first one and supports 1 DOFs feeders.

Photon is the only wired feeder and it uses the 1 DOF version.

Any similar looking feeder where the pick position is static should easily be able to use the 0 or 1 DOF algorithms.

For something like tray feeders which has 6 DOFs I have ideas but I didn't implemented any new algorithm yet.

---

The algorithm changed when using more than one nozzle per vision batch (2+ nozzles picking from the same feeders), rather having decreasing weights, it now average the error of all the nozzle tips.

---

I've added an UI:
Screenshot From 2026-09-24 03-56-12
It shows the current state, expose some config options, and additional debug info.

The Pick Offset Correction UI tab is reusable by multiple feeders.

---

You can see the whole package in action right there:

https://github.com/user-attachments/assets/49d2c8a2-f110-4d91-9530-d3b10fe4e276

---

There are tests who were written by the silicon intern, I've skimmed them, they look good.

---

I've added a machine listener that reset the correction state when homing the machine.

---

I would like to edit the wiki but I still do not have access to it AFAIK, it would be sad if that were the blocker.

---

**Imo this is good for review & merge.** thanks anyone taking a look

janm012012:
Hi Jorropo!

That's a very cool feature I've been looking for since a long time! I'm very happy to see, that you too the lead and implemented it. 

It would be cool - for me - to use it with a double nozzle machine and ReferencePushPullFeeders. This would require the algo to handle feed distances of multiple pitch distances like 4mm feed for two 2mm pitched picks of 0402 parts or one 8mm feed with two 4mm pitched picks for efficient 0603 handling. Handling picks and bottom vision for two nozzles in any order would be also required...

Did you added a listener to reset the controller in case the user manually updates e.g. the pick location?

Your work - once again - cries out for refactoring feeders into a base class where it could be made available for all of them at once.

Many thanks for your work and your patience for getting it merged!

Jan

Jorropo:
> It would be cool - for me - to use it with a double nozzle machine and ReferencePushPullFeeders. 

Ok noted, I'll submit a PR for you later today.

---

> Handling picks and bottom vision for two nozzles in any order would be also required...

This is already the case, the two vision results are averaged together and are considered one reading (I use this with two N045s).

---

> This would require the algo to handle feed distances of multiple pitch distances like 4mm feed for two 2mm pitched picks of 0402 parts or one 8mm feed with two 4mm pitched picks for efficient 0603 handling.

I am pretty sure the algo would already handle the double feed mode.
The algo's loop is feed → vision → loop.

So feed → feed → vision → vision → loop should be identical with feed × 2 → vision → vision → loop (it'll average the two visions results).

---

> Did you added a listener to reset the controller in case the user manually updates e.g. the pick location?

Right now I hardcoded `pickOffsetCorrection.reset()` calls in all of the feeder's methods where it make sense (setting the position, triggering a manual feed, ...).

So yes behavior, different implementation.

---

> Your work - once again - cries out for refactoring feeders into a base class where it could be made available for all of them at once.

I am not sure, different feeders would handle this very differently.
A rotated tray feeder would need to do an attempt at solving a matrix with 6 DOFs (translation, rotation, scale, skew) given the last N previous pick results to predict the positions of the next positions.

If anything we could have a stationary feeder base class for all the powered feeder (and pushpull feeder), but tbh I'm not a fan of big OOP and the injected code is easy to reason about.

Jorropo:
cc @janm012012 I had claude write an integration in the reference push pull feeder #1985 
I can't test it without modifying one of my two PnPs, I took a look at the code and it looks ok.

It's sad to not have nudging, and the strategy is a bit naive (it moves the whole feeder without ever trying to rotate the feed plane) but tbh the length of the farthest and closest component on the feed plane can't be that far otherwise any small error would make it unusable already so I think it should be an improvement over nothing even if it's not ideal.

PR #1915, opened 2025-11-07
camera batch operations

tobydickenson:
# Description
On [google groups](https://groups.google.com/d/msgid/openpnp/a63a86a4-b8b2-4a5e-ac9a-cf971a0e6c70n%40googlegroups.com?utm_medium=email&utm_source=footer) Micael wrote:

> Ideally, I'd like to find a way not to switch LED off from the previous shot, but then we need to move the control of the up light from the pipeline and get some knowledge about "entering pit/leaving pit" instead.

This change implements a new mechanism to defer turning lights off until the end of a batch of vision capture operations.

Batching is implemented for:
 * Board/panel fiducial scanning. It will leave the light on while scanning all fiducials,and turn it off after visiting the last fiducial.
 * Nozzle tip calibration
 * Bottom vision. Sometimes bottom vision needs multiple captures. This might be for composite vision, or because the first alignment showed an offset greater than the threshold and it therefore repeats the bottom vision from directly underneath the actual part center to eliminate parallax errors.
 * Job processor alignment stage. On a machine with two nozzles it leaves the bottom light on until all nozzles are aligned. 

This new mechanism does not interfere with anti-glare or any other mechanisms that explicitly turn a light off. It only affects the method for turning a light off *after a capture*.

# Justification

Micael wanted this because he is working towards having both nozzles running bottom vision without any machine movement.

I wanted this to improve the longevity of my ring light by keeping the lights turned off for most of the time. The original option to "turn the light off after capture" causes unwanted slowness, because vision would often have to wait through several frames of darkness.

# Concerns

1. The current implementation always enables this batching feature. Is there are need to make this optional? Are there any users who want the light to blink on and off during fiducial scanning, during nozzle tip calibration, or during multishot alignment?

2. Maybe logging is too verbose still. Should I change those `.info` to `.debug`?

# Instructions for Use

[this wiki](https://github.com/openpnp/openpnp/wiki/Setup-and-Calibration_Camera-Lighting) will need updating.

# Implementation Details
1. How did you test the change? -- Testing in a production job to follow next week.
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional  - justification for these changes. Changes to these packages require extensive review and testing. -- **yes** There is a new interface in `org.openpnp.spi`, and a new implementation of this interface in `org.openpnp.machine.reference`

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferencePnpJobProcessor.java
@@ -1546,6 +1546,33 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
             
             prerotateAllNozzles(alignLocator);
             
+            return new StartCameraBatchOperation(plannedPlacements);
+        }
+    }
+
+    /**
+     * Start a camera batch operation.
+     * Should any lights get turned on during an alignment, they remain on for the subsequent alignments.
+     */
+    protected class StartCameraBatchOperation implements Step {
+        protected List<PlannedPlacement> plannedPlacements;
+
+        protected StartCameraBatchOperation(List<PlannedPlacement> plannedPlacements) {
+            this.plannedPlacements = plannedPlacements;
+        }
+
+        public Step step() throws JobProcessorException {
+            if (cameraBatchOperationStarted) {
+                // unexpected!
+            } else {
+                CameraBatchOperation cbo = machine.getCameraBatchOperation();
+                if (cbo!=null)
+                {
+                    cbo.startBatchOperation("align step");
+                    cameraBatchOperationStarted = true;
+                }
+            }
+
             return new Align(plannedPlacements);
         }
     }
@@ -1561,28 +1588,7 @@ public class ReferencePnpJobProcessor extends AbstractPnpJobProcessor {
         @Override
         public Step stepImpl(PlannedPlacement plannedPlacement) throws JobProcessorException {
             if (plannedPlacement == null) {
-
-                if(cameraBatchOperationStarted)
-                {
-                    cameraBatchOperationStarted = false;
-                    try {
-                        machine.getCameraBatchOperation().endBatchOperation("align step");
-                    }
-                    catch (Exception e) {
-                        throw new JobProcessorException(machine, "Error turning lights off after vision.");
-                    }
-                }
-
-                return new OptimizeNozzlesForPlace(plannedPlacements);
-            }
-
-            if(!cameraBatchOperationStarted) {
-                CameraBatchOperation cbo = machine.getCameraBatchOperation();
-                if (cbo!=null)
-                {
-                    cbo.startBatchOperation("align step");
-                    cameraBatchOperationStarted = true;
-                }
+                return new EndCameraBatchOperation(plannedPlacements);
             }
 
             final Nozzle nozzle = plannedPlacement.nozzle;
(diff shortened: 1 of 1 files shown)


janm012012 (commented):
Hi Toby!

Besides the option questions you're discussing with Micael on the list, I'd like to add some comments here:

- Please review your coding style. You mixed if statements with the opening brace in the same line (the common style for OpenPnP) and in the following line. In addition, it seems common to have a space between if and '('.
- I don't understand the reason for `cameraBatchOperationStarted` if you provide `getCameraBatchOperation()`. You could just use the later as it appears paired in most placed and even doubled up in the `Align()`.
- You may consider adding two extra steps surrounding `Align` to start/end camera batch operation while keeping `Align` as is.
- If `endBatchOperation()` throws an exception, I'd prefer to read that "endBatchOperation" has caused it and not "Error turning lights off". This is your current usecase but not obvious to others and might change in the future.
- As usual, please consider logging at debug level. I don't think that we need all the information at info (for normal users all day use).
- your implementation of `nestingLevel` is not thread safe. I'd prefer if you could fix that. https://stackoverflow.com/questions/15174611/is-incrementing-an-integer-thread-safe-in-java might provide guidance. If you can't, please add a warning. This type of problems are almost impossible to find and can cause big trouble.

Many thanks for provide this PR! I think its an other very good opportunity to further transform OpenPnP from a single to a multi nozzle machine control software.

Jan

tobydickenson:
Hi Jan, thanks for the review

>     * I don't understand the reason for `cameraBatchOperationStarted` if you provide `getCameraBatchOperation()`. You could just use the later as it appears paired in most placed and even doubled up in the `Align()`.

This is a consequence of having the batch operation in its own subystem object, and would not be needed if methods like`cameraBatchOperationStarted` were added to the machine object. Its the same as how `getFeeder()` and `feeder.feed()` are often paired together. I think this is ok, accepting that java is verbose.

> >     * You may consider adding two extra steps surrounding `Align` to start/end camera batch operation while keeping `Align` as is.

Thats interesting. I will see how that looks.

>     * If `endBatchOperation()` throws an exception, I'd prefer to read that "endBatchOperation" has caused it and not "Error turning lights off". This is your current usecase but not obvious to others and might change in the future.

Yes that would be better

>     * As usual, please consider logging at debug level. I don't think that we need all the information at info (for normal users all day use).

Agreed

>     * your implementation of `nestingLevel` is not thread safe. I'd prefer if you could fix that. https://stackoverflow.com/questions/15174611/is-incrementing-an-integer-thread-safe-in-java might provide guidance. If you can't, please add a warning. This type of problems are almost impossible to find and can cause big trouble.

All access *should* be from the machine thread, so it *shouldnt* need synchronisation. But I agree this would be a sensibly defensive.

tobydickenson:
> > > * You may consider adding two extra steps surrounding `Align` to start/end camera batch operation while keeping `Align` as is.
> > > ```
> 
> Thats interesting. I will see how that looks.

Having done that, I note that we have quite a few classes that `implements Step` which receive a `List<PlannedPlacement> plannedPlacements` in their constructor for the sole purpose of passing it along the line.

It would make sense to refactor that into a common base class, but I will leave that for another PR.

tobydickenson:
All feedback has been addresses. I will test this on a real machine in the next few days then plan to merge.

vespaman:
@tobydickenson  
When I ran my last job, it looked like the nozzle tip calibration (after swap) could sometimes be messed up, the light went out very quickly, and therefore the calibration failed. By manually switching on the light, it would work. I meant to have a look at it, but got a large order, so I haven't had time to do any more test. Might be worth looking deeper into, if you are going to run it on your machine. Strangely, I don't think all nozzle tip calibration failed, but it was some time ago now, so I might remember incorrectly.

tobydickenson:
> @tobydickenson When I ran my last job, it looked like the nozzle tip calibration (after swap) could sometimes be messed up, the light went out very quickly, and therefore the calibration failed. 

Thanks for the report, but I couldnt reproduce this problem. Lights were doing the right thing during tip calibration. Lets merge for wider testing

PR #1916, opened 2025-11-11
Multiple bottom vision cameras

vespaman:
# Read This First

# Description
This PR contains changes to allow the use of multiple bottom vision cameras in OpenPnP, e.g. one per nozzle, in order to speed up/minimize the small moves otherwise needed. It is the outcome of this [thread](https://groups.google.com/g/openpnp/c/cBpsV-cbKQY) in the OpenPnP group.

The PR contains a change to getBottomVisionCamera() method, which is used in many places, so the PR looks bigger than it really is. The change basically includes a nozzle argument, so getBottomVisionCamera() can return the correct up-vision camera.

The PR also contains means to avoid the move when going for the next bottom vision picture taking, by introducing image displacement properties on cameras. Once displacement is enabled (Configuration.get().getMachine().isDisplacementEnabled()) in machine.xml, small moves(*) in the camera pit will be replaced by displacement coordinates, that will be picked up by findOffsetsPreRotate().

*Small move limits can be adjusted by editing machine.xml. 

In order to achieve no (or very small) Z move between the nozzles, a "balance level" concept has been introduced. This defaults to 0.0mm. This balance level is the Z level where both nozzles on a shared Z axis are on the same level - balanced.
It's use is to make sure Z is known for both nozzles on cam-type heads, and that focus can be achieved on other shared Z style heads, and will be the Z level that bottomvision will use when pulling the head over to the camera pit for picture taking.

Note: 
Since I am still learning java, I am sure things can be improved. I consider this PR the basics for multiple up cameras, which can be improved in many steps going forward, and maybe before it can be considered good enough for inclusion upstream.

# Justification
This change improves bottom vision performance by removing the small move, with its settle time.

# Instructions for Use
Prerequisites: Two (or more) similar bottom vision cameras, with same setup (lenses etc). Cameras should be physically adjusted to have their crosshair to be as close to their respective nozzle tip center as possible. Camera distance/lenses should be setup so good enough focus can be had for both a short and tall component at the same Z (see balance level above).
Prerotate must also be enabled for the components, otherwise there will always be a physical movement between the pictures.

So far, there's no user interface for this change. Users with only one bottom vision camera should not be affected.
If users has multiple up-cameras, the code will bind each camera-nozzle by the name of camera/nozzle. E.g. "Nozzle 2" will be bound to "Camera2b". I.e. the first number found in camera/nozzle will be used for paring.
Now the machine will use the paired camera for every operation on a specific nozzle.
Lastly, machine.xml must be searched for "displacement-enabled" and set to "true".

# Implementation Details
1. How did you test the change? Using the built in test function, and on my own (physical) machine. (2 bottom vision cameras, one cam-style, shared Z axis).
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` : Yes, I had to, since I refactored getBottomVisionCamera(), now with nozzle argument (or null for default camera).
4. Be sure to run `mvn test` before submitting the Pull Request. Done!

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/camera/ReferenceCamera.java
+++ b/src/main/java/org/openpnp/machine/reference/camera/ReferenceCamera.java
@@ -190,7 +190,7 @@ public abstract class ReferenceCamera extends AbstractBroadcastingCamera impleme
 
     private Actuator lightActuator;
 
-    protected boolean displacementActive = false;
+    private boolean displacementActive = false;
 
     public enum FocusSensingMethod {
         None,
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -34,7 +34,11 @@ import org.openpnp.model.Placement;
 import org.openpnp.model.VisionCompositing;
 import org.openpnp.model.VisionCompositing.Composite;
 import org.openpnp.model.VisionCompositing.Shot;
-import org.openpnp.spi.*;
+import org.openpnp.spi.NozzleTip;
+import org.openpnp.spi.Nozzle;
+import org.openpnp.spi.PartAlignment;
+import org.openpnp.spi.PropertySheetHolder;
+import org.openpnp.spi.Camera;
 import org.openpnp.util.MovableUtils;
 import org.openpnp.util.OpenCvUtils;
 import org.openpnp.util.Utils2D;
@@ -558,9 +562,9 @@ public class ReferenceBottomVision extends AbstractPartAlignment {
             if (displacementIsEnabled) {
                  // Adjust 'shot Z' to nozzle balance level
                 if (nozzle != null) {
-                    double balance = Configuration.get().getMachine().getBalanceLevel();
+                    double balance = Configuration.get().getMachine().getBalanceLevel().getValue();
                     shotLocationTemp = shotLocationTemp.derive(null, null, balance, null);
-                    Logger.debug("Setting shotLocation Z to nozzle balance level: {} -> {}", balance, shotLocationTemp.getUnits());
+                    Logger.debug("Setting shotLocation Z to nozzle balance level: {}{}", balance, shotLocationTemp.getUnits());
                 }
             }
             final Location shotLocation = shotLocationTemp;
@@ -568,6 +572,7 @@ public class ReferenceBottomVision extends AbstractPartAlignment {
             pipeline.new PipelineShot() {
                 @Override
                 public void apply() throws Exception {
+                    camera.setDisplacementActive(false);
                     if (nozzle.getLocation().getLinearLengthTo(camera.getLocation(nozzle))
                             .compareTo(camera.getRoamingRadius()
                                     .add(nozzleTip.getMaxPickTolerance())) > 0) {
@@ -582,24 +587,19 @@ public class ReferenceBottomVision extends AbstractPartAlignment {
                             double rotationDifference = Math.abs(currentLocation.getRotation() - shotLocation.getRotation());
                             double zDifference = Math.abs(currentLocation.getZ() - shotLocation.getZ());
 
-                            if ( distance < 0.5
-                                    && (rotationDifference < 0.1 || rotationDifference == 360)
-                                    && zDifference < 0.1) {
+                            if ( distance < Configuration.get().getMachine().getDisplacementMaxDistance().getValue()
+                                    && (rotationDifference < Configuration.get().getMachine().getDisplacementMaxRotation() || rotationDifference == 360)
+                                    && zDifference < Configuration.get().getMachine().getDisplacementMaxZDiff().getValue()) {
                                 Logger.debug("Skipped moveTo because distance ({}) < 0.5mm, rotation difference ({}) < 0.1 degree, and Z difference ({}) < 0.1mm", distance, rotationDifference, zDifference);
                                 double xOffset = shotLocation.getX() - currentLocation.getX();
                                 double yOffset = shotLocation.getY() - currentLocation.getY();
                                 camera.setDisplacements(xOffset, yOffset);
                                 camera.setDisplacementActive(true);
                                 Logger.debug("Saved offsets - X Offset: {}, Y Offset: {}", xOffset, yOffset);
-                            }
-                            else {
-                                camera.setDisplacementActive(false);
+                            } else {
+                                nozzle.moveTo(shotLocation);
                             }
                         } else {
-                            camera.setDisplacementActive(false);
-                        }
-
-                        if (!camera.isDisplacementActive()) {
                             nozzle.moveTo(shotLocation);
                         }
                     }
--- a/src/main/java/org/openpnp/util/VisionUtils.java
+++ b/src/main/java/org/openpnp/util/VisionUtils.java
@@ -3,8 +3,11 @@ package org.openpnp.util;
 import java.awt.Color;
 import java.awt.image.BufferedImage;
 import java.io.IOException;
-import java.util.*;
-
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Comparator;
+import java.util.Collections;
 import org.apache.commons.io.IOUtils;
 import org.openpnp.machine.reference.vision.ReferenceBottomVision;
 import org.openpnp.machine.reference.vision.ReferenceFiducialLocator;
--- a/src/main/java/org/openpnp/machine/reference/ReferenceMachine.java
+++ b/src/main/java/org/openpnp/machine/reference/ReferenceMachine.java
@@ -163,7 +163,17 @@ public class ReferenceMachine extends AbstractMachine {
     private boolean displacementEnabled = false;
 
     @Element(required = false)
-    private double balanceLevel = 0.0;
+    private Length balanceLevel = new Length( 0.0, LengthUnit.Millimeters);
+
+    @Element(required = false)
+    private Length displacementMaxDistance = new Length( 0.5, LengthUnit.Millimeters);
+
+    @Element(required = false)
+    private Length displacementMaxZDiff = new Length( 0.1, LengthUnit.Millimeters);
+
+    @Element(required = false)
+    private double displacementMaxRotation = 0.01;
+
 
     @Element(required = false)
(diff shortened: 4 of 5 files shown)


janm012012:
Hi vespaman!

Many thanks for contributing your multi-camera PR.

It is good, that you provide a dedicated enable flag for this feature and that you already exposed the balance point to the machine.xml. Please also expose the max. displacements. Might be useful for others to adjust them.

I'll put some more comments to your code.

Jan

vespaman:
Hi Jan, 
thanks for looking at my contribution. I'm sure there's a lot that can be improved!

> Please also expose the max. displacements. Might be useful for others to adjust them.

I agree X/Y distance might be useful to change. But (especially) rotation and perhaps Z, are they not too technical? 
(I don't mind adding them, just want to make sure we don't add unnecessary settings)

**Edit:** Perhaps this is the norm in OpenPnP, to have everything in machine.xml, even if there will never be any UI for it? If this is the case, I totally agree that also rotation and Z should be soft. 

> I'll put some more comments to your code.

Thanks!

 Micael

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -562,7 +575,33 @@ public void apply() throws Exception {
...
                     else {
-                        nozzle.moveTo(shotLocation);
+                        if ( displacementIsEnabled ) {
+                            Location currentLocation = nozzle.getLocation();
+                            Logger.debug("Nozzle current location: {}, Target shot location: {}", currentLocation, shotLocation);
+                            double distance = currentLocation.getLinearLengthTo(shotLocation).getValue();
+                            double rotationDifference = Math.abs(currentLocation.getRotation() - shotLocation.getRotation());
+                            double zDifference = Math.abs(currentLocation.getZ() - shotLocation.getZ());
+
+                            if ( distance < 0.5
```
janm012012: Please replace the numbers by variables giving them meaningful names and expose them to the machine.xml. In addition, please convert them to Length type (which includes it unit and supports automatic  unit conversion) and consider grouping them into a Location type (which contains x, y, z and rotation as Length).
The hard coding of units might be troublesome for non-SI-unit users...

src/main/java/org/openpnp/util/VisionUtils.java
```diff
@@ -3,11 +3,7 @@
 import java.awt.Color;
 import java.awt.image.BufferedImage;
 import java.io.IOException;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
janm012012: Please do not use asterisk, stick with the dedicated list.

src/main/java/org/openpnp/util/VisionUtils.java
```diff
@@ -146,10 +143,46 @@ public int compare(Location o1, Location o2) {
         });
         return locations;
     }
-    
-    public static Camera getBottomVisionCamera() throws Exception {
+
+    public static Integer findFirstNumber(String str) {
```
janm012012: Please see https://stackoverflow.com/questions/14974033/extract-digits-from-string-stringutils-java for a more java style to solve that.

janm012012 (commented):
After reviewing the code, I've a few more questions:

- Please check the nodes I've added.
- Please tell me where the xy-displacement is compensated. I can see, that you store it, but not where you use it.
- If you allow bottom vision to take place at a specified Z level (balanceLevel) you shall verify, that this is supported by the current setup. It requires 3d-UPP or advanced camera calibration.
- How do you compensate for the new Z level if you force it to balanceLevel? I don't see that you take the Z level into account anywhere.

Concerning the mount of getBottomVisionCamera() calls, you may - how that you caught all -  consider providing `getBottomVisionCamera()` again with just a default implementation to `getBottomVisionCamera(null)`. This shall remove a lot of the related changes from the final changeset.

Jan

janm012012:
Hi Micael!

> I agree X/Y distance might be useful to change. But (especially) rotation and perhaps Z, are they not too technical? (I don't mind adding them, just want to make sure we don't add unnecessary settings)
> 
> **Edit:** Perhaps this is the norm in OpenPnP, to have everything in machine.xml, even if there will never be any UI for it? If this is the case, I totally agree that also rotation and Z should be soft.
> 

Exposing them to the machine.xml allows to change them. It's cheap from the programmers perspective (a least cheaper then coding a UI), but as they are highly machine specific, especially as they are unit-related, users will need to change them. Also its not a big deal for the machine.xml to carry a few more parameters. And, if they are exposed, they can be changed quickly without having the create a new PR...

Jan

vespaman:
Hi Jan,
Thank you for your input!

>     * Please check the nodes I've added.

Will do! 
 
>     * Please tell me where the xy-displacement is compensated. I can see, that you store it, but not where you use it.

This is done at row 234;
                // If displacement is active, we need to adjust x and y offsets to match the displacement.
                if( camera.isDisplacementActive() ) {
                    Location displacement = nozzleLocation.derive(camera.getXDisplacement(), camera.getYDisplacement(),null,null);
                    nozzleLocation = nozzleLocation.subtract(displacement);
                    Logger.debug("Bottom vision displacement active, adjusting nozzle location by {}, new nozzleLocation {}", displacement, nozzleLocation);
                }

> 
>     * If you allow bottom vision to take place at a specified Z level (balanceLevel) you shall verify, that this is supported by the current setup. It requires 3d-UPP or advanced camera calibration.

I have no idea on how to do that, especially given all different setups. I am not sure I think this is very important (remember, this is not a Z that camera is using, this is only the top of the component)?

> 
>     * How do you compensate for the new Z level if you force it to balanceLevel? I don't see that you take the Z level into account anywhere.
 
I don't. 
I realized that the current bottomvision code caps the Z at 0, and does not do an further calculation on the height. (This is how I have been running my machine since about 2 years now). I think the distance to camera comes in handy here; the size difference becomes very small in relation to the sensor distance.

The fix, as I understand it, would be to, once 0 has been capped, add back the height of the component (backwards), but since this is a generic function, and I _really_ struggle with positive Z (e.g. Lumen), I simply do not know how this works in reality.

I do agree that this might be a nice improvement, esp. more expensive lens/sensor setups. But maybe we can see how this plays out? I think this is really only an issue on shared Z axis anyway. 
And regardless; after understanding how this is working today, I think this should be a separate change/PR/improvement. 

> Concerning the mount of getBottomVisionCamera() calls, you may - how that you caught all - consider providing `getBottomVisionCamera()` again with just a default implementation to `getBottomVisionCamera(null)`. This shall remove a lot of the related changes from the final changeset.
 
I did, but I arrived to the conclusion that it might be better to force people to have a think about what they want. Most of the time, they will probably want a nozzle argument. Especially, since this is an old function, that might be used out of habit. The null argument is only used at 6 places. I am flexible though, so if you insist, I'll make this change?

janm012012:
Hi Micael!

> I agree X/Y distance might be useful to change. But (especially) rotation and perhaps Z, are they not too technical? (I don't mind adding them, just want to make sure we don't add unnecessary settings)
> 
> **Edit:** Perhaps this is the norm in OpenPnP, to have everything in machine.xml, even if there will never be any UI for it? If this is the case, I totally agree that also rotation and Z should be soft.
> 

Exposing them to the machine.xml allows to change them. It's cheap from the programmers perspective (a least cheaper then coding a UI), but as they are highly machine specific, especially as they are unit-related, users will need to change them. Also its not a big deal for the machine.xml to carry a few more parameters. And, if they are exposed, they can be changed quickly without having the create a new PR...

Jan

vespaman:
Hi Jan,
Yes, I agree. I'll add them to machine.xml too! 
While testing further yesterday, I found the displacement sometimes malfunction (after merging with the new test base), so I am trying to understand why this is. It looks like the order of things are changed somehow. 
So give me some time, before coming back to your suggested changes.

src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
```diff
@@ -34,11 +34,7 @@
 import org.openpnp.model.VisionCompositing;
 import org.openpnp.model.VisionCompositing.Composite;
 import org.openpnp.model.VisionCompositing.Shot;
-import org.openpnp.spi.Camera;
-import org.openpnp.spi.Nozzle;
-import org.openpnp.spi.NozzleTip;
-import org.openpnp.spi.PartAlignment;
-import org.openpnp.spi.PropertySheetHolder;
+import org.openpnp.spi.*;
```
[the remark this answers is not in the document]
vespaman: It was actually the IDE that suggested this :-) But I'll change it!

src/main/java/org/openpnp/machine/reference/camera/ReferenceCamera.java
```diff
@@ -190,11 +190,22 @@ public abstract class ReferenceCamera extends AbstractBroadcastingCamera impleme
 
     private Actuator lightActuator;
 
+    protected boolean displacementActive = false;
```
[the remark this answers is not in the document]
vespaman: OK, done!

src/main/java/org/openpnp/util/VisionUtils.java
vespaman: I looked at that (many suggestions!). I realize this can be achieved with regular expressions, but I am not sure I like it? They are rather slow, and this function is called a couple of times on each bottom vision.
I also think it is beneficial to return on the first number, not just getting all the numbers from the string. (E.g. "1st nozzle Type 2" - I really only want '1'.  Probably this can be solved with regex as well, but then it gets more complex, and afaict this rules out some of the other suggestions in that thread. 
Please advice.

vespaman:
Hi Jan,
I don't know how this works, but it looks like the pull request was updated automatically. But the change dates are still from a couple of days ago.

 -  Micael

vespaman:
> While testing further yesterday, I found the displacement sometimes malfunction

Turned out to be one of my pre-PR cleanups that I cleaned a little bit too much. Simple fix.

quang301095:
If multiple cameras have been added, each camera will process the pipeline in parallel. can you?

janm012012:
> If multiple cameras have been added, each camera will process the pipeline in parallel. can you?

No, that's not supported yet. The pipeline still runs synchronously in the machine task. Please feel free to enhance that. Any PR is very much appreciated.

Jan

Issue #1920, opened 2025-12-05
Deleting the last part in the Parts table causes a hang.


kbralten:
Version 2.4
Windows 11, x64

Deleting the last part in the Parts list causes OpenPNP to hang. Tabbing away from and back to the parts tab creates an invalid UI.

<img width="1210" height="873" alt="Image" src="https://github.com/user-attachments/assets/16d3beb9-f94c-4f66-8d42-f10c9f128eb9" />
<img width="280" height="117" alt="Image" src="https://github.com/user-attachments/assets/8860101d-d989-4e5e-8f82-dc2eabd5702f" />
<img width="1210" height="873" alt="Image" src="https://github.com/user-attachments/assets/9ef87415-09a1-4082-bca5-fc4c2076a6c6" />

Deleting parts from anything other than the last row works.

kbralten:
This is already resolved in test.

tobydickenson:
#1906, right?

kbralten:
Yes, looks like #1906 fixed this, even if that wasn't the intention of the change. It's still silently absorbing the issue with the index. Thanks for digging out the right reference.

PR #1921, opened 2025-12-05
Fix hang when deleting last part from parts panel. openpnp/openpnp#1920

kbralten:
# Description
Correct openpnp/openpnp#1920 where PartsPanel hung when deleting the last part.

# Justification
Annoying UI glitch.

# Instructions for Use
Open the PartsPanel, select the last row from the parts list, click delete and acknowledge.

# Implementation Details
1. Minimal patch
2. Test exercises this behaviour
3. Guarded against a null `mainFrame` so the PartsPanel can be unit tested.


tobydickenson:
This looks good to me. @kbralten ready for merge?

kbralten:
The CI test smokes with a `java.awt.HeadlessException`. I don't want to rework the whole testsuite to let AWT initialize. Running the test locally passes and I've run this for the last week without issue.

Here's a patch to pull the test. 

Otherwise good to merge from my view.

PR #1928, opened 2025-12-17
Block keyboard shortcuts if a TextInput is focused

lukebayes:
When editing a text input (i.e., setting height, or constructing a vision pipeline), it's common to press CTRL+SHIFT+ARROW or CTRL+ARROW to select a word to the left or right. Unfortunately, these commands also trigger unexpected keyboard shortcuts which risk damaging machines and parts.

This change will block keyboard shortcuts if a text input has focus.

I tested the change by triggering the issue in the main build of openpnp and then verifying that it is not present this build. 

I ran `mvn test` and everything passed.

### DIFF
--- a/src/main/java/org/openpnp/gui/MainFrame.java
+++ b/src/main/java/org/openpnp/gui/MainFrame.java
@@ -574,11 +574,16 @@ public class MainFrame extends JFrame {
             @Override
             protected void dispatchEvent(AWTEvent event) {
                 if (event instanceof KeyEvent) {
-                    KeyStroke ks = KeyStroke.getKeyStrokeForEvent((KeyEvent) event);
-                    Action action = hotkeyActionMap.get(ks);
-                    if (action != null && action.isEnabled()) {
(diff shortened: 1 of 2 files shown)


tobydickenson:
Thanks for the PR. And thank you for checking on google groups first.

I do wonder whether this change will block too much. That is, whether an operator will get frustrated because he presses Ctrl+Shift+Arrow expecting the machine to move, but it gets blocked because of where the focus happens to be. I guess that is the fundamental problem with having duplicated keyboard shortcuts. Anyway, the best way to find out whether this is a problem is to merge it with the test branch.

Thank you for the contribution.

tobydickenson:
> merge it with the test branch

Aha! your PR is targeting `main` but it should be the `test` branch. Can you change that please.

lukebayes:
Thanks for checking it out!

Sorry about the main branch mix up. I've re-applied my changes onto the upstream test branch and force pushed it.

Please let me know if there's anything else I need to do.

PR #1930, opened 2025-12-17
photonfeeder: change the naming scheme in-order to order by slot

Jorropo:
# Description
Ordering by name now orders by physical slots; which is way more useful than by UUIDs.

# Justification
It allows users to order their photon feeders by physical slots.

# Instructions for Use
Go to the feeder tab,
Create a photon feeder,
Go to global config in the photon feeder,
Scan all feeders,
Back in the feeder spreadsheet click to order by name.

# Implementation Details
  I've ran it on my machine.
  I hope so.
  It is running on my computer.

### DIFF
--- a/src/main/java/org/openpnp/machine/photon/PhotonFeeder.java
+++ b/src/main/java/org/openpnp/machine/photon/PhotonFeeder.java
@@ -1,5 +1,11 @@
 package org.openpnp.machine.photon;
 
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+
+import javax.swing.Action;
+
 import org.openpnp.ConfigurationListener;
 import org.openpnp.gui.support.Wizard;
 import org.openpnp.machine.photon.exceptions.FeedFailureException;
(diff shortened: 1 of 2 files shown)


tobydickenson:
Hi, sorry for the slow reply, I have been away for a few days.

I am not keen on the string formatting implementation.
1. Using `String.format` to build another template string for `String.format` is problematic:
  a. It will be error-prone for translators.
  b. It is not _obviously_ evident that there is no vulnerability analogous to sql injection. I dont think there is, but it triggered my vulnerability spider-sense.
2. There definitely are cases where it is ok to use floating point operations to implement integer functions, but I dont think using log10 to calculate the number of decimal digits in as 8 bit number is one of them.

What about a function like this:

```
public String formatSlot(int slot) {
    if (maxFeederAddress>=0) {
        if (maxFeederAddress<10) {
            // If we have less than 10 slots then format them as a single digit decimal number
            return String.format("%01d",slot);
        }
        if (maxFeederAddress<100) {
            // If we have less than 100 slots then format them as a two digit decimal number
            return String.format("%02d",slot);
        }
    }
    // Otherwise a three digit decimal number
    return String.format("%03d",slot)
}
```

Jorropo:
Sure, thanks for the review. done !

PR #1934, opened 2025-12-18
Push pull delay

tobydickenson:
This is a provisional PR

# Description
This adds an optional delay to each step of ReferencePushPullFeeder motion.

# Justification
1. Sometimes it seems like the Marlin motion controller will not quite complete the full motion when given a continuous motion to move to a point, then turn around and move back in the opposite direction. Or maybe the problem is stretch and deflection of the 3d printed head assembly under acceleration loads. Either way, in practice I find that adding a tiny delay can increase reliablility of push/pull actions.
2. I have been experimenting with some push/pull actions which need a small (but not tiny) delay.

# Concerns
a. I find the PushPullFeeder gui intimidating at the best of times and I am aware that this change is only making that worse. I would welcome any suggestions on how to improve this.
b. PushPullFeeder gui has no translations; the new text introduced by this PR also does not.

# Outstanding Tasks
1. Testing on real machine. (although I have been using a gui-less version of this for a while)
2. rebase and tidy the commit comment.

# Instructions for Use

Draft wiki content, to be inserted after [motion path / locations](https://github.com/openpnp/openpnp/wiki/ReferencePushPullFeeder#motion-path--locations)

 If the delay value greater than zero, the machine will pause for the specified number of milliseconds on reaching this point. It pauses in both directions, forward and back. If the delay is zero then there is no pause, and the machine moves smoothly through each point.

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. -- described above
4. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? -- yes
6. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. -- test pass

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java
+++ b/src/main/java/org/openpnp/machine/reference/feeder/ReferencePushPullFeeder.java
@@ -160,6 +160,17 @@ public class ReferencePushPullFeeder extends ReferenceFeeder {
     protected boolean includedPull3 = false; 
 
     @Attribute(required = false)
+    protected int delay0 = 0;
+    @Attribute(required = false)
+    protected int delay1 = 0;
+    @Attribute(required = false)
+    protected int delay2 = 0;
+    @Attribute(required = false)
+    protected int delay3 = 0;
+    @Attribute(required = false)
+    protected int delay4 = 0;
+
+    @Attribute(required = false)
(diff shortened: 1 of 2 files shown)


tobydickenson:
image

tobydickenson:
I plan to merge this later this week

PR #1935, opened 2026-01-03
Estimate pcb z

kbralten:
# Description
Add a "focus method" that uses parallax to estimate Z height.
Also enhance the "Estimate Z-height of Object" menu to fall back to using the auto-focus method (and therefore the parallax calculation) if no secondary camera is available.

# Justification
This is a simple, if not very accurate, method of evaluating Z distances; this can be used as a base method to build automated checks for z height since the availability of a fixed height fixed focus camera is basically universal.

# Instructions for Use
This feature estimates the Z-height of parts or PCBs using parallax from a single down-looking perspective camera by moving the head in X/Y.

## Prerequisites

-   **Camera Type**: Must be a **Perspective** camera (telecentric lenses will not work).
-   **Movement**: The camera must be capable of moving in X and Y. Z movement is not required for calibration as it relies on objects of different heights.

## 1\. Enable the Feature

1.  Open OpenPnP and go to **Machine Setup**.
2.  Select your down-looking camera in the tree.
3.  On the **General** settings tab, find **Focus Sensing Method**.
4.  Select **

    Parallax** from the dropdown.
5.  Click **Apply**.

## 2\. Configuration

Navigate to the **Auto Focus** tab (this tab appears after you apply the "Parallax" method).

image

-   **Shift Distance**: The distance (in mm) the camera will move sideways to create parallax. Recommended: **2mm - 10mm**. Larger distances provide better resolution but keep the feature in the field of view.
-   **Feature Size**: The size (in pixels) of the square region used for template matching. Recommended: **64 - 128**. Larger sizes help when the camera zooms digitally or changes focus.
-   **Settle Time**: Milliseconds to wait after moving before capturing an image (to reduce vibration blur). Recommended: **200ms - 500ms**.
-   **Show Diagnostics**: Check this to see debug images of the template matching process in the log/debug view.

## 3\. Calibration

You must calibrate the "Focal Point Z" (the optical center of the camera) for accurate measurements.

This process involves measuring the parallax at two different heights. Since most top-down cameras have a fixed Z-height, you will need **two targets of different heights** (e.g., the PCB surface and a block placed on the PCB).

3.  **Calibration**:
    -   **Step 1**: Center the camera on a distinct, flat feature on the PCB surface (Low Z).
    -   Enter the Z-height of this surface in the **Surface Z (mm)** field. This should be the absolute Z coordinate (e.g., -23.5). If unknown, use the nozzle tip to touch the surface and read the Z position from the DRO.
    -   Click the **Measure Low Z** button. The "P1" field will update with the measured parallax.
    -   **Step 2**: Place a block of known height (e.g., 5mm or 10mm) on the PCB.
    -   Enter the height difference (thickness) of the block in the **Block Height (mm)** field (e.g. 5.0). This is relative to the first surface.
    -   Center the camera on a distinct feature on top of the block (High Z).
    -   Click the **Measure High Z** button.
    -   The system will calculate the Focal Point Z.
    -   Save your configuration.

## 4\. Testing & Usage

-   **Test Button**: Click **Test Part Height** in the wizard. The machine will perform a full measurement cycle and report the estimated Z-height of the surface currently under the camera.

# Implementation Details
1. How did you test the change? _Tested on two PNP machines with different camera setups. Sadly none have a secondary camera or actual autofocus so I can't test the previous paths._
2. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? _Yes🙏_ 
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. _No Changes_
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. _Green tests_.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/camera/ParallaxPartHeightProvider.java
+++ b/src/main/java/org/openpnp/machine/reference/camera/ParallaxPartHeightProvider.java
@@ -30,6 +30,9 @@ public class ParallaxPartHeightProvider implements FocusProvider {
     private int featureSize = 64;
 
     @Attribute(required = false)
+    private int framesToAverage = 1;
+
+    @Attribute(required = false)
     private double focalPointZ = 0.0;
     
     @Attribute(required = false)
@@ -51,11 +54,10 @@ public class ParallaxPartHeightProvider implements FocusProvider {
     public double measureParallaxShift(Camera camera, HeadMountable movable, Location startLocation) throws Exception {
          // 1. Capture image at start location
         movable.moveTo(startLocation);
-        BufferedImage templateImage = camera.settleAndCapture();
+        Mat templateMat = captureAndAverage(camera, framesToAverage);
         
         try {
             // 2. Create template from center
-            Mat templateMat = OpenCvUtils.toMat(templateImage);
             Mat template = cropCenter(templateMat, featureSize);
             
             if (showDiagnostics) {
@@ -73,8 +75,7 @@ public class ParallaxPartHeightProvider implements FocusProvider {
             movable.moveTo(shiftLocation);
             
             // 4. Capture search image
-            BufferedImage searchImage = camera.settleAndCapture();
-            Mat searchMat = OpenCvUtils.toMat(searchImage);
+            Mat searchMat = captureAndAverage(camera, framesToAverage);
     
             // 5. Find feature using Template Matching
             Point matchLoc = matchTemplate(searchMat, template);
@@ -143,7 +144,7 @@ public class ParallaxPartHeightProvider implements FocusProvider {
         return new Mat(img, new org.opencv.core.Rect(x, y, size, size));
     }
     
-    // Helper match template
+    // Helper match template with sub-pixel refinement
     private Point matchTemplate(Mat img, Mat templ) {
         int result_cols = img.cols() - templ.cols() + 1;
         int result_rows = img.rows() - templ.rows() + 1;
@@ -152,8 +153,44 @@ public class ParallaxPartHeightProvider implements FocusProvider {
         Imgproc.matchTemplate(img, templ, result, Imgproc.TM_CCOEFF_NORMED);
         MinMaxLocResult mmr = Core.minMaxLoc(result);
         
+        Point refined = subPixelRefine(result, mmr.maxLoc);
+        
         result.release();
-        return mmr.maxLoc;
+        return refined;
+    }
+    
+    private Point subPixelRefine(Mat result, Point maxLoc) {
+        int x = (int) maxLoc.x;
+        int y = (int) maxLoc.y;
+        
+        // Check bounds (need 1 pixel border)
+        if (x < 1 || x >= result.cols() - 1 || y < 1 || y >= result.rows() - 1) {
+            return maxLoc;
+        }
+        
+        // Extract 3x3 window around peak
+        // y0 is left/top, y1 is center (max), y2 is right/bottom
+        
+        // Refine X (horizontal)
+        float v_x_minus = (float) result.get(y, x - 1)[0];
+        float v_x_center = (float) result.get(y, x)[0];
+        float v_x_plus = (float) result.get(y, x + 1)[0];
+        
+        double deltaX = (v_x_minus - v_x_plus) / (2.0 * (v_x_minus - 2.0 * v_x_center + v_x_plus));
+        
+        // Refine Y (vertical)
+        float v_y_minus = (float) result.get(y - 1, x)[0];
+        float v_y_center = (float) result.get(y, x)[0];
+        float v_y_plus = (float) result.get(y + 1, x)[0];
+        
+        double deltaY = (v_y_minus - v_y_plus) / (2.0 * (v_y_minus - 2.0 * v_y_center + v_y_plus));
+        
+        // Sanity check deltas (should be within +/- 0.5)
+        if (Math.abs(deltaX) > 0.6 || Math.abs(deltaY) > 0.6) {
+             return maxLoc;
+        }
+        
+        return new Point(maxLoc.x + deltaX, maxLoc.y + deltaY);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


kbralten:
Expanding `FocusSensingMethod` seemed like the right path forward, but happy to rework this into another system if there's a better fit.

kbralten:
This is only accurate to about +/-1mm (maybe better, it's within 0.5mm in every test I've written down) but can produce silly values if used on a repeating pattern since the algo is based on pattern finding - e.g. it's simply awful if you try to use one pad from a TSSOP or something. I have been testing mostly with fiducials (or silk screen dots) but using one pad off a large electrolytic seems to be equally effective if not very common. I'm not 100% confident on it, but it also does OK with pockets from 8mm tape.

tobydickenson:
Hi, this is a nice idea.

The `Thread.sleep(ms)` needs to be `moveable.delay(ms)`. This will implement the machine coodination (waiting for all movement to complete before starting the delay).

The geometry is not favorable for getting an accurate measurement using this method. A`matchTemplate` position result is quantised to whole pixels (unlike the circle detection, which can give sub-pixel accuracy), and the geometry means that small errors in X give a larger error in Z. A single pixel error leads to a z error of (pixel width / tan(θ)). My machine has 12µm pixels and assuming a 10° parallax angle this gives 0.07mm resolution in Z measurement. Not ideal, but it should be usable.

I wonder it the accuracy can be improved by capturing more offset images, and averaging the result of multiple `matchTemplate` runs?

kbralten:
The geometry is overall not favourable for this and I'd be surprised if `matchTemplate` is accurate to a single pixel against the real world in this application either since we're inherently assuming there's parallax distortion from off-center imaging. In my testing I got less than 2 pixels of variation using the _image, nudge-back-and-forth, image_ test.

I'll test to see if repeated imaging gives a tighter average, but without digging in to `matchTemplate` I'd be concerned that it's coercing toward one side or the other or at the very least is doing it deterministically so we'd maybe see a doubling or resolution where the noise from repeated imaging gets us half-pixel resolution.

Honestly I was just aiming for "mms of resolution" and debated putting in forced rounding to avoid any assumptions parallax-based estimation was an alternative to a depth probe. My hope was to use this as a prelude to sanity checking Z inputs of boards/feeders etc. I'm just tired of forgetting to put in the z-height of boards (or forgetting the minus sign) and dropping parts from mid air or mistyping the z-height of feeders and having it smash the nozzle into it. Typoing the feeder height is what started this - I would have been happy with telling -21.5 from 0 but being able to tell -21.5 and -25.1 apart is fantastic for my use case.

Even if we can't improve accuracy, this should pretty reliably validate _is this really Z=-23, Z=0, or Z=23_ at least enough to prompt me before continuing.

Oh, and I'll get a commit in to fix `Thread.sleep`. I'm wondering now if that's even necessary, camera settling time should address that but I must have been struggling with that when I first implemented this.

kbralten:
> The `Thread.sleep(ms)` needs to be `moveable.delay(ms)`. This will implement the machine coodination (waiting for all movement to complete before starting the delay).

Done. Actually removed in favor of `settleAndCapture()`.

> I wonder it the accuracy can be improved by capturing more offset images, and averaging the result of multiple `matchTemplate` runs?

In my testing, no, the peak from `matchTemplate` tends to be strong enough you just get the same hot pixel each time.
But, we actually get a nice 2D grid matrix, so we can calculate a center from the weights around that pixel.
Once I added sub pixel calculation, I got marginally better results by averaging a few images, but it really depended on the hardware, on my higher resolution color camera it was useful, on a lower res global shutter grayscale camera it made no difference, so I made it configurable.

tonyluken:
I don't understand your comment 

> Also enhance the "Estimate Z-height of Object" menu to fall back to using the auto-focus method (and therefore the parallax calculation) if no secondary camera is available.

The "Estimate Z-height of Object" menu never required a secondary camera. It uses the information obtained from Advanced Camera calibration to perform what you seem to propose, i.e., use parallax to estimate Z coordinates.

tobydickenson:
I am still a little puzzled about the rationale for this feature....

If it could get 0.1mm accuracy then it would be very useful:
* It could be used for fine tuning the pick height. Some parts sit low in their pocket on the tape, so the nozzle needs to be a little lower to pick the part without it having to jump through the air by suction and possibly tumbling as it does so.
* Fine tuning the board height.

The existing feature that Tony mentioned gets 1mm accuracy for me, and I dont find that particularly useful.

> pretty reliably validate is this really Z=-23, Z=0, or Z=23

The top camera typically has a limited focal depth. If the image is in focus then you know the height within a few mm. If not, you cant really use the out-of-focus image to gauge anything about z.

kbralten:
> The "Estimate Z-height of Object" menu never required a secondary camera. It uses the information obtained from Advanced Camera calibration to perform what you seem to propose, i.e., use parallax to estimate Z coordinates.

Apparently I have mis understood "Estimate Z-height of Object". Let me go back and play with it. If it uses parallax then there's no value in _another_ parallax method.

> The existing feature that Tony mentioned gets 1mm accuracy for me, and I dont find that particularly useful.

That's interesting, because I reliably get 0.1mm level accuracy with my parallax estimation - or I get completely improbable estimates because off-axis imaging isn't reliable (e.g. shiny clear tape). So maybe there's a chance to port something learned.

I originally used this just to sanity check Z-axis inputs, but I accidentally found I can reliably fault on billboarded 0603s in the feeder.

If I'm having a dumb user moment with the existing functionality, then there isn't value here.

kbralten:
> The "Estimate Z-height of Object" menu never required a secondary camera. It uses the information obtained from Advanced Camera calibration to perform what you seem to propose, i.e., use parallax to estimate Z coordinates.

I had mentally confused it with detectPartHeight for uplooking cameras. There's probably a few bugs in the fallback logic then.

---

I went back and refreshed my memory...
 - As far as I can tell, the "Estimate Z-height of Object" relies on having 3D calibration complete and honestly I'd never done it because it's disabled when Advanced Calibration is active. It feels like it should be possible to derive this from Advanced calibration too, but I haven't debugged why that doesn't seem to happen. 
 - "Estimate Z-height of Object" relies on the user clicking-jogging-clicking and I wanted something automated but also something with better than "user clicking" level of precision.

I can see now that I've ended up duplicating the 3D units-per-pixel concept just with automated image recognition - which presumably can be worked out from Advanced Calibration anyway.

I think there's something to the _automated-jog-with-image-recognition parallax estimation_ because it's pretty accurate when given an easy to recognize target (it's pretty awful with a bad target).

I'm going to pull the PR for now though because there is an existing function, just not one I could make work well, so this is kind of a solution in search of a problem. If I get a chance I'll pull out just the automated-jog-with-image-recognition part and rework that to be a pure library function and resubmit that in hopes it fits a problem.

PR #1936, opened 2026-01-15
sot-23 vision pipeline

tobydickenson:
# Description
As discussed on google groups, some parts with 3 legs such as sot-23 footprint can appear triangular to bottom vision. This is not a great match for the vision pipeline which is oriented around finding rectangles.

Generally `MinAreaRect` will do the right thing, but sometimes it can do this:

image

This problem can be avoided by restricting the search angle to less than 20°.

This works because the false solution shown above is actually 360/3 = 120° out. Openpnp sees this as 30° pick alignment error (and openpn is ignorant of having the wrong major axis). Restricting the search angle to 20° allows a margin of safety of 10°. That is, the part would have to be picked with a misalignment of more than 10° before openpnp would select the false solution shown above.

# Justification

This is a common problem and there is a stock solution, so it makes sense to include this in openpnp default install.

# Instructions for Use

Needed on the wiki somewhere.

# FAQ

Should the same restriction be applied to other pipelines?

I dont think so. In the case of SOT-23 and other triangular parts, the `MinAreaRect` minimum shown above is a false minimum.  Other square parts do not have a false minimum solution in the same way.

It might make sense to add an _Alignment tolerance_ field that works like size checking: it raises an error and reject the part if was found to be picked at a funny angle. But that would be a different feature; this feature never raises an error.

# Timeline

I have been using a pipeline like this for sot-23 parts since #1667 fixed the search-angle code.

But this PR is untested, and I definitely wont have time to test before the next release.

# Implementation Details

1. How did you test the change? Be descriptive. Untested code will not be accepted. -- not yet tested
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. -- no

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
+++ b/src/main/java/org/openpnp/machine/reference/vision/ReferenceBottomVision.java
@@ -852,6 +852,7 @@ public class ReferenceBottomVision extends AbstractPartAlignment {
                 // Create other stock pipelines.
                 createStockVisionSettings(configuration, AbstractVisionSettings.STOCK_BOTTOM_RECTLINEAR_ID, "Rectlinear", "- Rectlinear Symmetry Bottom Vision Settings -");
                 createStockVisionSettings(configuration, AbstractVisionSettings.STOCK_BOTTOM_BODY_ID, "Body", "- Whole Part Body Bottom Vision Settings -");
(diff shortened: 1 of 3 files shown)


tobydickenson:
I am now testing this with the intention of merging next month

PR #1939, opened 2026-01-29
placement compression; parts can be placed slighltly below board z

tobydickenson:
# Description

This adds the Placement Compression field to `ReferenceMachine`. This is a Z offset applied to placement locations.

# Justification

This feature was discussed on [google groups](https://groups.google.com/d/msgid/openpnp/CALEDPUtGQs_46kxYZA7W7RVM-RyLK1zRrC-fpx_o7u9qx9j7CA%40mail.gmail.com?utm_medium=email&utm_source=footer)

There are various uses for aligning the nozzle during placement as if the part was being placed below the board Z:

1. When the vacuum turns on the surrounding air pressure pushes the
CP40 nozzle inward against the spring. If you set your z-locations with the
vacuum off, the nozzle tip will end up a little ways above those
locations with the vacuum on.

2. It is a margin of safety. If you have any errors in measuring part
height or board height, or any "bed leveling" imperfection, the
consequences for releasing the part too high are worse than for
releasing too low.

A commonly recommended hack is to set the board Z lower than reality. This is simple and effective, but causes undesirable parallax errors in fiducial location.

The real justification for this feature is that no-one will need that hack. No-one will need to deliberately enter an inaccurate board Z value.

image

# Other considerations

There was a suggestion of making this setting per-nozzle; but the job planner offers poor control over which nozzle is used for each placement.

There was a suggestion of making this setting per-part; but I  think the same could be achieved by setting a small part height

There was a suggestion of making this setting per-placement; but I think we already have that! The placement holds a `Location` which is X/Y/Z/C. The gui only offers to edit X, Y and rotation, and the Z field typically stays as zero, but I strongly  suspect setting a non-zero Z in the xml would be effective.

# Instructions for Use

A sentence to add to this [paragraph on the wiki](https://github.com/openpnp/openpnp/wiki/User-Manual)

The Z coordinate should be set to the Z coordinate of the nozzle tip when it is just barely touching the upper surface of the physical board when it is on the machine. Part height is added to this when placing a part so that the nozzle tip stops at the right height above the board. *The Placement Compression field (from the Machine Setup) is subtracted, so this field can be used to compensate if your nozzle contracts slighly under vacuum pressure*

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. -- not tested yet
3. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. -- **YES** the machine interface gains `getPlacementCompression()`


tobydickenson:
On review I think this would better as *Placement offset* not *placement compression*. It's the same thing but negative, to follow the conventional orientation of z axis things.

tobydickenson:
A good observation on the lumenpnp discord:

> can you set it per nozzle ?
> a wider nozzle compress the spring less
> since less surface area is closed

tobydickenson:
This PR is stalled. There has been discussion here and on lumenpnp discord of "good idea but it needs to be per-nozzle" or "good idea but it needs to be per-part". Or "per job because it depends on paste type' and we never saw any agreement for just the simple feature of a single z offset for placement.

PR #1941, opened 2026-02-08
Use advanced camera calibration to populate the 3D Calibration data.

kbralten:
# Description
automatic enablement of 3D calibration when advanced calibration completes. The feature automatically:

 - Enables the "3D Calibration?" checkbox
 - Populates Units Per Pixel values at two Z heights
 - Sets the Default Working Plane Z value
 - Handles edge cases gracefully

# Justification
As a new user, I didn't realize I was operating without 3D Calibration and then couldn't figure out how why it was disabled; the Advanced Camera Calibration path is a more user friendly and automated approach to capturing the 3D calibration data.

I implemented this by populating 3D Calibration, maybe the right solution is to patch all users of 3D calibration data but I wasn't as confident I'd find all of them.

# Instructions for Use
1. Perform Advanced Calibration. Notice that the 3D Calibration data is prepopulated for you and 3D Calibration is enabled. You can now use features like "Estimate Z Height".
2. If you don't want to repeat 3D Calibration, you can check "Skip New Collection And Reprocess Prior Collection" and your previously captured calibration data will be used to populate the 3D Calibration data.
3. If you disable Advanced Calibration, the 3D calibration settings remain enabled even if advanced calibration is later disabled, allowing for independent manual adjustment if desired.

# Implementation Details
1. How did you test the change? _Tested by repeatedly running Advanced Calibration both in full and by re-applying prior collection. Values match the Advanced Calibration results and are proximate (but I suspect more accurate) than those captured with the 3D Calibration method._
4. Did you follow the [coding style](https://github.com/openpnp/openpnp/wiki/Developers-Guide#coding-style)? _Yes_
5. If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages you will need to add additional justification for these changes. Changes to these packages require extensive review and testing. _No changes_
6. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. _tests green_.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/camera/calibration/AdvancedCalibration.java
+++ b/src/main/java/org/openpnp/machine/reference/camera/calibration/AdvancedCalibration.java
@@ -1473,4 +1473,79 @@ public class AdvancedCalibration extends LensCalibrationParams {
         camera.clearCalibrationCache();
         camera.captureTransformed();
     }
+    
+    /**
+     * Enables 3D calibration on the camera and sets the Units Per Pixel values from the  
+     * advanced calibration data. This should be called after successful advanced calibration
+     * completion to automatically configure the basic 3D calibration settings.
+     * 
+     * @param camera - the ReferenceCamera to configure
+     */
+    public void enable3DCalibration(ReferenceCamera camera) {
+        if (savedTestPattern3dPointsList == null || savedTestPattern3dPointsList.length < 2) {
+            // Need at least 2 Z heights for 3D calibration
+            Logger.debug("Advanced calibration does not have data at multiple Z heights, skipping 3D calibration auto-enable");
+            return;
+        }
+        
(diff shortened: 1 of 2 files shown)


tonyluken:
I think you have a fundamental misunderstanding. When Advanced Calibration is active, it provides all the functionality of 3D Calibration regardless of the settings in the 3D calibration section. It purposefully does not alter the values on the 3D calibration section so that if one disables Advanced Calibration, OpenPnP goes back to using the original 3D calibration data that was there before Advanced Calibration was run.

kbralten:
I wondered about that - that maybe the intention was to prefer Advanced Calibration in every point of use rather than back-porting to older data structures.
In that case see #1945 which overrides _Estimate Z Height of Object_ to use the Advanced Calibration data when available (and the camera is a ReferenceCamera).

PR #1948, opened 2026-03-10
AsyncGcodeDriver: Fix Marlin position synchronization issues causing machine crashes

Angelic47:
# Description
Marlin firmware occasionally prints unexpected or unsolicited position reports (for example, after executing a `G92` command).  
In `GcodeAsyncDriver`, this behavior causes the driver to synchronize with outdated position data, leading to severe synchronization errors and potential physical crashes.  

This PR introduces a targeted fix by explicitly forcing a position re-verification (`getReportedLocation`) if the firmware is identified as Marlin,  
ensuring the driver catches the correct, final position before draining the command queue.

# Justification
The existing OpenPnP Solutions suggest disabling `reportedLocationConfirmation` for Marlin to avoid issues.  
However, this does not address the root cause.  

Even with `reportedLocationConfirmation` set to `false`, the Reader Thread logic ([GcodeDriver:L1561](https://github.com/openpnp/openpnp/blob/main/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java#L1561)) unconditionally calls `processPositionReport(line);`, processing any unsolicited position report sent by Marlin.  
Because the existing `GcodeAsyncDriver` logic does not explicitly force a new `M114` to fetch the *true* final position at the end of the wait, Marlin's premature position reports are erroneously accepted as the final state.  

Consequently, Marlin users are effectively unable to use `GcodeAsyncDriver` safely, 
as machine crashes will almost inevitably occur at random moments.  

Additionally, the `reportedLocationConfirmation` routine is fundamentally unreliable for Marlin and must be avoided.  
Marlin tends to trigger position reports *before* the physical motion actually completes, which prematurely terminates the driver's wait.  
While existing Solutions already suggest disabling this, I have added explicit in-line comments in the code to clarify exactly why this routine is dangerous for Marlin.  

# Instructions for Use
Users do not need to perform any manual configuration or extra steps.  
This fix works automatically out-of-the-box for all `GcodeAsyncDriver` users.  
As long as the `FIRMWARE_NAME` property is correctly set and contains "Marlin", the driver will automatically enforce the position re-verification mechanism to ensure safety.  

# Implementation Details
1. **How did you test the change?** 
   - I initially experienced multiple severe head crashes on my OpenPnP machine when migrating to `GcodeAsyncDriver`. By analyzing the logs, I found that Marlin's position reports were breaking the command completion wait, causing the machine to execute subsequent commands at the wrong physical locations.
   - Following existing Solutions and historical issues/PRs, I disabled `reportedLocationConfirmation`. However, the crashes persisted. Log analysis revealed that Marlin's unsolicited position reports were still being accepted by the Reader Thread and erroneously used as the final position, breaking synchronization.
   - I investigated the `GcodeAsyncDriver` implementation and identified the root cause. As a temporary workaround, I appended an `M114` command to my `MOVE_TO_COMPLETE_COMMAND` to force a position refresh.
   - After extensive long-term testing on my physical machine, this workaround completely resolved the synchronization issues, and the random crashes entirely ceased.
   - Finally, I translated this proven workaround into a formal fix within the `GcodeAsyncDriver` source code, adding the targeted position re-verification step for Marlin and explanatory comments.
2. **Did you follow the coding style?** Yes.
3. **If you made changes in the `org.openpnp.spi` or `org.openpnp.model` packages...** N/A (No changes made to these packages).
4. **Be sure to run `mvn test` before submitting...** Yes, tests passed locally.

### DIFF
--- a/src/main/java/org/openpnp/machine/reference/driver/GcodeAsyncDriver.java
+++ b/src/main/java/org/openpnp/machine/reference/driver/GcodeAsyncDriver.java
@@ -381,9 +381,20 @@ public class GcodeAsyncDriver extends GcodeDriver {
             if (reportedLocationConfirmation) {
                 // Then make sure we get a uniquely recognizable confirmation. 
                 // Confirmation is signaled with a position report.
+
+                // NOTICE: This routine `reportedLocationConfirmation` CANNOT be used for Marlin firmware. 
+                // Marlin may report its position before the motion actually completes, 
+                // prematurely terminating the wait and causing synchronization issues.
                 getReportedLocation(timeout);
             }
             else {
+                // For Marlin firmware, we MUST re-verify the reported position. 
+                // This works around Marlin's unexpected position reports (e.g., after G92) that can cause synchronization issues.
+                if (getFirmwareProperty("FIRMWARE_NAME", "").contains("Marlin")) {
+                    Logger.trace("{} re-verifying position for Marlin firmware.", getName());
+                    getReportedLocation(timeout);
+                }
+
                 drainCommandQueue(timeout);
             }
             Logger.trace("{} confirmation complete.", getName());


tobydickenson:
Hi Angelic47, thank you for the detailed write up and concise code.

I am a Marlin user, and the original developer who identified that location confirmation was a problem, and added the I&S steps to disable it on Marlin.

I agree with the comment part of this PR.

Which version of Marlin are you using? I am using the version distributed by opulo with lumenpnp, which identifies as
```
FIRMWARE_NAME:Marlin bugfix-2.1.x (Mar 15 2024 14:51:01) SOURCE_CODE_URL:github.com/MarlinFirmware/Marlin PROTOCOL_VERSION:1.0 MACHINE_TYPE:LumenPnP REV04 EXTRUDER_COUNT:0 AXIS_COUNT:6 UUID:0045004E0200C00044B844B5
```

I am using Marlin with AsyncGcodeDriver without noticing any problems, and plenty of other lumenpnp users are too. So I am interested in your analysis about how this can cause machine crashes.

I do not follow the logic in the second half of this PR. You added some code to `waitForCompletion` in the `else` branch which handles `reportedLocationConfirmation` being disabled. This `else` branch uses `drainCommandQueue` for synchronisation.

You have added code to fetch another report location. I do not understand why that helps.

Your comments say:

```
// This works around Marlin's unexpected position reports (e.g., after G92) that can cause synchronization issues.
```

If location confirmation is disabled, gcode async driver does not use position reports for synchronisation.

Angelic47:
Hi, Sorry for my late reply.

Here's the situation. First, I think I need to provide some additional explanation about the root cause of the issue:

When `reportedLocationConfirmation` is disabled, AsyncGcodeDriver does NOT actually ignore the Position Reports from the Marlin Firmware.

On the contrary -- Marlin Firmware's Position Reports are always processed, regardless of the state of `reportedLocationConfirmation`.

See the code at [GcodeDriver.java#L1561](https://github.com/openpnp/openpnp/blob/5bd404cfc70f34103a3ca0fbb6b50c2b465f407c/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java#L1561):
```java
    protected void processResponse(Line line) {
        String regex = getCommand(null, CommandType.COMMAND_CONFIRM_REGEX);
        if (regex != null && line.getLine().matches(regex)) {
            receivedConfirmationsQueue.add(line);
        }
        regex = getCommand(null, CommandType.COMMAND_ERROR_REGEX);
        if (regex != null && line.getLine().matches(regex)) {
            errorResponse = line;
        }
        processPositionReport(line); // Where the Position Report is processed unconditionally
    }
```

`GcodeDriver.java` might seem confusing, but in fact, `AsyncGcodeDriver` extends the `GcodeDriver` class, and its Reader Thread is entirely implemented by `GcodeDriver`.

So the conclusion is: regardless of the state of `reportedLocationConfirmation`, Position Reports from Marlin Firmware are always processed.

----

Now let's take a look at what `getReportedLocation` actually does:

In fact... it also comes from `GcodeDriver.java` through class extends, located at [GcodeDriver.java#L779](https://github.com/openpnp/openpnp/blob/5bd404cfc70f34103a3ca0fbb6b50c2b465f407c/src/main/java/org/openpnp/machine/reference/driver/GcodeDriver.java#L779):
```java
    @Override
    public AxesLocation getReportedLocation(long timeout) throws Exception {
        String command = getCommand(null, CommandType.GET_POSITION_COMMAND);
        if (command == null) {
            throw new Exception(getName()+" configuration error: missing GET_POSITION_COMMAND.");
        }
        if (getCommand(null, CommandType.POSITION_REPORT_REGEX) == null) {
            throw new Exception(getName()+" configuration error: missing POSITION_REPORT_REGEX.");
        }

        // TODO: true queued reporting. For now it is sufficient to poll one for one.
        reportedLocationsQueue.clear();
        sendGcode(command, -1);
        if (timeout == -1) {
            timeout = infinityTimeoutMilliseconds;
        }
        AxesLocation lastReportedLocation = reportedLocationsQueue.poll(timeout, TimeUnit.MILLISECONDS);
        if (lastReportedLocation != null) {
            Logger.trace("{} got lastReportedLocation {}", getName(), lastReportedLocation);
            return lastReportedLocation;
        }
        // Timeout expired.
        throw new Exception(getName()+" timeout waiting for response to " + command);
    }
```

So the naming of this function is actually somewhat misleading.

It doesn't retrieve the position from a previously reported Position Report from Marlin Firmware, instead, it sends a specific Gcode command to request a Position Report, and waits for the response.

This function has **blocking** behavior. So it was used as the implementation mechanism for `reportedLocationConfirmation`: "when a Location Report arrives, we consider the preceding series of GCode commands to be completed."

As we already know this approach is unreliable with Marlin Firmware, because Marlin's Position Reports don't always arrive at the expected time, they can also appear at random or unpredictable moments in between.

**But here's the thing:** 

Looking at it this way, the original design intent of `reportedLocationConfirmation` was to use Position Reports to confirm that the preceding GCode commands had finished executing, and it has **nothing** to do with whether Position Reports themselves are processed or not.

Therefore, disabling `reportedLocationConfirmation` does **NOT** cause Position Reports to be ignored, it simply means we no longer use Position Reports to confirm the execution status of GCode commands.

And this is where things become Crystal Clear.

There is actually a very dangerous **timing/sequencing logic error** here: 

> If Marlin Firmware's Position Report is reported mid-execution, it gets processed and becomes something like a "latch" inside GCodeAsyncDriver: *it stores a stale, outdated coordinate rather than the current one*.  
> Then, when the commands finish executing and `drainCommandQueue` completes, the signal is sent, and when OpenPnP plans the next batch of commands, this stale coordinate gets applied to OpenPnP's internal state.

OpenPnP then plans the next set of commands **based on** this outdated coordinate, causing the entire coordinate system to go haywire, ultimately resulting in a Machine Crash.

----

So what I actually did is: *use `getReportedLocation` to re-send `M114`*.   
That's all there is to it, no ambiguity here.

This way, `M114` will always be the last command in every batch of GCode commands, ensuring that regardless of when Marlin Firmware reports its Position Report, we always get a fresh, reliable Position Report at the end of command execution.

This corrects the timing/sequencing logic error described above.  

As for why `drainCommandQueue` still needs to be kept?

Very easy to understand. 
As we discussed above: although `getReportedLocation` is a blocking function, if we use it to block as the signal for command completion, it can be interrupted at any random moment by Marlin Firmware's Position Report.

So `drainCommandQueue` must be retained here: it serves as the catch-all guarantee that all commands have finished executing, regardless of when Marlin Firmware's Position Report arrives.

In other words, each does its own job: `getReportedLocation` is repurposed solely as an `M114` sending function, while `drainCommandQueue` still basically handles the command completion signal.

----

As for my Marlin firmware, I'm using my own build compiled back in 2022.

But I later found that this issue is actually unrelated to the firmware version, because the problem still exists in the latest firmware.

Let's take a look at [G92.cpp](https://github.com/MarlinFirmware/Marlin/blob/5b46ba9923d0a48332520e7a7b5358a6c24ae5ce/Marlin/src/gcode/geometry/G92.cpp#L127) and see how Marlin Firmware handles the G92 command:
```cpp
void GcodeSuite::G92() {

  // ......
  // ......

  // L116

  #if ENABLED(CNC_COORDINATE_SYSTEMS)
    // Apply Workspace Offset to the active coordinate system
    if (WITHIN(active_coordinate_system, 0, MAX_COORDINATE_SYSTEMS - 1))
      coordinate_system[active_coordinate_system] = motion.workspace_offset;
  #endif

  if (sync_XYZE) motion.sync_plan_position();
  #if HAS_EXTRUDERS
    else if (sync_E) motion.sync_plan_position_e();
  #endif

  IF_DISABLED(DIRECT_STEPPING, motion.report_position()); // <-- Position Report is triggered here after processing G92
}
```

`DIRECT_STEPPING` is a special feature toggle in Marlin Firmware, and it is disabled in virtually all common Marlin firmware configurations.

Therefore, after processing the G92 command, Marlin Firmware always triggers a Position Report.

So this issue has actually always existed in Marlin Firmware, and it is independent of the firmware version.

As for why LumenPnP is unaffected -- the reason is that LumenPnP actually uses GcodeDriver, not AsyncGcodeDriver.
Let's take a look at [machine.xml#L434](https://github.com/opulo-inc/lumenpnp/blob/93f8124f24e04245a856cc3c9731e8de32d38726/openpnp/machine.xml#L434):
```xml
    <!-- ... before ... -->
    <!-- ... before ... -->
      <drivers>
         <driver class="org.openpnp.machine.reference.driver.GcodeDriver" id="DRV16982438146c1dd4" name="GcodeDriver" motion-control-type="ConstantAcceleration" communications="serial" connection-keep-alive="true" sync-initial-location="true" allow-unhomed-motion="false" units="Millimeters" max-feed-rate="0" backlash-offset-x="-1.0" backlash-offset-y="-1.0" backlash-offset-z="0.0" backlash-offset-r="0.0" non-squareness-factor="0.0" backlash-feed-rate-factor="0.1" timeout-milliseconds="10000" connect-wait-time-milliseconds="6000" dollar-wait-time-milliseconds="50" visual-homing-enabled="true" backslash-escaped-characters-enabled="false" remove-comments="false" compress-gcode="false" compression-excludes="[]&quot;" logging-gcode="false" supporting-pre-move="false" using-letter-variables="true" infinity-timeout-milliseconds="60000">
    <!-- ... after ... -->
    <!-- ... after ... -->
```

So LumenPnP does not actually use AsyncGcodeDriver, at least not in its default configuration.

LumenPnP likely ran into this issue but couldn't identify the root cause, so they chose to fall back to GcodeDriver as a workaround.

----

So in summary, I added a single line of `getReportedLocation` to ensure that the last command in every batch of GCode commands is always `M114`, guaranteeing that we obtain a final, up-to-date Position Report upon command completion, thus resolving the timing/sequencing error in the synchronization.

I believe this approach should be logically sound, it should not introduce new issues, nor cause any other synchronization errors.

tobydickenson:
I suspect this report is AI nonsense. There are lots of words here, but none of them explain how any problem can cause a crash. "the entire coordinate system to go haywire" does not explain the failure mechanism.

Angelic47:
> I suspect this report is AI nonsense. There are lots of words here, but none of them explain how any problem can cause a crash. "the entire coordinate system to go haywire" does not explain the failure mechanism.

So using AI as grammarly means I'm AI, and the report itself is AI nonsense?
That would be funny, isn't it?

You may suspect this is fully AI-generated, that's totally fine. I won't argue any about it.

But I do feel sad that I put into writing all of this is dismissed with a single line. 
All of these was discovered & written by myself, starting by several damaged Juki 503 nozzles, and digging through the logs line by line to trace the cause. 

Yeah I can't through a screen to provide any information about I'm human or not. And I don't want to.
The length of my analysis is actually my sin in the age of AI.

**If you believe this is not a valid report, feel free to close it. I just leave the content here for anyone who might find it useful in the future. It's already enough for me.**

Angelic47:
Honestly, If there’s actually an AI so magic there and has capable to fix these edge-case bugs, please share the provider and your aff link. 
I’d love to just retire, okay? Debugging these issues manually makes me headache.

Angelic47:
> If Marlin Firmware's Position Report is reported mid-execution, it gets processed and becomes something like a "latch" inside GCodeAsyncDriver: it stores a stale, outdated coordinate rather than the current one.
> Then, when the commands finish executing and drainCommandQueue completes, the signal is sent, and when OpenPnP plans the next batch of commands, this stale coordinate gets applied to OpenPnP's internal state.
> OpenPnP then plans the next set of commands based on this outdated coordinate, causing the entire coordinate system to go haywire, ultimately resulting in a Machine Crash.

**PLEASE Read my posts CAREFULLY first.**

If you don't understand, use an AI summary by yourself to /explain what it means.  
  
**TLDR: Timing logic error.** Old position is reported & used, then caused a crash next time.  
Any other problems? I think it's clear enough.  
  
Logs have ~1M and are nearly unreadable.  
I think there is no need to post here. You don't want to see that.  
  
**To reproduce**, just add a **G92** to your A/B axis rotating command and see what happens, that is enough.

tobydickenson:
Hi Angelic47,

I still dont understand how the existing code is going wrong.

We know that Marlin sends spurious position reports. The **timing** of those spurious reports means that we can't use the "location confirmation" feature with Marlin. I&S already takes care of preventing this.

Your patch fetches a new **value** for the position report, but I dont see where the position report value (spurious or otherwise) gets used. Given how openpnp backlash compensation works, the position report values are not particularly useful. Openpnp java is the authority on logical machine position, not the controller running the gcode. I might be wrong, but I believe your patch re-fetches a value which is never used.

It sounds like you understand a failure mechanism, but you have not communicated it here. You wrote "OpenPnP then plans the next set of commands based on this outdated coordinate". It would be helpful if you could point out where openpnp incorrectly calculates commands based on the (possibly incorrect) position report.

Angelic47:
Hi tobydickenson,

I think maybe I missed something, but the position reported by marlin **does** get used.  

The log is uploaded here, for your reference.  

[openpnp.log](https://github.com/user-attachments/files/28056770/openpnp.log)

In this log file, the position is reported at Line 9510; and GcodeDriver's Reader Thread then confirms the position at Line 9512.  
The position was then used at Line 9528.  

For convenience, I attached the relevant log lines here:

```log
...

// L9508: G92 Sent to Marlin
2026-02-26 22:35:19.246 GcodeAsyncDriver$WriterThread TRACE: [GcodeAsyncDriver:ttyACM0] >> G92     B0.5313 ; reset coordinates

...

// L9510: Marlin processed G92 and reports the position
2026-02-26 22:35:19.246 GcodeDriver$ReaderThread TRACE: [GcodeAsyncDriver:ttyACM0] << X:353.09 Y:256.13 Z:0.00 A:179.20 B:0.53 Count X:70443 Y:50977 Z:-2 A:-8007 B:-9595

...

// L9512: GcodeDriver's Reader Thread confirms the position
2026-02-26 22:35:19.246 GcodeDriver TRACE: Position report: X:353.09 Y:256.13 Z:0.00 A:179.20 B:0.53 Count X:70443 Y:50977 Z:-2 A:-8007 B:-9595

...
...

// Lots of G1 commands to move the machine

...
...

// L9528: The position is getting used by AbstractMotionPlanner
2026-02-26 22:35:19.671 AbstractMotionPlanner DEBUG: Reported location changes current location from (x:353.088612, y:256.128360, ZN:46.559345, C1:179.295535, C2:0.531270) to (x:353.090000, y:256.130000, ZN:0.000000, C1:179.200000, C2:0.530000)

...
```

Please notice here **`ZN:46.559345` becomes `ZN:0.000000`**, which is an absolutely **outdated position** (which from L9512 and **before** the G1 commands) .

Then my machine crashed at Line 9551, which is a head-moving command without lifting the head up:

```log
2026-02-26 22:35:19.685 GcodeAsyncDriver DEBUG: [GcodeAsyncDriver:ttyACM0] commandQueue offer >> G1 X380.7784 Y256.1508  A0.5313  F24000.00
```

The nozzle moved quickly without any Z lift, smashing my C&R strips and damaging the nozzle.

By searching the strings, I found the position is getting used at [AbstractMotionPlanner.java#L997](https://github.com/openpnp/openpnp/blob/5bd404cfc70f34103a3ca0fbb6b50c2b465f407c/src/main/java/org/openpnp/machine/reference/driver/AbstractMotionPlanner.java#L997). 

I hope this log can help you understand the issue better.  
Maybe my setup is a bit special, or maybe I missed some configuration. But I do think there are some issues with the position report handling here.

tobydickenson:
please can you share your machine.xml

tobydickenson:
> please can you share your machine.xml

@Angelic47 any update on this please? can you share your machine.xml? I still suspect the underlying problem is a configuration error.

PR #1950, opened 2026-03-20
change various length columns to be right aligned, and not use space padding

tobydickenson:
# Description

This changes the `displayAligned` parameter of `LengthCellValue` to avoid using space character padding for alignment. There is a corresponding change to make those columns right aligned, so that the decimal points all line up. 

This leads to a much more compact column presentation. Previously numeric cells were padded with 5 characters to the left of the decimal point, and 5 characters to the right. Such cells can not be made particularly narrow, which is a problem when working with large tables on small screens.

@tonyluken originally wrote this so hopefully he can review.

# Justification
This was discussed on [google groups](https://groups.google.com/d/msgid/openpnp/9e1b21aa-461a-433a-89d5-476b3d4bb525n%40googlegroups.com?utm_medium=email&utm_source=footer)

**Before:** This screenshot show some wasted space in the placement X and Y columns. The board X, Y and Z columns have used ellipses to obliterate the useful data, while preserving space for all the padding spaces.
image

**After:** A compact layout with no wasted padding
image

# Instructions for Use
There is no change to usage.

# Implementation Details
    * viewed my production board data to confirm everything looks ok
    * edited some x/y/z/rotation data in some boards
    * edited some other field on a board.
6. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. -- tests pass

### DIFF
--- a/src/main/java/org/openpnp/gui/tablemodel/PlacementsHolderLocationsTableModel.java
+++ b/src/main/java/org/openpnp/gui/tablemodel/PlacementsHolderLocationsTableModel.java
@@ -75,8 +75,8 @@ public class PlacementsHolderLocationsTableModel extends AbstractObjectTableMode
             LengthCellValue.class, Side.class, LengthCellValue.class, LengthCellValue.class,
             LengthCellValue.class, RotationCellValue.class, Boolean.class, Boolean.class};
 
-    private int[] columnAlignments = new int[] {LEFT, LEFT, CENTER, CENTER, CENTER, CENTER, 
-            CENTER, CENTER, CENTER, CENTER, CENTER};
+    private int[] columnAlignments = new int[] {LEFT, LEFT, RIGHT, RIGHT, CENTER, RIGHT,
+            RIGHT, RIGHT, RIGHT, CENTER, CENTER};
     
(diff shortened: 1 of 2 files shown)


tobydickenson:
I am not sure whether to suggest a further change. Openpnp uses `MonospacedFontTableCellRenderer` for length values. But we *probably* dont **need** a monospace font here. The default system font probably works sufficiently well to get the decimal points lined up because the digits have uniform width. 

That saves a few percent horizontal space. It would look like this:

image

tonyluken:
I'm good with this change. I think the only reason I did it the way I did was so that the decimal points would stay aligned regardless of whether the column was aligned left, right, or centered. WRT the mono space font - are you positive everyone's default system font has uniform digit widths? I just don't know.

tobydickenson:
> WRT the mono space font - are you positive everyone's default system font has uniform digit widths? I just don't know.

Maybe not pixel perfect. I know Arial font has kerning for adjacent `1` digits that can cause a tiny irregularity on some systems. I'm not sure if that matters.

tonyluken:
> Maybe not pixel perfect. I know Arial font has kerning for adjacent 1 digits that can cause a tiny irregularity on some systems.

In that case, I'd recommend keeping the monospaced font so that we're sure things remain aligned.

PR #1952, opened 2026-03-25
Job Planner TravelCost optimisations

tobydickenson:
# Description

On a big job yesterday I noticed some behaviour that suggests the system was cpu-bound during the job planning stage. So I looked at the inner loops, and spotted a couple of easy optimisation opportunities.

This is purely a saving in cpu cycles. Job planner and machine behaviour should be unchanged. Details are in the commit comments.

Unit test still pass.

I have not yet proven this helps in practice, so this is a provisional PR for now.

# Justification

Less cpu for the same behaviour

# Instructions for Use

No changes

# Implementation Details
1. How did you test the change? Be descriptive. Untested code will not be accepted. -- this is still outstanding
4. Be sure to run `mvn test` before submitting the Pull Request. If the tests do not pass the Pull Request will not be accepted. -- tests still pass


tobydickenson:
The benefit is negligible. PR withdrawn

PR #1977, opened 2026-08-23
Flatlaf update

dkonigsberg:
# Description
This updates the FlatLaf library, and associated modules, to their latest versions. It also adds additional themes that were added in recent versions of FlatLaf as well as the extra IntelliJ themes pack provided by that project.

# Justification
OpenPnP currently uses a very old version of FlatLaf.  Since then, the upstream project has made various minor cosmetic improvements to there themes (such as fixing the visibility of table header dividers), and also added several additional themes.

# Instructions for Use
Simply rebuild OpenPnP and notice new options in the Appearance Settings dialog.

# Implementation Details
1. Change as been tested on local machine, but I encourage others to give it a try.
2. This was a few minor updates to `pom.xml` and a few additional lines in `ThemesSettingsPanel`.  Fortunately, the upstream project provided a built-in list of IntelliJ themes so that was a relatively easy addition.
3. This change has been split into 3 separate commits, making it easier to separate the various elements of it if the project maintainers don't want to pick up all of it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -212,17 +212,17 @@
 		<dependency>
 			<groupId>com.formdev</groupId>
 			<artifactId>flatlaf</artifactId>
-			<version>1.0</version>
+			<version>3.7.2</version>
 		</dependency>
 		<dependency>
 			<groupId>com.formdev</groupId>
 			<artifactId>flatlaf-extras</artifactId>
-			<version>1.0</version>
+			<version>3.7.2</version>
 		</dependency>
 		<dependency>
 			<groupId>com.formdev</groupId>
 			<artifactId>svgSalamander</artifactId>
-			<version>1.1.2.4</version>
+			<version>1.1.4</version>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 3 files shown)


tobydickenson:
Thank you for the PR.

I know nothing of flataf, and still very little about java swing, but this looks good to me.

> OpenPnP currently uses a very old version of FlatLaf.  Since then, the upstream project has made various minor cosmetic improvements

Running old and new version side by side, I could not see any difference with this. This is not necessarily a problem of course. This was on ubuntu 26.04. Can you highlight same flataf changes that affect openpnp? If there are changes, it would be good to highlight them in the changelog for this openpnp version.

> fixing the visibility of table header dividers

I'm not sure exactly what this refers to, but I could see no difference in table headers.

dkonigsberg:
Here's a side-by-side example from one of my systems:
image
(Its possible that this isn't visible on all systems or themes or DPI levels, but I've generally noticed that table header dividers are often invisible in OpenPnP currently.)

PR #1978, opened 2026-08-31
Add feeder firmware version query for Photon feeders

dkonigsberg:
# Description
The Opulo Photon feeders have a command that can be used to query their firmware version, implemented in a slightly awkward way under their "vendor" command code.  This change implements that command code, and wraps it with all the necessary logic to actually do a version string query.  It then runs these commands as part of a feeder search/find, and shows them in the related feeder info panel.

# Justification
Lately Opulo has been releasing regular firmware updates for their Photon feeders as they refine their next version.  As its common to have many such feeders, some on and some off the PnP machine, it can be quite easy to lose track of what firmware versions you have installed on each.  As such, I really wanted an easy way to just query all my feeders to find out.

# Instructions for Use
The version query happens implicitly as part of the feeder search/find process, and is displayed as an additional row in each feeder's "Info" panel.

# Implementation Details
1. Added a `VendorOptions` class to wrap the "vendor" command provided by the Photon feeders
2. Added code to `PhotonFeeder` to wrap this command with the necessary setup and loops to fully tease out the Photon firmware version string.  This can take multiple commands, but in practice never takes more than one as the version string is always of reasonable length.
3. Added code to `PhotonFeeder` to run this query as part of the feeder search/find processes.
4. Updated `PhotonFeederTest` and related classes to do very rudimentary testing of this new operation, by adding mocked commands to existing unit tests that implicitly trigger it.

The unit tests pass and I've tested this change on my own LumenPnP system.

Its possible there are additional concerns with how I've implemented this, so feel free to review in case any of those are apparent.  However, I wanted to toss this out there in case it is useful to anyone else.

### DIFF
--- a/src/main/java/org/openpnp/machine/photon/PhotonFeeder.java
+++ b/src/main/java/org/openpnp/machine/photon/PhotonFeeder.java
@@ -41,6 +41,8 @@ public class PhotonFeeder extends ReferenceFeeder {
 
     protected Integer slotAddress = null;
 
+    protected String firmwareVersion = null;
+    
     @Attribute(required = false)
     protected int partPitch = 4;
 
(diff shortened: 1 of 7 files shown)


tobydickenson:
Implementation looks good to me.

A screenshot on the PR is always appreciated.

When creating a PR there is a template of questions to fill in. Please can you update this PR with that template. (I should have raised this for your previous PR too. No worries on that one, but please do update this one)

dkonigsberg:
Screenshot showing the new "Firmware Version" field:
Screenshot From 2026-08-31 10-48-01

I thought I did fill in the template, unless there's some nuance to it that I missed or some boilerplate text I shouldn't have replaced.

tobydickenson:
There is a section that begins "Fill out all the details below. All sections below are required", and it is helpful to have those questions and answers on the PR.  #1976  was the PR preceding this one for example.

No worries - this one is merged

